PR #417, opened 2022-05-13
Datamodel support for icing and aero

lagacep-ans:
Exposure of icing/aero datamodels through the session.icing and session.aero attributes.

New launcher fluent_icing=True , fluent_aero=True  arguments, to launch fluent in that mode.
Usage requirement a fluent v231 with the -flicing / -flaero command line support (PR pending)

If used with v222 or a fluent unable to provided the service, the icing / aero entries will not be offered in the session.


lagacep-ans:
Old initial implementation of a few months ago, deleting this branch - superseded by a newer PR

PR #607, opened 2022-07-11
Tidy up descriptions a little bit

dnwillia-work:
I had added these with generic descriptions a while back and thought they can be clarified now.

### DIFF
--- a/examples/00-fluent/README.txt
+++ b/examples/00-fluent/README.txt
@@ -1,3 +1,6 @@
-Full Examples Using PyFluent
+Complete Simulation Examples
 ============================
-These examples demonstrate how to use Fluent capabilities from Python.
+The following examples demonstrate how to use Fluent capabilities from Python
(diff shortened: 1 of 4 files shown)


examples/00-fluent/README.txt
```diff
@@ -1,3 +1,6 @@
-Full Examples Using PyFluent
-============================
-These examples demonstrate how to use Fluent capabilities from Python.
\ No newline at end of file
+Complete Workflow Examples
```
smf2001, line 1: I would avoid the general use of 'workflow' in the title such as this as they have a specific meaning in meshing (watertight and fault-tolerant) and the solver (turbo workflow). The "Complete Workflow Examples" title may imply that this is an example of using the meshing or a solver workflow when it's more of an 'end-to-end' process. I would suggest something like "Complete" or "Comprehensive"  "Examples of Using PyFluent"

smf2001 (approved):

PR #648, opened 2022-07-20
Support 22.2 and 23.1 backends

mkundu1:
Packages and documentation are not uploaded for 23.1 backend. I've just disabled the verbose logging during doc build - will check how much API information is exposed from logs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,15 +91,14 @@ jobs:
     runs-on: [self-hosted, pyfluent]
     strategy:
       matrix:
-        #image-tag: [latest, solver-latest]
-        image-tag: [latest]
+        image-tag: [latest, solver-latest]
 
     steps:
       - uses: actions/checkout@v3
 
       # used for documentation deployment
       - name: Get Bot Application Token
-        if: github.event_name == 'push' && contains(github.ref, 'refs/tags')
+        if: github.event_name == 'push' && contains(github.ref, 'refs/tags') && matrix.image-tag == 'latest'
         id: get_workflow_token
         uses: peter-murray/workflow-application-token-action@v1
         with:
@@ -190,6 +189,7 @@ jobs:
           FLUENT_IMAGE_TAG: ${{ matrix.image-tag }}
 
       - name: Upload HTML Documentation
+        if: matrix.image-tag == 'latest'
         uses: actions/upload-artifact@v3
         with:
           name: HTML-Documentation-tag-${{ matrix.image-tag }}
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -91,15 +91,14 @@ jobs:
     runs-on: [self-hosted, pyfluent]
     strategy:
       matrix:
-        #image-tag: [latest, solver-latest]
-        image-tag: [latest]
+        image-tag: [latest, solver-latest]
```
dnwillia-work, line 94: OK, so `latest` would just be the cad+solver image and `solver-latest` would be the solver only image.  So, staring at these changes what it would do is run the CI/CD 2x right?

dnwillia-work: i.e. I think the `latest` image is constructed using this Dockerfile:

https://github.com/pyansys/pyfluent/blob/main/docker/fluent/Dockerfile.meshing

You can see on line 6 that it is adding the CAD content to the `solver-latest` image which is built using this file

https://github.com/pyansys/pyfluent/blob/main/docker/fluent/Dockerfile

dnwillia-work: I think the second run with `solver-latest` would (or should) not actually be able to build the documentation because that requires the cad import right.  The PR CI/CD with the `solver-latest` image did not rebuild the examples:

https://github.com/pyansys/pyfluent/runs/7433287000?check_suite_focus=true

I don't think it would have worked if this actually ran.

mkundu1: Yeah, I misunderstood the tag names, we don't have the 23.1 image yet. I'll update the .yml to upload the artifacts for 22.2 image only.

.github/workflows/ci.yml
```diff
@@ -91,15 +91,14 @@ jobs:
     runs-on: [self-hosted, pyfluent]
     strategy:
       matrix:
-        #image-tag: [latest, solver-latest]
-        image-tag: [latest]
+        image-tag: [v22.2.0]
```
mkundu1, line 94: Add v23.1.0 later

.github/workflows/ci.yml
```diff
@@ -212,8 +212,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        #image-tag: [latest, solver-latest]
-        image-tag: [latest]
+        image-tag: [v22.2.0]
```
mkundu1, line 215: Add v23.1.0 later

seanpearsonuk (approved):

dnwillia-work (approved):

PR #651, opened 2022-07-21
Use static info for tuigen

mkundu1:
For 23.1, `GetStaticInfo` rpc will be used for tuigen. I've also reduced the codegen frequency in CI. For 22.2, we shall always restore the generated code from cache (changes are not in the current PR).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,8 +142,12 @@ jobs:
             src/ansys/fluent/core/meshing/tui.py
             src/ansys/fluent/core/solver/settings
             src/ansys/fluent/core/solver/tui.py
-          key: API-Code-v${{ env.API_CODE_CACHE }}-${{ steps.version.outputs.PYFLUENT_VERSION }}-${{ matrix.image-tag }}-${{ hashFiles('codegen/**') }}-${{ github.sha }}
-          restore-keys: API-Code-v${{ env.API_CODE_CACHE }}-${{ steps.version.outputs.PYFLUENT_VERSION }}-${{ matrix.image-tag }}-${{ hashFiles('codegen/**') }}
+            doc/source/api/core/meshing/tui
+            doc/source/api/core/meshing/datamodel
+            doc/source/api/core/solver/tui
+            doc/source/api/core/solver/datamodel
+          key: API-Code-v${{ env.API_CODE_CACHE }}-${{ steps.version.outputs.PYFLUENT_VERSION }}-${{ matrix.image-tag }}-${{ hashFiles('codegen/**') }}
+          restore-keys: API-Code-v${{ env.API_CODE_CACHE }}-${{ steps.version.outputs.PYFLUENT_VERSION }}-${{ matrix.image-tag }}
 
       - name: Login to GitHub Container Registry
         if: steps.cache-api-code.outputs.cache-hit != 'true'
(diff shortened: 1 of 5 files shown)


src/ansys/fluent/core/services/error_handler.py
```diff
@@ -14,7 +12,6 @@ def func(*args, **kwargs) -> Callable:
         try:
             return f(*args, **kwargs)
         except grpc.RpcError as ex:
-            LOG.error("GRPC_ERROR: %s", ex.details())
```
mkundu1, line 17: Already re-throwing, so no much point in additionally logging it

src/ansys/fluent/core/services/datamodel_tui.py
```diff
@@ -197,6 +202,45 @@ def get_doc_string(self, include_unavailable: bool = False) -> str:
...
 
+    def get_static_info(self) -> Dict[str, Any]:
+        """Get static info at menu level.
+
+        Returns
+        -------
+        DataModelProtoModule.StaticInfo
+            static info
+        """
+        if hasattr(DataModelProtoModule, "GetStaticInfoRequest"):
```
mkundu1, line 213: returns the info dictionary either using the new `GetStaticInfo` rpc (for 23.1) or recursively retrieving child item info (for 22.2)

src/ansys/fluent/core/launcher/fluent_container.py
```diff
@@ -57,7 +57,7 @@ def start_fluent_container(mounted_from: str, mounted_to: str, args: List[str])
                 "-e",
                 "FLUENT_LAUNCHED_FROM_PYFLUENT=1",
                 "ghcr.io/pyansys/pyfluent",
-                "-g",
+                "-gu",
```
mkundu1, line 60: Apparently this is now necessary, otherwise getting an error while retrieving child-names at `display/set/picture/driver`. Earlier probably it was not going upto that level.

mkundu1, line 60: Runing the container in `-gu`

.github/workflows/ci.yml
```diff
@@ -257,8 +261,8 @@ jobs:
             doc/source/api/core/meshing/datamodel
             doc/source/api/core/solver/tui
             doc/source/api/core/solver/datamodel
-          key: API-Code-v${{ env.API_CODE_CACHE }}-${{ steps.version.outputs.PYFLUENT_VERSION }}-${{ matrix.image-tag }}-${{ hashFiles('codegen/**') }}-${{ github.sha }}
-          restore-keys: API-Code-v${{ env.API_CODE_CACHE }}-${{ steps.version.outputs.PYFLUENT_VERSION }}-${{ matrix.image-tag }}-${{ hashFiles('codegen/**') }}
+          key: API-Code-v${{ env.API_CODE_CACHE }}-${{ steps.version.outputs.PYFLUENT_VERSION }}-${{ matrix.image-tag }}-${{ hashFiles('codegen/**') }}
```
mkundu1, line 264: This should improve the caching strategy. Right now the codegen is running too frequently - with every new PR.

.github/workflows/ci.yml
```diff
@@ -143,8 +143,12 @@ jobs:
             src/ansys/fluent/core/meshing/tui.py
             src/ansys/fluent/core/solver/settings
             src/ansys/fluent/core/solver/tui.py
-          key: API-Code-v${{ env.API_CODE_CACHE }}-${{ steps.version.outputs.PYFLUENT_VERSION }}-${{ matrix.image-tag }}-${{ hashFiles('codegen/**') }}-${{ github.sha }}
-          restore-keys: API-Code-v${{ env.API_CODE_CACHE }}-${{ steps.version.outputs.PYFLUENT_VERSION }}-${{ matrix.image-tag }}-${{ hashFiles('codegen/**') }}
+            doc/source/api/core/meshing/tui
+            doc/source/api/core/meshing/datamodel
+            doc/source/api/core/solver/tui
+            doc/source/api/core/solver/datamodel
+          key: API-Code-v${{ env.API_CODE_CACHE }}-${{ steps.version.outputs.PYFLUENT_VERSION }}-${{ matrix.image-tag }}-${{ hashFiles('codegen/**') }}
```
mkundu1, line 150: For 22.2, we will use `key: API-Code-v${{ env.API_CODE_CACHE }}-${{ steps.version.outputs.PYFLUENT_VERSION }}-${{ matrix.image-tag }}` as key here.

.github/workflows/ci.yml
dnwillia-work: Yeah, was there a reason you put `github.sha` in the key before?

.github/workflows/ci.yml
dnwillia-work: OK, that's fine.  Though, if we wanted to force regenerating it for some reason which part of the key would we update?

Also I guess the idea with hashing the codegen package is that if anything changes in the there then it should force a regeneration?

dnwillia-work (approved):

.github/workflows/ci.yml
mkundu1: @dnwillia-work I originally copied this from [pymapdl](https://github.com/pyansys/pymapdl/blob/main/.github/workflows/ci.yml#L182).

seanpearsonuk (approved):

PR #660, opened 2022-07-25
Replaced PYPI_TOKEN and pull_fluent_image.py

hpohekar:
Replaced PYPI_TOKEN in ci.yml.
Replaced pull_fluent_image.py with pull_fluent_image.sh


.ci/pull_fluent_image.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
dnwillia-work, line 1: You should use: 
```suggestion
#!/bin/sh
```
for best portability.

hpohekar: Hi Dan,
Yes, sure. 
Thank you.

seanpearsonuk (approved):

dnwillia-work (approved):

PR #663, opened 2022-07-25
Removed pylint

hpohekar:
Removed pylint and few more changes.

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -448,7 +448,6 @@ class NamedObject(SettingsBase[DictStateType], Generic[ChildTypeT]):
     def _create_child_object(self, cname: str):
         ret = self._objects.get(cname)
         if not ret:
-            # pylint: disable=no-member
             cls = self.__class__.child_object_type
             ret = self._objects[cname] = cls(cname, self)
         return ret
@@ -571,7 +570,6 @@ class ListObject(SettingsBase[ListStateType], Generic[ChildTypeT]):
     query_names = []
 
     def _update_objects(self):
-        # pylint: disable=no-member
         cls = self.__class__.child_object_type
         self._setattr("_objects", [cls(str(x), self) for x in range(self.get_size())])
 
@@ -619,7 +617,7 @@ class Map(SettingsBase[DictStateType]):
 
 
 def _get_new_keywords(obj, kwds):
-    newkwds = {}
+    newkwds = dict()
     for k, v in kwds.items():
         if k in obj.argument_names:
             ccls = getattr(obj, k)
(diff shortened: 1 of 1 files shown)


dnwillia-work (approved):

seanpearsonuk (approved):

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -925,7 +922,6 @@ def get_root(flproxy) -> Group:
         cls = settings.root
     except Exception:
```
dnwillia-work, line 923: Looking at the complete code for this function it seems not right:
```
    try:
        from ansys.fluent.core.solver import settings

        if settings.SHASH != _gethash(obj_info):
            LOG.warning(
                "Mismatch between generated file and server object "
                "info. Dynamically created settings classes will "
                "be used."
            )
        cls = settings.root
    except Exception:
        cls = get_cls("", obj_info)
```
It raises a `RuntimeError` which will force it to go into the Exception handler case right?  If so, should line 923 not be
```suggestion
    except RuntimeError:
```
as it is now, all exceptions are caught and it's not clear from the code in the try block if all possible exceptions could be thrown.

hpohekar: Hi Dan, Yeah, sure. Thank you.

mkundu1 (approved):

prmukherj (approved):

src/ansys/fluent/core/solver/flobject.py
dnwillia-work: If you want to merge this code as-is and do it later it's fine.  Just to be clear.

I only noticed this because I ran `pylint` on the module locally and it highlights the issue right.

dnwillia-work (approved):

src/ansys/fluent/core/solver/flobject.py
hpohekar: Hi Dan, 
I am thinking to implement the suggestion simultaneously for both Fluent and pyfluent. 
Thank you.

hpohekar: Hi Dan,
Sean and I have gone through the code, we haven't found other specific spots.  We have replaced Exception with RuntimeError as per above suggestion. Can you please mention the other specific spots?
Thank you.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -619,7 +617,7 @@ class Map(SettingsBase[DictStateType]):
 
 
 def _get_new_keywords(obj, kwds):
-    newkwds = {}
+    newkwds = dict()
```
dnwillia-work, line 620: So... I know I suggested this change but I think it can be reverted.   See the PyFluent dev channel for a link I shared.

src/ansys/fluent/core/solver/flobject.py
dnwillia-work: How about at lines 699, 714, 724, 855 and 924 (924 is the one that I questioned the most since the try block specifically throws a RuntimeError).  In all those cases the implementation will capture every possible exception but what is best is if you specifically catch the exceptions that can actually occur at those locations.  It's ok with me if you want to leave this for now and come back to it later, just something that got pointed out by `pylint`.

hpohekar: Hi Dan, Thank you.

PR #679, opened 2022-07-27
Feat/project as case-file path

prmukherj:
Now project directory can be passed as input to the case reader as well.
Example:
**"......\Static_Mixer_Parameter\Static_Mixer_Parameters.cffdb\Static_Mixer_Parameters-Solve\Static_Mixer_Parameters.cas.h5"**

For the above case the inputs might be:
**"......\Static_Mixer_Parameter"** if '.flprj' file is present in the same directory, or it can also be **"......\Static_Mixer_Parameter\Static_Mixer_Parameters.cffdb"**


src/ansys/fluent/core/filereader/casereader.py
```diff
@@ -170,3 +175,35 @@ def _get_processed_string(input_string: bytes) -> str:
...
+        The directory containing the case file
+
+    Returns
+    -------
+    case file path (str)
+    """
+    file_list = list(
+        itertools.chain(
+            *(
+                glob.glob(project_dir_path + r"\**\**-Solve\*.%s" % ext)
```
mkundu1, line 197: I think the case-filepath information should be extracted from the .flprj file. @lagacep-ans Do we have a document on .flprj layout?

mkundu1 (approved):

hpohekar (approved):

src/ansys/fluent/core/filereader/casereader.py
```diff
@@ -78,6 +81,8 @@ class CaseReader:
 
     def __init__(self, case_filepath: str):
 
+        if Path(case_filepath).suffix in [".flprj", ".flprz"]:
```
seanpearsonuk, line 84: @prmukherj I would prefer it if a project file path were indicated by a separate argument rather than case_filepath having multiple meanings.

prmukherj: Actually, the approach being followed here is that if the provided path is project file path then it'll extract the case file path and carry on or else the provided path might be a case file path.
So, shall we ask the user to input either a case file-path or project-file path?

src/ansys/fluent/core/filereader/casereader.py
```diff
@@ -170,3 +175,35 @@ def _get_processed_string(input_string: bytes) -> str:
     rp_vars_str = codecs.decode(input_string, errors="ignore")
     string_identifier = "(37 ("
     return string_identifier + rp_vars_str.split(string_identifier)[1]
+
+
+def _get_case_filepath(project_dir_path: str) -> str:
+    """Gets path of any file having extension either of .cas, .cas.gz or.
+
+    .cas.h5 and returns the same.
```
seanpearsonuk, line 183:
```suggestion
    """Gets case file path within the provided project directory path.
```

(@prmukherj I think there's an extra unneeded empty line, and maybe we don't need to document it as it stands. )

src/ansys/fluent/core/filereader/casereader.py
```diff
@@ -170,3 +175,35 @@ def _get_processed_string(input_string: bytes) -> str:
...
+                glob.glob(project_dir_path + r"/**/**-Solve/*.%s" % ext)
+                for ext in ["cas", "cas.h5", "cas.gz"]
+            )
+        )
+    )
+    if len(file_list) < 1:
+        raise RuntimeError(f"No case files are present in: {dirname(file_list[0])}")
+    elif len(file_list) > 1:
+        raise RuntimeError(
+            f"More than one case file is present in: {dirname(file_list[0])}"
```
seanpearsonuk, line 204:
```suggestion
        raise RuntimeError(f"No case files are present in: {project_dir_path}")
    elif len(file_list) > 1:
        raise RuntimeError(
            f"More than one case file is present in: {project_dir_path}"
```

src/ansys/fluent/core/filereader/casereader.py
seanpearsonuk: @prmukherj I had followed the approach; I just though it preferable not to bend the meanings of arguments. If reusing the same general argument, then I would prefer to change it's name to something generic (just filepath), but I prefer it if we are explicit about the meanings of things (i.e. have two distinct arguments).

seanpearsonuk (approved):

seanpearsonuk:
https://github.com/pyansys/pyfluent/issues/674

PR #685, opened 2022-07-28
Remove 23.1 doc builds from PR CI/CD pipeline

dnwillia-work:
Elijah reported a couple doc build failures this afternoon:

https://github.com/pyansys/pyfluent/pull/680/checks
https://github.com/pyansys/pyfluent/pull/664/checks

The problem seemed to be that the 'ansys.fluent.core.settings' package was not properly installed into the Python environment at the build step 'Install again after codegen'.  At that point we've already installed the package during the previous step labelled 'Install PyFluent', but without the generated code, and pip does not seem to overwrite it with the updated wheel.   Setting up `make install` to force re-installation does the trick.

I thought what we had agreed about the 23.1 doc builds was that we should just be building it once a day in the nightly doc pipeline, so added it there and removed it from the PR pipeline.     The development documentation at `dev.fluentdoc.pyansys.com` is updated in that pipeline using  22.2.

The PR pipeline is still running the tests 2x with both 22.2 and 23.1, which doubles the testing time:

image

It might be prudent to move the 23.1 PR testing to an additional nightly build pipeline as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -194,7 +194,7 @@ jobs:
       - name: Install again after codegen
         run: |
           rm -rf dist
-          make install
+          make install > /dev/null
 
       - name: Cache examples
         uses: actions/cache@v3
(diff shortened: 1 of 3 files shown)


dnwillia-work:
@mkundu1 @seanpearsonuk  This afternoon Elijah pointed out that the doc builds were failing on 23.1 so I posted this PR to remove it from the PR doc build pipeline but the issue in the failed builds he shared is also happening with 22.2.  I have verified that.  The issue seems to be that the generated 'settings' API code is not found during the doc build process for creating the settings rst files.  I've verified that:

- The cached auto-generated settings module is properly restored into the 'src' directory by the "Cache API Code" step.
- The settings module is built and packaged properly into the wheel file during the step "Install again after codegen"
- That before the documentation is built that the ansys.fluent.core.solver package in the virtual environment _does not_ contain the settings module.

So, what seems to be happening is that the `pip install dist/*.whl` command executed by `make install` during the "Install again after codegen" build step is not actually installing the updated wheel file.

mkundu1 (approved):

egravenh (approved):

PR #689, opened 2022-08-02
Handling ImportError in flobject.py

hpohekar:


mkundu1 (approved):

prmukherj (approved):

dnwillia-work (approved):

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -921,7 +921,7 @@ def get_root(flproxy) -> Group:
             )
             raise RuntimeError("Mismatch in hash values")
         cls = settings.root
-    except RuntimeError:
+    except (ImportError, RuntimeError):
```
dnwillia-work, line 924: So should this be capturing RuntimeError?  It will swallow the one that is  raised three lines above when the hashes do not match.

dnwillia-work: @hpohekar I guess what I'm asking is if `RuntimeError` should even be there since it's explicitly `raise`d in the try block.

hpohekar: Hello @dnwillia-work 

Can we do like this?

```python
    obj_info = flproxy.get_static_info()
    cls = get_cls("", obj_info)
    
    try:
        from ansys.fluent.core.solver import settings
    
        if settings.SHASH != _gethash(obj_info):
            LOG.warning(
                "Mismatch between generated file and server object "
                "info. Dynamically created settings classes will "
                "be used."
            )
        else:
            cls = settings.root
    except ImportError:
        pass
        
    root = cls()
    root.set_flproxy(flproxy)
    root._setattr("_static_info", obj_info)
    return root
```

dnwillia-work: Close. I think the code should still raise the RuntimeError as before but should not catch it.

mkundu1: Note that the original design is if the files are not generated (the path ansys/fluent/core/solver/settings won't exist) or there is a mismatch between the generated files and server object-info (SHA mismatch), it should create the classes at runtime (`get_cls` call).

PR #696, opened 2022-08-08
Disable flserver datamodel in Fluent

mkundu1:
Reading cases with large no. of zones from PyFluent consumes too much time/memory as flserver datamodel is enabled in Fluent. Setting the env. var. `APP_LAUNCHED_FROM_CLIENT` (added for lbapp/gpuapp) in Fluent disables the flserver datamodel.


lagacep-ans:
> Reading cases with large no. of zones from PyFluent consumes too much time/memory as flserver datamodel is enabled in Fluent. Setting the env. var. `APP_LAUNCHED_FROM_CLIENT` (added for lbapp/gpuapp) in Fluent disables the flserver datamodel.

flserver will be required for icing/aero, but we can disable this line when fluent is launched with icing/aero as an option.
... i had some work related to that in https://github.com.mcas.ms/pyansys/pyfluent/pull/417 put this is months old and i'll need to revisit it soon.

gyeole (approved):

seanpearsonuk (approved):

PR #701, opened 2022-08-10
Feat/accessor methods on settings object

prmukherj:


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -244,37 +276,43 @@ class Real(SettingsBase[RealType]):
...
     """A ``String`` object representing a string value setting."""
 
     _state_type = str
 
 
-class Filename(SettingsBase[str]):
+class Filename(SettingsBase[str], Textual):
     """A ``Filename`` object representing a file name."""
 
     _state_type = str
```
seanpearsonuk, line 288: @prmukherj I was tempted to say that we could simply derive this from String and the body of the class would just be 
```
    pass
```
but you have preserved the prior format, and it's better not to have concrete inheritance anyway.

seanpearsonuk (approved):

hpohekar (approved):

seanpearsonuk:
https://github.com/pyansys/pyfluent/issues/618

PR #702, opened 2022-08-11
Provide users option to set fluent version

prmukherj:
Users / Developers will have option to either provide the fluent.exe file path or set the Ansys version manually.

Sample use case:
```
1. 
>>> pyfluent.set_ansys_version(version=pyfluent.FluentVersion.version_23R1)
>>> meshing = pyfluent.launch_fluent(precision="double", processor_count=2, mode="meshing")
2.
>>> pyfluent.set_fluent_path("---path---\fluent.exe")  # provide path in form of string or pathlib.Path
>>> solver = pyfluent.launch_fluent(precision="double", processor_count=2, mode="solver")
```
3.
In a certain case where both "set_ansys_version()" and "set_fluent_path()" are used together, then the path set to fluent.exe will have precedence.

### DIFF
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -25,8 +25,39 @@ import ansys.platform.instancemanagement as pypim
 
 _THIS_DIR = os.path.dirname(__file__)
 _OPTIONS_FILE = os.path.join(_THIS_DIR, "fluent_launcher_options.json")
-FLUENT_VERSION = "22.2"
-PIM_FLUENT_PRODUCT_VERSION = FLUENT_VERSION.replace(".", "")
+FLUENT_VERSION = ["22.2"]
+PIM_FLUENT_PRODUCT_VERSION = FLUENT_VERSION[0].replace(".", "")
+FLUENT_EXE_PATH = []
+
+
+class FluentVersion(Enum):
+    """Contains the standard ansys / fluent release."""
+
+    version_22R2 = "22.2"
+    version_23R1 = "23.1"
+
+
+def set_fluent_path(fluent_exe_path: Union[str, Path]) -> None:
+    """Lets the user set the fluent installation path manually.
+
+    This supersedes the fluent path set in the environment variable
+    """
+    FLUENT_EXE_PATH.append(str(fluent_exe_path))
+
+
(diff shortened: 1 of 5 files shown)


src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -52,7 +52,12 @@ def set_fluent_path(fluent_exe_path: Union[str, Path]) -> None:
 
     This supersedes the fluent path set in the environment variable
     """
-    FLUENT_EXE_PATH.append(str(fluent_exe_path))
+    if str(fluent_exe_path).endswith("fluent.exe"):
```
seanpearsonuk, line 55: @prmukherj or instead of creating a `str`, create a Path, access its `name` and do a more precise comparison.

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -25,8 +25,52 @@
...
+    """Lets the user set the fluent version manually.
+
+    Only works if the provided ansys version is installed and the
+    environment variables are updated properly. This supersedes the
+    fluent path set in the environment variable
+    """
+    if type(version) in [float, str]:
+        version = FluentVersion.get_version(str(version))
+    if version in FluentVersion or str(version) in FluentVersion.value:
+        FLUENT_VERSION[0] = version.value
```
seanpearsonuk, line 73: @prmukherj PIM_FLUENT_PRODUCT_VERSION depends on FLUENT_VERSION. So we should update it too.

src/ansys/fluent/core/launcher/launcher.py
prmukherj: Added a Path.exists() check which should take care of it along with the sting check. Should be fine? Or should we do something else?

src/ansys/fluent/core/launcher/launcher.py
prmukherj: ok, done. Thank you

src/ansys/fluent/core/launcher/launcher.py
prmukherj: Thank you for the suggestion. It's implemented and seems much better.

seanpearsonuk (approved):

hpohekar (approved):

PR #725, opened 2022-08-16
update dir, remove some redundant code

seanpearsonuk:
- update `__dir__` in a few places because some elements are missing where they are attributes of other attributes. The original PR uses `__getattr__` in those places which leaves things missing from `dir()` results. 
- remove some redundant code in meshing/workflow.py
- - add \_\_call\_\_ in MeshingWorkflow class

### DIFF
--- a/src/ansys/fluent/core/meshing/workflow.py
+++ b/src/ansys/fluent/core/meshing/workflow.py
@@ -17,20 +17,11 @@ def _new_command_for_task(task, meshing):
 
 class MeshingWorkflow:
     class Task(PyCallableStateObject):
-        class Args(PyCallableStateObject):
-            def __init__(self, task):
-                self._task = task
-                self._args = task.Arguments
-
-            def __getattr__(self, attr):
-                return getattr(self._args, attr)
-
         def __init__(self, meshing, name):
             self._workflow = meshing._workflow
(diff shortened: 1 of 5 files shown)


src/ansys/fluent/core/meshing/workflow.py
```diff
@@ -50,6 +41,11 @@ def _command(self):
         def __getattr__(self, attr):
             return getattr(self._task, attr)
 
+        def __dir__(self):
+            return sorted(
+                set(list(self.__dict__.keys()) + dir(type(self)) + dir(self._task))
```
seanpearsonuk, line 46: add this object's attributes, the class attributes (which brings in methods), and attributes inherited from the task

src/ansys/fluent/core/meshing/workflow.py
```diff
@@ -59,3 +55,8 @@ def task(self, name):
 
     def __getattr__(self, attr):
         return getattr(self._workflow, attr)
+
+    def __dir__(self):
+        return sorted(
+            set(list(self.__dict__.keys()) + dir(type(self)) + dir(self._workflow))
```
seanpearsonuk, line 61: similar logic to the above

PR #730, opened 2022-08-17
Dependabot for gh-actions wrongly configured

RobPasMue:
It looks like unless we define it at "/" it will not work. Solved it at PyDisco already and it is now working fine.


RobPasMue:
Somebody here to approve this PR? @seanpearsonuk @mkundu1 @ajain-work  Thanks in advance!

mkundu1 (approved):

ajain-work (approved):

RobPasMue:
For some reason CI is failing, but it seems to be a general thing in all pipelines

PR #734, opened 2022-08-18
Quantity (RealwithUnit) unit converter using pint module

hpohekar:
Source code - quantity.py
Test cases - test_quantity.py
Pint is added in install requirements in setup.py

RealWithUnits class is renamed to Quantity.

Description of unit names and symbol is available on following link -
https://github.com/hgrecco/pint/blob/master/pint/default_en.txt

Pint module documentation is available on following link -
https://pint.readthedocs.io/en/0.19/index.html


seanpearsonuk:
@mkundu1 tests are failing in both @hpohekar and @prmukherj 's PR.

Running API codegen
Unable to find image 'ghcr.io/pyansys/pyfluent:latest' locally
latest: Pulling from pyansys/pyfluent
Digest: sha256:fd0fd01135e7b19e471990f5dbaf173ed11a8453bb0af301d7f0286bff48d5ca
Status: Downloaded newer image for ghcr.io/pyansys/pyfluent:latest
89795f6c0debabd55a3a5ae8af8b59073c03874db39d19006242adf0636d84bb
503253dc9983c5f259cdc870566ecc8a54a3baf9b0cc9ecc655efea33031bac5
  File "/home/runner/work/pyfluent/pyfluent/codegen/allapigen.py", line 8, in <module>
    tuigen.generate()
  File "/home/runner/work/pyfluent/pyfluent/codegen/tuigen.py", line 345, in generate
    TUIGenerator(meshing=True).generate()
  File "/home/runner/work/pyfluent/pyfluent/codegen/tuigen.py", line 313, in generate
    info = PyMenu(self._service, self._main_menu.path).get_static_info()
  File "/home/runner/work/pyfluent/pyfluent/src/ansys/fluent/core/services/datamodel_tui.py", line 216, in get_static_info
    response = self._service.get_static_info(request)
  File "/home/runner/work/pyfluent/pyfluent/src/ansys/fluent/core/services/error_handler.py", line [15](https://github.com/pyansys/pyfluent/runs/7903509930?check_suite_focus=true#step:11:16), in func
    raise RuntimeError(ex.details()) from None
RuntimeError
make: *** [Makefile:[30](https://github.com/pyansys/pyfluent/runs/7903509930?check_suite_focus=true#step:11:31): api-codegen] Error 1
Error: Process completed with exit code 2.

Laurence has posted new docker images.

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,229 @@
...
+    ("rad/s", "rps"),
+    ("rad/s", "cps"),
+    ("radian/s", "Hz"),
+    ("radian/s", "hertz"),
+    ("radian/s", "rpm"),
+    ("radian/s", "rps"),
+    ("radian/s", "cps"),
+)
+
+restricted_unit_expansion = {
```
seanpearsonuk, line 48: Later version would build this list automatically using a standard list of prefixes.

tests/test_quantity.py
```diff
@@ -0,0 +1,146 @@
+from pytest import approx
+
+import ansys.fluent.core.quantity as q
+
```
seanpearsonuk, line 4: for some non-matching units conversions, you can check for exceptions
https://stackoverflow.com/questions/23337471/how-to-properly-assert-that-an-exception-gets-raised-in-pytest

You could test Hz etc

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,229 @@
...
+        restricted_units_dict[key] = list(set(restricted_units_dict[key]))
+    return restricted_units_dict
+
+
+restricted_units = build_restricted_conversions(
+    restricted_conversions, restricted_unit_expansion
+)
+
+
+class Quantity:
```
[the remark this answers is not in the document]
hpohekar: Yes, sure. Thank you for sharing.

src/ansys/fluent/core/quantity.py
hpohekar: Yes, sure. Now we have confirmed the usage of standard prefixes while writing test cases so we can generate list using them.

tests/test_quantity.py
hpohekar: Thank you for sharing. We will try to check all the possible restricted conversions while writing new test cases.

src/ansys/fluent/core/quantity.py
hpohekar: Done.

tests/test_quantity.py
hpohekar: Done.

tests/test_quantity.py
```diff
@@ -0,0 +1,398 @@
+import pytest
+
+import ansys.fluent.core.quantity as q
+
+
+def test_viscosity():
+    v = q.Quantity(1, "P")  # poise
+    conversion_output = v["kg m^-1 s^-1"]
```
dnwillia-work, line 8: you might add one to try `Pa s`.

tests/test_quantity.py
```diff
@@ -0,0 +1,398 @@
...
+    assert conversion_output._real_value == 0.1
+
+
+def test_volume():
+    v = q.Quantity(1, "gal")
+    conversion_output = v["m^3"]
```
dnwillia-work, line 14: Maybe it's just me but I'd prefer the API to be a bit more explicit `vm3 = v.convert("m^3")`

tests/test_quantity.py
```diff
@@ -0,0 +1,398 @@
...
+def test_youngs_modulus():
+    ym = q.Quantity(1, "lbf ft^-2")
+    conversion_output = ym["N m^-2"]
+    assert conversion_output._real_value == pytest.approx(47.89, 0.1)
+
+
+def test_temperature():
+    tempC = q.Quantity(1, "degC")  # celsius
+    conversion_output_1 = tempC["degK"]
+    assert conversion_output_1._real_value == 274.15
```
dnwillia-work, line 27: Why do you have to access a private variable to test this?  Would you not have an API `conversion_output_1.value()` or a public property `conversion_output_1.value`.  I don't think you should name it `real_value` since it might imply later that there is the possibility of a `complex_value`.

tests/test_quantity.py
hpohekar: Done.

tests/test_quantity.py
hpohekar: Done.

tests/test_quantity.py
hpohekar: _real_value changed to value.

tests/test_quantity.py
seanpearsonuk: > Maybe it's just me but I'd prefer the API to be a bit more explicit `vm3 = v.convertTo("m^3")`. For me the `[]` syntax implies an array or dictionary element access.

@dnwillia-work agreed, we had unanimously agreed against dict style access here. We had decided to use the same naming as in pint which is just `to`. 

One important aspect of the naming is that whatever we choose must clearly indicate that the current object is unaffected and that a new object is returned. 

Your suggestion would need to be `convert_to`. @hpohekar, I see that it's gone in as `convertTo`.

dnwillia-work: OK great.  If `pint` already provides a convention for what this method should be called I'd stick with that.

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,176 @@
+from itertools import permutations
+
+import pint
+from pint import Unit
+
+unit = pint.UnitRegistry(autoconvert_offset_to_baseunit=True, system="SI")
```
dnwillia-work, line 6: You might want to rename this to `unit_system` or `base_units` or since you are forcing it to be SI base units `base_units_SI`.

tests/test_quantity.py
hpohekar: Done. We have replaced ```converTo``` with ```to```.

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,184 @@
...
+
+    All the instances of this class are converted to base SI units
+    system. Any conversion between "Hz", "hertz", "rad/s", "radian/s",
+    "rpm", "rps", "cps" is restricted.
+    """
+
+    def __init__(self, real_value, units_string):
+        self.value = real_value
+        self.unit = units_string
+        self._quantity = quantity(self.value, self.unit)
```
dnwillia-work, line 92: Does `self._quantity` also cache the `real_value`?  If so, do you need the redundant copy?

src/ansys/fluent/core/quantity.py
hpohekar: We have replaced ```units``` with ```base_units```. We are trying to find a way to change base unit system between SI, US, cgs, mks as per our need by keeping same unit registry. So we have kept SI as base system for now.

src/ansys/fluent/core/quantity.py
hpohekar: Yes it does. We are using  ```self.value``` and ```self.unit``` in ```__float__```,  ```__str__``` and ```__repr__``` methods. 
also if we directly use ```real_value``` to instantiate ```self._quanity``` it modifies it's precision according to pint's default format and we can not represent our original Quantity instance as per the discussion in proposal. Hence we need the redundant copy as well.

dnwillia-work: Yeah, I guess I just wondered if you could replace the cached copy with
```
def value(self):
    return self._quantity.magnitude
```
but if the `quantity` is modifying the passed value then maybe it's fine.

src/ansys/fluent/core/quantity.py
dnwillia-work: Yeah, I see that documented here:  https://pint.readthedocs.io/en/stable/systems.html 

Very strange that they do not document this keyword argument.  I can't seem to find the pre-defined unit systems they support either.

tests/test_quantity.py
```diff
@@ -0,0 +1,590 @@
...
+
+
+def test_rpm():
+    x = q.Quantity(1, "rpm")
+    conversion_output = x.to("rad s^-1")
+    assert conversion_output.value == pytest.approx(0.1047198, 0.000001)
+
+
+def test_particles_per_gm():
+    x = q.Quantity(1, "1^15particles g^-1")
```
seanpearsonuk, line 479: @hpohekar what happens when you multiply two quantities of this type together?

hpohekar: We are getting following result after multiplying 2 quantities of this type together.

image

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,184 @@
+from itertools import permutations
+
+import pint
+from pint import Unit
+
+base_units = pint.UnitRegistry(autoconvert_offset_to_baseunit=True)
+base_units.default_system = "SI"
+
+# Unit definitions
+base_units.define("1.e15-particles/kg = 1 kg^-1 = 1^15particles kg^-1")
```
seanpearsonuk, line 10: @hpohekar I'm not convinced that we should cover this kind of "special" stuff from Fluent. Not only does it seem special, but considering what we can already see being returned with units from the settings API right now:
```
>>> s.setup.named_expressions()
{'parameter_1': {'definition': '1.5 [m/s]', 'description': '', 'input_parameter': False, 'output_parameter': False}}
```
that is not based on Fluent units so those special units will not be seen there, and are not supported there. The proposal is to only use standard units and to at least be self-consistent. Let's eliminate this from the implementation.

src/ansys/fluent/core/quantity.py
hpohekar: Even if we set ```us.default_system = None```, it ultimately converts to SI units after calling ```q.(1, "cm").to_base_units()```.

image

src/ansys/fluent/core/quantity.py
hpohekar: Yes, sure.

tests/test_quantity.py
seanpearsonuk: per gram per kilogram doesn't give me a good feeling. Hopefully this is a side effect of supporting the special units above, which we are now removing. Can you check the something like that doesn't appear when e.g. multiply two "g^-1" quantities?

hpohekar: No this is not appearing while dealing with standard units.

image

src/ansys/fluent/core/quantity.py
dnwillia-work: Yes, agree with Sean on this one.  Let's keep this module as generic as possible.

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+restricted_units = build_restricted_conversions(
+    restricted_conversions, restricted_unit_expansion
+)
+
+
+class Quantity:
+
+    """This class instantiates physical quantities using their real values and
```
dnwillia-work, line 81: Since this class is wrapping a `pint` `Quantity` I would appreciate that the documentation describes how it _extends_ or builds on the functionality of `pint`

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,184 @@
...
+    units.
+
+    All the instances of this class are converted to base SI units
+    system. Any conversion between "Hz", "hertz", "rad/s", "radian/s",
```
dnwillia-work, line 85: What does 'restricted' mean?

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,184 @@
+from itertools import permutations
+
+import pint
+from pint import Unit
+
+base_units = pint.UnitRegistry(autoconvert_offset_to_baseunit=True)
```
dnwillia-work, line 6: So, now that I've stared at `pint` a bunch more probably it's best to just name this `_ureg`, as they call it usually in the `pint` documentation, and make it a private variable.  I don't think there is a reason to make it public.

src/ansys/fluent/core/quantity.py
hpohekar: The pint is able to do the conversion between above mentioned quantities but with a wrong scale factor. As we had gone through the derivation / relationship of those quantities, and we found that it's based on formula rather than a appropriate scale factor. So even if pint allows to convert them from one another produces wrong results. Hence we have restricted these conversions.

hpohekar: Yes, sure.

seanpearsonuk: As well as working on the description, let's make sure that the way we are prohibiting conversions is generalizing in the required way. We are prohibiting conversions between frequency and angular velocity via control over conversions between units of those quantities. What does that mean for the treatment of radians, radians/s^2, ... ? 

As for why we don't allow the conversions, it's not that the scale factor is wrong per se; it's that we will not define a general conversion between these distinct sets of units. By raising exceptions, we rule out any ambiguities/inconsistencies. We are also consistent with the core C/C++ units library.

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,184 @@
...
+    def __repr__(self):
+        return f'(Quantity ({self.value}, "{self.unit}"))'
+
+    def to(self, to_unit):
+
+        """This method checks the compatibility between current instance unit
+        and user provided unit, if both of them are compatible, then only it
+        performs required conversion otherwise raises a Value Error."""
+
+        if (
```
seanpearsonuk, line 111: @hpohekar in fact, what we are going to need to do here is to get the dimensions of each units. Each dimensions can be represented as a numerical sequence.  I hope that pint will give us a method to get the dimensions from the units and that that dimensions object includes a field for angle. If so, we only need to compare those two dimensions objects. I'm sure that pint can provide that. That will be rigorous. The current approach cannot unfortunately extend to cover every possibility.

hpohekar: Yes pint supports method for dimensionality.

image

seanpearsonuk: @hpohekar can you please check also the dimensions for radians. I'm sure it's going to be correct. It means all that complicated, incomplete code can go away.

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,184 @@
...
+        restricted_units_dict[key] = list(set(restricted_units_dict[key]))
+    return restricted_units_dict
+
+
+restricted_units = build_restricted_conversions(
+    restricted_conversions, restricted_unit_expansion
+)
+
+
+class Quantity:
```
h-krishnan, line 79: Would it make sense to derive this class from `float` and store the SI value in `float`? That way, we could easily construct numpy arrays etc.

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,184 @@
...
+            temp = quantity(self._base_si_quantity / other, self.unit)
+        return Quantity(temp.magnitude, temp.units)
+
+    def __add__(self, other):
+        if self.unit in self._restricted_conversions.keys():
+            raise ValueError("This arithmetic operation is restricted")
+
+        if isinstance(other, Quantity):
+            temp = self._base_si_quantity + other._quantity
+        elif isinstance(other, int) or isinstance(other, float):
```
h-krishnan, line 153: Do we not need to check that the object is dimensionless before allowing the addition operation with float or int?

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,184 @@
...
+            temp = quantity(
+                self._base_si_quantity.magnitude - other, self._base_si_quantity.units
+            )
+        return Quantity(temp.magnitude, temp.units)
+
+    def __rsub__(self, other):
+        if self.unit in self._restricted_conversions.keys():
+            raise ValueError("This arithmetic operation is restricted")
+
+        if isinstance(other, Quantity):
```
h-krishnan, line 178: Code will never reach here

src/ansys/fluent/core/quantity.py
hpohekar: Replacing ```base_units``` with ```_ureg``` is possible. 

but making it private variable is not possible.

We have tried this - 

image

and by doing this every time new instance is created with new registry and we can not operate instances of 2 different registries together. 

we get this error - 

image

It's mentioned here - 
https://pint.readthedocs.io/en/stable/tutorial.html#tutorial

image

src/ansys/fluent/core/quantity.py
hpohekar: image

src/ansys/fluent/core/quantity.py
hpohekar: Yes, pint supports method to check whether the object is dimensionless or not and returns boolean as follows -

image

hpohekar: Yes, all tests are still passing by implementing following change in ```__add__, __radd__, __sub__, __rsub__``` methods.

image

image

src/ansys/fluent/core/quantity.py
hpohekar: Done.

src/ansys/fluent/core/quantity.py
hpohekar: Done.

src/ansys/fluent/core/quantity.py
hpohekar: Renamed ```base_units``` to ```ureg```.

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,197 @@
...
+from pint import Unit
+
+ureg = pint.UnitRegistry(autoconvert_offset_to_baseunit=True)
+ureg.default_system = "SI"
+
+
+quantity = ureg.Quantity
+
+
+restricted_units = ["Hz", "hertz", "rad/s", "radian/s", "rpm", "rps", "cps"]
```
seanpearsonuk, line 13: This as a flat list does not cover what we need since it prevents legal conversions also. We would need to structure this into list of pairs of lists.

However, we need to have a solution that scales properly (including as yet unimagined unit systems). The required approach is to compute the dimensions from the units and compare those. However @hpohekar informs me that pint does not include angle in its dimensions definitions. One option is to drop the requirement for this restriction altogether and just live with the pint behaviour.

src/ansys/fluent/core/quantity.py
hpohekar: Quantity is derived from float.

src/ansys/fluent/core/quantity.py
hpohekar: There is no reserved field for angle in dimensions and hence we can not compare 2 dimensions.

src/ansys/fluent/core/quantity.py
hpohekar: Documentation is now updated.

src/ansys/fluent/core/quantity.py
hpohekar: Yes, I agree with this, because ```rad/s, radian/s``` can also be given as ```rad s^-1, radian s^-1```. We can not cover every possible variation of same unit by going this approach.

hpohekar: I have also gone through the conversion table from ```units.scm``` and there is no such conversion made between Hz, rad/s.  

image

Also by default pint is correctly providing conversion between ```rpm``` and ```rad/s``` which we have restricted above.

image

image

Hence I think instead of restricting conversions manually it's better to stick with default pint behaviour.

dnwillia-work: Yes, that's fine by me.  @seanpearsonuk maybe submit an issue to the pint repo for discussion?  Interesting how even the internet does not get these details right.  :-).

dnwillia-work: @hpohekar Curious... What was the rationale to restrict converting rpm and radians/s?   `1 rev = 2*pi` radians right, they are dimensionally consistent.

tests/test_quantity.py
```diff
@@ -0,0 +1,560 @@
...
+def test_rad_radian():
+    f = q.Quantity(1, "rad/s")
+    with pytest.raises(Exception) as e_info:
+        conversion_output = f.to("radian/s")
+
+
+def test_rad_rpm():
+    f = q.Quantity(1, "rad/s")
+    with pytest.raises(Exception) as e_info:
+        conversion_output = f.to("rpm")
```
dnwillia-work, line 169: Regarding my question in the other file, I never noticed this test before.  Why is this raising an Exception?

src/ansys/fluent/core/quantity.py
hpohekar: Yes, that's right. They are dimensionally consistent. Actually we wanted to restrict conversion between Hz, hertz, rpm but by taking all permutations together it formed this pair. 

image


image

tests/test_quantity.py
hpohekar: It's raising exception because it is disallowed right now.

dnwillia-work: Right, ok, but I'm asking what the rationale is to disallow it.

hpohekar: Yes, it should be allowed but wrong pair of ```rad/s``` and ```rpm``` is formed due to considering all flat list permutations.

src/ansys/fluent/core/quantity.py
hpohekar: Removed restricted code and related test cases and updated the documentation accordingly.

tests/test_quantity.py
hpohekar: This test case is removed now.

tests/test_quantity.py
```diff
@@ -0,0 +1,308 @@
...
+def test_viscosity():
+    v = q.Quantity(1, "P")  # poise
+    conversion_output = v.to("kg m^-1 s^-1")
+    assert conversion_output.__float__() == 0.1
+
+
+def test_dynamic_viscosity():
+    vd = q.Quantity(1, "Pa s")
+    conversion_output = vd.to("P")
+    assert conversion_output.__float__() == 10.0
```
dnwillia-work, line 15: Why can't this be `conversion_output == 10.0`?

hpohekar: Because ```conversion_output``` is quantity instance with value and magnitude. 
Now we have derived the Quantity from the float, hence we have used ```conversion_output.__float__() == 10.0```  

image

dnwillia-work: OK, that's fine, but it seems a little strange to me to call the `__float()__` method directly like that.

hpohekar: Then should we use ```conversion_output.real == 10.0``` while testing?

dnwillia-work: I don't see why not. 

You might also consider implementing  `__eq__()` then this might work too:

`assert vd == q.Quantity(1, 'Pa s)`

hpohekar: Yes, sure. We will implement above method as well. Thank you for the suggestion.

hpohekar: image

hpohekar: Test also passed for above method - 

image

image

hpohekar: Replaced ```conversion_output.__float__()``` with ```conversion_output.real```. 
Added equal check method and it's test case.

hpohekar: > OK, that's fine, but it seems a little strange to me to call the `__float()__` method directly like that. I guess if we are accessing the value I kind of preferred the previous interface, i.e. `conversion_output.value`.

Done. 

image

Now we can access value of quantity instance using following ways.

image

dnwillia-work (approved):

seanpearsonuk (approved):

seanpearsonuk:
@hpohekar you can merge this now if you're ready

hpohekar:
> @hpohekar you can merge this now if you're ready

The c++ side unit conversion testing is not completed yet.

PR #735, opened 2022-08-18
Feat/enable command args for datamodel se

prmukherj:

PR #737, opened 2022-08-19
Feat/execute tui and journaling commands

prmukherj:
Sample Use Case:

session = pyfluent.launch_fluent(mode="meshing")

session.setup_python_console_in_tui()
session.start_journal("-----file_path-----")

session.stop_journal()

This is the basic version of how it should look like. Still needs to include port information.

### DIFF
--- a/src/ansys/fluent/core/session.py
+++ b/src/ansys/fluent/core/session.py
@@ -1,5 +1,6 @@
 """Module containing class encapsulating Fluent connection and the Base
 Session."""
+import json
 from typing import Any
 import warnings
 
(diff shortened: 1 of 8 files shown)


prmukherj:
This can be reviewed once the tests are fixed as part of the PR #735

prmukherj:
This PR can be reviewed once the tests are fixed as part of PR #735

mkundu1 (approved):

src/ansys/fluent/core/session.py
```diff
@@ -126,6 +127,30 @@ def exit(self) -> None:
         """Close the Fluent connection and exit Fluent."""
         self.fluent_connection.exit()
```
seanpearsonuk, line 129: @prmukherj scheme_eval returns a value. Ideally, start_journal should return a value if we can do that. Then that value can be passed to stop_journal, right?

cc: @mkundu1

hpohekar (approved):

seanpearsonuk (approved):

PR #750, opened 2022-08-23
Feat/add commands to meshing object

prmukherj:
```
>>> meshing_session = launch_fluent(mode='meshing')
>>> meshing = meshing_session.meshing
>>> workflow = meshing_session.workflow
>>> meshing.execute_tui(...)
>>> meshing.switch_to_solver()
```
This is to provide exact consistency with the Py Console in Fluent


mkundu1 (approved):

src/ansys/fluent/core/meshing/meshing.py
```diff
@@ -0,0 +1,33 @@
+import json
+
+
+class MeshingMeshing:
```
prmukherj, line 4: This name will be changed to Meshing

hpohekar (approved):

src/ansys/fluent/core/meshing/meshing.py
```diff
@@ -0,0 +1,33 @@
...
+        """Executes a tui command."""
+        self._fluent_connection.scheme_eval.scheme_eval(
+            f'(tui-menu-execute {json.dumps(command)} "")'
+        )
+
+    def switch_to_solver(self):
+        """Switch to solver session."""
+        from ansys.fluent.core.session_solver import Solver
+
+        self._tui.switch_to_solution_mode("yes")
```
seanpearsonuk, line 20: @prmukherj We are missing an opportunity to delegate here, and so we have duplication between here and session_meshing.py.  The duplicated methods are slightly different so we get slightly different behaviour, depending on how we call switch_to_solver. If we are deleting the session's switch_to_solver upon session.switch_to_solver being called then let's make sure that we we do that no matter how switch_to_solver is invoked, and also lets delete both the switch_to_solver methods.

src/ansys/fluent/core/meshing/meshing.py
```diff
@@ -0,0 +1,33 @@
+import json
+
+
+class Meshing:
+    def __init__(self, meshing, tui, fluent_connection):
+        self._meshing = meshing
+        self._tui = tui
+        self._fluent_connection = fluent_connection
+
+    def execute_tui(self, command: str) -> None:
```
seanpearsonuk, line 10: should also delegate execute_tui in order to avoid duplication

seanpearsonuk (approved):

src/ansys/fluent/core/meshing/meshing.py
prmukherj: Presently, in this approach, the "switch_to_solver()" does not delete the session attributes specific to the Meshing Session. Some time was spent to address this issue and come to a solution but no flawless solution was discovered. For now this duplication is as it is, and later maybe we can come up with a more efficient technique? @seanpearsonuk  and @mkundu1

PR #754, opened 2022-08-23
Add icing session type with datamodel

lagacep-ans:
New **SolverIcing** session type, similar to **Solver** but with the flserver datamodel enabled and linked to the **icing** attribute.  Fluent is launched with the `-flicing` command line option.

To use the icing datamodel, a datamodelgen is required with v231/fluent/ pointed by PYFLUENT_FLUENT_ROOT

Usage scenario is as:

```
session = pyfluent.launch_fluent(mode="solver-icing")
csim = session.icing
# Below code is the same as icing client python journaling
csim.Ice.IceConditions.IcingAirTemperature = 275.
csim.Solution.IceRun.Calculate()
```


codegen/datamodelgen.py
```diff
@@ -81,15 +81,17 @@ def _build_command_docstring(name: str, info: Any, indent: str):
 
 
 class DataModelStaticInfo:
-    def __init__(self, rules: str, mode: str):
+    def __init__(self, rules: str, mode: str, rules_save_name: str = ""):
```
lagacep-ans, line 84: Different datamodel for icing/aero or plain flserver datamodel. 
The rules_save_name permits to store out to different datamodels for the same remote structure.

codegen/datamodelgen.py
```diff
@@ -118,6 +121,9 @@ def _populate_static_info(self):
         run_solver_mode = any(
             info.mode == "solver" for _, info in self._static_info.items()
         )
+        run_icing_mode = "PYFLUENT_FLUENT_ROOT" in os.environ.keys() and any(
```
lagacep-ans, line 124: How can we know that we will be connecting to a v231 ?  Right now the hardcoded state is AWP_ROOT222 I think,
so we can assume that using PYFLUENT_FLUENT_ROOT is "probably" v231.

codegen/datamodelgen.py
```diff
@@ -134,6 +140,29 @@ def _populate_static_info(self):
...
+                        if (
+                            len(
+                                info.static_info.singletons["Case"]
+                                .singletons["App"]
+                                .singletons
+                            )
+                            == 0
+                        ):
+                            print(
+                                "Information: Icing settings not generated ( R23.1+ is required )\n"
```
lagacep-ans, line 158: If v222 is used, this printout will show  (only if PYFLUENT_FLUENT_ROOT was defined) , as the icing datamodel is not available.

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -141,7 +144,6 @@ def _get_subprocess_kwargs_for_fluent(env: Dict[str, Any]) -> Dict[str, Any]:
         kwargs.update(shell=True, start_new_session=True)
     fluent_env = os.environ.copy()
     fluent_env.update({k: str(v) for k, v in env.items()})
-    fluent_env["APP_LAUNCHED_FROM_CLIENT"] = "1"  # disables flserver datamodel
```
lagacep-ans, line 144: Moved out of this function to be applied if the solver mode warrants it

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -353,6 +355,8 @@ def launch_fluent(
             mode = LaunchModes.get_mode(mode)
         new_session = mode.value[1]
         meshing_mode = mode.value[2]
+        for k, v in mode.value[3]:
```
lagacep-ans, line 358: Apply the last item in the session definition tuple, to the launcher command line environment
SOLVER_ICING = ("solver-icing", SolverIcing, False, **[("fluent_icing", True)]**)

will end up binding to "-flicing" fluent command line

codegen/datamodelgen.py
seanpearsonuk: It's not hardcoded to 222 any more.  One can call set_fluent_version(23.1).

lagacep-ans: set_ansys_version() yes

but for codegen, either we use an environment variable or add a command line argument.
maybe

python codegen/allapigen.py --ansysVersion=23.1

seanpearsonuk (approved):

src/ansys/fluent/core/session_solver_icing.py
```diff
@@ -0,0 +1,36 @@
...
+    def __init__(
+        self,
+        fluent_connection: _FluentConnection,
+    ):
+        super(SolverIcing, self).__init__(fluent_connection=fluent_connection)
+        self._flserver_root = None
+
+    @property
+    def _flserver(self):
+        """root settings object."""
```
dnwillia-work, line 25: I guess this is not accurate

codegen/datamodelgen.py
dnwillia-work: You mean if you do nothing then when Fluent is launched 22.2 is launched by default, but you want to be able to run all this and have launch_fluent use 23.1 instead?

lagacep-ans: @seanpearsonuk @dnwillia-work added optional command line arguments to codegen, for convenience.

codegen/allapigen.py
```diff
@@ -1,9 +1,34 @@
...
 import tuigen
 
+from ansys.fluent.core.launcher.launcher import FLUENT_VERSION, set_ansys_version
+
 if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Generate python code from Fluent APIs"
+    )
+    parser.add_argument(
+        "--ansys-version",
```
lagacep-ans, line 16: new command line arguments.   to be able to run icing (assuming later on i point to the proper fluent connection) i would then need to 

`python codegen/allapigen.py --ansys-version=23.1`

src/ansys/fluent/core/session_solver_icing.py
lagacep-ans: fixed the comment, its the datamodel root.

FYI icing is using both the datamodel (for all its icing settings), and the settings api for regular fluent settings access , report definitions, etc.

dnwillia-work (approved):

mkundu1:
I gave the CI a rerun as I've renamed one of the CI job.

PR #759, opened 2022-08-24
Enhancement to avoid repetitions

akulshre-qa:
Added a new function download_input_file() to avoid repetition. 
We can directly download file by just passing file and folder name so we do not need to write separate fixture for each required file.


seanpearsonuk (approved):

tests/util/fixture_fluent.py
```diff
@@ -4,87 +4,119 @@
...
+    if ".msh" in full_file_name:
+        return "mesh"
+    if ".cas" in full_file_name:
+        return "case"
+    if ".dat" in full_file_name:
+        return "data"
+    if (
+        ".fmd" in full_file_name
+        or ".scdoc" in full_file_name
+        or ".pmdb" in full_file_name
```
mkundu1, line 17: from pathlib import Path
ext = Path(full_file_name).suffix
if ext in (".fmd", ".scdoc", ".pmdb"):
etc.

mkundu1 (approved):

prmukherj (approved):

PR #778, opened 2022-08-25
Add modules for job scheduler support.

dnwillia-work:
Introducing support for SLURM, UGE, LSF, PBS and WindowsHPC.  This PR just adds the modules required to parse the environment and drive the machine lists, with core counts, that can be passed to Fluent via `-cnf`.  It's not yet connected to `launch_fluent`.

A lot of the original code was all camel case.  I converted many (maybe all) the API functions, input parameters and output variables to snake, but there are still local variables using camel case.


tests/test_machines.py
```diff
@@ -0,0 +1,377 @@
+"""
```
dnwillia-work, line 1: It's using the `unittest` module because these tests were already implemented before.  If you feel strongly I will refactor it to use `pytest`,  but it is working as-is.

src/ansys/fluent/core/launcher/machinelist.py
```diff
@@ -0,0 +1,467 @@
...
+variables, respectively.
+"""
+import csv
+import os
+import subprocess
+
+from ansys.fluent.core.launcher.machines import Machine, MachineList
+
+
+def loadMachines(machineDict=None, hostInfo=None, ncores=None):
```
dnwillia-work, line 15: So... the return content of this would be used to construct a `-cnf` string to pass to Fluent.

tests/test_machines.py
dnwillia-work, line 1: It's using the `unittest` module because these tests were already implemented before.  If you feel strongly I will refactor it to use `pytest`,  but it is working as-is.

src/ansys/fluent/core/launcher/load_machines.py
```diff
@@ -0,0 +1,468 @@
...
+the PE_HOSTFILE, LSB_MCPU_HOSTS, PBS_NODEFILE and SLURM_JOB_NODELIST
+variables, respectively.
+"""
+import csv
+import os
+import subprocess
+
+from ansys.fluent.core.launcher.machine_list import Machine, MachineList
+
+
```
dnwillia-work, line 14: So... the return value of load_machines would be used to construct a `-cnf` string to pass to Fluent.

seanpearsonuk: @dnwillia-work should we put these two new files in a subdirectory such as scheduling or job_scheduling?

src/ansys/fluent/core/launcher/load_machines.py
```diff
@@ -0,0 +1,468 @@
...
+             {'machine-name' : <m-name-2>, 'core-count' : <int>},
+             ... ]
+    host_info : str
+        Optional host file name or list of machines and cores as a string
+        separated by commas and colons.
+        Example 1:  'M0:3,M1:2'
+        Example 2:  'M0,M0,M0,M1,M1'
+    ncores : int
+        Optional total core count.
+        If provided without machineDict, sets the core count for local parallel.
```
mkundu1, line 33: machine_info instead of machineDict?

mkundu1 (approved):

src/ansys/fluent/core/launcher/load_machines.py
dnwillia-work: Yeah sure, I can move them

seanpearsonuk: Either way, I will approve...

Many thanks for adding this, @dnwillia-work

seanpearsonuk (approved):

src/ansys/fluent/core/launcher/load_machines.py
dnwillia-work: It's done.

PR #780, opened 2022-08-26
Try running CI in github runners except the nightly CI

mkundu1:


mkundu1:
Doc build is hanging in github runner

PR #782, opened 2022-08-26
Session upload and download methods on session objects.

hpohekar:
This  feature is for internal use only.
Added _Uploader class to session.py.
Added upload and download methods to both Session and _BaseSession class.
Added test cases for both Session and _BaseSession class in test_uploader.py.
Tested code in Ansys Lab and working fine.

### DIFF
--- a/src/ansys/fluent/core/session.py
+++ b/src/ansys/fluent/core/session.py
@@ -433,4 +433,4 @@ class _Uploader:
         if self.file_service.file_exist(file_name):
             self.file_service.download_file(file_name, local_file_path)
         else:
-            raise FileNotFoundError("Uploaded remote file name is different.")
+            raise FileNotFoundError("Remote file does not exist.")
--- a/src/ansys/fluent/core/session.py
+++ b/src/ansys/fluent/core/session.py
@@ -65,6 +65,7 @@ class _BaseSession:
     def __init__(self, fluent_connection: _FluentConnection):
         self.fluent_connection = fluent_connection
         self.scheme_eval = self.fluent_connection.scheme_eval
+        self._uploader = None
 
     @classmethod
     def create_from_server_info_file(
@@ -170,6 +171,18 @@ class _BaseSession:
             )
         )
 
+    def _upload(self, file_path: str, remote_file_name: str = None):
+        """Uploads a file on the server."""
+        if not self._uploader:
+            self._uploader = _Uploader(self.fluent_connection._remote_instance)
+        return self._uploader.upload(file_path, remote_file_name)
+
+    def _download(self, file_name: str, local_file_path: str = None):
+        """Downloads a file from the server."""
+        if not self._uploader:
+            self._uploader = _Uploader(self.fluent_connection._remote_instance)
+        return self._uploader.download(file_name, local_file_path)
+
 
 class Session:
     """Instantiates a Fluent connection. This is a deprecated class. This has
@@ -238,6 +251,8 @@ class Session:
         self.solver = Session.Solver(
             self._datamodel_service_tui, self._settings_service
         )
+
+        self._uploader = None
 
     @classmethod
     def create_from_server_info_file(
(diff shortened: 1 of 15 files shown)


tests/test_uploader.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+
+from ansys.fluent.core import launch_fluent
+
+
+def test_base_session():
+    base_session = launch_fluent(cleanup_on_exit=False, mode="meshing")
+    with pytest.raises(Exception) as e_info:
+        base_session._upload(
+            "D:/PythonPractise/RealWithUnits/ConversionTable.py", "testNow.py"
```
mkundu1, line 10: Is the path `D:/..` from local machine? This won't accesible from linux docker container. Plase use a path relative to `pyfluent.EXAMPLES_PATH`

src/ansys/fluent/core/session.py
```diff
@@ -348,3 +375,63 @@ def root(self):
...
+        Download a file from the server.
+    """
+
+    def __init__(self, pim_instance):
+        self.pim_instance = pim_instance
+
+        try:
+            upload_server = self.pim_instance.services["http-simple-upload-server"]
+        except AttributeError:
+            print("\nPIM is not installed or not authorized.\n")
```
mkundu1, line 414: `LOG.error` would be preferred.

tests/test_uploader.py
hpohekar: Ok. Thank you.

src/ansys/fluent/core/session.py
hpohekar: Ok. Thank you.

hpohekar: Done.

tests/test_uploader.py
hpohekar: Done.

mkundu1 (approved):

tests/test_uploader.py
```diff
@@ -0,0 +1,24 @@
...
+        base_session._upload(
+            pyfluent.EXAMPLES_PATH + "/mixing_elbow.py", "test_upload_download.py"
+        )
+        base_session._download("test_upload_download.py", pyfluent.EXAMPLES_PATH)
+    base_session.exit()
+
+
+def test_session(new_mesh_session):
+    session = new_mesh_session
+    with pytest.raises(Exception) as e_info:
```
seanpearsonuk, line 19: if asserting an exception will be raised, assert it for each method that's expected to raise.

src/ansys/fluent/core/session.py
```diff
@@ -348,3 +376,61 @@ def root(self):
...
+        expanded_file_path = os.path.expandvars(file_path)
+        upload_file_name = remote_file_name or os.path.basename(expanded_file_path)
+        self.file_service.upload_file(expanded_file_path, upload_file_name)
+
+    def download(self, file_name: str, local_file_path: str = None):
+        """Downloads a file from the server."""
+        if self.file_service.file_exist(file_name):
+            self.file_service.download_file(file_name, local_file_path)
+        else:
+            raise FileNotFoundError("Uploaded remote file name is different.")
```
seanpearsonuk, line 436: Can we just say that the remote file does not exist?

src/ansys/fluent/core/session.py
```diff
@@ -319,6 +335,18 @@ def __dir__(self):
             )
         )
```
seanpearsonuk, line 337: are `_upload` and `_download` simply non-throwing no-op methods if pim is unconfigured? That is the requirement.

tests/test_uploader.py
hpohekar: Done.

src/ansys/fluent/core/session.py
hpohekar: Yes, we can. Done.

src/ansys/fluent/core/session.py
hpohekar: Yes.  
In following usage, we do not have single dependency on PIM or it's instance, it's doesn't matter whether PIM is installed or not. 

Here ```self._remote_instance``` is act as actual ```self._pim_instance```.

```
from ansys.fluent.core import launch_fluent

meshing_session = launch_fluent(cleanup_on_exit=False, meshing_mode=True)
pyfluent.session.Session._uploader=None
pyfluent.session.Session._upload=_upload
pyfluent.session.Session._download=_download
meshing_session._upload('upload/exhaust_system11.ipynb', remote_file_name='exh6.ipynb')
meshing_session._download('exh6.ipynb', local_file_path='download')
```

seanpearsonuk (approved):

PR #789, opened 2022-08-26
fix for meshingworkflow class

seanpearsonuk:
If not properly wrapping TaskObject type itself in the new workflow wrapper then we can get the new CommandArguments functionality via that route as opposed to calling the new task method.

### DIFF
--- a/src/ansys/fluent/core/meshing/workflow.py
+++ b/src/ansys/fluent/core/meshing/workflow.py
@@ -16,6 +16,28 @@ def _new_command_for_task(task, meshing):
 
 
 class MeshingWorkflow:
+    class TaskContainer(PyCallableStateObject):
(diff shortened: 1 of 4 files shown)


src/ansys/fluent/core/meshing/workflow.py
```diff
@@ -16,6 +16,28 @@ def __init__(self, task_name):
 
 
 class MeshingWorkflow:
+    class TaskContainer(PyCallableStateObject):
+        def __init__(self, meshing):
+            self._meshing_container = meshing
+            self._workflow = meshing._workflow
```
seanpearsonuk, line 22: rm unused attrs here

prmukherj (approved):

mkundu1 (approved):

PR #794, opened 2022-08-28
Get current fluent mode (meshing / solver)

prmukherj:
Gets the current running mode of fluent session. ("meshing" or "solver")

### DIFF
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -433,7 +433,7 @@ def launch_fluent(
                 pyfluent.EXAMPLES_PATH, pyfluent.EXAMPLES_PATH, args
             )
             return new_session(
-                _FluentConnection(
+                fluent_connection=_FluentConnection(
                     port=port,
                     cleanup_on_exit=cleanup_on_exit,
                     start_transcript=start_transcript,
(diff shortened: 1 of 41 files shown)


mkundu1 (approved):

seanpearsonuk (approved):

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -267,6 +267,7 @@ def launch_fluent(
     case_filepath: str = None,
     meshing_mode: bool = None,
     mode: Union[LaunchModes, str, None] = None,
+    server_info_filepath: str = None,
```
mkundu1, line 270: The password can also be passed as `launch_fluent` argument. It will try to deduce the ip/port/password in the following order:

1. Accept the ip/port/password if they have been passed to `launch_fluent` explictly.
2. If some information are missing after 1, extract them from server-info file if it has been passed to `launch_fluent`.
3. If ip/port are still missing, try to get them from env. var.
4. If something is still missing, raise exception.

I'm fine to skip the warning `"The server-info file was not parsed because ip and port were provided."` if that makes the code too complex.

prmukherj: The current implementation looks better?

seanpearsonuk (approved):

mkundu1 (approved):

PR #797, opened 2022-08-29
Fix doc deployment and an issue related to uninstallation

mkundu1:


.github/workflows/ci.yml
```diff
@@ -221,7 +221,7 @@ jobs:
           retention-days: 7
 
       - name: Deploy
-        if: github.event_name == 'push' && contains(github.ref, 'refs/tags') && matrix.image-tag == 'v22.2.0'
+        if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v') && !contains(github.ref, 'dev') && matrix.image-tag == 'v22.2.0'
```
mkundu1, line 224: Deploy only when a release tag is pushed

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -971,7 +971,7 @@ def get_root(flproxy, version: str = "") -> Group:
             )
             raise RuntimeError("Mismatch in hash values")
         cls = settings.root
-    except (ImportError, RuntimeError):
+    except Exception:
```
mkundu1, line 974: The empty settings directory may remain in site-packages after uninstallation, then it throws AttributeError from `settings.SHASH` from within the try block. Catching the general Excepton here as we need to catch too many different types.

prmukherj (approved):

hpohekar (approved):

src/ansys/fluent/core/solver/flobject.py
dnwillia-work: Yes, but then you break the `raise RuntimeError(...)` three lines above no?

dnwillia-work (approved):

src/ansys/fluent/core/solver/flobject.py
mkundu1: I think the function should still proceed to `get_cls` when the hash-value is mismatched as then the settings classes can be built in-memory from the metadata coming from the running server session. I'll log a warning in this case so that the message is still printed.

PR #801, opened 2022-08-30
resets the global version variable after test

prmukherj:
Fix the "test_manual_fluent_version_setting()" to reset the version of fluent after the test.

### DIFF
--- a/tests/test_launcher.py
+++ b/tests/test_launcher.py
@@ -19,6 +19,10 @@ def test_manual_fluent_version_setting():
     with pytest.raises(RuntimeError):
         pyfluent.set_ansys_version(22.1)
 
+    # Resets the global variable to its original state
+    pyfluent.set_ansys_version(version=pyfluent.FluentVersion.version_22R2)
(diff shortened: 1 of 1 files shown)


tests/test_launcher.py
```diff
@@ -19,6 +19,10 @@ def test_manual_fluent_version_setting():
     with pytest.raises(RuntimeError):
         pyfluent.set_ansys_version(22.1)
 
+    # Resets the global variable to its original state
+    pyfluent.set_ansys_version(version=pyfluent.FluentVersion.version_22R2)
```
mkundu1, line 23: What if we are originally running Fluent 23.1?

prmukherj: ohk. I'll update it accordingly.
Thank you

mkundu1 (approved):

hpohekar (approved):

seanpearsonuk (approved):

lagacep-ans (approved):
Thanks for the quick fix

PR #804, opened 2022-08-30
Fix root access with old-style session

mkundu1:
The following code was failing:
```
session = pyfluent.launch_fluent()
session.solver.root
```

This seems to be broken after merging the segregration of generated code.


mkundu1:
Will add a unittest with the code which was failing

hpohekar (approved):

seanpearsonuk (approved):

PR #809, opened 2022-09-01
Add markers for 22.2 or 23.1 tests

mkundu1:
This will remove the usage of `FLUET_IMAGE_TAG` env. var. from pytest so that users can run them easily.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -231,7 +231,7 @@ jobs:
           CLEAN: true
 
   build-test:
-    name: Unit Testing
+    name: Build and Unit Testing
     needs: test-import
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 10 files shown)


codegen/tuigen.py
```diff
@@ -80,7 +80,7 @@ def _get_tui_docdir(mode: str):
 
 def _copy_tui_help_xml_file(version: str):
     if os.getenv("PYFLUENT_LAUNCH_CONTAINER") == "1":
-        image_tag = os.getenv("FLUENT_IMAGE_TAG", "v23.1.0")
+        image_tag = os.getenv("FLUENT_IMAGE_TAG", "v22.2.0")
```
mkundu1, line 83: Correct default for users

prmukherj (approved):

.github/workflows/ci.yml
```diff
@@ -340,7 +340,7 @@ jobs:
           make install > /dev/null
 
       - name: 22.2 Unit Testing
-        run: make unittest
+        run: make unittest-dev-222
```
dnwillia-work, line 343: Is there a reason to have a specific versioned name in this case? It would be nice if the generic non-versioned name was still there and executes the latest supported release.

mkundu1: `make unittest` will run the default set for users which is dev unittests with 22.2 image (we will change that to 23.1 image after 23.1 Fluent release). As we are running both in CI, explicitly added the version for clarity.

hpohekar (approved):

akulshre-qa (approved):

PR #845, opened 2022-09-06
User friendly settings assignments

hpohekar:
Now we can assign settings in following ways.

```setup.models.energy.set_state(enabled=True, viscous_dissipation=False)```

```setup.models.energy(enabled=True, viscous_dissipation=False)```

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -236,9 +236,7 @@ class SettingsBase(Base, Generic[StateT]):
 
     def set_state(self, state: StateT = None, **kwargs):
         """Set the state of the object."""
-        if state:
-            return self.flproxy.set_var(self.path, self.to_scheme_keys(state))
-        elif kwargs:
+        if kwargs:
             return self.flproxy.set_var(self.path, self.to_scheme_keys(kwargs))
         else:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by hpohekar** [comment]
> Apart from following test case (line 409-410), all other are passing.
> 
> image
> 
> image
> 
> image

Following are the usage of new settings assignment.

image

As mentioned before, we have observed following behavior. 

settings assignment to ```True``` is working but assignment to ```False``` is not working.

image

**Review by prmukherj — APPROVED** [review]

File: src/ansys/fluent/core/solver/flobject.py [hunk 1]
```diff
@@ -234,9 +234,14 @@ def get_state(self) -> StateT:
         """Get the state of the object."""
         return self.to_python_keys(self.flproxy.get_var(self.path))
 
-    def set_state(self, state: StateT):
+    def set_state(self, state: StateT = None, **kwargs):
         """Set the state of the object."""
-        return self.flproxy.set_var(self.path, self.to_scheme_keys(state))
+        if state:
```
seanpearsonuk, line 239: Is it necessary to check `state` if we are going to end up doing the same in the else block? Isn't `kwargs` the only check to do?

seanpearsonuk (approved):

PR #856, opened 2022-09-08
Updated static info -> available accessor methods to low level objects

prmukherj:
This will work after server side changes are incorporated. Hence, this PR will be merged after the docker image is updated.


src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -867,14 +827,40 @@ def __init__(
...
         PyParameter.__init__(self, service, rules, path)
 
 
+class AccessorModes(Enum):
+    """Provides the standard Fluent launch modes."""
+
+    # Tuple:   Name, Solver object type, Meshing flag, Launcher options
+    TEXT = (["String", "ListString", "String List"], PyTextualCommandArgumentsSubItem)
+    NUMBER = (
+        ["Real", "Int", "ListReal", "Real List", "Integer"],
```
seanpearsonuk, line 844: @prmukherj Is it correct to include the numerical list types here. The reason for the Numerical class is for min and max which only apply to the single valued numerical types.

seanpearsonuk: @prmukherj are the tests failures there simply because we are waiting for the Fluent docker update?

prmukherj: Yes, the test failures are due to the pending update in the server side. Regarding your previous comment, we can keep numerical list as PyParamter only.

mkundu1 (approved):

seanpearsonuk (approved):

PR #869, opened 2022-09-11
Activate Job Scheduler in launch_fluent

dnwillia-work:
- `processor_count` is removed from the options .json since it's handled by the scheduler module.  Scheduler machine allocations are clamped to this value if it is provided to `launch_fluent` and is _less_ than the total allocated cores.
- `load_machines` always returns a `MachineList` object now.  Previously returned a vanilla list object in some cases which made testing a bit inconvenient.
- If `-tN` or `-cnf=<list>` are provided as `addtional_arguments` to `launch_fluent` then `load_machines` is skipped, allowing for library users to over-ride the automatic treatment.
- Fixed a bug in the scheduler unit tests.  CCP_NODES was not cleaned up in the environment which can lead to subsequent test failures.

### DIFF
--- a/src/ansys/fluent/core/launcher/fluent_launcher_options.json
+++ b/src/ansys/fluent/core/launcher/fluent_launcher_options.json
@@ -23,10 +23,6 @@
         },
         "fluent_format": "{}"
     },
-    "processor_count": {
-        "type": "int",
-        "fluent_format": " -t{}"
-    },
     "journal_filename": {
         "type": "str",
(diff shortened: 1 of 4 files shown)


ypatel-qa (approved):

seanpearsonuk (approved):

dnwillia-work:
Still debugging why the codegen step in the unit tests build will not run.  It seems to run fine locally so maybe I've broken something to do with connection to the container.

PR #877, opened 2022-09-13
setup for fluent

hpohekar:


src/ansys/fluent/core/__main__.py
```diff
@@ -0,0 +1,29 @@
...
+
+from ansys.fluent.core.launcher.launcher import launch_fluent
+from ansys.fluent.core.session_solver import Solver
+
+
+def setup_for_fluent(version: str, mode: str):
+    """Uses global PyConsole objects."""
+    session = launch_fluent(version=version, mode=mode)
+    if mode == "meshing":
+        globals()["meshing"] = session
```
seanpearsonuk, line 12: Is this quite right? On Fluent meshing console side, `meshing` is the meshing state engine which is equivalent to `session.meshing`.

src/ansys/fluent/core/__main__.py
```diff
@@ -0,0 +1,29 @@
...
+        globals()["workflow"] = session.workflow
```
seanpearsonuk, line 13: In addition we need `PartManagement` and `PMFileManagement`.  Also in either mode we need preferences.

hpohekar: Done. Thank you.

src/ansys/fluent/core/__main__.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/__main__.py
```diff
@@ -9,8 +9,11 @@ def setup_for_fluent(version: str, mode: str):
...
     if mode == "meshing":
-        globals()["meshing"] = session
+        globals()["meshing"] = session.meshing
         globals()["workflow"] = session.workflow
+        globals()["PartManagement"] = session.PartManagement
+        globals()["PMFileManagement"] = session.PMFileManagement
+        globals()["preferences"] = session.preferences
         globals()["solver"] = Solver(fluent_connection=session.fluent_connection)
     elif mode == "solver":
         globals()["solver"] = session
```
seanpearsonuk, line 19:
```suggestion
        globals()["solver"] = session
        globals()["solver"] = session.preferences
```

hpohekar: Done. Thank you.

src/ansys/fluent/core/__main__.py
```diff
@@ -0,0 +1,33 @@
+import os
+import sys
+
+from ansys.fluent.core.launcher.launcher import launch_fluent
+from ansys.fluent.core.session_solver import Solver
+
+
+def setup_for_fluent(version: str, mode: str):
```
mkundu1, line 8: If we want to move the logic within ansys.fluent.core, we can write a function `get_globals_for_fluent(version: str, mode: str)` which returns a dictionary and inside `setup_for_fluent` we can update the global dict: `globals().update(get_globals_for_fluent(version, mode)`.

dnwillia-work:
It's not entirely clear to me what this is for.

hpohekar:
> It's not entirely clear to me what this is for.

We want to use meshing, workflow, solver, PartManagement,, PMFileManagement and preferences as  global PyConsole objects while executing python journal. To get this to work seamlessly in PyFluent, we have implemented setup_for_fluent.

dnwillia-work:
> > It's not entirely clear to me what this is for.
> 
> We want to use meshing, workflow, solver, PartManagement,, PMFileManagement and preferences as global PyConsole objects while executing python journal. To get this to work seamlessly in PyFluent, we have implemented setup_for_fluent.

So, that you don't need to type `session.solver.<service>.x.y.z` and you can instead type `<service>.x.y.z`?

hpohekar:
> > > It's not entirely clear to me what this is for.
> > 
> > 
> > We want to use meshing, workflow, solver, PartManagement,, PMFileManagement and preferences as global PyConsole objects while executing python journal. To get this to work seamlessly in PyFluent, we have implemented setup_for_fluent.
> 
> So, that you don't need to type `session.solver.<service>.x.y.z` and you can instead type `<service>.x.y.z`?
> 
> If so, I'm confused. The thing is that these objects are specific to a session right.

You're right, just one change, we don't need to type ```session.solver.<service>.x.y.z```, we can type ```solver.<service>.x.y.z```, ```meshing.<service>.x.y.z```, ```workflow.<service>.x.y.z```, ```preferences.<service>.x.y.z``` etc in python journal.

seanpearsonuk:
@hpohekar based on recent discussions, we need to add `tui` to the globally exposed `meshing` object. We can discuss this.

dnwillia-work:
> > > > It's not entirely clear to me what this is for.
> > > 
> > > 
> > > We want to use meshing, workflow, solver, PartManagement,, PMFileManagement and preferences as global PyConsole objects while executing python journal. To get this to work seamlessly in PyFluent, we have implemented setup_for_fluent.
> > 
> > 
> > So, that you don't need to type `session.solver.<service>.x.y.z` and you can instead type `<service>.x.y.z`?
> > If so, I'm confused. The thing is that these objects are specific to a session right.
> 
> You're right, just one change, we don't need to type `session.solver.<service>.x.y.z`, we can type `solver.<service>.x.y.z`, `meshing.<service>.x.y.z`, `workflow.<service>.x.y.z`, `preferences.<service>.x.y.z` etc in python journal.

Which means this is useful in the scenario where you want to interact with a single Fluent session right?

hpohekar:
> > > > > It's not entirely clear to me what this is for.
> > > > 
> > > > 
> > > > We want to use meshing, workflow, solver, PartManagement,, PMFileManagement and preferences as global PyConsole objects while executing python journal. To get this to work seamlessly in PyFluent, we have implemented setup_for_fluent.
> > > 
> > > 
> > > So, that you don't need to type `session.solver.<service>.x.y.z` and you can instead type `<service>.x.y.z`?
> > > If so, I'm confused. The thing is that these objects are specific to a session right.
> > 
> > 
> > You're right, just one change, we don't need to type `session.solver.<service>.x.y.z`, we can type `solver.<service>.x.y.z`, `meshing.<service>.x.y.z`, `workflow.<service>.x.y.z`, `preferences.<service>.x.y.z` etc in python journal.
> 
> Which means this is useful in the scenario where you want to interact with a single Fluent session right?

Yes, right.

hpohekar:
> @hpohekar based on recent discussions, we need to add `tui` to the globally exposed `meshing` object. We can discuss this.
> 
> cc: @mkundu1

Yes, sure.

mkundu1 (approved):

src/ansys/fluent/core/__main__.py
mkundu1, line 8: @hpohekar Let's change the signature to `setup_for_fluent(*args, **kwargs)`, so we can update the Fluent code without affecting PyFluent.

hpohekar: Yes, sure.

hpohekar: Done.

prmukherj (approved):

seanpearsonuk (approved):

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -298,6 +299,8 @@ def launch_fluent(
 
     Parameters
     ----------
+    product_version: str, optional
+        Sets the product version. Options are ``"22.2"`` and ``"23.1"``.
```
mkundu1, line 303: @hpohekar From the code below, it seems to take strings `22.2.0` or `23.1.0` only. If we are adding a `product_version` argument, it should accept the version in the same fashion as in `set_ansys_version` for consistency.

Or, we can just call `set_ansys_version` before `launch_fluent` in `setup_for_fluent` without chaning `launch_fluent`'s signature.

hpohekar: Right. We will update this now.

hpohekar: Done.

hpohekar: Both ```__main__.py and launch_fluent``` updated now.

mkundu1 (approved):

PR #890, opened 2022-09-15
Feat/transfer mesh

seanpearsonuk:
utility for mesh transfer between Fluent instances. I have tested this in Ansys Lab and on the desktop.

### DIFF
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -420,7 +420,14 @@ def launch_fluent(
             if (show_gui is False) or (
                 show_gui is None and (os.getenv("PYFLUENT_SHOW_SERVER_GUI") != "1")
             ):
-                launch_string += " -hidden"
+                aargs = additional_arguments
+                if (
+                    "-g " not in aargs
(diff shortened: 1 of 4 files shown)


src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -420,7 +420,14 @@ def launch_fluent(
...
             ):
-                launch_string += " -hidden"
+                aargs = additional_arguments
+                if (
+                    "-g " not in aargs
+                    and "-gu " not in aargs
+                    and not aargs.endswith("-g")
+                    and not aargs.endswith("-gu")
+                ):
+                    launch_string += " -hidden"
```
seanpearsonuk, line 430: only allow hidden if another mode hasn't been chosen

src/ansys/fluent/core/session.py
```diff
@@ -450,13 +448,15 @@ def __init__(self, pim_instance):
```
seanpearsonuk, line 448: upload and download need to be no-ops, not throwing functions if not enabled.

prmukherj (approved):

hpohekar (approved):

PR #907, opened 2022-09-20
Fix / nightly docs build

prmukherj:


examples/00-fluent/mixing_elbow_settings_api.py
```diff
@@ -180,7 +180,8 @@
 # Solve for 150 iterations.
 
 solver.solution.run_calculation.iterate.get_attr("arguments")
-solver.solution.run_calculation.iterate(number_of_iterations=150)
```
seanpearsonuk, line 183: Hi @prmukherj. How does the fix work? If this line doesn't work, is that a bug and is an issue raised?  Thanks.

prmukherj: @seanpearsonuk , I am checking why it is failing. Locally I could reproduce the issue and the solution that I have posted works locally. If it fixes the nightly docs build, we can investigate further I think.

prmukherj: @seanpearsonuk and @mkundu1 , Thank you for clarifying this. It has been fixed. This looks good?

seanpearsonuk (approved):

examples/00-fluent/mixing_elbow_settings_api.py
```diff
@@ -180,7 +180,10 @@
...
+if float(solver.get_fluent_version().split(".")[0]) >= 23.0:
```
mkundu1, line 183: `session.get_fluent_version() >= "23.1.0"` would be sufficeint

mkundu1 (approved):

examples/00-fluent/mixing_elbow_settings_api.py
prmukherj: Thank you for the suggestion. Done.

PR #914, opened 2022-09-20
Testing/reg fix 01

akulshre-qa:
Fixing tests failing in nightly CI 

Nightly CI using local branch is running at:
https://github.com/pyansys/pyfluent/actions/runs/3091301375


ypatel-qa (approved):

tests/test_meshingmode/test_meshing_launch.py
```diff
@@ -57,7 +57,7 @@ def test_launch_pure_meshing(load_mixing_elbow_pure_meshing):
     with open(file_path) as fp:
         for count, line in enumerate(fp):
             pass
-    assert count == 17
+    assert count == 18
```
seanpearsonuk, line 60: @akulshre-qa 
As a general principle, I would prefer to avoid asserting such a thing, which is only weakly correlated with the correct functioning of the journal recording.  The assertion could succeed while journaling has failed and vice versa. In more concrete terms, we are currently changing the implementation, which will increase the number of lines in the file.

tests/test_pro/test_pro_post.py
```diff
@@ -34,10 +38,13 @@ def test_pro_exp(launch_fluent_solver_3ddp_t2):
...
     solver.execute_tui(r"""/solve/monitors/residual/plot? no """)
     solver.solution.initialization.hybrid_initialize()
-    solver.file.write(file_type="case-data", file_name="out/pro_poly_ini")
+    solver.file.write(
+        file_type="case-data", file_name=os.path.join(out, "pro_poly_ini")
+    )
     solver.mesh.check()
     solver.execute_tui(
-        r"""/file/write-profile "out/test1" symmetry bottom front left right top () pressure velocity-magnitude x-wall-shear pressure-coefficient x-velocity quit """
+        r"""/file/write-profile "%s" symmetry bottom front left right top () pressure velocity-magnitude x-wall-shear pressure-coefficient x-velocity quit """
```
dnwillia-work, line 46: Is there a reason you are using ‘execute_tui’ and not the TUI API for all these.  Generally in PyFluent I would have though the latter is the go to if the settings api does not have what you want.

tests/test_meshingmode/test_meshing_launch.py
akulshre-qa: Hi @seanpearsonuk thanks for the feedback. I just removed this assertion. We are anyway validating this recorded journal by replaying it back in the same session.

tests/test_solvermode/test_general.py
```diff
@@ -71,7 +72,10 @@ def test_solver_import_mixingelbow(load_mixing_elbow_mesh):
     with open(file_path) as fp:
         for count, line in enumerate(fp):
             pass
-    assert count == 12
+    if platform.system() == "Windows":
```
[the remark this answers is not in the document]
akulshre-qa: This is also the same. I was validating number of lines in recorded journal. They were differing by one line on Windows and Linux so I kept this condition. But as you suggested in you previous comment , I have removed this assertion.

tests/test_pro/test_pro_post.py
akulshre-qa: Hi @dnwillia-work the plan was to keep everything with settings API and in the place where they were not available I used execute TUI. We keep on revisiting them once settings API for those lines are available and update the tests.

dnwillia-work: OK, I just thought it's better to use the TUI command, `solver.tui`, interface since it's documented right.  `execute_tui` is undocumented.

seanpearsonuk: @dnwillia-work yes, that's the point I wanted to make as well.

akulshre-qa: @seanpearsonuk 
These journals were automatically generated using tui to python conversion tool. If we want to use only solver.tui then can development please help updating the conversion tool accordingly?

dnwillia-work: Right.  That explains why it’s like that then.

mkundu1 (approved):

prmukherj (approved):

tests/test_pro/test_pro_post.py
dnwillia-work: So, you will update this to use `solver.tui`?

akulshre-qa: Hi @dnwillia-work it can be done but it will require some time to update all the places to use solver.tui as we need to discover equivalent commands. 
Also as these tests are also running in FLUTE regression where old journals are being converted with each run so our plan was to use those latest converted journal (whenever available) to update remaining commands in github tests. 
Adding @ypatel-qa if we need to prioritize this work.

akulshre-qa: Hi @dnwillia-work and @ypatel-qa I'm merging this PR for now. If further improvement is needed we can create a new PR.

PR #920, opened 2022-09-22
Fix/error msg for unsuccessful fluent launch

prmukherj:
The following error was handled to give a proper-error msg:
error_to_be_fixed

### DIFF
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -446,7 +446,9 @@ def launch_fluent(
                     break
                 if start_timeout == 0:
                     LOG.error("The launch process has been timed out.")
-                    break
+                    raise RuntimeError(
+                        "The fluent connection could not be established."
+                    )
                 time.sleep(1)
                 start_timeout -= 1
                 LOG.info(
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -445,10 +445,7 @@ def launch_fluent(
                     LOG.info("Fluent process is successfully launched.")
                     break
                 if start_timeout == 0:
-                    LOG.error("The launch process has been timed out.")
-                    raise RuntimeError(
-                        "The fluent connection could not be established."
-                    )
+                    raise RuntimeError("The launch process has been timed out.")
                 time.sleep(1)
                 start_timeout -= 1
                 LOG.info(
--- a/src/ansys/fluent/core/session.py
+++ b/src/ansys/fluent/core/session.py
@@ -25,14 +25,11 @@ except Exception:
 def parse_server_info_file(filename: str):
     with open(filename, encoding="utf-8") as f:
         lines = f.readlines()
-    try:
-        ip_and_port = lines[0].strip().split(":")
-        ip = ip_and_port[0]
-        port = int(ip_and_port[1])
-        password = lines[1].strip()
-        return ip, port, password
-    except Exception:
-        raise RuntimeError("The fluent connection could not be established.")
+    ip_and_port = lines[0].strip().split(":")
+    ip = ip_and_port[0]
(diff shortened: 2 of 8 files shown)


tests/test_session.py
```diff
@@ -202,3 +203,11 @@ def test_get_fluent_mode(new_mesh_session):
     assert session.fluent_connection.get_current_fluent_mode() == "meshing"
     session = session.switch_to_solver()
     assert session.fluent_connection.get_current_fluent_mode() == "solver"
+
+
+def test_unsuccessful_fluent_connection():
+    # start-timeout is intentionally provided to be 2s for the connection to fail
+    logging.disable(logging.CRITICAL)
```
mkundu1, line 210: why this is needed?

src/ansys/fluent/core/session.py
```diff
@@ -25,11 +25,14 @@
...
-    port = int(ip_and_port[1])
-    password = lines[1].strip()
-    return ip, port, password
+    try:
+        ip_and_port = lines[0].strip().split(":")
+        ip = ip_and_port[0]
+        port = int(ip_and_port[1])
+        password = lines[1].strip()
+        return ip, port, password
+    except Exception:
```
seanpearsonuk, line 34: @prmukherj the correct catch-all type is `BaseException`.

src/ansys/fluent/core/session.py
```diff
@@ -25,11 +25,14 @@
 def parse_server_info_file(filename: str):
```
seanpearsonuk, line 25: @prmukherj when this function is called the connection has already timed out. Currently, we simply log an error, break out of the connection loop and call this function. Instead we should simply raise an exception at the point of the error (failed connection). We don't need to change this code.

seanpearsonuk (changes requested):
Please try the changes mentioned in my comment.

tests/test_session.py
prmukherj: @mkundu1, it was logging and error of time-out.

src/ansys/fluent/core/session.py
mkundu1: Note that, catching `BaseExecption` will shallow Ctrl+C. Catching `Exception` is generally fine for all coding logic level exception.

tests/test_session.py
prmukherj: this was given just so that line is not logged during testing.

src/ansys/fluent/core/session.py
prmukherj: @seanpearsonuk, Thank you for the suggestion. This makes more sense. Changes applied.

tests/test_session.py
mkundu1: Ok, probably a good idea to enable it after the test

prmukherj: @mkundu1, ohk, thank you Mainak. This has been applied.

seanpearsonuk: @prmukherj with the new fix in place, we no longer need this.

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -446,7 +446,9 @@ def launch_fluent(
                     break
                 if start_timeout == 0:
                     LOG.error("The launch process has been timed out.")
```
seanpearsonuk, line 448: @prmukherj do we need to log the error as well as raise?

tests/test_session.py
prmukherj: This test has been skipped altogether now. We can run this only locally.

src/ansys/fluent/core/launcher/launcher.py
prmukherj: @seanpearsonuk, the Runtime error msg has been updated with the error log msg and the log has been deleted.

seanpearsonuk (approved):

mkundu1 (approved):

PR #922, opened 2022-09-22
Fix/defensive check for attrs

prmukherj:


prmukherj:
@seanpearsonuk, this can be a fix from the client side as of now. This is working as well.

seanpearsonuk (approved):

mkundu1 (approved):

PR #926, opened 2022-09-23
Refactor launch_fluent()

prmukherj:


src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -272,6 +275,99 @@ def launch_remote_fluent(
     )
 
 
+def get_session_info(
```
mkundu1, line 278: Should all the new methods be private?

mkundu1 (approved):

src/ansys/fluent/core/launcher/launcher.py
prmukherj: @mkundu1, yes, right. Will do it.

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -407,30 +497,20 @@ def launch_fluent(
             )
         )
     if start_instance:
```
seanpearsonuk, line 499: @prmukherj maybe you're already doing it. lines 491 to 499 can collapse to `if _start_instance(start_instance):`

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -407,30 +497,20 @@ def launch_fluent(
...
-            ):
-                aargs = additional_arguments
-                if (
-                    "-g " not in aargs
-                    and "-gu " not in aargs
-                    and not aargs.endswith("-g")
-                    and not aargs.endswith("-gu")
-                ):
-                    launch_string += " -hidden"
+            launch_string = _update_launch_string_wrt_gui_options(
```
seanpearsonuk, line 511: @prmukherj can call one top-level function to update the launch string. That can delegate to `_update_launch_string_wrt_gui_options` but also execute other code before that (i.e. the above steps).

prmukherj: @seanpearsonuk, I am still thinking o this :)

src/ansys/fluent/core/launcher/launcher.py
prmukherj: @seanpearsonuk, this has been applied already. Thanks for pointing this out. Will push it shortly

hpohekar (approved):

seanpearsonuk (approved):

PR #942, opened 2022-09-28
Fix/User creatable in flobject

prmukherj:


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -865,7 +865,7 @@ def get_cls(name, info, parent=None):
                     dct["__doc__"] = f"'{pname.strip('_')}' child."
 
         include_child_named_objects = info.get("include_child_named_objects", False)
-        user_creatable = info.get("user_creatable", True)
+        user_creatable = info.get("user-creatable?", True)
```
mkundu1, line 868: @prmukherj I think the original code is fine according to the previous line. You wrote `user-creatable` in the chat, looking at which I thiught it should be `user-creatable?` as per scheme naming convention. The attribute name is `user_creatable` in python or `user-creatable?` in scheme.

seanpearsonuk: @prmukherj based on @mkundu1's comment, we can close this PR now. Thanks.

PR #956, opened 2022-09-30
Feat/write transcript to file

prmukherj:
We can write transcripts to file by passing it as an argument.
solver.start_transcript(r"FilePath")

Please note: This file path is an optional argument. So, if we call session.start_solver() that will also work and if we pass a file path as an argument to this method then it will write the transcript to the file path provided.

### DIFF
--- a/src/ansys/fluent/core/fluent_connection.py
+++ b/src/ansys/fluent/core/fluent_connection.py
@@ -271,7 +271,7 @@ class _FluentConnection:
                 os.remove(file_path)
         _FluentConnection._transcript_data_filepath = file_path
         self._transcript_thread = threading.Thread(
-            target=self._process_transcript,
+            target=_FluentConnection._process_transcript,
             args=(self._transcript_service,),
         )
 
--- a/src/ansys/fluent/core/fluent_connection.py
+++ b/src/ansys/fluent/core/fluent_connection.py
@@ -25,6 +25,8 @@ from ansys.fluent.core.services.transcript import TranscriptService
 from ansys.fluent.core.solver.events_manager import EventsManager
 from ansys.fluent.core.solver.monitors_manager import MonitorsManager
 
+lock = threading.Lock()
+
 
 def _get_max_c_int_limit() -> int:
     """Get the maximum limit of a C int.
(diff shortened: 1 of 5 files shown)


hpohekar (approved):

seanpearsonuk:
@prmukherj in the description, did you mean `session.start_transcript()` instead of `session.start_solver()` ?

src/ansys/fluent/core/session.py
```diff
@@ -122,13 +124,22 @@ def id(self) -> str:
         """Return the session id."""
         return self.fluent_connection.id
 
-    def start_transcript(self) -> None:
+    def start_transcript(self, file_path: str = None) -> None:
         """Start streaming of Fluent transcript."""
```
seanpearsonuk, line 128: @prmukherj please can you document the new argument, which will be useful in the documentation

src/ansys/fluent/core/session.py
```diff
@@ -122,13 +124,22 @@ def id(self) -> str:
...
-        self.fluent_connection.start_transcript()
+        if file_path:
+            if Path(file_path).exists():
+                os.remove(file_path)
+            self.execute_tui(f'(cx-start-transcript "{file_path}")')
```
seanpearsonuk, line 132: This is good for client and server on the same machine. As a user here, I would expect the file to be written to the specified path on my local machine, but that's not the case if the server is remote. So you really want to delegate to the fluent connection here. We can request it to write the transcript to the specified path. We could also provide the option to pass in a callback to receive the transcript strings.

seanpearsonuk (changes requested):
Please refer to comments

src/ansys/fluent/core/session.py
prmukherj: Sure, will do it

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -59,6 +62,68 @@ def run(self) -> None:
             cb()
 
 
+class AppendToFile:
```
seanpearsonuk, line 65: Just one comment I think: we could put all the transcript-related code in a new transcript.py.

prmukherj: Done, Thank you.

src/ansys/fluent/core/transcript.py
```diff
@@ -0,0 +1,68 @@
...
+        For eg.: add_transcript_callback(print) prints the transcript on
+        the interpreter screen.
+        """
+        with lock:
+            self._transcript_callbacks[self._transcript_callback_id] = (
+                callback_fn,
+                keep_new_lines,
+            )
+            returned_callback_id = self._transcript_callback_id
+            self._transcript_callback_id = self._transcript_callback_id + 1
```
ajain-work, line 33: You can also use `itertools.count()` and next .

prmukherj: @ajain-work, Thank you for the suggestion. This has been implemented.

src/ansys/fluent/core/transcript.py
```diff
@@ -0,0 +1,71 @@
...
+        the interpreter screen.
+        """
+        with lock:
+            self._transcript_callbacks[self._transcript_callback_id] = (
+                callback_fn,
+                keep_new_lines,
+            )
+            returned_callback_id = self._transcript_callback_id
+            self._transcript_callback_id = next(self._id_count)
+        if len(self._transcript_callbacks) == 1:
```
seanpearsonuk, line 36: The size of `self._transcript_callbacks` can change after the lock is released at the end of the with block, which would affect this logic.

src/ansys/fluent/core/transcript.py
```diff
@@ -0,0 +1,71 @@
...
+        self._transcript_callback_id = next(self._id_count)
+
+    def add_transcript_callback(self, callback_fn: Callable, keep_new_lines=False):
+        """Initiates a fluent transcript streaming depending on the
+        callback_fn.
+
+        For eg.: add_transcript_callback(print) prints the transcript on
+        the interpreter screen.
+        """
+        with lock:
```
seanpearsonuk, line 29: This acquires the lock for all instances of `Transcript` since the lock exists at module rather than object level. Is there a reason for that?

src/ansys/fluent/core/transcript.py
```diff
@@ -0,0 +1,71 @@
...
+            self._transcript_thread = threading.Thread(
+                target=self._process_transcript,
+                args=(self.transcript_service,),
+            )
+            self._transcript_thread.start()
+        return returned_callback_id
+
+    def remove_transcript_callback(self, callback_id):
+        """Stops each transcript streaming based on the callback_id."""
+        del self._transcript_callbacks[callback_id]
```
seanpearsonuk, line 46: this will also affect the execution of `_process_transcript`.

src/ansys/fluent/core/transcript.py
```diff
@@ -0,0 +1,71 @@
...
+        transcript = ""
+        while True:
+            try:
+                response = next(responses)
+                transcript += response.transcript
+                if transcript[-1] == "\n":
+                    for (
+                        callback_function,
+                        keep_new_lines,
+                    ) in self._transcript_callbacks.values():
```
seanpearsonuk, line 64: Here we don't have to acquire the lock so we are free to iterate over the values that can be affected meanwhile inside `add_transcript_callback`.

src/ansys/fluent/core/transcript.py
```diff
@@ -0,0 +1,71 @@
...
+        For eg.: add_transcript_callback(print) prints the transcript on
+        the interpreter screen.
+        """
+        with lock:
+            self._transcript_callbacks[self._transcript_callback_id] = (
+                callback_fn,
+                keep_new_lines,
+            )
+            returned_callback_id = self._transcript_callback_id
+            self._transcript_callback_id = next(self._id_count)
```
seanpearsonuk, line 35: as discussed we can eliminate self._transcript_callback_id

seanpearsonuk (approved):

PR #962, opened 2022-10-01
Fix 22.2 doc build

mkundu1:
Fluent Sync PR 348906

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -833,7 +833,7 @@ class _NonCreatableNamedObjectMixin(
         child.set_state(value)
 
 
-def get_cls(name, info, parent=None):
+def get_cls(name, info, parent=None, version=None):
     """Create a class for the object identified by "path"."""
     try:
         if name == "":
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -870,6 +870,8 @@ def get_cls(name, info, parent=None):
         user_creatable = info.get("user-creatable?", False) or info.get(
             "user_creatable", False
         )
+        if version == "222":
+            user_creatable = True
```
mkundu1, line 874: `user-creatable?` attribute doesn't exist in 222 settings API, so the `True` is the correct default for 22.2.

mkundu1: whereas `False` is the 23.1 default, as `user-creatable? = #t` is only marked where it is true.

prmukherj (approved):

hpohekar (approved):

src/ansys/fluent/core/solver/flobject.py
seanpearsonuk: We should update the Fluent version of the file even if the new code is redundant there.

seanpearsonuk (approved):

PR #995, opened 2022-10-14
Added check for -g and -gu for windows platform

hpohekar:
image

image

### DIFF
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -11,7 +11,7 @@ import platform
 import subprocess
 import tempfile
 import time
-from typing import Any, Dict, List, Union
+from typing import Any, Dict, Union
 import warnings
 
 from ansys.fluent.core.fluent_connection import _FluentConnection
@@ -299,9 +299,11 @@ def _is_windows():
     return platform.system() == "Windows"
 
 
-def _raise_exception_g_gu_in_windows_os(launch_string: Union[str, List[str]]) -> None:
+def _raise_exception_g_gu_in_windows_os(additional_arguments: str) -> None:
     """If -g or -gu is passed in Windows OS, the exception should be raised."""
-    if _is_windows() and (("-g" in launch_string) or ("-gu" in launch_string)):
+    if _is_windows() and (
+        ("-g" in additional_arguments) or ("-gu" in additional_arguments)
+    ):
         raise ValueError("'-g' and '-gu' is not supported on windows platform.")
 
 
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -302,13 +302,11 @@ def _update_launch_string_wrt_gui_options(
     if (show_gui is False) or (
         show_gui is None and (os.getenv("PYFLUENT_SHOW_SERVER_GUI") != "1")
     ):
-        if (
-            "-g " not in additional_arguments
-            and "-gu " not in additional_arguments
-            and not additional_arguments.endswith("-g")
-            and not additional_arguments.endswith("-gu")
-        ):
+        if not {"-g", "-gu"} & set(additional_arguments.split()):
```
prmukherj, line 305: @hpohekar, please look into the new location of the code and test it. Thank you.

prmukherj (approved):

mkundu1 (approved):

seanpearsonuk:
@hpohekar Unfortunately e.g., `s=pf.launch_fluent(mode="solver", show_gui=True, additional_arguments="-g")` is still allowed and it starts in `-g` mode. The check needs to be more comprehensive.

seanpearsonuk (changes requested):
As per previous comment. Perhaps we could also add a test for the expected behaviour.

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -302,13 +302,11 @@ def _update_launch_string_wrt_gui_options(
     if (show_gui is False) or (
```
prmukherj, line 302: @seanpearsonuk, this was wrongly modified as  we had an assumption that -g and -gu is only used id show_gui=False. Thank you for clarifying it.

hpohekar:
> @hpohekar Unfortunately, e.g., `s=pf.launch_fluent(mode="solver", show_gui=True, additional_arguments="-g")` is still allowed and it starts in `-g` mode. The check needs to be more comprehensive.

Done. Thank you.

hpohekar:
> As per previous comment. Perhaps we could also add a test for the expected behaviour.

Done. Thank you.

src/ansys/fluent/core/launcher/launcher.py
hpohekar: Right, Thank you.

hpohekar:
> @hpohekar, please look at the current implementation. @seanpearsonuk, the tests should run locally without any issues, however though we have a case that should launch fluent in '-g' mode in linux as well. Let's see if the tests behave as we expect them to :).
> 
> >

Yes, looks good. Thank you.

mkundu1 (approved):

seanpearsonuk (approved):

PR #1012, opened 2022-10-19
Removed pint dependency

hpohekar:


mkundu1 (approved):

seanpearsonuk (approved):

seanpearsonuk:
@hpohekar @mkundu1 should we rename quantity.py -> _quantity.py? Should we put a try - except ImportError around import pint in that module and show an appropriate message indicating need to install pint>0.18?

prmukherj (approved):

hpohekar:
> @hpohekar @mkundu1 should we rename quantity.py -> _quantity.py? Should we put a try - except ImportError around import pint in that module and show an appropriate message indicating need to install pint>0.18?

Yes, we can rename it to `_quantity.py` also if needed we can put try-except `ImportError` because anyhow if `pint` is not there ultimately it will give standard `ModuleNotFoundError` as error message.

PR #1024, opened 2022-10-21
Real Quantity with Units Module 

hpohekar:
Developed Real Quantity with Units Module.

Checked all test cases from C++ side units library and working fine.


src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,50 @@
...
+
+    Returns
+    -------
+    Quantity instance.
+
+    All the instances of this class are converted to base SI units system to have
+    consistency in all arithmetic operations.
+    """
+
+    def __new__(self, real_value, units_string):
```
seanpearsonuk, line 38: Is this usual / required?

hpohekar: Yes, because without this method we are not able to inherit float class properly.

seanpearsonuk: OK, I think you should rename `self` as `cls` though, as that is what it is, right?

seanpearsonuk (approved):

src/ansys/fluent/core/quantity.py
hpohekar: Done. Thank you.

jorgepiloto:
Pinging here @RobPasMue. He's feedback may be useful for this, as he implemented a units module on top of [pint](https://pint.readthedocs.io/en/stable/) for an internal project.

hpohekar:
> Pinging here @RobPasMue. He's feedback may be useful for this, as he implemented a units module on top of [pint](https://pint.readthedocs.io/en/stable/) for an internal project.

Thank you.

hpohekar:
> Developed Real Quantity with Units Module.
> 
> Checked all test cases from C++ side units library and working fine.
> 
> Only individual temperature conversion test case is remaining and will handle this after discussion.

We have considered following approach and handled it with separate class methods.

image

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,562 @@
...
+
+    All the instances of this class are converted to base SI units system to have
+    consistency in all arithmetic operations.
+    """
+
+    def __new__(cls, real_value, units_string):
+        return float.__new__(cls, real_value)
+
+    def __init__(self, real_value, unit_str):
+        float.__init__(real_value)
```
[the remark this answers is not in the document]
h-krishnan: I feel we need to convert the value to SI units here before initializing `float`.

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,562 @@
...
+        return temp_quantity
+
+    def _get_si_unit(self, other, func):
+        curr_dim = self.get_dimensions_list()
+        other_dim = other.get_dimensions_list()
+        temp_dim = [func(curr_dim[i], other_dim[i]) for i in range(len(other_dim))]
+        temp_unit = get_si_unit_from_dim(temp_dim)
+        return temp_unit
+
+    def pow(self, exponent):
```
h-krishnan, line 493: Should this be `__pow__`?

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,562 @@
...
+            isinstance(other, int) or isinstance(other, float)
+        ):
+            temp_value = self.si_value - other
+        else:
+            raise ValueError(f"Quantity{(self.value, self.unit)} is not dimensionless.")
+        return Quantity(temp_value, self.si_unit)
+
+    def __rsub__(self, other):
+        if isinstance(other, Quantity):
+            temp_value = other.si_value - self.si_value
```
h-krishnan, line 555: The above will never be true?

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,562 @@
...
+            temp_value = self.si_value - other.si_value
+        elif self.is_dimension_less() and (
+            isinstance(other, int) or isinstance(other, float)
+        ):
+            temp_value = self.si_value - other
+        else:
+            raise ValueError(f"Quantity{(self.value, self.unit)} is not dimensionless.")
+        return Quantity(temp_value, self.si_unit)
+
+    def __rsub__(self, other):
```
h-krishnan, line 553: Couldn't the implementation of this just be `return `Quantity(other, "") - self`?

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,562 @@
...
+        "F": 255.92777777777778,
+        "R": 0.5555555555555556,
+    }
+
+    _offset_dict = {
+        "K": {"C": 274.15, "F": 255.92777777777778, "R": 0.5555555555555556},
+        "C": {"K": -272.15, "F": -17.2222222222222, "R": -272.59444444444443},
+        "F": {"C": 33.79999999999993, "K": -457.87, "R": -458.67},
+        "R": {"C": 493.4699999999999, "F": 460.66999999999996, "K": 1.7999999999999998},
+    }
```
h-krishnan, line 153: Do we want to store the above in a separate file?

src/ansys/fluent/core/quantity.py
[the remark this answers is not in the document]
hpohekar: > `super().__init__(real_value)`?

No. `float.__init__(real_value)` works fine.

`super().__init__(real_value)` gives following error.

image

Thank you.

src/ansys/fluent/core/quantity.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/quantity.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/quantity.py
hpohekar: Not sure yet. Needs discussion about handling this from and to individual temperature conversion corner cases. Thank you.

src/ansys/fluent/core/quantity.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/quantity.py
hpohekar: Right. Thank you.

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,554 @@
...
+
+    _offset_conversions = {
+        "K": 1,
+        "C": 274.15,
+        "F": 255.92777777777778,
+        "R": 0.5555555555555556,
+    }
+
+    _offset_dict = {
```
dnwillia-work, line 148: Any reason you mark all the members with `_` prefix?  Usually this means private right but below you are using them directly.

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,554 @@
...
+        "M": 10**6,
+        "G": 10**9,
+        "T": 10**12,
+        "P": 10**15,
+        "E": 10**18,
+        "Z": 10**21,
+        "Y": 10**24,
+    }
+
+    _dimension_order = OrderedDict(
```
dnwillia-work, line 79: The second entry of each tuple seems to be a short name for each dimensional quantity.  I think you would want to avoid using actual units for that.  eg: `K`, `mol`, `cd`, `A` and `sr` should probably be something else no?

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,554 @@
...
+    @value.setter
+    def value(self, val):
+        self._value = val
+
+    @property
+    def dimension(self):
+        return self._dimension
+
+    @property
+    def si_value(self):
```
dnwillia-work, line 445: Maybe it's just me but I think that this property and the next one should not be part of this interface.  The unit system should be a separate concept from the quantity.

tests/test_quantity.py
```diff
@@ -0,0 +1,437 @@
...
+
+def test_value_unit_4():
+    v = q.Quantity(5.7, "")
+    assert v.value == pytest.approx(5.7, DELTA)
+    assert v.unit == ""
+
+
+def test_dims_5():
+    v = q.Quantity(1.0, "kPa")
+    assert v.get_dimensions_list() == [1, -1, -2, 0, 0, 0, 0, 0, 0, 0]
```
dnwillia-work, line 34: I think dimensions should be `float` not `int`

tests/test_quantity.py
```diff
@@ -0,0 +1,437 @@
...
+
+def testing_dimensions():
+    print(f"{'*' * 25} {testing_dimensions.__name__} {'*' * 25}")
+
+    def dim_test(unit_str, dim_list):
+        qt = q.Quantity(10, unit_str)
+        print(f"{unit_str} : {qt.get_dimensions_list()}")
+        assert qt.get_dimensions_list() == dim_list
+
+    dim_test("m", [0, 1, 0, 0, 0, 0, 0, 0, 0, 0])
```
dnwillia-work, line 277: You should add tests for things like `s^0.5`.

src/ansys/fluent/core/quantity.py
hpohekar: Yes, the reason is user should not be able to access all of these members and other point is I don't want to keep all of these as part of Quantity class hence written as separate class and accessed directly.

src/ansys/fluent/core/quantity.py
hpohekar: The `si_value` and `si_unit` is added after carefully considering  following points -

1. As per the proposal, we want to consider every scalar as SI value only.
2. We want to convert every quantity to SI before sending to fluent.
3. `pint` module also maintains SI value and SI unit for each quantity.
4. Also maintaining SI value and SI unit avoids code duplication in arithmetic operator's overloading. 
5. I think current one is the simplest way to handle all above mentioned points.

src/ansys/fluent/core/quantity.py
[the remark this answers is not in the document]
hpohekar: > I feel we need to convert the value to SI units here before initializing `float`. Consider `sin(90[deg])` vs. `sin(pi/2)`. Both should return same value.

Trigonometric functions are not any physical quantity and I think we are dealing with only valid physical quantities in our `Quantity` class.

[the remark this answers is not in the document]
hpohekar: Do we have any use case of this except for angle ?

tests/test_quantity.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/quantity.py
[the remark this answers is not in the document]
hpohekar: > I feel we need to convert the value to SI units here before initializing `float`. Consider `sin(90[deg])` vs. `sin(pi/2)`. Both should return same value.

Do we have any use case of this except for angle ?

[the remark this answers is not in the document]
hpohekar: > I feel we need to convert the value to SI units here before initializing `float`. Consider `sin(90[deg])` vs. `sin(pi/2)`. Both should return same value.

image

tests/test_quantity.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/quantity.py
dnwillia-work: Normally if they are part of the public interface of the class you would not prefix them with an underscore.  Unless I am misunderstanding the convention.  That certainly is the case for modules.

src/ansys/fluent/core/quantity.py
dnwillia-work: For chemical amount I think you want capital `N` no? 

https://en.wikipedia.org/wiki/Dimensional_analysis

src/ansys/fluent/core/quantity.py
dnwillia-work: I'm not certain I'm convinced.  I guess I just think that the unit system should be 100% detangled from the quantity type.  It's two separate concepts.  I think you can achieve point 4 still and not have the `si_unit` and `si_value` functions as public properties.

src/ansys/fluent/core/quantity.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/quantity.py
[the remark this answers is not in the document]
h-krishnan: @hpohekar What I am suggesting is that the actual "float" value is always SI. This way, you can directly pass the Quantity object to mathematical functions.

src/ansys/fluent/core/quantity.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/quantity.py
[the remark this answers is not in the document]
hpohekar: @h-krishnan Done. Thank you.

image

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,553 @@
...
+        "l": (0.001, "m^3"),
+        "gal": (0.0037854117839999993, "m^3"),
+        "BTU": (1055.056, "J"),
+    }
+
+    multipliers = {
+        "d": 10**-1,
+        "c": 10**-2,
+        "m": 10**-3,
+        "μ": 10**-6,
```
h-krishnan, line 60: Do we support this character?

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,553 @@
...
+                )
+                unit_str = remove_multiplier(unit_str) if has_multiplier else unit_str
+
+            term_power *= power
+            unit_dim = self._get_dim(unit_str, term_power)
+
+            if unit_dim != None:
+                self._add_or_update_dimension(unit_dim)
+
+    def get_dimensions_dict(self):
```
h-krishnan, line 349: Maybe just `as_dict()`?

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,553 @@
...
+    def __repr__(self):
+        return f'(Quantity ({self.value}, "{self.unit}"))'
+
+    def __mul__(self, other):
+        if isinstance(other, Quantity):
+            temp_value = self.si_value * other.si_value
+            temp_unit = self._get_si_unit(other, lambda x, y: x + y)
+            return Quantity(temp_value, temp_unit)
+        elif isinstance(other, int) or isinstance(other, float):
+            temp = Quantity(self.si_value * other, self.si_unit)
```
h-krishnan, line 512: Shouldn't we retain the units in case of multiplication with scalar?

h-krishnan: I feel + and - should retain the unit of the left operand.
Multiplication with scalar should retain the unit.
For Quantity(1, 'g')*Quantity(1, 'm'), my expectation is Quantity(1, 'g m')

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,553 @@
...
+    def __truediv__(self, other):
+        if isinstance(other, Quantity):
+            temp_value = self.si_value / other.si_value
+            temp_unit = self._get_si_unit(other, lambda x, y: x - y)
+            return Quantity(temp_value, temp_unit)
+        elif isinstance(other, int) or isinstance(other, float):
+            temp = Quantity(self.si_value / other, self.si_unit)
+            return temp
+
+    def __add__(self, other):
```
h-krishnan, line 527: Why is this working?
``` python
>>> a
(Quantity (0.5, "cm s^-1"))
>>> v
(Quantity (0.005, "m"))
>>> v + a
(Quantity (0.01, "m"))
>>>
```

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,553 @@
...
+        return Quantity(new_si_value, new_si_unit)
+
+    def __float__(self):
+        return float(self.si_value)
+
+    def __str__(self):
+        return f'({self.value}, "{self.unit}")'
+
+    def __repr__(self):
+        return f'(Quantity ({self.value}, "{self.unit}"))'
```
h-krishnan, line 504: Do we need the additional ( ) wrapping Quantity?

hpohekar: No. Done. Thank you.

src/ansys/fluent/core/quantity.py
hpohekar: It's ASCII character right ? 

`ALT + 230`

image

src/ansys/fluent/core/quantity.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/quantity.py
hpohekar: `si_value` and `si_unit` functions are now private properties.

src/ansys/fluent/core/quantity.py
hpohekar: Now we are avoiding `+ , -` between incompatible dimensions. Thank you.

tests/test_quantity.py
```diff
@@ -0,0 +1,484 @@
...
+    assert convert.value == pytest.approx(11356.5713242, DELTA)
+    assert convert.unit == "J m^-2"
+
+
+def test_prop_constants_40():
+    c = q.Quantity(10.5, "s^-0.5")
+    x = q.Quantity(5.7, "m")
+    t = q.Quantity(4.8, "s")
+
+    y = c * x * t
```
mkundu1, line 287: I'm getting `'m'` as `y.unit` here

src/ansys/fluent/core/quantity.py
dnwillia-work: Probably it's better to use unicode: 

https://en.wikipedia.org/wiki/List_of_Unicode_characters#Greek_and_Coptic

Mu seems to be 956.

I think the ASCII characters beyond 128 depend on the loaded character set which is locale and maybe platform dependent.

tests/test_quantity.py
hpohekar: Right, we should get `y = (287.28, "m^1.0 s^0.5")`. It's fixed now. Thank you.

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,550 @@
...
+        unit_str = dim_to_unit_map[key]
+        spacechar = " " if len(si_unit) > 0 else ""
+        if power > 0.0 or power < 0.0:
+            si_unit += spacechar + unit_str + "^" + str(power)
+        elif power == 1.0:
+            si_unit += spacechar + unit_str
+    return si_unit
+
+
+class Quantity(float):
```
mkundu1, line 389: Should it implement `__eq__` and maybe other comparison operators?

tests/test_quantity.py
hpohekar: Yes, right ideally it should `y.unit = "m s^0.5"`, but since we have converted dimensions from `int` to `float` this issue arise.

but we will handle this too. Thank you.

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,549 @@
...
+        "Z": 10**21,
+        "Y": 10**24,
+    }
+
+    dimension_order = OrderedDict(
+        [
+            ("Mass", "M"),
+            ("Length", "L"),
+            ("Time", "T"),
+            ("Temperature", "Temp"),
```
dnwillia-work, line 84: I guess in principle you could use the unicode symbol for temperature as the short form?  Θ

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,601 @@
...
+        if power > 0.0 or power < 0.0:
+            si_unit += spacechar + unit_str + "^" + str(power)
+        elif power == 1.0:
+            si_unit += spacechar + unit_str
+    return si_unit
+
+
+class UnitSystem:
+
+    _dim_to_unit_sys_map = {
```
dnwillia-work, line 394: This is cool but could you restructure it so the keys are the other way around:
```
"SI" : {"M" : "kg", "L" : "m" ... }
"CGS" : { "M" : "g" : "L" : "cm" ... }
```
It's a bit easier to see what units make up a `UnitSystem` then

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,601 @@
...
+
+        if self._unit_system not in UnitSystem._supported_unit_sys:
+            raise ValueError(
+                "Unsupported unit system, only 'SI', 'CGS', 'BTU' is allowed."
+            )
+
+    def _get_unit_from_dim(self, dim_list):
+        unit = ""
+
+        for key, power in zip(_UnitsTable.dimension_order.values(), dim_list):
```
dnwillia-work, line 420: if you make the change above then I think this can be:
```
base_units = UnitSystem._dim_to_unit_sys_map[self._unit_system]
for key, power in zip(_UnitsTable.dimension_order.values(), dim_list):
    unit_str = base_units[key]
```

```

File: src/ansys/fluent/core/quantity.py [hunk 22]
```diff
@@ -0,0 +1,601 @@
...
+        elif unit_str in _UnitsTable.derived_units_with_conversion_factor:
+            _, unit_str = _UnitsTable.derived_units_with_conversion_factor[unit_str]
+            self._parser(unit_str, power)
+        else:
+            raise ValueError("Not implemented")
+
+
+def get_si_unit_from_dim(dim_list):
+    si_unit = ""
+    dim_to_unit_map = {
```
dnwillia-work, line 370: If you restructure `UnitSystem` as I suggested, then you can get this information from there I think.

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,601 @@
...
+        "N": "mol",
+        "J": "cd",
+        "Angle": "radian",
+        "SAngle": "sr",
+        "": "",
+    }
+    for key, power in zip(_UnitsTable.dimension_order.values(), dim_list):
+        unit_str = dim_to_unit_map[key]
+        spacechar = " " if len(si_unit) > 0 else ""
+        if power > 0.0 or power < 0.0:
```
dnwillia-work, line 385: I think you want to reverse this logic because 1.0 > 0.0 right so you miss the special case.
```
        if power == 1.0:
            si_unit += spacechar + unit_str
        elif power != 0.0:
            si_unit += spacechar + unit_str + "^" + str(power)
```
and you can test once for not being zero instead of 2x.

src/ansys/fluent/core/quantity.py
h-krishnan: I am not able to load this file in Python.
```
e:\temp>c:\Python310\python.exe quantity.py
SyntaxError: Non-UTF-8 code starting with '\xb5' in file e:\temp\quantity.py on line 62, but no encoding declared; see https://python.org/dev/peps/pep-0263/ for details
```
I think the following line needs to be added at the top of the file:
# -*- coding: latin-1 -*-

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,601 @@
...
+        elif self.is_dimension_less() and (
+            isinstance(other, int) or isinstance(other, float)
+        ):
+            temp_value = self._si_value - other
+        else:
+            raise ValueError("Incompatible dimensions.")
+        return Quantity(temp_value, self._si_unit)
+
+    def __rsub__(self, other):
+        return self.__sub__(other)
```
h-krishnan, line 598: This is not correct. You need to do `return Quantity(other) - self`

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,601 @@
...
+            temp_value = self._si_value * other._si_value
+            temp_unit = self._get_si_unit(other, lambda x, y: x + y)
+            return Quantity(temp_value, temp_unit)
+        elif isinstance(other, int) or isinstance(other, float):
+            return Quantity(self._si_value * other, self._si_unit)
+
+    def __rmul__(self, other):
+        return self.__mul__(other)
+
+    def __truediv__(self, other):
```
h-krishnan, line 559: Don't we also require `__rtruediv__`?

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,601 @@
...
+            temp_value = self._si_value - other
+        else:
+            raise ValueError("Incompatible dimensions.")
+        return Quantity(temp_value, self._si_unit)
+
+    def __rsub__(self, other):
+        return self.__sub__(other)
+
+    def __eq__(self, other):
+        return self._si_value == other._si_value and self._si_unit == other._si_unit
```
h-krishnan, line 601: We will also need to add other operators such as `__ge__`, `__gt__` and their `__r*` equivalents.

h-krishnan, line 601: If the quantity is dimensionless, we may want to support comparison with float as well.

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,601 @@
...
+    def _get_si_unit(self, other, func):
+        curr_dim = self.get_dimensions_list()
+        other_dim = other.get_dimensions_list()
+        temp_dim = [func(curr_dim[i], other_dim[i]) for i in range(len(other_dim))]
+        temp_unit = get_si_unit_from_dim(temp_dim)
+        return temp_unit
+
+    def __pow__(self, exponent):
+        new_dims = list(
+            map(lambda x: x * exponent if x != 0 else x, self.get_dimensions_list())
```
h-krishnan, line 536: Why the check for `x != 0`?

src/ansys/fluent/core/quantity.py
hpohekar: We have added `# -*- coding: latin-1 -*-` at the top of the file.

src/ansys/fluent/core/quantity.py
hpohekar: `return Quantity(other, "") - self` causes recursion error.

seanpearsonuk: @hpohekar can you identify a relevant failing test with the current code and make the necessary changes to pass the test.

src/ansys/fluent/core/quantity.py
seanpearsonuk: @hpohekar please consider those as backlog items: you can simply add empty methods to keep track of it. Please focus on testing the current code.

src/ansys/fluent/core/quantity.py
seanpearsonuk: Yes, that should be redundant.

src/ansys/fluent/core/quantity.py
hpohekar: @h-krishnan has helped me to resolve this error. It's resolved now. Thank you.

src/ansys/fluent/core/quantity.py
seanpearsonuk: `elif power == 1` is unreachable. Is this tested?

src/ansys/fluent/core/quantity.py
hpohekar: Done . Thank you.

src/ansys/fluent/core/quantity.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/quantity.py
hpohekar: Updated mu symbol by copying from character map.

image

src/ansys/fluent/core/quantity.py
hpohekar: Updated ohm symbol by copying from character map.

Don't know how to get this symbol -> Θ.

image

src/ansys/fluent/core/quantity.py
```diff
@@ -376,7 +376,7 @@ def get_si_unit_from_dim(dim_list):
         "M": "kg",
         "L": "m",
         "T": "s",
-        "Temp": "K",
+        "?": "K",
```
h-krishnan, line 379: @dnwillia-work @seanpearsonuk @hpohekar One issue with using characters such as these or even "mu" is that they are not readily accessible from the keyboard and one will have to copy them from some place.

src/ansys/fluent/core/quantity.py
h-krishnan:
``` python
if power == 1:
  unit += spacechar + base_units[key]
elif power != 0:
  unit += spacechar + base_units[key] + "^" + str(int(power) if power.is_integer() else power)
```

h-krishnan:
Some issues:
``` python
>>> a = Quantity (3.0, "m")
>>> c = Quantity(4.0, "")
>>> a + "a" # wrong error message
  File "quantity.py", line 588, in __add__
    raise ValueError("Incompatible dimensions.")
ValueError: Incompatible dimensions.
>>> c + 2  # adding float to dimensionless number should be supported
  File "quantity.py", line 589, in __add__
    return Quantity(temp_value, self._si_unit)
  File "quantity.py", line 483, in __init__
    self._dimension = Dimension(unit_str)
  File "quantity.py", line 322, in __init__
    self._parser(unit_str)
  File "quantity.py", line 353, in _parser
    unit_dim = self._get_dim(unit_str, term_power)
  File "quantity.py", line 370, in _get_dim
    raise ValueError("Not implemented")
ValueError: Not implemented
>>>
```

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,620 @@
...
+        "psf": "lbf ft^-2",
+        "ohm": "kg m^2 s^-3 A^-2",
+        "Hz": "s^-1",
+    }
+
+    derived_units_with_conversion_factor = {
+        "l": (0.001, "m^3"),
+        "gal": (0.0037854117839999993, "m^3"),
+        "BTU": (1055.056, "J"),
+        "C": (lambda c: c + 273.15, "K"),
```
seanpearsonuk, line 56: as I discussed with @hpohekar, we don't need any specific functions for these: only generic functions based on conversion tables. The code will be updated.

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,620 @@
...
+        "F": 255.92777777777778,
+        "R": 0.5555555555555556,
+    }
+
+    offset_dict = {
+        "K": {"C": 274.15, "F": 255.92777777777778, "R": 0.5555555555555556},
+        "C": {"K": -272.15, "F": -17.2222222222222, "R": -272.59444444444443},
+        "F": {"C": 33.79999999999993, "K": -457.87, "R": -458.67},
+        "R": {"C": 493.4699999999999, "F": 460.66999999999996, "K": 1.7999999999999998},
+    }
```
seanpearsonuk, line 159: as discussed with @hpohekar we don't need this special code

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,620 @@
...
+        "slugmol": 14593.9,
+        "cd": 1,
+        "sr": 1,
+        "radian": 1,
+        "degree": 0.017453292519943295,
+        "K": 1,
+        "C": 1,
+        "F": 0.5555555555555556,
+        "R": 0.5555555555555556,
+    }
```
seanpearsonuk, line 120: As discussed with @hpohekar we don't need to see any conversion data defined elsewhere, especially not inside functions. We can define factor and optional offset in each case, so temperature (and anything else where relative scales are used) can live together with everything else.

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,620 @@
...
+        "cd": "cd",
+        "sr": "sr",
+        "radian": "radian",
+        "degree": "radian",
+        "K": "K",
+        "C": "K",
+        "F": "K",
+        "R": "K",
+        "": "",
+    }
```
seanpearsonuk, line 145: This map should be redundant if we define everything properly.

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,620 @@
+# -*- coding: latin-1 -*-
+from collections import OrderedDict
+
+
```
seanpearsonuk, line 4: In general we need to re-work the fundamental configuration data structures, but initial focus needs to be on covering the current code with tests.

src/ansys/fluent/core/quantity.py
dnwillia-work: Yeah, it's true.  I was struggling with this yesterday.  I would say it's only an issue if the characters leak through the API and become part of the interface.   If it's just internal implementation, then I'm ok with it.

dnwillia-work: Here's a commit where I fixed up the question marks.

https://github.com/pyansys/pyfluent/commit/f10020b9c78ad368c18599226e58760636acd3d9

It seems OK.  I think it's important to do this in an editor where the encoding is UTF-8.

src/ansys/fluent/core/quantity.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/quantity.py
hpohekar: Thank you very much @dnwillia-work , @h-krishnan .

src/ansys/fluent/core/quantity.py
hpohekar: Thank you very much @dnwillia-work .

src/ansys/fluent/core/quantity.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/quantity.py
h-krishnan: The issue is not whether it appears OK in editors, but how users will enter such a symbol in the first place as it is not readily available in the keyboard.

dnwillia-work: Yes, got it, hence my first comment where I said "it's only an issue if the characters leak through the API ..." .  They do not leak through right now but, that said, probably it's best to stick to keyboard only characters.

hpohekar: @dnwillia-work @h-krishnan @seanpearsonuk This is resolved now.

The important point is that both `C++ units library` and `pint` uses `u` for `µ`.

So we have replaced `µ` with `u`.

`Theta` and `ohm` sign replaced by `Temp` and `SAngle` respectively.

src/ansys/fluent/core/quantity.py
hpohekar: Code is updated and as per discussion, used this with different logic.

src/ansys/fluent/core/quantity.py
hpohekar: Removed. Thank you.

src/ansys/fluent/core/quantity.py
hpohekar: Now we are supporting all temperature conversions.  We will definitely improve this one. Thank you.

src/ansys/fluent/core/quantity.py
hpohekar: Yes, we will restructure this. Thank you.

src/ansys/fluent/core/quantity.py
hpohekar: @dnwillia-work @seanpearsonuk @h-krishnan @mkundu1 

Now we support all temperature conversions as well.  

We will re-work on data structures and now focused on covering all possible testing.

Thank you.

tests/test_quantity.py
hpohekar: > Thanks, maybe we can make it `y = (287.28, "m s^0.5")` for brevity.

@mkundu1 The power issue is resolved now. Thank you.

image

h-krishnan:
``` python
>>> q
Quantity (5.0, "m")
>>> -q  # should return Quantity(-5.0, "m")
-5.0
>>>
```

hpohekar:
> Some issues:
> 
> ```python
> >>> a = Quantity (3.0, "m")
> >>> c = Quantity(4.0, "")
> >>> a + "a" # wrong error message
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "quantity.py", line 588, in __add__
>     raise ValueError("Incompatible dimensions.")
> ValueError: Incompatible dimensions.
> >>> c + 2  # adding float to dimensionless number should be supported
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "quantity.py", line 589, in __add__
>     return Quantity(temp_value, self._si_unit)
>   File "quantity.py", line 483, in __init__
>     self._dimension = Dimension(unit_str)
>   File "quantity.py", line 322, in __init__
>     self._parser(unit_str)
>   File "quantity.py", line 353, in _parser
>     unit_dim = self._get_dim(unit_str, term_power)
>   File "quantity.py", line 370, in _get_dim
>     raise ValueError("Not implemented")
> ValueError: Not implemented
> >>>
> ```

Done. Thank you.

image



image

hpohekar:
> ```python
> >>> q
> Quantity (5.0, "m")
> >>> -q  # should return Quantity(-5.0, "m")
> -5.0
> >>>
> ```

Done. Thank you.

image

h-krishnan:
What is the float or si_value of `Quantity(1, 'C")`?

src/ansys/fluent/core/quantity.py
```diff
@@ -0,0 +1,732 @@
...
+            isinstance(other, Quantity)
+            and self.get_dimensions_list() == other.get_dimensions_list()
+            and self._si_unit == other._si_unit
+        ):
+            return float(self) > float(other)
+        elif (
+            self.is_dimension_less()
+            and (not isinstance(other, Quantity))
+            and isinstance(other, (float, int))
+        ):
```
h-krishnan, line 731: If dimensions don't match, you should throw an error.

h-krishnan: Maybe in the next PR, you can include all comparison functions (__gt__, __ge__, __lt__, __le__, __eq__, __neq__) and their __r* counterparts (__rgt__, etc.)

hpohekar: > If dimensions don't match, you should throw an error.

Yes, Thank you.

hpohekar: > Maybe in the next PR, you can include all comparison functions (**gt**, **ge**, **lt**, **le**, **eq**, **neq**) and their __r* counterparts (**rgt**, etc.)

Yes sure, we will add remaining functions in next PR.

src/ansys/fluent/core/quantity.py
hpohekar: Yes sure.

hpohekar:
> What is the float or si_value of `Quantity(1, 'C')`?

@h-krishnan @seanpearsonuk @dnwillia-work 

One Celsius degree is an interval of 1 K, and zero degrees Celsius is 273.15 K.

Above point is mentioned on official website of US government - 
https://www.nist.gov/pml/owm/si-units-temperature#:~:text=One%20Celsius%20degree%20is%20an,on%20the%20Fahrenheit%20temperature%20scale.

Hence there are 2 parts now. 

1st part - representation of Celsius in terms of Kelvin (includes SI multiplier as 1) 

This gives float or si_value and si_unit of `Quantity(1, 'C')` as (1.0, K)

2nd part - conversion of Celsius to Kelvin (includes addition of an offset 273.15)

This gives conversion `Quantity(1, 'C')` to Kelvin as  (274.15, "K")

It is also defined in `units.cfx` as follows -

image

Therefore we have also defined this mapping at 2 places in `quantity.py`.

1st is in `conversion_map` as follows - 

image

2nd is in `temperature_conversions` as follows -

image

`C++ units library` does the same thing, so we are consistent with `units.cfx` as well.

We are also consistent with all conversions mentioned on following website - 
https://www.engineeringtoolbox.com/specific-heat-capacity-converter-d_673.html

We have covered and verified all possible conversions from above website and `UnitSystems.ccl` (including tempratues variance which involves `K^2`) in  `test_tempk_48 to test_temp_54`.

Thank you.

hpohekar:
> Pinging here @RobPasMue. He's feedback may be useful for this, as he implemented a units module on top of [pint](https://pint.readthedocs.io/en/stable/) for an internal project.

@jorgepiloto @h-krishnan @seanpearsonuk @dnwillia-work 

We have found 1 bug in `pint` module. 

It does not convert temperatures properly. 

See following case - Converting `(-40000, milli-kelvin)` to `micro-degree-Celsius`

image

`C++ units library` gives this - 

image

and this current `Real Quantity with Units module` is also gives same result in comparison to `C++ units library` - 

image

src/ansys/fluent/core/quantity.py
hpohekar: @h-krishnan @seanpearsonuk @dnwillia-work @mkundu1 

I have noted this points. In previous version using `pint` and in current version on `C++` side we are converting all quantities to SI before performing any mathematical operations, because in any version whether it is 'pint' based or `C++`,  we are doing this for consistency. So may be we can have discussion about this and then we can decide.

Thank you.

seanpearsonuk:
-40K makes sense only as a temperature _difference_ so the conversion to relative scales should be considered undefined. While pint seems to be inconsistent between C and microC, neither result is worse than the other.

mkundu1 (approved):

dnwillia-work (approved):

h-krishnan:
> > What is the float or si_value of `Quantity(1, 'C')`?
> 
> @h-krishnan @seanpearsonuk @dnwillia-work @mkundu1
> 
> `---------------------------------------------`
> 
> The degree Celsius is itself used in SI units alongside the Kelvin -
> 
> https://en.wikipedia.org/wiki/Celsius
> 
> "The degree Celsius is the unit of [temperature] on the Celsius scale, one of 2 [temperature scales] used in the [International System of Units] alongside the [Kelvin scale]" (mentioned on Wikipedia)
> 
> `---------------------------------------------`
> 
> One Celsius degree is an interval of 1 K, and zero degrees Celsius is 273.15 K.
> 
> Above point is mentioned on official website of US government - https://www.nist.gov/pml/owm/si-units-temperature#:~:text=One%20Celsius%20degree%20is%20an,on%20the%20Fahrenheit%20temperature%20scale.
> 
> Hence there are 2 parts now.
> 
> 1st part - representation of Celsius in terms of Kelvin (includes SI multiplier as 1)
> 
> This gives float or si_value and si_unit of `Quantity(1, 'C')` as (1.0, K)
> 
> 2nd part - conversion of Celsius to Kelvin (includes addition of an offset 273.15)
> 
> This gives conversion `Quantity(1, 'C')` to Kelvin as (274.15, "K")
> 
> It is also defined in `units.cfx` as follows -
> 
> image
> 
> Therefore we have also defined this mapping at 2 places in `quantity.py`.
> 
> 1st is in `conversion_map` as follows -
> 
> image
> 
> 2nd is in `temperature_conversions` as follows -
> 
> image
> 
> `C++ units library` does the same thing, so we are consistent with `units.cfx` as well.
> 
> We are also consistent with all conversions mentioned on following website - https://www.engineeringtoolbox.com/specific-heat-capacity-converter-d_673.html
> 
> We have covered and verified conversions from above website, `C++ units library` and `UnitSystems.ccl` (including temperatures variance which involves `K^2`) in `test_tempk_48 to test_temp_54` (from `test_quantity.py`).
> 
> Thank you.

My opinion:
If we say `Quantity(1, "C")` refers to temperature difference always, then the float value should be 1. If we however say that `Quantity(1, "C")` refers to the absolute temperature, the float value should `274.15`. 
cc @seanpearsonuk @dnwillia-work

hpohekar:
> > > What is the float or si_value of `Quantity(1, 'C')`?
> > 
> > 
> > @h-krishnan @seanpearsonuk @dnwillia-work @mkundu1
> > `---------------------------------------------`
> > The degree Celsius is itself used in SI units alongside the Kelvin -
> > https://en.wikipedia.org/wiki/Celsius
> > "The degree Celsius is the unit of [temperature] on the Celsius scale, one of 2 [temperature scales] used in the [International System of Units] alongside the [Kelvin scale]" (mentioned on Wikipedia)
> > `---------------------------------------------`
> > One Celsius degree is an interval of 1 K, and zero degrees Celsius is 273.15 K.
> > Above point is mentioned on official website of US government - https://www.nist.gov/pml/owm/si-units-temperature#:~:text=One%20Celsius%20degree%20is%20an,on%20the%20Fahrenheit%20temperature%20scale.
> > Hence there are 2 parts now.
> > 1st part - representation of Celsius in terms of Kelvin (includes SI multiplier as 1)
> > This gives float or si_value and si_unit of `Quantity(1, 'C')` as (1.0, K)
> > 2nd part - conversion of Celsius to Kelvin (includes addition of an offset 273.15)
> > This gives conversion `Quantity(1, 'C')` to Kelvin as (274.15, "K")
> > It is also defined in `units.cfx` as follows -
> > image
> > Therefore we have also defined this mapping at 2 places in `quantity.py`.
> > 1st is in `conversion_map` as follows -
> > image
> > 2nd is in `temperature_conversions` as follows -
> > image
> > `C++ units library` does the same thing, so we are consistent with `units.cfx` as well.
> > We are also consistent with all conversions mentioned on following website - https://www.engineeringtoolbox.com/specific-heat-capacity-converter-d_673.html
> > We have covered and verified conversions from above website, `C++ units library` and `UnitSystems.ccl` (including temperatures variance which involves `K^2`) in `test_tempk_48 to test_temp_54` (from `test_quantity.py`).
> > Thank you.
> 
> My opinion: If we say `Quantity(1, "C")` refers to temperature difference always, then the float value should be 1. If we however say that `Quantity(1, "C")` refers to the absolute temperature, the float value should `274.15`. cc @seanpearsonuk @dnwillia-work

@h-krishnan @seanpearsonuk @dnwillia-work @mkundu1 

The main point related to this is, the `C++ units library` uses same unit for both absolute temperature and temperature difference so we also follows this. (from `UnitSystems.ccl`)

image

image

image

image

image

image

image

`C++ units library` gives this - 

image

Current `Quantity` module gives this -

image

`pint` handles `degree_Celsius` differently, it uses `delta_degree_Celsius` unit for temperature difference as follows - 

image

`pint` does not allow to create `degree Celsius` quantity as `absolute temperature`, only allows to treat it as `temperature difference`

If we try to construct it as  `absolute temperature`, it gives following error. 

image

So we summarize above in following points -

1.  `Quantity(1, "C")` in both `C++ units library` and `Quantity` module refers to temperature difference always, so we the float value is 1.
3. `Quantity(1, "C")` in `pint` also always refers to temperature difference, but represented with `delta_degree_Celsius` unit. 

The conclusion is as follows - 
The `C++ units library`, `Quantity` module and `pint`, all of them always refers 'Quantity(1, "C")` as `temperature difference` only.

h-krishnan:
@hpohekar @seanpearsonuk @dnwillia-work So, what does `Quantity(2, "C")*Quantity(3, "m")` give? In the absence of any other context, is `Quantity(2, "C")` a temperature difference or a temperature? I feel it should be "temperature" in which case `Quantity(2, "C")` should be 275.15.

hpohekar:
> @hpohekar @seanpearsonuk @dnwillia-work So, what does `Quantity(2, "C")*Quantity(3, "m")` give? In the absence of any other context, is `Quantity(2, "C")` a temperature difference or a temperature? I feel it should be "temperature" in which case `Quantity(2, "C")` should be 275.15.

@h-krishnan @dnwillia-work @seanpearsonuk 

By referring terms (`temperature difference` and `absolute temperature`) used in following comment,

image

I have explained that by the value which we 

> @hpohekar @seanpearsonuk @dnwillia-work So, what does `Quantity(2, "C")*Quantity(3, "m")` give? In the absence of any other context, is `Quantity(2, "C")` a temperature difference or a temperature? I feel it should be "temperature" in which case `Quantity(2, "C")` should be 275.15.

@h-krishnan @dnwillia-work @seanpearsonuk 

Right, in absence of any other context, currently we can not differentiate between whether the quantity is `temperature` or `temperature difference` and root cause of this is the use of same unit for both `temperature` and `temperature difference`.

We will have to handle this case differently.

Thank you.

PR #1034, opened 2022-10-26
Fixed set_state to be applicable only to CommandArguments

prmukherj:


src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -849,6 +860,7 @@ class AccessorModes(Enum):
         ["Bool", "Logical", "Logical List"],
         PyParameterCommandArgumentsSubItem,
     )
+    MODELOBJECT = (["ModelObject"], PySingletonCommandArgumentsSubItem)
     GENERIC = ([], PyCommandArgumentsSubItem)
```
prmukherj, line 864: @seanpearsonuk , I don't think so that this is required anymore. In that case can we remove it? If so, do we need to raise an exception in the else case @ line 873?

seanpearsonuk: @prmukherj yes we can and I'll anticipate your changes along lines we discussed. Thanks

prmukherj: Yes @seanpearsonuk, those changes are already in place. Thank you

seanpearsonuk (approved):

hpohekar (approved):

PR #1047, opened 2022-11-01
Fix issues after Fluent exit

mkundu1:


src/ansys/fluent/core/fluent_connection.py
```diff
@@ -318,7 +318,10 @@ def _exit(
     ) -> None:
         if channel:
             if cleanup_on_exit:
-                scheme_eval.exec(("(exit-server)",))
+                try:
+                    scheme_eval.exec(("(exit-server)",))
```
mkundu1, line 322: This will throw if Fluent server has already exited.

prmukherj (approved):

hpohekar (approved):

ajain-work (approved):

PR #1067, opened 2022-11-06
Added methods to calculate force and moment.

prmukherj:


tests/test_reduction.py
```diff
@@ -226,6 +226,52 @@ def _test_area_integrated_average(solver1, solver2):
...
+def _test_moment(solver):
+    solver.solution.initialization.hybrid_initialize()
+    solver.setup.named_expressions["test_expr_1"] = {}
+    solver.setup.named_expressions[
+        "test_expr_1"
+    ].definition = "Moment(Force(['wall']),['wall'])"
+    expr_val_1 = solver.setup.named_expressions["test_expr_1"].get_value()
+
+    red_moment = reduction.moment(
+        expr="Force(['wall'])", locations=[solver.setup.boundary_conditions.wall]
```
prmukherj, line 267: @seanpearsonuk, for now the moment can be calculated by passing the expression like: "Force(['wall'])", as shown in the test case. However, we can also do something to process out the reduction string from reduction.force(<locations>), and then pass that sting to 'expr' of reduction.moment(). For now the later has not been applied for keeping the behavior analogous to others where 'expr' was a string. Please let me know what you think and whether the later should be applied (as it looks interesting :)).
Thank you.

seanpearsonuk: @prmukherj above usage looks good. expr argument can in general be any expression accepted as valid by the backend implementation so expr can be an API object too. I don't know if "process out the reduction string from reduction.force()" works or not.

seanpearsonuk (approved):
looks good

hpohekar (approved):

mkundu1 (approved):

PR #1075, opened 2022-11-08
Real Quantity with Units module comparison operators

hpohekar:


src/ansys/fluent/core/quantity.py
```diff
@@ -717,16 +717,98 @@ def __eq__(self, other):
...
-    def __ge__(self, other):
-        if (
-            isinstance(other, Quantity)
-            and self.get_dimensions_list() == other.get_dimensions_list()
-            and self._si_unit == other._si_unit
+    def __gt__(self, other):
+        if isinstance(other, Quantity) and (
+            self.get_dimensions_list() != other.get_dimensions_list()
         ):
+            raise ValueError("Incompatible dimensions.")
```
seanpearsonuk, line 724: Future refactor: write a single method that checks that the dimensions are compatible and raises an exception if they are not. The method can accept float arguments which are handled as dimensionless quantities. 
```
# throws
def validate_matching_dimensions(self, other):
```

src/ansys/fluent/core/quantity.py
```diff
@@ -717,16 +717,98 @@ def __eq__(self, other):
...
+
+    def __ge__(self, other):
+        if isinstance(other, Quantity) and (
+            self.get_dimensions_list() != other.get_dimensions_list()
+        ):
+            raise ValueError("Incompatible dimensions.")
+        elif isinstance(other, Quantity):
+            return float(self) >= float(other)
+        elif (
+            self.is_dimension_less()
```
seanpearsonuk, line 744: I think the name should be `is_dimensionless`.

src/ansys/fluent/core/quantity.py
```diff
@@ -717,16 +717,98 @@ def __eq__(self, other):
...
-        if (
-            isinstance(other, Quantity)
-            and self.get_dimensions_list() == other.get_dimensions_list()
-            and self._si_unit == other._si_unit
+    def __gt__(self, other):
+        if isinstance(other, Quantity) and (
+            self.get_dimensions_list() != other.get_dimensions_list()
         ):
+            raise ValueError("Incompatible dimensions.")
+        elif isinstance(other, Quantity):
```
seanpearsonuk, line 725: After we have done *all* the validation at the start of each method, the rest of the method can become something quite simple (we are on safe ground). So, the right-hand side can be converted to float regardless: we don't need to check what it is.

So, each method is something like:
```
validate_matching_dimensions(other)
return float(self) <op> float(other)
```

tests/test_quantity.py
```diff
@@ -520,6 +520,94 @@ def test_power_56():
...
+    with pytest.raises(ValueError) as e_info:
+        assert x >= z
+        assert x >= y
+        assert 5.0 >= r
+
+    with pytest.raises(TypeError) as e_info:
+        assert x >= 5.0
+
+
+def test_gt_59():
```
dnwillia-work, line 538: Did you want to test that the operator actually works?  Here you are testing the exceptions so far, which is good...

hpohekar: We have already confirmed the operator's working, so only testing the exceptions here. Thank you.

dnwillia-work: OK, staring at the unit tests I cannot see which unit test confirms they are working.  Am I blind?

hpohekar: I mean Krishnan and I, checked the both left and right equivalents locally, so I have not explicitly added them in these unit tests.

But now I will add them too.

Thank you.

hpohekar: Done. Thank you.

src/ansys/fluent/core/quantity.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/quantity.py
hpohekar: Done. Thank you very much.

src/ansys/fluent/core/quantity.py
hpohekar: Done. Thank you very much.

src/ansys/fluent/core/quantity.py
```diff
@@ -708,7 +708,7 @@ def __eq__(self, other):
         if isinstance(other, Quantity):
```
h-krishnan, line 708: Maybe you can use `validate_matching_dimensions` here too

src/ansys/fluent/core/quantity.py
```diff
@@ -693,7 +693,7 @@ def __sub__(self, other):
             temp_value = self._si_value - other._si_value
```
h-krishnan, line 693: Maybe you can use `validate_matching_dimensions` for `__add__` and `__sub__` as well.

src/ansys/fluent/core/quantity.py
hpohekar: Right. Done. Thank you.

src/ansys/fluent/core/quantity.py
hpohekar: Right. Done. Thank you.

src/ansys/fluent/core/quantity.py
hpohekar: Right. Done. Thank you.

prmukherj (approved):

h-krishnan (approved):

PR #1079, opened 2022-11-09
find_children() in flobject

prmukherj:


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -243,6 +244,32 @@ def set_state(self, state: StateT = None, **kwargs):
...
+                    cls.child_object_type, path, identifier, _list_of_children
+                )
+        if issubclass(cls, Group):
+            SettingsBase._get_child_path(cls, path, identifier, _list_of_children)
+
+    @staticmethod
+    def _get_child_path(cls, path, identifier, _list_of_children):
+        for name in cls.child_names:
+            path.append(name)
+            if identifier == "" or fnmatch.fnmatch(name, identifier):
```
mkundu1, line 266: I think we can remove the `identifier == ""` condition. `"*"` will return everything.

prmukherj: Thank you. Done that.

seanpearsonuk: @prmukherj: please could you add a test so that I can fully understand the behaviour. Thank you.

prmukherj: Sure @seanpearsonuk, doing that currently. Will commit it soon. Thank you

prmukherj: Tests have been put in place.

seanpearsonuk:
Adding @h-krishnan to get his feedback. Thanks, @prmukherj.

seanpearsonuk (approved):

mkundu1 (approved):

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -243,6 +244,33 @@ def set_state(self, state: StateT = None, **kwargs):
         else:
             return self.flproxy.set_var(self.path, self.to_scheme_keys(state))
 
+    def find_child(self, identifier="*"):
```
h-krishnan, line 247: Shouldn't this be called `find_children` if it is returning multiple children?

Should we rename `identifier` as `pattern` if it is a wild-card pattern?

Does this recurse to all children and return all children matching that pattern? If so, do we need a way to disable recursion?

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -243,6 +244,33 @@ def set_state(self, state: StateT = None, **kwargs):
...
+        """Lists path of all the child objects matching an identifier."""
```
h-krishnan, line 248: `Returns` instead of `Lists`?

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -243,6 +244,33 @@ def set_state(self, state: StateT = None, **kwargs):
...
+        _list_of_children = []
```
h-krishnan, line 249: Local variable need not/should not be prefixed with _.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -243,6 +244,33 @@ def set_state(self, state: StateT = None, **kwargs):
...
+
+    @staticmethod
+    def _list_children(cls, identifier, path, _list_of_children):
+        if issubclass(cls, NamedObject):
+            if hasattr(cls.child_object_type, "child_names"):
+                SettingsBase._get_child_path(
+                    cls.child_object_type, path, identifier, _list_of_children
+                )
+        if issubclass(cls, Group):
+            SettingsBase._get_child_path(cls, path, identifier, _list_of_children)
```
h-krishnan, line 261: Shouldn't the above be defined in the derived classes?

Why is `ListObject` missing here?

src/ansys/fluent/core/solver/flobject.py
prmukherj: find_children sounds good, we can update that and the identifier to pattern as well. Thank you for the suggestion.
Yes, it currently recurse to all children and return all children matching that pattern, @seanpearsonuk, do we need this behavior or stop the recursion?

h-krishnan: Should we return paths or the objects themselves? What can users do with the path other than get the object with it?

mkundu1: Please note that the current implementation is based on static information. It doesn't resolve the named objects nor filter the inactive children. This is to help users to write scripts before reading the case or setting up the model. Maybe a more appropriate name would be `search_child_paths` for this function.

h-krishnan: So, this is for use in the editor and not in actual scripts?

prmukherj: just to clarify on the above discussion:
>>> solver.setup.boundary_conditions.velocity_inlet.find_child("vmag*")
['phase/vmag', 'vmag']

This is the current behavior.

src/ansys/fluent/core/solver/flobject.py
prmukherj: Done, Thank you

src/ansys/fluent/core/solver/flobject.py
prmukherj: Done, thank you

src/ansys/fluent/core/solver/flobject.py
prmukherj: ListObject has been included as well. Thank you.

src/ansys/fluent/core/solver/flobject.py
prmukherj: All the suggestions provided by @h-krishnan have been implemented. The ListObjects are covered by this as well now. Can this go in as the first version of find_child() and later we do any modifications as and when required? What do you think @h-krishnan , @seanpearsonuk ?

seanpearsonuk: Hi @prmukherj, just recalling the `find_children` suggestion from @h-krishnan.

prmukherj: Yes @seanpearsonuk, this sounds more appropriate and has been renamed accordingly. Thank you for pointing it out.

h-krishnan (approved):

PR #1089, opened 2022-11-09
Resolved Temperature Conversion issue

hpohekar:
@h-krishnan @seanpearsonuk @dnwillia 

`------------------------------------------------------------------------------------------------------------------`

image


prmukherj (approved):

h-krishnan (commented):
Is this correct?
``` python
>>> a = Quantity(2, "C^-1")
>>> float(a)
2.0
```

tests/test_quantity.py
```diff
@@ -803,6 +827,20 @@ def testing_properties():
...
-
-    c = q.Quantity(1.0, "C")
-    print(float(c))
+    print("")
+
+    c = q.Quantity(1.0, "C^1")
+    m = q.Quantity(1.0, "m")
+    print(f'c = q.Quantity(1.0, "C^1") = {c}')
+    print("")
+    print(f"float(c) = {float(c)}")
```
dnwillia-work, line 836: Seems like most of these print statements can be replaced by unit tests plus assertions.  From this code I can't actually tell what each individual operation will actually do.

seanpearsonuk: Yes, print statements OK while debugging but not left in for the merge.

tests/test_quantity.py
```diff
@@ -803,6 +827,20 @@ def testing_properties():
     print(
         f"User unit: {x._unit.user_unit}, multiplier: {x._unit.si_factor}, reduced_si_unit: {x._unit.si_unit}, si_value: {x._si_value}"
     )
-
-    c = q.Quantity(1.0, "C")
-    print(float(c))
+    print("")
+
+    c = q.Quantity(1.0, "C^1")
```
seanpearsonuk, line 832: On a different topic, does pint use `"C"` for degrees Celcius?

src/ansys/fluent/core/quantity.py
```diff
@@ -333,6 +338,12 @@ def _compute_multipliers(self, unit_str, power):
                     _UnitsTable.derived_units[unit_str], term_power
                 )
 
+    def _compute_offset(self, unit_str):
+        if unit_str == "C" or unit_str == "C^1":
```
seanpearsonuk, line 342: or `if unit_str in ("C", "C^1")`

Later the logic will have to be less tied to the specific units, and even beyond that, to the string representation of those units.

src/ansys/fluent/core/quantity.py
```diff
@@ -333,6 +338,12 @@ def _compute_multipliers(self, unit_str, power):
...
+            self._si_offset = 273.15
```
seanpearsonuk, line 343: I don't see any changes where we are removing `273.15` (nor other numbers essentially derived from that offset), so we need to keep in mind the need to refactor subsequently.

tests/test_quantity.py
```diff
@@ -146,12 +146,16 @@ def test_to_20():
     with pytest.raises(ValueError) as e_info:
         convert = v.to("radian s^-1")
 
+    assert e_info.value.args[0] == "Incompatible conversion from Hz : to radian s^-1"
```
seanpearsonuk, line 149: This kind of test would be more stable if the implementation used a new exception type (which could inherit from ValueError) with the two unit strings as construction arguments. Then you can easily construct the expected exception in the test for comparison.

tests/test_quantity.py
```diff
@@ -535,9 +539,13 @@ def test_ge_57():
         assert x >= y
         assert 5.0 >= r
 
+    assert e_info.value.args[0] == "Incompatible dimensions."
```
seanpearsonuk, line 542: Similar as above applies here.

src/ansys/fluent/core/quantity.py
```diff
@@ -532,13 +543,12 @@ def __init__(self, real_value, unit_str):
         self._value = float(real_value)
         self._unit = Unit(unit_str)
         self._dimension = Dimension(unit_str)
-        self._si_value = self._unit.si_factor * self._value
+        self._si_value = self._unit.si_factor * self._value + self._unit.si_offset
```
seanpearsonuk, line 546: We're also doing some specific stuff like this (in lambdas for temperature conversions). That other code still stands currently?

tests/test_quantity.py
hpohekar: We are going to remove all code from `if __name__ == "__main__":` onwards later on. It's temporary for now. Thank you.

src/ansys/fluent/core/quantity.py
hpohekar: Yes, the upcoding code is already updated.

src/ansys/fluent/core/quantity.py
hpohekar: Yes, it's a part of our main core `to` method and we have added above code to handle special case of `si_value` of degree Celsius and Fahrenheit.

hpohekar:
> Is this correct?
> 
> ```python
> >>> a = Quantity(2, "C^-1")
> >>> float(a)
> 2.0
> ```

image

seanpearsonuk:
> Is this correct?
> 
> ```python
> >>> a = Quantity(2, "C^-1")
> >>> float(a)
> 2.0
> ```

If we're talking about something "per celsius" then what does that mean? I would saying it's pretty trashy to begin with, but if we want to allow it, then we need to understand what can possibly be meant, and I believe that the offset has no role here because we are talking about "per _degree of_ celsius", implying that the above behaviour is correct (or at least the least incorrect). However, there is a concern that this behaviour could lead to inconsistencies and therefore bugs. E.g. can we write a test that takes a celsius quantity in a cycle that changes its value (inverting twice for instance)?

dnwillia-work:
> > Is this correct?
> > ```python
> > >>> a = Quantity(2, "C^-1")
> > >>> float(a)
> > 2.0
> > ```
> 
> If we're talking about something "per celsius" then what does that mean? I would say it's pretty trashy to begin with, but if we want to allow it, then we need to understand what can possibly be meant, and I believe that the offset has no role here because we are talking about "per _degree of_ celsius", implying that the above behaviour is correct (or at least the least incorrect). However, there is a concern that this behaviour could lead to inconsistencies and therefore bugs. E.g. can we write a test that takes a celsius quantity in a cycle that changes its value (inverting twice for instance)?

Yeah, a better example might be `cp = q.Quantity(4184.0, 'J kg^-1 C^-1')` which is equivalent to `cp = q.Quantity(4184.0, 'J kg^-1 K^-1')`.  I think it means that `C^-1` is implying a mathematical operation so it should be a temperature difference.

src/ansys/fluent/core/quantity.py
hpohekar: Yes, we will definitely refactor all code. thank you. Right now code is updated and passing all test cases. Thank you.

tests/test_quantity.py
hpohekar: Done. Thank you.

tests/test_quantity.py
hpohekar: Done. Thank you.

h-krishnan:
> > > Is this correct?
> > > ```python
> > > >>> a = Quantity(2, "C^-1")
> > > >>> float(a)
> > > 2.0
> > > ```
> > 
> > 
> > If we're talking about something "per celsius" then what does that mean? I would say it's pretty trashy to begin with, but if we want to allow it, then we need to understand what can possibly be meant, and I believe that the offset has no role here because we are talking about "per _degree of_ celsius", implying that the above behaviour is correct (or at least the least incorrect). However, there is a concern that this behaviour could lead to inconsistencies and therefore bugs. E.g. can we write a test that takes a celsius quantity in a cycle that changes its value (inverting twice for instance)?
> 
> Yeah, a better example might be `cp = q.Quantity(4184.0, 'J kg^-1 C^-1')` which is equivalent to `cp = q.Quantity(4184.0, 'J kg^-1 K^-1')`. I think it means that `C^-1` is implying a mathematical operation so it should be a temperature difference. This use case comes from `Q = m*cp*deltaT` so `cp = Q / (m*deltaT)` i.e. by definition the temperature contribution to `cp` is physically a temperature difference. I think probably for other situations where you have C^-1 as some kind of slope 'per unit temperature' it's got to be a temperature difference.

We then need a rule that if unit is exactly "K", "R", "C" or "F" in `Quantity(value, unit)`, the unit quantity is "Temperature". In all other situations where these units appear (including "C^-1", the unit quantity is "Temperatue Difference".

hpohekar:
> > > > Is this correct?
> > > > ```python
> > > > >>> a = Quantity(2, "C^-1")
> > > > >>> float(a)
> > > > 2.0
> > > > ```
> > > 
> > > 
> > > If we're talking about something "per celsius" then what does that mean? I would say it's pretty trashy to begin with, but if we want to allow it, then we need to understand what can possibly be meant, and I believe that the offset has no role here because we are talking about "per _degree of_ celsius", implying that the above behaviour is correct (or at least the least incorrect). However, there is a concern that this behaviour could lead to inconsistencies and therefore bugs. E.g. can we write a test that takes a celsius quantity in a cycle that changes its value (inverting twice for instance)?
> > 
> > 
> > Yeah, a better example might be `cp = q.Quantity(4184.0, 'J kg^-1 C^-1')` which is equivalent to `cp = q.Quantity(4184.0, 'J kg^-1 K^-1')`. I think it means that `C^-1` is implying a mathematical operation so it should be a temperature difference. This use case comes from `Q = m*cp*deltaT` so `cp = Q / (m*deltaT)` i.e. by definition the temperature contribution to `cp` is physically a temperature difference. I think probably for other situations where you have C^-1 as some kind of slope 'per unit temperature' it's got to be a temperature difference.
> 
> We then need a rule that if unit is exactly "K", "R", "C" or "F" in `Quantity(value, unit)`, the unit quantity is "Temperature". In all other situations where these units appear (including "C^-1", the unit quantity is "Temperature Difference".

Yes, this is perfect.

tests/test_quantity.py
hpohekar: > On a different topic, does pint use `"C"` for degrees Celsius? And if so, how does it cover coulomb? Similar question regarding the latter for this code.

@seanpearsonuk `C++ units library` uses `coulomb` for coulomb.

image

hpohekar:
> > > > Is this correct?
> > > > ```python
> > > > >>> a = Quantity(2, "C^-1")
> > > > >>> float(a)
> > > > 2.0
> > > > ```
> > > 
> > > 
> > > If we're talking about something "per celsius" then what does that mean? I would say it's pretty trashy to begin with, but if we want to allow it, then we need to understand what can possibly be meant, and I believe that the offset has no role here because we are talking about "per _degree of_ celsius", implying that the above behaviour is correct (or at least the least incorrect). However, there is a concern that this behaviour could lead to inconsistencies and therefore bugs. E.g. can we write a test that takes a celsius quantity in a cycle that changes its value (inverting twice for instance)?
> > 
> > 
> > Yeah, a better example might be `cp = q.Quantity(4184.0, 'J kg^-1 C^-1')` which is equivalent to `cp = q.Quantity(4184.0, 'J kg^-1 K^-1')`. I think it means that `C^-1` is implying a mathematical operation so it should be a temperature difference. This use case comes from `Q = m*cp*deltaT` so `cp = Q / (m*deltaT)` i.e. by definition the temperature contribution to `cp` is physically a temperature difference. I think probably for other situations where you have C^-1 as some kind of slope 'per unit temperature' it's got to be a temperature difference.
> 
> We then need a rule that if unit is exactly "K", "R", "C" or "F" in `Quantity(value, unit)`, the unit quantity is "Temperature". In all other situations where these units appear (including "C^-1", the unit quantity is "Temperatue Difference".

Done. Thank you.

tests/test_quantity.py
```diff
@@ -7,7 +7,7 @@
 DELTA = 1.0e-5
 
 
-class QuantityErrors(Exception):
```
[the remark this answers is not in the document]
hpohekar: Done. Thank you.

tests/test_quantity.py
```diff
@@ -143,14 +143,20 @@ def test_to_19():
...
+        assert e.value_error == q.QuantityError("Hz", "radian s^-1").value_error
 
 
 def test_to_21():
     v = q.Quantity(1.0, "radian s^-1")
-    with pytest.raises(ValueError) as e_info:
-        convert = v.to("Hz")
+    try:
+        v.to("Hz")
+        raise q.QuantityError("radian s^-1", "Hz")
```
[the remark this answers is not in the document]
seanpearsonuk: @hpohekar rather than comparing the exceptions directly, you can compare the exception data. Comparing exceptions is not reliable. That's the only thing you need to change - up to now you were comparing the message strings.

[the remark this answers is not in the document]
seanpearsonuk:
```

>>> class UnitConversionError(ValueError):
...   def __init__(self, from_unit, to_unit):
...     self.from_unit = from_unit
...     self.to_unit = to_unit
       # ...
...

>>> with pytest.raises(UnitConversionError) as ee:
...   raise UnitConversionError("mm", "lb")
...
>>> ee.value.to_unit
'lb'
>>> ee.value.from_unit
'mm'
```


```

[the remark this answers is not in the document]
hpohekar: Done. Thank you.

seanpearsonuk (approved):

PR #1092, opened 2022-11-10
Session clean up

prmukherj:


tests/test_fluent_session.py
```diff
@@ -28,7 +28,7 @@ def print_transcript(transcript):
     print_transcript.called = False
     print_transcript.transcript = None
 
-    session.add_transcript_callback(print_transcript)
+    session._transcript.register_callback(print_transcript)
```
seanpearsonuk, line 31: @prmukherj is the `_transcript` object meant to be public - that would be my assumption. If so, shouldn't we remove the underscore?

prmukherj: yes, it s public. Implemented. Thank you.

seanpearsonuk (approved):

ajain-work (approved):

mkundu1 (approved):

hpohekar (approved):

PR #1104, opened 2022-11-17
Fix noise issue in workflow grpc calls

prmukherj:
…ally.

### DIFF
--- a/src/ansys/fluent/core/services/datamodel_se.py
+++ b/src/ansys/fluent/core/services/datamodel_se.py
@@ -662,16 +662,17 @@ class PyCommand:
         response = self.service.create_command_arguments(request)
         return response.commandid
 
-    def _generate_static_info(self):
-        request = DataModelProtoModule.GetStaticInfoRequest()
-        request.rules = self.rules
-        response = self.service.get_static_info(request)
-        PyCommand._stored_static_info = response.info
+    def _get_static_info(self):
+        if self.rules not in PyCommand._stored_static_info.keys():
+            request = DataModelProtoModule.GetStaticInfoRequest()
+            request.rules = self.rules
+            response = self.service.get_static_info(request)
+            PyCommand._stored_static_info[self.rules] = response.info
+        return PyCommand._stored_static_info[self.rules]
 
     def new(self):
         try:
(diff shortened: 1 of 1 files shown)


hpohekar (approved):

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -661,23 +662,24 @@ def _create_command_arguments(self):
         response = self.service.create_command_arguments(request)
         return response.commandid
 
-    def _get_static_info(self):
+    def _generate_static_info(self):
         request = DataModelProtoModule.GetStaticInfoRequest()
         request.rules = self.rules
         response = self.service.get_static_info(request)
-        return response.info
+        PyCommand._stored_static_info = response.info
```
seanpearsonuk, line 669: @prmukherj 

this code needs to be changed so that the above container is a dict where each key is a rules name. Otherwise we are trying to share the same info for every rules name.

Also, new doesn't really need to change. get_static_info can encapsulate the instantiation of each static info.

prmukherj: Thank you for the feedback @seanpearsonuk. It has been implemented, please let me know what you think of the current implementation.

seanpearsonuk (approved):

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -662,10 +663,12 @@ def _create_command_arguments(self):
         return response.commandid
 
     def _get_static_info(self):
-        request = DataModelProtoModule.GetStaticInfoRequest()
-        request.rules = self.rules
-        response = self.service.get_static_info(request)
-        return response.info
+        if self.rules not in PyCommand._stored_static_info.keys():
```
seanpearsonuk, line 666: @prmukherj please could you add a comment here to the effect that we know that the system has already obtained these same infos remotely in another context and we need to coordinate the code so that the global infos are commonly available in a ll contexts (without additional remote calls).

prmukherj: @seanpearsonuk , Done. Thank you

PR #1108, opened 2022-11-22
Added watch_health module in health check()

prmukherj:
Please note: The watch health functionality is active and can be used, however it needs some server side changes which are implemented only 23R2 onwards. So, we are using the earlier technique for connecting to the fluent session. In future we might look into the prospect of shifting to this new efficient method.
Thank you.

### DIFF
--- a/src/ansys/fluent/core/fluent_connection.py
+++ b/src/ansys/fluent/core/fluent_connection.py
@@ -155,14 +155,7 @@ class _FluentConnection:
 
         self.health_check_service = HealthCheckService(self._channel, self._metadata)
 
-        try:
-            self.health_check_service.wait_for_server(timeout=start_timeout)
-        except RuntimeError as e:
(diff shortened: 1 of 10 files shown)


src/ansys/fluent/core/fluent_connection.py
```diff
@@ -156,11 +156,10 @@ def __init__(
 
         self.health_check_service = HealthCheckService(self._channel, self._metadata)
 
-        counter = 0
-        while not self.health_check_service.is_serving:
-            time.sleep(1)
-            counter += 1
-            if counter > start_timeout:
+        start_time = time.time()
+        while not self.health_check_service.watch_health() == "SERVING":
```
seanpearsonuk, line 160: code to be updated as described

seanpearsonuk (approved):

mkundu1 (approved):

PR #1109, opened 2022-11-22
get_object_names as list

hpohekar:


prmukherj (approved):

seanpearsonuk (approved):

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -573,7 +573,7 @@ def items(self):
 
     def get_object_names(self):
         """Object names."""
-        return self.flproxy.get_object_names(self.path)
+        return list(self.flproxy.get_object_names(self.path))
```
seanpearsonuk, line 576: n.b. the Fluent version is already returning a list here, but it's a harmless update on the Fluent side.

hpohekar: Ok. Thank you.

PR #1110, opened 2022-11-22
get_object_names as list

hpohekar:


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -573,7 +573,12 @@ def items(self):
 
     def get_object_names(self):
         """Object names."""
-        return list(self.flproxy.get_object_names(self.path))
+        object_names = (
+            self.flproxy.get_object_names(self.path)
+            if isinstance(self.flproxy.get_object_names(self.path), list)
```
seanpearsonuk, line 578: It might be better to store the result of `self.flproxy.get_object_names(self.path)` in a local variable.

prmukherj (approved):

src/ansys/fluent/core/solver/flobject.py

seanpearsonuk (approved):

PR #1122, opened 2022-11-30
unit_quantity_map method

hpohekar:
We can construct unit and quantity simultaneously using `unit_quantity_map` from settings API.


tests/test_quantity.py
```diff
@@ -766,6 +766,35 @@ def test_temp_addition_69():
...
+        "Mass": 1,
+        "Velocity": 2.5,
+        "Current": 3,
+        "Light": 2,
+        "Epsilon Flux Coefficient": 2,
+    }
+
+    api_test = q.Quantity(10.5, quantity_map=quantity_map_from_settings_API)
+    assert api_test.value == 10.5
+    assert api_test.unit == "kg m^2.5 s^-2.5 A^3 cd^2 kg^2 m^-4 s^-4"
```
seanpearsonuk, line 780: Ideally, terms with the same units should be collected together. For instances, metres and seconds appear twice each.

hpohekar: by collecting terms with the same units together will only reduce the unit and need to have another loop and power sum method for each base and it will make this operation too much expensive.

Currently each term in `final_unit` (`"kg m^2.5 s^-2.5 A^3 cd^2 kg^2 m^-4 s^-4"`) returned by `get_unit_from_map` method is handled one by one through parser.

So there is no need to collect terms with same units together.

hpohekar: > Ideally, terms with the same units should be collected together. For instances, metres and seconds appear twice each.

One another point is, this returned unit is not reduced to SI fundamentals units and hence we will have to consider both SI bases and derived units bases like `N (newton), T (Tesla), H (Henry), Wb (weber)` etc.

hpohekar: @seanpearsonuk  Done. Thank you.

tests/test_quantity.py
```diff
@@ -766,6 +767,47 @@ def test_temp_addition_69():
     assert t.type == "Temperature"
 
 
+def test_quantity_map_70():
+    quantity_map_from_settings_API = {
```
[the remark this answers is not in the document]
seanpearsonuk: @dnwillia-work This code isn't aiming to cover the possibility of non-SI units coming from the APIs. A parameter object will be composed of a value plus an optional quantity map, which would be of the form shown here. That quantity map will determine the SI units associated with the value, while also retaining full quantity information.

seanpearsonuk (approved):

mkundu1 (approved):

PR #1144, opened 2022-12-02
Fully resolve the field data error handling issue

### LINKED ISSUE #1113: Handle bad field queries more gracefully
Showing low-level errors from nodes when we can trap the error at high-level (user passes in wrong name). We could even say: You requested "turbulent-kinetic-energy". Did you mean "turb-kinetic-energy"?

seanpearsonuk:
Trying to add a bit of safety around field data calls. Up to now, invalid calls (invalid field name, location name) crash Fluent, so I am checking field_info for the validity of names up front, and cleanly raising exceptions in the client as appropriate. An extra feature is that the raised exception can contain hints for closely matching names, which should help many users - we get a lot of queries about what is the right field name to use.

This is similar to the original PR (https://github.com/pyansys/pyfluent/pull/1119), and the second one also but this one takes care of the data-valid? chack successfully - at last. So this PR resolves  #1113.

Outstanding issues:

- the field names from field info are not necessarily all "active". If there is no data then non-mesh variable access will still tend to crash Fluent. Perhaps we can pass an extra argument to the field info queries to request active-only (can this be easily implemented?)
- alignment of field names with variable names across the API. Judging by the queries I receive, I suspect that there is some misalignment, which would be the cause for the misuse
- potential reuse of the introduced close matches function across the APIs?

### DIFF
--- a/tests/test_field_data.py
+++ b/tests/test_field_data.py
@@ -1,6 +1,6 @@
 import os
 
-os.environ["PYFLUENT_FLUENT_ROOT"] = r"C:\ANSYSDev\ANSYSDev\vNNN\fluent"
+os.environ["PYFLUENT_FLUENT_ROOT"] = r"C:\ANSYSDev\ANSYSDev\vNNN\fluent"  # ???
 
 import numpy as np
 import pytest
--- a/tests/test_field_data.py
+++ b/tests/test_field_data.py
@@ -1,7 +1,3 @@
-import os
-
-os.environ["PYFLUENT_FLUENT_ROOT"] = r"C:\ANSYSDev\ANSYSDev\vNNN\fluent"  # ???
-
 import numpy as np
 import pytest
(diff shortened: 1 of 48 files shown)


tests/test_field_data.py
```diff
@@ -1,6 +1,6 @@
 import os
 
-os.environ["PYFLUENT_FLUENT_ROOT"] = r"C:\ANSYSDev\ANSYSDev\vNNN\fluent"
+os.environ["PYFLUENT_FLUENT_ROOT"] = r"C:\ANSYSDev\ANSYSDev\vNNN\fluent"  # ???
```
mkundu1, line 3: Should remove this

mkundu1 (approved):

tests/test_field_data.py
seanpearsonuk: I agree, done

prmukherj (approved):

hpohekar (approved):

PR #1150, opened 2022-12-05
Handling objects_handle.py dependencies

hpohekar:
Eliminated the dependency on visualization library in objects_handle.py from pyfluent-ui repo, while using the object definitions from visualization library.


src/ansys/fluent/core/graphics_definitions/post_data_extractor.py
```diff
@@ -0,0 +1,408 @@
+"""Module providing data extractor APIs."""
+
```
ajain-work, line 2: should not be moved.

src/ansys/fluent/core/graphics_definitions/post_windows_manager.py
```diff
@@ -0,0 +1,174 @@
+"""Module providing PostWindow and PostWindowManager abstract classes.
```
ajain-work, line 1: Should not be moved.

src/ansys/fluent/core/graphics_definitions/post_data_extractor.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/graphics_definitions/post_windows_manager.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/graphics_definitions/pyvista/pyvista_objects.py
```diff
@@ -0,0 +1,276 @@
+"""Module providing visualization objects for PyVista."""
```
seanpearsonuk, line 1: In pyfluent-ui we are using `Graphics` from here while we are not using pyvista. Similarly from the other package we are using `Plot` but not matplotlib. So, in moving this, I believe that we should also move `Graphics` and `Plot` out of such namespaces, so that everything is clearer in the client code.

src/ansys/fluent/core/graphics_definitions/matplotlib/plotter_defns.py
```diff
@@ -0,0 +1,223 @@
+"""Module providing matplotlib plotter functionality."""
+
+from typing import List, Optional
+
+import matplotlib.pyplot as plt
```
seanpearsonuk, line 5: This introduces matplotlib "dependency" in core. We are bringing some shared object definitions from visualization package to here so that we can use them in ui. Is there code in this module that needs to be used outside of visualization. If so, it implies that we need to break this module up and keep the matplotlib-dependent stuff in visualization.

src/ansys/fluent/core/graphics_definitions/pyvista/pyvista_windows_manager.py
```diff
@@ -0,0 +1,694 @@
+"""Module for pyVista windows management."""
+import itertools
+import threading
+from typing import List, Optional, Union
+
+import numpy as np
+import pyvista as pv
```
seanpearsonuk, line 7: similar point for pyvista as regarding matplotlib

prmukherj:
@seanpearsonuk, @ajain-work and @hpohekar -> closing this PR as this is no longer relevant. Please confirm once.
Thank you.

PR #1151, opened 2022-12-05
Feat/field allowed

seanpearsonuk:


src/ansys/fluent/core/fluent_connection.py
```diff
@@ -62,26 +62,14 @@ def run(self) -> None:
 
 
 class _IsDataValid:
-    def __init__(self, events_manager):
```
seanpearsonuk, line 65: https://github.com/pyansys/pyfluent/issues/1156

mkundu1 (approved):

prmukherj (approved):

PR #1133, opened 2022-12-01
MAINT: migrate tools configuration to pyproject.toml

jorgepiloto:
Fix #1130 by declaring tools configuration inside the `pyproject.toml` file.

- [x] Migrate black
- [x] Migrate isort
- [x] Migrate codespell
- [x] Migrate pylint
- [x] Migrate docformatter

PR #1152, opened 2022-12-06
Update .gitignore, Remove .pylintrc

dnwillia-work:
Looks like the autogenerated documentation is coming out in different directories now, so ignore those.

Pylintrc is not really use since we use the flake8 linting and black


jorgepiloto (approved):

jorgepiloto:
I will adapt #1133 to this PR.

seanpearsonuk (approved):

Issue #1149, opened 2022-12-05
API to return number of iterations defined in case file


tejalprabhu01:
### 📝 Description of the feature

I would like to request for an API that returns number of iterations defined in the case file. 
This is required to write a custom journal script by optiSLang that calls iterate command. This command requires number of iterations as argument. We would like to give the default number of iterations as argument here. 

@seanpearsonuk @mkundu1

seanpearsonuk:
Implemented as iter_count()

Issue #1148, opened 2022-12-05
API to get value of parameter using CaseReader


tejalprabhu01:
### 📝 Description of the feature

It would be good to have an API to get value and unit separately from the CaseReader. Currently spit(" ") needs to be called to split value and unit, which leads to errors when there are spaces in the unit.
There is already a way to get units for Output parameters. I would like to request a similar API for InputParamters
For e.g. 
```
>>> reader = CaseReader(case_filepath=r"D:\Release23R2\Fluent\Unitless\StaticMixer2.cas.h5")
>>> for p in reader.input_parameters():
...    print(p.name)
...    print(p.value)
```
inlet_main_v
5 [m s^-1]

@seanpearsonuk @mkundu1

PR #1154, opened 2022-12-06
case reader updates

seanpearsonuk:
Resolves a couple of requests posted via issues for optiSLang integration
#1148 
#1149


tests/test_casereader.py
```diff
@@ -15,9 +19,9 @@ def call_casereader(
     if expected is not None:
         assert reader.precision() == expected["precision"]
         assert reader.num_dimensions() == expected["num_dimensions"]
-        assert {p.name: p.value for p in reader.input_parameters()} == expected[
-            "input_parameters"
-        ]
+        assert {
+            p.name: (p.number, p.units) for p in reader.input_parameters()
```
seanpearsonuk, line 23: @tejalprabhu01 one needs to access `number` and `units` attributes of the input parameter object. `value` is retaining its original meaning (it is used as such in the pyfluent-parametric library).

prmukherj (approved):

tejalprabhu01 (approved):
Thank you for the APIs @seanpearsonuk

PR #1168, opened 2022-12-07
Fix for start_instance argument and some missing changes from 232 CI PR

mkundu1:
1. Fix for `start_instance` argument which fixes the issue with `server_info_filepath` argument.
2. Some changes I forgot to commit with the 23.2 CI PR.

### DIFF
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -377,10 +377,10 @@ def _get_running_session_mode(
     return session_mode.value[1]
 
 
-def _start_instance(start_instance: bool = None):
+def _start_instance(start_instance: Union[bool, None]):
     """Sets up how to start an instance of fluent."""
-    if not start_instance:
-        start_instance = bool(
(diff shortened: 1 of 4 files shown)


src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -377,10 +377,10 @@ def _get_running_session_mode(
     return session_mode.value[1]
 
 
-def _start_instance(start_instance: bool = None):
+def _start_instance(start_instance: Union[bool, None]):
     """Sets up how to start an instance of fluent."""
-    if not start_instance:
```
mkundu1, line 382: If `start_instance=False` is passed, the previous code was setting it to True. We should try to evaluate `start_instance` from environment only if it is not passed through `launch_fluent`.

This fixes the issue with `server_info_filename` argument. Following will now work:
```
>>> solver = pyfluent.launch_fluent(start_instance=False, server_info_filepath=<server_info_filepath>)
```

hpohekar (approved):

PR #1180, opened 2022-12-12
Get fluent version using string_eval

hpohekar:
We have made `self.version` as a data member and assigned `string_eval` it to get fluent version.

### DIFF
--- a/src/ansys/fluent/core/fluent_connection.py
+++ b/src/ansys/fluent/core/fluent_connection.py
@@ -285,7 +285,7 @@ class _FluentConnection:
 
     def get_fluent_version(self):
         """Gets and returns the fluent version."""
-        return ".".join(map(str, self.scheme_eval.scheme_eval("(cx-version)")))
+        return self.scheme_eval.string_eval("(cx-version)")
 
     def exit(self) -> None:
         """Close the Fluent connection and exit Fluent."""
--- a/src/ansys/fluent/core/fluent_connection.py
+++ b/src/ansys/fluent/core/fluent_connection.py
@@ -214,10 +214,10 @@ class _FluentConnection:
         )
 
         self.journal = Journal(self.scheme_eval)
+        self.version = self.scheme_eval.string_eval("(cx-version)")
 
         try:
-            version = self.scheme_eval.string_eval("(cx-version)")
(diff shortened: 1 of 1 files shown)


prmukherj (approved):

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -285,7 +285,7 @@ def check_health(self) -> str:
 
     def get_fluent_version(self):
         """Gets and returns the fluent version."""
-        return ".".join(map(str, self.scheme_eval.scheme_eval("(cx-version)")))
+        return self.version
```
seanpearsonuk, line 288: Looks like this is now going to return '(23 2 0)' rather than '23.2.0'. To rectify this, see above.

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -214,10 +214,10 @@ def __init__(
         )
 
         self.journal = Journal(self.scheme_eval)
+        self.version = self.scheme_eval.string_eval("(cx-version)")
```
seanpearsonuk, line 217: Let's process this right away. Essentially we need to apply ".".join(version.strip("()").split()). i.e. everything except the slice.

seanpearsonuk, line 217: Has this been pulled out of the try block deliberately?

mkundu1:
@hpohekar I've incorporated the fix in #1182, you may close this PR.

src/ansys/fluent/core/fluent_connection.py
hpohekar: > @hpohekar I've incorporated the fix in #1182, you may close this PR.

@mkundu1 Yes, sure. Thank you.

hpohekar: > Has this been pulled out of the try block deliberately?

@seanpearsonuk  Yes, but later I saw your message in teams that we want to keep current behavior. So this change is not useful now.

src/ansys/fluent/core/fluent_connection.py
hpohekar: Yes, we have closed this PR now.

PR #1182, opened 2022-12-12
Fix scheme_eval call

mkundu1:


src/ansys/fluent/core/services/scheme_eval.py
```diff
@@ -213,9 +213,13 @@ class SchemeEval:
         value
     """
 
-    def __init__(self, service: SchemeEvalService, version=None):
+    def __init__(self, service: SchemeEvalService):
         self.service = service
-        self.version = version
+        try:
+            version = self.string_eval("(cx-version)")
```
seanpearsonuk, line 219: The part that @hpohekar will work on (or yourself?) is to consolidate the the remote calls to evaluate cx-version. (N.b we are still calling it in fluent_connection.py.)

seanpearsonuk (approved):

seanpearsonuk (approved):

prmukherj (approved):

hpohekar (approved):

PR #1194, opened 2022-12-14
Added test for topy

hpohekar:
Added basic test for `topy`.

Refactored code in `launcher.py`

### DIFF
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -567,22 +567,18 @@ def launch_fluent(
                 kwargs.update(cwd=cwd)
             if topy:
                 if isinstance(topy, str):
-                    journal_name = topy.split(".")[0]
-                    journal_ext = topy.split(".")[1]
-                    launch_string += f' -i {journal_name}.{journal_ext} -command="(api-start-python-journal \\\"\\\"{journal_name}.py\\\"\\\")"'  # noqa: E501
+                    name = topy.split(".")[0]
+                    ext = topy.split(".")[1]
+                    launch_string += f' -i {name}.{ext} -command="(api-start-python-journal \\\"\\\"{name}.py\\\"\\\")"'  # noqa: E501
                 if isinstance(topy, list):
                     all_scm_journal_names = ""
                     all_py_journal_names = ""
                     name_list = []
                     for journals in topy:
-                        if os.path.exists(journals):
-                            journal_name = journals.rsplit("\\")[-1].split(".")[0]
-                            journal_ext = journals.rsplit("\\")[-1].split(".")[1]
-                        else:
-                            journal_name = journals.rsplit(".")[0]
-                            journal_ext = journals.rsplit(".")[1]
-                        name_list.append(journal_name)
-                        all_scm_journal_names += f' -i {journal_name}.{journal_ext}'
+                        name = Path(journals).stem.split('.')[0]
+                        ext = journals.rsplit(".")[1]
+                        name_list.append(name)
+                        all_scm_journal_names += f' -i {name}.{ext}'
                     all_py_journal_names += '_'.join(name_list)
                     launch_string += f'{all_scm_journal_names} -command="(api-start-python-journal \\\"\\\"{all_py_journal_names}.py\\\"\\\")"'  # noqa: E501
             subprocess.Popen(launch_string, **kwargs)
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -575,8 +575,12 @@ def launch_fluent(
                     all_py_journal_names = ""
                     name_list = []
                     for journals in topy:
-                        journal_name = journals.rsplit(".")[0]
-                        journal_ext = journals.rsplit(".")[1]
+                        if os.path.exists(journals):
+                            journal_name = journals.rsplit("\\")[-1].split(".")[0]
+                            journal_ext = journals.rsplit("\\")[-1].split(".")[1]
+                        else:
+                            journal_name = journals.rsplit(".")[0]
+                            journal_ext = journals.rsplit(".")[1]
                         name_list.append(journal_name)
                         all_scm_journal_names += f' -i {journal_name}.{journal_ext}'
                     all_py_journal_names += '_'.join(name_list)
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -421,6 +421,18 @@ def _generate_launch_string(
     return launch_string
 
 
+def scm_to_py(topy):
+    if not isinstance(topy, (str, list)):
+        raise TypeError("Journal name should be of str or list type.")
+    launch_string = ""
+    if isinstance(topy, str):
+        topy = [topy]
+    fluent_jou_arg = "".join([f'-i "{journal}" ' for journal in topy])
+    py_jou = "_".join([Path(journal).stem for journal in topy])
+    launch_string += f' {fluent_jou_arg} -command="(api-start-python-journal \\\"\\\"{py_jou}.py\\\"\\\")"'  # noqa: E501
+    return launch_string
+
+
 #   pylint: disable=unused-argument
 def launch_fluent(
     product_version: str = None,
@@ -566,12 +578,8 @@ def launch_fluent(
             if cwd:
                 kwargs.update(cwd=cwd)
             if topy:
-                if isinstance(topy, str):
-                    topy = [topy]
-                if isinstance(topy, list):
-                    fluent_jou_arg = "".join([f'-i "{journal}" ' for journal in topy])
-                    py_jou = "_".join([Path(journal).stem for journal in topy])
-                    launch_string += f' {fluent_jou_arg} -command="(api-start-python-journal \\\"\\\"{py_jou}.py\\\"\\\")"'  # noqa: E501
+                launch_string += scm_to_py(topy)
+
             subprocess.Popen(launch_string, **kwargs)
 
             _await_fluent_launch(server_info_filepath, start_timeout, sifile_last_mtime)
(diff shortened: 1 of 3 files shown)


prmukherj (approved):

tests/test_topy.py
```diff
@@ -0,0 +1,99 @@
+import os
+
+import ansys.fluent.core as pyfluent
+
+
+def test_single_jou():
+    file_path = os.path.join(pyfluent.EXAMPLES_PATH, "jou1.jou")
```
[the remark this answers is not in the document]
hpohekar: Done. Thank you.

tests/test_session.py
```diff
@@ -233,7 +233,13 @@ def test_create_session_from_launch_fluent_by_setting_ip_and_port_env_var(
 @pytest.mark.fluent_231
 def test_execute_tui_commands(new_mesh_session, tmp_path=pyfluent.EXAMPLES_PATH):
     session = new_mesh_session
-    file_path = os.path.join(tmp_path, "sample_py_journal.txt")
+
```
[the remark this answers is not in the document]
hpohekar: You're welcome. @prmukherj

tests/test_topy.py
```diff
@@ -0,0 +1,136 @@
+import os
+import tempfile
+
+import ansys.fluent.core as pyfluent
+
+
+def test_single_jou(with_launching_container):
+
```
seanpearsonuk, line 8: @hpohekar are we asserting that the Python file has been created?

hpohekar: @seanpearsonuk Yes, we are reading generated file contents by `returned = f.readlines()` and by using `assert returned` we are confirming that `returned` is non empty.

hpohekar: @seanpearsonuk Yes, perfect catch! Now we are asserting that Python file has been generated for 1st and 2nd test.

We can not assert this for multiple journals because then it gives `invalid_argument error` for `os` as it can not read the generated file path due to concatenation of 2 file paths together.

hpohekar: We are asserting that the Python file has been created for single journal name. Now we are resolving issues related to journal names as list of strings. We will update this code accordingly soon.

hpohekar: We have updated the code now.

seanpearsonuk: Thanks @hpohekar I was looking at a version that didn't have the Python file checks - this looks good now :)

tests/test_topy.py
```diff
@@ -0,0 +1,168 @@
...
+    assert returned
+
+    gen_file_path = f'{file_path.split(".")[0]}.py'
+
+    with open(gen_file_path) as file:
+        gen_returned = file.readlines()
+
+    assert gen_returned
+
+    while os.path.exists(gen_file_path):
```
seanpearsonuk, line 32: We should always make such loops finite even when we know that it's not necessary.

hpohekar: Yes, sure. Thank you.

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -575,8 +575,12 @@ def launch_fluent(
                     all_py_journal_names = ""
                     name_list = []
                     for journals in topy:
-                        journal_name = journals.rsplit(".")[0]
-                        journal_ext = journals.rsplit(".")[1]
+                        if os.path.exists(journals):
+                            journal_name = journals.rsplit("\\")[-1].split(".")[0]
```
seanpearsonuk, line 579: @hpohekar 

The "\\\\" seems like a WIndows assumption. Shall we use something designed for the job that works with any os, e.g. os.path.split()? 

Also these two lines are unnecessarily duplicated - the rhs of a single statement can just be `name, ext = `.

Also there's no need for the if/else here. It works just the same.

hpohekar: Ok. We will update this as well.

seanpearsonuk: (I was still editing the comment - please check all points and let me know if anything is unclear)

hpohekar: > (I was still editing the comment - please check all points and let me know if anything is unclear)

Yes, sure. Thank you.

hpohekar: > [Path(<path>).stem](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.stem) can also be used here. We probably don't need to extract the extension here.

I see. Thank you.

tests/test_topy.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/launcher/launcher.py

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -567,18 +567,18 @@ def launch_fluent(
                 kwargs.update(cwd=cwd)
             if topy:
                 if isinstance(topy, str):
-                    journal_name = topy.split(".")[0]
-                    journal_ext = topy.split(".")[1]
-                    launch_string += f' -i {journal_name}.{journal_ext} -command="(api-start-python-journal \\\"\\\"{journal_name}.py\\\"\\\")"'  # noqa: E501
+                    name = topy.split(".")[0]
+                    ext = topy.split(".")[1]
```
seanpearsonuk, line 571: or something like:
```
name, ext = topy.split(".")[:2]
```
because of the duplication
?

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -567,18 +567,18 @@ def launch_fluent(
...
                 if isinstance(topy, list):
                     all_scm_journal_names = ""
                     all_py_journal_names = ""
                     name_list = []
                     for journals in topy:
-                        journal_name = journals.rsplit(".")[0]
-                        journal_ext = journals.rsplit(".")[1]
-                        name_list.append(journal_name)
-                        all_scm_journal_names += f' -i {journal_name}.{journal_ext}'
+                        name = Path(journals).stem.split('.')[0]
```
seanpearsonuk, line 578: Why is _journals_ pluralised?

hpohekar: Because it's list of journals.

seanpearsonuk, line 577: Why are we splitting the stem?

src/ansys/fluent/core/launcher/launcher.py

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -567,18 +567,18 @@ def launch_fluent(
...
+                        ext = journals.rsplit(".")[1]
```
seanpearsonuk, line 578: Why a separate method for the extension.

Also this code should be consolidated with the earlier code. All can be done in a single way in one function..

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -567,18 +567,18 @@ def launch_fluent(
                 kwargs.update(cwd=cwd)
             if topy:
                 if isinstance(topy, str):
-                    journal_name = topy.split(".")[0]
-                    journal_ext = topy.split(".")[1]
-                    launch_string += f' -i {journal_name}.{journal_ext} -command="(api-start-python-journal \\\"\\\"{journal_name}.py\\\"\\\")"'  # noqa: E501
+                    name = topy.split(".")[0]
+                    ext = topy.split(".")[1]
+                    launch_string += f' -i {name}.{ext} -command="(api-start-python-journal \\\"\\\"{name}.py\\\"\\\")"'  # noqa: E501
                 if isinstance(topy, list):
```
seanpearsonuk, line 572: do you mean elif?

src/ansys/fluent/core/launcher/launcher.py
seanpearsonuk: `topy` is the list that we are iterating over. journals is one element in that list - a string, which is the name or path of a journal.

src/ansys/fluent/core/launcher/launcher.py
hpohekar: `Path.stem` does not give file name with extension.

src/ansys/fluent/core/launcher/launcher.py
hpohekar: Yes. Done. Thank you.

src/ansys/fluent/core/launcher/launcher.py
hpohekar: Removed plurals. Thank you.

hpohekar: Removed splitting. Thank you.

hpohekar:
Thank you very much @mkundu1 for refactoring this code.

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -567,20 +567,11 @@ def launch_fluent(
...
-                    name_list = []
-                    for journals in topy:
-                        journal_name = journals.rsplit(".")[0]
-                        journal_ext = journals.rsplit(".")[1]
-                        name_list.append(journal_name)
-                        all_scm_journal_names += f' -i {journal_name}.{journal_ext}'
-                    all_py_journal_names += '_'.join(name_list)
-                    launch_string += f'{all_scm_journal_names} -command="(api-start-python-journal \\\"\\\"{all_py_journal_names}.py\\\"\\\")"'  # noqa: E501
+                    topy = [topy]
+                elif isinstance(topy, list):
```
mkundu1, line 571: This should be an if, otherwise the code won't work when the passed topy is a string.

hpohekar:
@mkundu1 We have updated the `topy` argument code using a function and added type check in function.

We have also tested following scenarios (Journal name and path containing space etc.) locally and all of them are working fine.

image

mkundu1 (approved):

PR #1197, opened 2022-12-14
FIX: deploy dev docs using desired image

jorgepiloto:
This pull-request adds a conditional to deploy development documentation when using a particular image in the nighly-build, since HTML doc artifacts are generated only for v23.1.0.

### DIFF
--- a/.github/workflows/nightly-doc-build.yml
+++ b/.github/workflows/nightly-doc-build.yml
@@ -80,6 +80,7 @@ jobs:
       - name: "Deploy development documentation"
         if: matrix.image-tag == env.DOC_DEPLOYMENT_IMAGE_TAG
         uses: pyansys/actions/doc-deploy-dev@v2
+        if: matrix.image-tag == 'v23.1.0'
         with:
             cname: ${{ env.DOCUMENTATION_CNAME }}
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-doc-build.yml
```diff
@@ -77,6 +77,7 @@ jobs:
 
       - name: "Deploy development documentation"
         uses: pyansys/actions/doc-deploy-dev@v2
+        if: matrix.image-tag == 'v23.1.0'
```
dnwillia-work, line 80: @mkundu1 @seanpearsonuk We should only be deploying documentation to built against v22.2 right.  i.e. the doc here:  

https://dev.fluent.docs.pyansys.com/

should be only generated with v22.2 because that's all customers have right.

dnwillia-work: So, I think this should say `if matrix.image-tag == 'v22.2.0'`

mkundu1: I have added the check against 22.2 in #1199

jorgepiloto:
Closing this in favor of #1199. Thanks for working on this, @mkundu1 👍🏽

PR #1199, opened 2022-12-14
Deploy doc for 22.2

mkundu1:
We should deploy docs for Fluent 22.2 version as it is currently available to customers.


seanpearsonuk (approved):

dnwillia-work:
Looks like maybe the doc build failed because one of the tests is expecting 23.1 settings API (if I read it right).

hpohekar (approved):

PR #1208, opened 2022-12-15
Consistent Fluent Version format.

prmukherj:
There are some inconsistencies how Fluent version is formatted at various places in the pyfluent code:

get_fluent_version() returns string "x.y.z"
launch_fluent accepts string "x.y"
launch_remote_fluent accepts string "xy"
set_ansys_version accepts string "x.y", float x.y, or an enum.

This has been addressed as part of this PR.

**Please Note: This PR work is still in progress and it would be helpful to get your views on this. Thank you.**

Current modification handles both cases even when the user enters "x.y.0" or "x.y".


src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -69,11 +61,18 @@ def set_ansys_version(version: Union[str, float, FluentVersion]) -> None:
...
-        FLUENT_VERSION[0] = version.value
-        PIM_FLUENT_PRODUCT_VERSION[0] = FLUENT_VERSION[0].replace(".", "")
+    if isinstance(version, (float, str)):
+        version = str(version)
+    elif isinstance(version, FluentVersion):
+        version = version.value
+
+    if len(version.split(".")) == 2:
+        version += ".0"
+    if version not in _SUPPORTED_FLUENT_VERSIONS:
```
seanpearsonuk, line 71: Note that the inability to instantiate a `FluentVersion` indicates the same. i.e. `FluentVersion(version)` raises a ValueError if `version` is invalid. That means that we can eliminate `_SUPPORTED_FLUENT_VERSIONS`. For printing the values, this is the list: [member.value for member in FluentVersion]

prmukherj: Thank you @seanpearsonuk, sounds good.

prmukherj: Done.

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -536,6 +545,10 @@ def launch_fluent(
     machines and core counts are queried from the scheduler environment and
     passed to Fluent.
     """
+    if "meshing_mode" in kwargs.keys():
```
prmukherj, line 548: For this check a new argument of **kwargs has been introduced. One concern here is that the user can pass anything as kwargs, which will/might raise exceptions down the line, during fluent launch. There can be another technique here, though not as clean as the above technique. That is to accept the "meshing_mode" argument as it is and raise the error if the "meshing_mode" is not none, and delete the meshing mode from the list of arguments while displaying? Please let me know your thoughts on this, Thank you.

mkundu1 (approved):

hpohekar (approved):

src/ansys/fluent/core/launcher/launcher.py
prmukherj: Thank you @mkundu1 for your inputs on this. I am resolving this conversation and merging the PR.

PR #1210, opened 2022-12-16
FIX: nigthly build syntax

jorgepiloto:
As per the action logs, https://github.com/pyansys/pyfluent/actions/runs/3710172176/workflow, a syntax error was introduced in #1199. This pull-request fixes the YML file controlling the nightly build workflow.

### DIFF
--- a/.github/workflows/nightly-doc-build.yml
+++ b/.github/workflows/nightly-doc-build.yml
@@ -7,6 +7,7 @@ on:
 
 env:
   DOCUMENTATION_CNAME: 'fluent.docs.pyansys.com'
+  DOC_DEPLOYMENT_IMAGE_TAG: v22.2.0
 
 jobs:
   nightly_docs_build:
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-doc-build.yml
```diff
@@ -7,6 +7,7 @@ on:
 
 env:
   DOCUMENTATION_CNAME: 'fluent.docs.pyansys.com'
```
dnwillia-work, line 9: Jorge, I guess this CNAME is working with the versioned documentation right?  Previously we would have deployed it to `dev.fluent.docs.pyansys.com` and now it would be hosted at this CNAME instead.

dnwillia-work (approved):

.github/workflows/nightly-doc-build.yml
jorgepiloto: That's right, @dnwillia-work. Once the migration to multi-documentation is complete, everything is hosted under the same domain, in this case, "fluent.docs.pyansys.com".

mkundu1 (approved):

PR #1215, opened 2022-12-22
Unit Quantity Attribute Test

hpohekar:
Settings API currently provides `units-quantity` attributes for some real parameters (under `materials` and `boundary_conditions`). These are string format, so we are testing our `Quantity` code against this since "velocity" is equivalent to {"velocity":1}.

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -735,6 +735,7 @@ class Command(Action):
         return self.flproxy.execute_cmd(self._parent.path, self.obj_name, **newkwds)
 
     def _get_value_quantity_map(self, **kwds):
+        """Value and quantity_map of units-quantity."""
         newkwds = _get_new_keywords(self, kwds)
         results_list = self.flproxy.execute_cmd(self._parent.path, self.obj_name, **newkwds)
         unit_quantity_map = {}
@@ -748,10 +749,12 @@ class Command(Action):
         return [value, unit_quantity_map]
 
     def value(self, **kwds):
+        """Value of units-quantity."""
         value_map = Command._get_value_quantity_map(self, **kwds)
         return value_map[0]
 
     def quantity_map(self, **kwds):
+        """Quantity map of units-quantity."""
         value_map = Command._get_value_quantity_map(self, **kwds)
         return value_map[1]
 
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -296,8 +296,7 @@ class Real(SettingsBase[RealType], Numerical):
         """Get the units-quantity of the object."""
         val = self.get_attr("units-quantity", str)
         if isinstance(val, str):
-            val = val.replace("'", '')
-            val = ' '.join(word.capitalize() for word in val.split())
+            val = ' '.join(word.capitalize() for word in val.lstrip("'").split("-"))
         return {val: 1.0} if isinstance(val, str) else val
 
     _state_type = RealType
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -292,6 +292,19 @@ class Real(SettingsBase[RealType], Numerical):
     expression values.
     """
 
+    def value(self):
+        """Get the value of the object."""
+        val = self.get_attr("value", (float, int))
+        return val
+
+    def quantity_map(self) -> Dict[str, float]:
+        """Get the units-quantity of the object."""
+        val = self.get_attr("units-quantity", str)
+        if isinstance(val, str):
+            val = val.replace("'", '')
+            val = ' '.join(word.capitalize() for word in val.split())
+        return {val: 1.0} if isinstance(val, str) else val
+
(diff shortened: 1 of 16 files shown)


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -292,6 +292,19 @@ class Real(SettingsBase[RealType], Numerical):
...
+        """Get the value of the object."""
+        val = self.get_attr("value", (float, int))
+        return val
+
+    def quantity_map(self) -> Dict[str, float]:
+        """Get the units-quantity of the object."""
+        val = self.get_attr("units-quantity", str)
+        if isinstance(val, str):
+            val = val.replace("'", '')
+            val = ' '.join(word.capitalize() for word in val.split())
```
mkundu1, line 305: Do we need to convert a symbol-string to PascalCase here (`"'heat-flux"` -> `"Heat Flux"`)?

`val = ' '.join(x.capitalize() for x in val.lstrip("'").split("-"))`

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -292,6 +292,19 @@ class Real(SettingsBase[RealType], Numerical):
     expression values.
     """
 
+    def value(self):
+        """Get the value of the object."""
+        val = self.get_attr("value", (float, int))
```
mkundu1, line 297: I don't think we should evaluate the `value` attribute directly. `__call__()` or `get_state` will return this?

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -727,6 +740,27 @@ def __call__(self, **kwds):
...
+                quantity = ' '.join(word.capitalize() for word in quantity.split())
+                value = real[0]
+                unit_quantity_map[quantity] = 1.0
+        return [value, unit_quantity_map]
+
+    def value(self, **kwds):
+        value_map = Command._get_value_quantity_map(self, **kwds)
+        return value_map[0]
+
+    def quantity_map(self, **kwds):
```
mkundu1, line 760: It is probably more useful to have this method on the returned object of command/query execution rather than on the command/query object itself.

src/ansys/fluent/core/solver/flobject.py
hpohekar: We have used PascalCase in `quantity_map_method` because it's also used in CFX units library.

From `quantity_map_method` -

image

From CFX - 

image

mkundu1: Yeah, I got that. But the current code doesn't split `"heat-flux"` by `"-"`.

hpohekar: Right. Got it. It's corrected now. Thank you very much.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -727,6 +740,27 @@ def __call__(self, **kwds):
         newkwds = _get_new_keywords(self, kwds)
         return self.flproxy.execute_cmd(self._parent.path, self.obj_name, **newkwds)
 
+    def _get_value_quantity_map(self, **kwds):
+        newkwds = _get_new_keywords(self, kwds)
+        results_list = self.flproxy.execute_cmd(self._parent.path, self.obj_name, **newkwds)
```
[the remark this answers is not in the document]

src/ansys/fluent/core/solver/flobject.py
hpohekar: Yes, we will discuss this soon.

hpohekar:
@h-krishnan Could you please elaborate how we want to associate a Unit Quantity Map ( { Unit Quantity : power }) with every real value solver setting or expression/report definition value.

src/ansys/fluent/core/solver/flobject.py
seanpearsonuk: @hpohekar it would be good to resolve this issue.

hpohekar: @seanpearsonuk @mkundu1 We already tried this for `report_definitions`. The `report_definitions` returns `list` type of object, then we will have to modify default built-in python data type to add `quantity_map` as an attribute to `list` type.

image


image

Similar to this we will have to implement for other data types.

hpohekar: @seanpearsonuk @mkundu1  but as per our latest discussion we do not want to depend on return data type of commands and query.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -292,6 +292,13 @@ class Real(SettingsBase[RealType], Numerical):
     expression values.
     """
 
+    def quantity_map(self) -> Dict[str, float]:
+        """Get the units-quantity of the object."""
+        val = self.get_attr("units-quantity", str)
+        if isinstance(val, str):
+            val = ' '.join(word.capitalize() for word in val.lstrip("'").split("-"))
```
seanpearsonuk, line 299: @hpohekar we won't merge the version with capitalize. We are using a different set of names from the settings API. Either we conform to those names (either directly or via some adapter) or the settings API provides the names in our (Ansys) format.

PR #1225, opened 2023-01-03
Fluent version markers

mkundu1:
This will fix the issue in running tests with some Fluent versions.

### DIFF
--- a/.github/workflows/nightly-test-run.yml
+++ b/.github/workflows/nightly-test-run.yml
@@ -112,24 +112,18 @@ jobs:
           PYFLUENT_START_INSTANCE: 0
           FLUENT_IMAGE_TAG: v22.2.0
 
-      - name: Upload 22.2 Coverage Results to Codecov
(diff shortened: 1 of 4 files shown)


hpohekar (approved):

seanpearsonuk (approved):

prmukherj:
@mkundu1, if it is not very urgent can we merge this PR after the release? Actually, it might affect the existing tests and we might need to change some markers as well.
Thank you.

mkundu1:
> @mkundu1, if it is not very urgent can we merge this PR after the release? Actually, it might affect the existing tests and we might need to change some markers as well. Thank you.

Yes sure. Note that it won't require changing any existing markers unless some tests are wrongly marked, but there is no urgency to check that before the release.

mkundu1:
https://github.com/pyansys/pyfluent/actions/runs/3906173342/jobs/6674043923

Nightly test run is passed.

PR #1235, opened 2023-01-05
FieldMethod infrastructure help

hpohekar:
Resolved interference of `FieldMethod` infrastructure with `help().`

image

`------------------------------------------------------------------------------------------------------------------------`

image

### DIFF
--- a/src/ansys/fluent/core/services/field_data.py
+++ b/src/ansys/fluent/core/services/field_data.py
@@ -2,6 +2,7 @@
 import difflib
 from enum import IntEnum
 from functools import partial, reduce
+import pydoc
 from typing import Callable, Dict, List, Optional, Tuple, Union
 
 import grpc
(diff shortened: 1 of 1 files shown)


src/ansys/fluent/core/services/field_data.py
```diff
@@ -12,6 +13,13 @@
 from ansys.fluent.core.services.error_handler import catch_grpc_error
 from ansys.fluent.core.services.interceptors import TracingInterceptor
 
+
+def override_help_text(func, func_to_be_wrapped):
+    func.__doc__ = "\n" + pydoc.text.document(func_to_be_wrapped)
```
mkundu1, line 18: Is it possible to set this within _FieldMethod.__init__?

seanpearsonuk: @mkundu1 That was what we were trying to do initially, and could not get it to update, as it is read-only. Whether or not such a change can work within the current code, perhaps we could alternatively still streamline this code a bit by adding one more level of indirection.

mkundu1 (approved):

prmukherj (approved):

seanpearsonuk (approved):

src/ansys/fluent/core/services/field_data.py
hpohekar: Thank you.

PR #1254, opened 2023-01-11
Implementing pre-commit.ci

### LINKED ISSUE #1253: Implementing pre-commit.ci
Hi @seanpearsonuk @dnwillia!

We have been recently implementing a tool called ``pre-commit.ci`` in several PyAnsys repositories. This tool is characterized for the following:

* It allows us to "get rid" of the pre-commit stage of our own workflows in favor of using a third-party tool
* This tool then runs pre-commit on its own cached resources, allowing for faster run times of pre-commit
* If the tool detects a pre-commit failure, it creates a commit to try and fix the issue encountered (this only occurs if the pre-commit hook that failed is able to reformat)
* It maintains the pre-commit hook versions updated (which by the way, dependabot does not).
* It also provides the repository with a new badge for its main README informing of the status of pre-commit on its latest run.

PyMAPDL, PyAEDT and the PyAnsys metapackage, among others, have already migrated to this tool. If you agree as well on its use, I'd be happy to implement it on your repositories:

* PyFluent
* PyFluentVisualization
* PyFluentParametric

RobPasMue:
Closes #1253

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,26 +26,6 @@ env:
   DOCUMENTATION_CNAME: 'fluent.docs.pyansys.com'
 
 jobs:
-  stylecheck:
-    name: Style Check
-    runs-on: ubuntu-latest
-
-    steps:
-      - uses: actions/checkout@v3
-
-      - name: Set up Python
-        uses: actions/setup-python@v4
-        with:
-          python-version: 3.9
-
-      - name: Install pre-commit
-        run: |
-          python -m pip install pre-commit
-
-      - name: Run pre-commit
-        if: always()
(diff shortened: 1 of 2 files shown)


dnwillia-work:
@RobPasMue  Looks like maybe there is a corrupt Python environment on the runner?

RobPasMue:
> @RobPasMue Looks like maybe there is a corrupt Python environment on the runner?

Yes, I talked with @prmukherj about this some hours ago. I recommend cleaning up the runners cache and rebooting the machine if possible. Have you got access to the machine?

prmukherj (approved):

mkundu1 (approved):

PR #1273, opened 2023-01-16
Fixed Built Doc Files from zipping.

alikasim72:
As the upload-artifact action already zips the file, the prior manual compression in the file is unnecessary and leads to "double compression" which is only uncompressed once during deployment stage. Hence, the lines of manual compression were omitted to fix this issue.


mkundu1:
Uploading the doc artifact now takes [13 min](https://github.com/pyansys/pyfluent/actions/runs/3933547412/jobs/6727373010#step:15:20) instead of [35 sec](https://github.com/pyansys/pyfluent/actions/runs/3906690005/jobs/6675139815#step:16:20). Not zipping the doc artifact fixes the deployment, so we can zip it when we don't need to deploy.

jorgepiloto:
Pinging here @MaxJPRey. I know PyAEDT run into a similar issue at some point but after checking their CI/CD pipelines it looks like they are no longer compressing the artifact.

I guess we could add an input parameter in pyansys actions to unzip the different artifacts.

MaxJPRey:
> Pinging here @MaxJPRey. I know PyAEDT run into a similar issue at some point but after checking their CI/CD pipelines it looks like they are no longer compressing the artifact.
> 
> I guess we could add an input parameter in pyansys actions to unzip the different artifacts.

Agreed to expose the argument to unzip the content.

jorgepiloto:
I just opened https://github.com/pyansys/actions/issues/143 to implement this.

jorgepiloto:
Now that the decompress logic got added in the pyansys/actions, could you use the `decompress_artifact` input variable, @alikasim72?

alikasim72:
Updated the code by adding back the compression code lines and setting the `decompress_artifact` to True.

mkundu1 (approved):

.github/workflows/ci.yml
```diff
@@ -192,14 +192,15 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: HTML-Documentation-tag-${{ env.DOC_DEPLOYMENT_IMAGE_TAG }}
-          path: doc.zip
+          path: doc.zip
```
dnwillia-work, line 195: maybe remove whitespace?

jorgepiloto, line 195:
```suggestion
          path: doc.zip
```

jorgepiloto (changes requested):

jorgepiloto (approved):

PR #1280, opened 2023-01-17
Fluent exe path order

mkundu1:


src/ansys/fluent/core/__init__.py
```diff
@@ -9,10 +9,9 @@
 from ansys.fluent.core.launcher.launcher import (  # noqa: F401
     FluentVersion,
     LaunchModes,
-    get_ansys_version,
```
mkundu1, line 12: I have exposed only the setters from the module. We may not know ansys_version in some situations before actually launching Fluent, e.g., when `set_fluent_exe_path()` is set. After Fluent is launched, user should use `solver.get_fluent_version()` to know the version.

prmukherj (approved):

hpohekar (approved):

seanpearsonuk (approved):

PR #1293, opened 2023-01-20
Format lists within pydoc

mkundu1:
2023-01-20 11_09_40-Untitled2 ipynb - JupyterLab

### DIFF
--- a/src/ansys/fluent/core/utils/fldoc.py
+++ b/src/ansys/fluent/core/utils/fldoc.py
@@ -1,0 +1,28 @@
+import pprint
+import pydoc
+import sys
+
+
+def docother(self, object, name=None, mod=None, parent=None, maxlen=None, doc=None):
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/utils/fldoc.py
```diff
@@ -0,0 +1,28 @@
+import pprint
+import pydoc
+import sys
+
+
+def docother(self, object, name=None, mod=None, parent=None, maxlen=None, doc=None):
+    """Produce text documentation for a data object."""
+    if isinstance(object, list):
```
mkundu1, line 8: This special case is the fix - exactly as the pyconsole fix. The else part is a copy from cpython source.

prmukherj (approved):

hpohekar (approved):

seanpearsonuk (approved):

PR #1301, opened 2023-01-24
Improve Error message when launch fluent fails.

prmukherj:
When launch_fluent fails, we should notify the launch path and launcher arguments with which Fluent was attempted to launch.


prmukherj:
image
Sample output.

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -564,6 +564,11 @@ def launch_fluent(
             return new_session.create_from_server_info_file(
                 server_info_filepath, cleanup_on_exit, start_transcript
             )
+        except Exception as ex:
+            identifier = "fluent.exe"
+            print("Fluent Launch Path: ", launch_string.split(identifier)[0] + identifier)
```
seanpearsonuk, line 569: rather than printing, we can add this info to the exception, which guarantees it gets propagated in any context

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -564,6 +564,11 @@ def launch_fluent(
...
+            print("Launcher Arguments: ", launch_string.split(identifier)[1].split())
+            raise Exception(ex)
```
seanpearsonuk, line 571: Is it idiomatic to wrap the original exception? If so, what is the purpose?

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -564,6 +568,11 @@ def launch_fluent(
             return new_session.create_from_server_info_file(
                 server_info_filepath, cleanup_on_exit, start_transcript
             )
+        except Exception as ex:
+            identifier = "fluent.exe"
+            details = "\n" + "Fluent Launch Path: " + launch_string.split(identifier)[0] + identifier
```
seanpearsonuk, line 573: Looks like launch_string is the only argument we need to pass to the LaunchFluentError constructor. Then we can put this code there and have just one simple line inside the except block.

prmukherj: Done, Thank you for the suggestion. Looks much better now.

seanpearsonuk (approved):

hpohekar (approved):

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -405,6 +405,15 @@ def scm_to_py(topy):
     return launch_string
 
 
+class LaunchFluentError(Exception):
+
+    def __init__(self, launch_string):
+        identifier = "fluent.exe"
```
mkundu1, line 411: This won't work for Linux.

I feel it is sufficient to show just the whole `launch_string`, no need to break inside it

seanpearsonuk: Good point @mkundu1

prmukherj: Thank you @mkundu1. This has been updated.

PR #1308, opened 2023-01-25
Check for file modification in test

mkundu1:
The test `test_start_transcript_file_write` is randomly failing probably due to server to holding the file while the client tries to read it. Change to check for file modification instead of file content.

### DIFF
--- a/tests/test_session.py
+++ b/tests/test_session.py
@@ -281,19 +281,18 @@ def test_start_transcript_file_write(new_mesh_session, tmp_path=pyfluent.EXAMPLE
     )
     os.close(fd)
 
+    prev_stat = Path(file_path).stat()
+    prev_mtime = prev_stat.st_mtime
+    prev_size = prev_stat.st_size
+
     session = new_mesh_session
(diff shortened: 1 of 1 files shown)


seanpearsonuk (approved):

hpohekar (approved):

prmukherj (approved):

tests/test_session.py
```diff
@@ -281,19 +281,18 @@ def test_start_transcript_file_write(new_mesh_session, tmp_path=pyfluent.EXAMPLE
...
     os.close(fd)
 
+    prev_stat = Path(file_path).stat()
+    prev_mtime = prev_stat.st_mtime
+    prev_size = prev_stat.st_size
+
     session = new_mesh_session
     session.transcript.start(file_path)
     session = session.switch_to_solver()
     session.transcript.stop()
```
prmukherj, line 292: Thank you @mkundu1, for this solution.

prmukherj: We can do similar check for journal writing as well I think.

PR #1310, opened 2023-01-27
Feat/workflow wrapper extensions

seanpearsonuk:
An incremental improvement to the workflow interface: 
- provide information about the task hierarchy
-- which tasks are directly up/downstream of a given task, which is implied by task inputs/outputs. The methods are get_direct_upstream_tasks(), get_direct_downstream_tasks()
-- get all ordered children of either the workflow or of a given task. By recursion, the full hierarchy can by obtained. The method is ordered_children() (also inactive_ordered_children())
- remove the class nesting in workflow.py because of documentation difficulties
- add test for the new functionality. Also update existing tests, reflecting a recent API change in the backend, which is effective in 2023R2

N.b. I have deliberately avoided changing rst files to avoid conflicts with @hpohekar.

### DIFF
--- a/src/ansys/fluent/core/workflow.py
+++ b/src/ansys/fluent/core/workflow.py
@@ -1,3 +1,5 @@
+from typing import List
+
 from ansys.fluent.core.services.datamodel_se import PyCallableStateObject
 
 
@@ -55,7 +57,7 @@ class WorkflowWrapper:
                 return []
             attrs = set(attrs)
             tasks = [
-                task for task in self._command_source._all_task_objects() if
+                task for task in self._command_source.top_level_task_objects() if
                 task.name() != self.name()
             ]
             matches = []
@@ -94,6 +96,9 @@ class WorkflowWrapper:
                         type_ , id_ = k.split(':')
                         if type_ == "TaskObject":
                             return id_
+
+        def get_idx(self):
+            return int(self.get_id()[len("TaskObject"):])
 
         @property
         def CommandArguments(self):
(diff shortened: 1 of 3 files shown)


src/ansys/fluent/core/workflow.py
```diff
@@ -15,99 +17,265 @@ def __init__(self, task_name):
...
+    def _task_by_id_impl(self, task_id, workflow_state):
+        task_key = "TaskObject:" + task_id
+        task_state = workflow_state[task_key]
+        return self.task(task_state["_name_"])
+
+    def _task_by_id(self, task_id):
+        workflow_state = self._workflow_state()
+        return self._task_by_id_impl(task_id, workflow_state)
+
+    def _is_downstream(self, task, upstreams):
```
seanpearsonuk, line 270: optimise

tests/test_meshing_workflow.py
```diff
@@ -260,18 +260,17 @@ def test_accessors_for_argument_sub_items(new_mesh_session):
     ).CommandArguments.CadImportOptions.OneZonePer.is_read_only()
     assert (
         w.task(
-            "Import Geometry"
-        ).CommandArguments.CadImportOptions.OneZonePer.default_value()
-        == "Body"
+            "Generate the Volume Mesh"
+        ).CommandArguments.VolumeFillControls.Type.default_value()
+        == "Cartesian"
```
seanpearsonuk, line 265: I have chosen a different parameter here because the other one has changed in the meshing application itself, and we don't want to handle the differences between the versions here.

hpohekar:
@seanpearsonuk 

Following change might resolve one of the issue related to api doc generation.

`raise_method_not_yet_implemented_exception()` -> `_raise_method_not_yet_implemented_exception()`

because for `raise_method_not_yet_implemented_exception()` also, the sphinx was generating a separate `rst` file.

mkundu1 (approved):

PR #1320, opened 2023-02-01
Fix is_active checks

mkundu1:
Fluent bugs 774607, 774755, 774903

For all these defects, `is_active()` returns a list from Fluent for which we need to cast to boolean type in PyFluent. There are many such occurrences in Fluent code, it is probably not possible to find and change the Fluent code everywhere.

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -157,7 +157,7 @@ class Base:
                 attr_type_or_types = (attr_type_or_types,)
             if isinstance(val, attr_type_or_types):
                 return val
-            if val is not None and issubclass(bool, attr_type_or_types):  # cast to bool for boolean attributes
+            if val is not None and any(issubclass(x, bool) for x in attr_type_or_types):  # cast to bool for boolean attributes
                 return bool(val)
             return None
         return val
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -153,10 +153,12 @@ class Base:
             val = attrs[attr]
 
         if attr_type_or_types:
-            if not type(attr_type_or_types) == tuple:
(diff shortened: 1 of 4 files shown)


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -153,10 +153,12 @@ def get_attr(self, attr, attr_type_or_types=None) -> Any:
             val = attrs[attr]
 
         if attr_type_or_types:
-            if not type(attr_type_or_types) == tuple:
+            if not isinstance(attr_type_or_types, tuple):
                 attr_type_or_types = (attr_type_or_types,)
-            if type(val) in attr_type_or_types:
+            if isinstance(val, attr_type_or_types):
                 return val
+            if bool in attr_type_or_types:  # cast to bool for boolean attributes
```
mkundu1, line 160: we can do the same for other datatypes

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -153,10 +153,12 @@ def get_attr(self, attr, attr_type_or_types=None) -> Any:
             val = attrs[attr]
 
         if attr_type_or_types:
-            if not type(attr_type_or_types) == tuple:
+            if not isinstance(attr_type_or_types, tuple):
```
mkundu1, line 156: I feel we should avoid passing tuple here. A given attribute of a given type should be of definite datatype. This will also make type casting easier if we decide to extend it for other datatypes.

seanpearsonuk (approved):

tests/test_flobject.py
```diff
@@ -408,7 +408,6 @@ def test_primitives():
     assert r.g_1.b_3() is False
     r.g_1.s_4 = "foo"
     assert r.g_1.s_4() == "foo"
-    return True
```
mkundu1, line 411: https://docs.pytest.org/en/7.2.x/deprecations.html#returning-non-none-value-in-test-functions

PR #1324, opened 2023-02-02
Patch_Release 0.11.1

alikasim72:


.github/workflows/ci.yml
```diff
@@ -374,7 +374,6 @@ jobs:
   release:
     name: Release
     if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
-    needs: build-test
```
jorgepiloto, line 377: This line is needs to be restored. Otherwise, we may release a version that does not build properly and thus, can not be installed by users.

jorgepiloto (changes requested):

.github/workflows/ci.yml
```diff
@@ -370,11 +370,83 @@ jobs:
             dist/*.whl
             dist/*.tar.gz
           retention-days: 7
+          
+  test:
```
jorgepiloto, line 374: I can see in line 340 there are some other unit tests going on. Hence, I am not sure if this is a duplicated job.

Pinging here @mainakk do provide more insights about it.

jorgepiloto (changes requested):

.github/workflows/ci.yml
alikasim72: Talked with @mkundu1 and resolved the issue, yea I accidently duplicated the tests. From 0.12 the pyfluent team separated the testing and building and while pulling from the latest ci_cd file I also added the separate test section along with the build_test. Should be fixed now with only one build_test job.

mkundu1 (approved):

PR #1327, opened 2023-02-06
Split reading case from launch fluent process.

prmukherj:
We split things up inside launch_fluent to invoke the API after the server is launched.


seanpearsonuk:
@prmukherj we can also consider read journal, and also case & data, for which we would require a new launch_fluent argument.

hpohekar (approved):

seanpearsonuk (approved):

PR #1330, opened 2023-02-07
Add allowed_values() for rp-vars.

prmukherj:
Allowed values for rp-vars provide guidance for users. If the user provides a wrong input we raise a runtime error with near matches of the input as well.


hpohekar (approved):

src/ansys/fluent/core/rpvars.py
```diff
@@ -4,10 +4,12 @@
 variables like rpvars but instead through the high-level object-based
 interfaces: solver settings objects and task-based meshing workflow.
 """
-
-from typing import Any
+from typing import Any, List
 
 import ansys.fluent.core.filereader.lispy as lispy
+from ansys.fluent.core.services.field_data import allowed_name_error_message
```
seanpearsonuk, line 10: We ought to move allowed_name_error_message out of field_data to a common place

tests/test_rp_vars.py
```diff
@@ -53,4 +53,23 @@ def test_get_all_rp_vars(new_solver_session_no_transcript) -> None:
...
+def test_rp_vars_allowed_values(new_solver_session_no_transcript) -> None:
+    solver = new_solver_session_no_transcript
+    rp_vars = solver.rp_vars
+
+    assert rp_vars("number-of-iterations") == 0
+
+    with pytest.raises(RuntimeError) as msg:
+        rp_vars("number-of-iterat")
+
+    assert msg.value.args[0] == "number-of-iterat is not an allowed  name.\n" \
```
seanpearsonuk, line 70: noting the double space. Is this because you are not specifying the argument.  

We should pass the argument e.g., "variable".

Also the implementation of allowed_name_error_message should check the string to decide if the extra space is needed

tests/test_rp_vars.py
```diff
@@ -53,4 +53,23 @@ def test_get_all_rp_vars(new_solver_session_no_transcript) -> None:
...
+                                "The most similar names are: number-of-iterations, " \
+                                "number-of-time-steps, lb/number-of-timesteps, " \
+                                "number-of-samples, gpuapp/total-number-of-subiterations."
+
+    assert "number-of-iterations" in rp_vars.allowed_values()
```
seanpearsonuk, line 75: more precisely it should be rp_vars.var.allowed_values()

src/ansys/fluent/core/rpvars.py
prmukherj: yes, shall we create a new file and move it or is there a common place for all these?

src/ansys/fluent/core/rpvars.py
```diff
@@ -48,13 +50,23 @@ def __call__(self, var: str = None, val: Any=None) -> Any:
             self._get_var(var) if var else self._get_vars()
         )
 
+    def allowed_values(self) -> List[str]:
+        if not _allowed_rpvars_values:
+            _allowed_rpvars_values.append(lispy.parse(self._eval_fn("(cx-send '(map car rp-variables))")))
```
mkundu1, line 55: Instead of having a list within list here, you can do the following:
```
class RPVars:
    _allowed_values = None
    @property
    def allowed_values(self) -> List[str]:
        if not RPVars._allowed_values:
            RPVars._allowed_values =  lispy.parse(self._eval_fn("(cx-send '(map car rp-variables))"))
        return RPVars._allowed_values
    def _get_var(self, var: str):
        if var not in self.allowed_values:
            ...
```

tests/test_rp_vars.py
prmukherj: @seanpearsonuk, could we have a short discussion on this?

src/ansys/fluent/core/rpvars.py
prmukherj: Done, Thank you

tests/test_rp_vars.py
prmukherj: Done, Thank you :)

src/ansys/fluent/core/rpvars.py
```diff
@@ -48,13 +50,23 @@ def __call__(self, var: str = None, val: Any=None) -> Any:
...
 
+    def allowed_values(self) -> List[str]:
+        if not RPVars._allowed_values:
+            RPVars._allowed_values = lispy.parse(self._eval_fn("(cx-send '(map car rp-variables))"))
+        return RPVars._allowed_values
+
     def _get_var(self, var: str):
+        if not RPVars._allowed_values:
+            RPVars._allowed_values = self.allowed_values()
+        if var not in RPVars._allowed_values:
```
mkundu1, line 61: I think `if var not in self.allowed_values():` is sufficient. `RPVars._allowed_values` will be populated by `self.allowed_values()`.

mkundu1 (approved):

tests/test_rp_vars.py
seanpearsonuk: @prmukherj only just seeing your response here. `allowed_values()` is customarily called on the argument, which is `var` here.

PR #1343, opened 2023-02-09
Updated optislang integration test to use case reader

cj-hodgson:


tests/integration/test_optislang/test_optislang_integration.py
```diff
@@ -38,66 +38,60 @@ def test_simple_solve(load_mixing_elbow_param_case_dat):
...
-        assert output_parameters[0] == output_parameters2[0]
-        solver_session.exit()
+    # Step 6: Read the data again from the case and data file
+    solver_session.tui.file.read_case_data(case_path)
+    from ansys.fluent.parametric.parameters import InputParameters, OutputParameters
+    inputs_table = InputParameters(solver_session)
+    outputs_table = OutputParameters(solver_session) 
+
+    output_units = outputs_table["outlet_temp-op"].split(' ')[1].strip('[]')
+    assert input_parameters["inlet2_temp"][0] == inputs_table["inlet2_temp"]
```
tejalprabhu01, line 92: input_parameters consists of the original value of parameters and inputs_table contains the value that was set before solving. why would we expect these two values to be the same? Rather following check can be performed:
assert inputs_table["inlet2_temp"] = 500 ---> this is to confirm that the value with which we solved the case file is as expected

tejalprabhu01 (changes requested):

PR #1357, opened 2023-02-14
Support light mode while reading case within launch_fluent

mkundu1:


seanpearsonuk (approved):

hpohekar (approved):

prmukherj (approved):

tests/test_session.py
```diff
@@ -304,15 +304,20 @@ def test_solverworkflow_in_solver_session(new_solver_session):
...
+    solver.setup.models.energy.enabled = False
+    old_fluent_connection_id = id(solver.fluent_connection)
+    while id(solver.fluent_connection) == old_fluent_connection_id:
         time.sleep(1)
-    assert new_solver_session.setup.models.energy.enabled() == False
-    pyfluent.USE_LIGHT_IO = False
+    try:
+        solver.health_check_service.wait_for_server(10)
+    except Exception:
+        assert False
```
seanpearsonuk, line 320: @mkundu1 in follow-up work, we could simply provide a mechanism to allow registration of a callback that would replace these details. What do you think?

tests/test_session.py
```diff
@@ -304,15 +304,20 @@ def test_solverworkflow_in_solver_session(new_solver_session):
...
+def test_read_case_using_light_io(with_launching_container):
     import_filename = examples.download_file(
         "mixing_elbow.cas.h5", "pyfluent/mixing_elbow"
     )
-    pyfluent.USE_LIGHT_IO = True
-    new_solver_session.read_case(file_name=import_filename)
-    new_solver_session.setup.models.energy.enabled = False
-    old_fluent_connection_id = id(new_solver_session.fluent_connection)
-    while id(new_solver_session.fluent_connection) == old_fluent_connection_id:
+    solver = pyfluent.launch_fluent(case_filepath=import_filename, use_light_io=True)
```
seanpearsonuk, line 312: @mkundu1 I'm wondering if we could rename `use_light_io` to simply `lightweight_mode` to de-emphasise the IO aspect?

tests/test_session.py
mkundu1: Thanks @seanpearsonuk , will do as a follow-up

PR #1372, opened 2023-02-19
Suggest closest matching names for config-vars error.

prmukherj:
If the config-vars gets a key-error, we return the closest matching names along with the error message.

### DIFF
--- a/src/ansys/fluent/core/filereader/case.py
+++ b/src/ansys/fluent/core/filereader/case.py
@@ -25,6 +25,7 @@ from typing import List
 import h5py
 
 from . import lispy
+from ..allowed_name_error_msg import allowed_name_error_message
 
 
 class InputParameter:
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/filereader/case.py
```diff
@@ -99,7 +100,11 @@ def __call__(self, name: str = ""):
         if not name:
             error_name = self._path[:-1] if self._path else self._path
             raise RuntimeError(f"Invalid variable {error_name}")
-        return self._variables[name]
+        try:
+            return self._variables[name]
+        except KeyError:
+            raise RuntimeError(allowed_name_error_message(
```
mkundu1, line 106: I think we can throw [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) here for out-of-bound arguments.

mkundu1 (approved):

hpohekar (approved):

seanpearsonuk (approved):

PR #1380, opened 2023-02-21
Settings API Matching Names Exception

hpohekar:
image

`----------------------------------------------------------------------------`

```
 >>> solver.setup.model

 model is a child of viscous.radiation.species.structure                                               <<<--------

  File "D:\pyfluent\src\ansys\fluent\core\solver\flobject.py", line 476, in __getattribute__
    return super().__getattribute__(name)
AttributeError: 'setup' object has no attribute 'model'. Did you mean: 'models'?

The above exception was the direct cause of the following exception:

  File "D:\pyfluent\except.py", line 9, in <module>
    solver.setup.model
  File "D:\pyfluent\src\ansys\fluent\core\solver\flobject.py", line 479, in __getattribute__
    raise AttributeError(
AttributeError: model is not an allowed Settings objects name.                                       <<<--------
The most similar names are: models.. Did you mean: 'models'?                                        <<<--------
>>>

```

`-------------------------------------------------------------------------------------------`

```
>>> solver.setup.model
  File "D:\pyfluent\src\ansys\fluent\core\solver\flobject.py", line 460, in __getattribute__
    return super().__getattribute__(name)
AttributeError: 'setup' object has no attribute 'model'. Did you mean: 'models'? 

The above exception was the direct cause of the following exception:

  File "D:\pyfluent\src\ansys\fluent\core\solver\flobject.py", line 462, in __getattribute__
    raise AttributeError(
AttributeError: model is not an allowed Settings objects name.
The most similar names are: models.. Did you mean: 'models'?
>>>
```
`----------------------------------------------------------------------------------------------`

```
>>> solver.setup.boundary_conditions.velocity_inlet['inlet-0']
  File "D:\pyfluent\src\ansys\fluent\core\solver\flobject.py", line 681, in __getitem__
    raise KeyError(
KeyError: 'inlet-0 is not an allowed Settings objects name.\nThe most similar names are: hot-inlet.'  <-------------------
>>>
```

`----------------------------------------------------------------------------------------------`

```
>>> solver.setup.models.viscous.model = "k_epsilon"

Error: Value is not allowed
Error Object: ("inviscid" "laminar" "k-epsilon" "k-omega" "mixing-length" "spalart-allmaras" "k-kl-w" "transition-sst" "reynolds-stress" "scale-adaptive-simulation" "detached-eddy-simulation" "large-eddy-simulation")
  File "D:\pyfluent\src\ansys\fluent\core\solver\flobject.py", line 474, in __setattr__
    return getattr(self, name).set_state(value)
  File "D:\pyfluent\src\ansys\fluent\core\solver\flobject.py", line 261, in set_state
    return self.flproxy.set_var(self.path, self.to_scheme_keys(state))
  File "D:\pyfluent\src\ansys\fluent\core\services\settings.py", line 91, in _fn
    return fn(self, *args, **kwds)
  File "D:\pyfluent\src\ansys\fluent\core\services\settings.py", line 156, in set_var
    self._service_impl.set_var(request)
  File "D:\pyfluent\src\ansys\fluent\core\services\error_handler.py", line 15, in func
    raise RuntimeError(ex.details()) from None
RuntimeError: Value is not allowed
Error Object: (inviscid laminar k-epsilon k-omega mixing-length spalart-allmaras k-kl-w transition-sst reynolds-stress scale-adaptive-simulation detached-eddy-simulation large-eddy-simulation)

The above exception was the direct cause of the following exception:

  File "D:\pyfluent\src\ansys\fluent\core\solver\flobject.py", line 479, in __setattr__
    raise allowed_values_error(name, value, values) from ex
ValueError: k_epsilon is not an allowed model name.
The most similar names are: k-epsilon.  <-------------------
>>>
```


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -584,8 +589,13 @@ def get_object_names(self):
...
     def __getitem__(self, name: str) -> ChildTypeT:
-        if name not in self.get_object_names():
-            raise KeyError(name)
+        try:
+            if name not in self.get_object_names():
+                raise KeyError(name)
+        except KeyError:
+            keys = self.get_object_names()
+            print(f"KeyError: '{name}'")
+            print(f"Available keys are - {keys}")
```
prmukherj, line 598: @hpohekar, I think here we can use the "allowed_name_error_msg.py" to display the names that are closely matching the name passed by user.

hpohekar: Yes we have integrated both approaches. Thank you.

prmukherj (approved):

seanpearsonuk:
@hpohekar, I just noticed that allowed_name_error_**msg**.py is misnamed.

src/ansys/fluent/core/solver/allowed_name_error_msg.py
```diff
@@ -19,3 +19,9 @@ def allowed_name_error_message(
     if matches:
         message += f"The most similar names are: {', '.join(matches)}."
     return message
+
```
seanpearsonuk, line 22: @hpohekar, as discussed, we don't want to have this file in here. We can simply add a file allowed_name_error_message.py with single line:
```
from ..allowed_name_error_message import allowed_name_error_message
```
In both versions of flobject.py:
```
from .allowed_name_error_message import allowed_name_error_message
```
We need t ou update the filename also, as mentioned just now.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -456,10 +457,26 @@ def __getattribute__(self, name):
...
                 raise RuntimeError(f"'{self.path}' is currently not active")
-        return super().__getattribute__(name)
+        try:
+            return super().__getattribute__(name)
+        except AttributeError as ex:
+            attributes = [
+                attribute
+                for attribute in self.__dict__.keys()
+                if not attribute.startswith("_")
+            ]
```
seanpearsonuk, line 467: @hpohekar looking at the top of this method, it seems that `super().__getattribute__("child_names")` is the relevant list here. If we find ourselves writing code like the above then we need to step back.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -456,10 +457,26 @@ def __getattribute__(self, name):
...
+                allowed_name_error_message("Settings objects", name, attributes)
+            ) from ex
 
     def __setattr__(self, name: str, value):
-        return getattr(self, name).set_state(value)
+        try:
+            return getattr(self, name).set_state(value)
+        except BaseException as ex:
+            if hasattr(getattr(self, name), "allowed_values"):
+                values = getattr(self, name).allowed_values()
```
seanpearsonuk, line 477:
```suggestion
            allowed = getattr(getattr(self, name), "allowed_values", None):
            if allowed:
                values = allowed()
```

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -456,10 +457,26 @@ def __getattribute__(self, name):
...
+                if value not in values:
```
seanpearsonuk, line 478:
```suggestion
            allowed = getattr(getattr(self, name), "allowed_values", None):
            if allowed and value not in allowed():
```

src/ansys/fluent/core/solver/allowed_name_error_msg.py
mkundu1: How about, add this line in src\ansys\fluent\core\solver\__init__.py:
```
from ansys.fluent.core import allowed_name_error_message
```
In flobject.py:
```
from allowed_name_error_message import allowed_name_error_message, closest_allowed_names
```
Add allowed_name_error_message.py at the same level of flobject.py in Fluent.

seanpearsonuk: @mkundu1 I tried that version first but got an import error (partial / cyclic import)

mkundu1: Yeah, this won't work, we need to create the file.

src/ansys/fluent/core/solver/flobject.py

src/ansys/fluent/core/solver/flobject.py

src/ansys/fluent/core/solver/allowed_name_error_msg.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/solver/flobject.py
hpohekar: @seanpearsonuk 

above change gives following error - 

image

So we changed it as follows -

image

src/ansys/fluent/core/solver/flobject.py
hpohekar: all of the following are same i.e `attributes, self.child_names, super().__getattribute__("child_names")`

image

`------------------------------------------------------------------`

image

mkundu1:
I'm getting this message

model_msg

hpohekar:
> I'm getting this message
> 
> model_msg

Right, because right now we are appending it to empty string. We will append it to list.

Thank you.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -456,10 +457,31 @@ def __getattribute__(self, name):
         if name in super().__getattribute__("child_names"):
             if not self.is_active():
                 raise RuntimeError(f"'{self.path}' is currently not active")
-        return super().__getattribute__(name)
+        try:
+            return super().__getattribute__(name)
+        except AttributeError as ex:
+            parents = []
+            for parent in self.get_active_child_names():
```
seanpearsonuk, line 464: parent is apparently synonymous with child in this nomenclature

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -456,10 +457,31 @@ def __getattribute__(self, name):
...
+                        parents.append(parent)
+                except AttributeError:
+                    pass
+            if len(parents):
+                print(f"\n {name} is a child of {parents} \n")
+            raise AttributeError(
+                allowed_name_error_message(
+                    "Settings objects", name, super().__getattribute__("child_names")
+                )
+            ) from ex
```
seanpearsonuk, line 476: This except block is quite complicated and the reader needs to scrutinize it in some detail to see what is going on - that's bad for maintenance.
Better in general:
```
try:
    do_stuff()
except ParticularError as ex:
    handle_particular_error_in_explicitly_described_way(ex, args)
```
 
- Note that `handle_particular_error_in_explicitly_described_way` can now also be tested in isolation.
- By isolating the handling in a single function, it can easily be switched off

seanpearsonuk:
> 

a dot path would be best

seanpearsonuk:
It would be reassuring to see some unit tests here. Note for example, @mkundu1 's comment above about the string formatting. How can we unit test this? By breaking the code up into small units and testing them individually. This is tiresome, and even slows the code down, but if we don't do it then we are relying on reviewers to run the whole code (which doesn't stop the code getting broken in future).

hpohekar:
> > 
> 
> a dot path would be best

@seanpearsonuk any thoughts on this ?

image

hpohekar:
> I'm getting this message
> 
> model_msg

image

src/ansys/fluent/core/solver/flobject.py
hpohekar: Done. Thank you.

hpohekar:
> It would be reassuring to see some unit tests here. Note for example, @mkundu1 's comment above about the string formatting. How can we unit test this? By breaking the code up into small units and testing them individually. This is tiresome, and even slows the code down, but if we don't do it then we are relying on reviewers to run the whole code (which doesn't stop the code getting broken in future).

@seanpearsonuk 

image

hpohekar:
> > 
> 
> a dot path would be best

Done. Thank you.

image

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -452,14 +453,42 @@ def get_active_query_names(self):
...
 
     def __setattr__(self, name: str, value):
-        return getattr(self, name).set_state(value)
+        try:
+            return getattr(self, name).set_state(value)
```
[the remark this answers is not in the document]
hpohekar: It raises Run Time Error and gives information about error object.

image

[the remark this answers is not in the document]
hpohekar: @h-krishnan right. We are handling this now separately. Thank you.

h-krishnan (approved):

src/ansys/fluent/core/solver/allowed_name_error_message.py
```diff
@@ -0,0 +1,4 @@
+from ..allowed_name_error_msg import (  # noqa: F401
```
h-krishnan, line 1: Should we have just one file error_msg.py that contains classes for different errors, instead of allowed_name_error_msg.py, allowed_name_error_message.py etc.?

But why this file, in any case?

hpohekar: Yes, definitely. We will keep single `error_msg.py` file.

PR #1384, opened 2023-02-21
Use random port in unittests

mkundu1:
I was seeing random issues from these tests, most likely due to sharing the same port number.


mkundu1:
Changes merged with https://github.com/pyansys/pyfluent/pull/1378

Issue #1333, opened 2023-02-07
Intellisense for latest release (0.12.3) not working with VS code


snallamoansys:
### 🔍 Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### 🐞 Description of the bug

I am not seeing the intellisense working in visual studio code. Anything else needs to be done? For previous release (0.11.0), it works fine as shown here. But for 0.12.3, nothing appears after solver.setup.

![image](https://user-images.githubusercontent.com/99287292/217265891-b77d5ecb-6ecd-4b48-8b5d-844cda0de477.png)

![image](https://user-images.githubusercontent.com/99287292/217265670-9d182017-1f71-4438-9203-594ebd0cee00.png)

### 📝 Steps to reproduce

In Visual Studio Code, activate virtual environment with 0.12.3 version of ansys-fluent-core installed and load any case file such as elbow case file and try to access the settings api as shown in the picture

import ansys.fluent.core as pyfluent
fluent = pyfluent.launch_fluent(mode='solver',show_gui=True)
tui=fluent.tui
fluent.file.read(file_name=r'elbow.cas.h5',file_type='case')

### 💻 Which operating system are you using?

Windows

### 🐍 Which Python version are you using?

3.9

### 📦 Installed packages

```shell
ansys-api-fluent==0.3.5
ansys-api-platform-instancemanagement==1.0.0b3
ansys-fluent-core==0.12.3
ansys-fluent-parametric==0.6.0
ansys-fluent-visualization==0.6.0
ansys-platform-instancemanagement==1.0.2
appdirs==1.4.4
asttokens==2.2.1
backcall==0.2.0
certifi==2022.12.7
charset-normalizer==3.0.1
click==8.1.3
colorama==0.4.6
comm==0.1.2
contourpy==1.0.7
cycler==0.11.0
dash==2.7.1
dash-bootstrap-components==1.3.0
dash-core-components==2.0.0
dash-html-components==2.0.0
dash-table==5.0.0
debugpy==1.6.5
decorator==5.1.1
entrypoints==0.4
executing==1.2.0
Flask==2.2.2
fonttools==4.38.0
googleapis-common-protos==1.58.0
grpcio==1.51.1
h5py==3.7.0
idna==3.4
imageio==2.24.0
importlib-metadata==6.0.0
ipykernel==6.20.2
ipympl==0.9.2
ipython==8.8.0
ipython-genutils==0.2.0
[...]

PR #1397, opened 2023-02-24
Add typing stub files

mkundu1:
Added the type-hints in separate stub files as those depend on the generated API code. If these were added in actual source, it will fail due to ImportError when codegen is not run.

This PR will fix static intellisense issues in vscode etc.


tests/test_session.py
```diff
@@ -274,6 +274,7 @@ def test_get_fluent_mode(new_mesh_session):
 
 @pytest.mark.dev
 @pytest.mark.fluent_232
+@pytest.mark.skip("Failing in github")
```
mkundu1, line 277: This test often fails in github probably due to the issue demonstrated in https://github.com/pyansys/pyfluent/pull/1302

dnwillia-work:
Is this for #1333?

dnwillia-work (approved):

prmukherj (approved):

PR #1398, opened 2023-02-25
Change doc file name in PR pipeline

dnwillia:
Doc uploads will work properly, already fixed like this in nightly build the same way.


dnwillia:
Never mind about this for now.  It's not quite working.

PR #1399, opened 2023-02-25
Remove path from zipped doc archive

dnwillia-work:
I think this is why the `dev` link is broken, the `doc/_build/html` part of the path is messing up finding the index.html.


dnwillia-work:
Force merging to see if it's fixed...

PR #1416, opened 2023-03-05
Testing/increasing coverage

Adam-Boutin:
I have added 4 extra tests to tests/test_scheduler.py. These tests are as follows:

test_machine_info:
This makes sure that calling load_machines() with machine_info gets correctly sorted through the elif block.  

test_restrict_machines:
This test checks that if test_restrict_machines() is called directly it will not alter a machine list that has less than or equal to the ncores specified.

test_lsb_mcpu:
simply makes sure that "LSB_MCPU_HOSTS" is checked for a possible host list.

test_no_machine_name:
Checks to make sure that _parse_host_info will throw the 'Problem with machine list format.' exception if there is a blank machine name.

Please let me know if there is anything incorrect, poor logic, or unneeded tests here.
If this is accepted this will be my first contribution. I would be happy to receive any feedback and look forward to working on more testing.

### DIFF
--- a/.github/workflows/nightly-doc-build.yml
+++ b/.github/workflows/nightly-doc-build.yml
@@ -7,7 +7,7 @@ on:
 
 env:
   DOCUMENTATION_CNAME: 'fluent.docs.pyansys.com'
-  DOC_DEPLOYMENT_IMAGE_TAG: v22.2.0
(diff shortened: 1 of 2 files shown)


mkundu1 (approved):
Looks good, thanks

dnwillia-work (approved):

seanpearsonuk (approved):

PR #1418, opened 2023-03-07
Feat/cache

seanpearsonuk:


tests/test_datamodel_service.py
```diff
@@ -140,3 +145,95 @@ def test_add_on_affected_at_type_path(new_mesh_session):
...
+
+    stream.register_callback(cb)
+    stream_size = len(cb.states)
+    assert stream_size == 0
+
+    import_filename = examples.download_file(
+        "mixing_elbow.cas.h5", "pyfluent/mixing_elbow"
+    )
+    solver.file.read_case(file_name=import_filename)
+    sleep(5)
```
mkundu1, line 220: The sleep calls are needed as the stream is received in a different thread

src/ansys/fluent/core/data_model_cache.py
```diff
@@ -0,0 +1,51 @@
+from collections import defaultdict
```
seanpearsonuk, line 1: @mkundu1 in this module we'll need a function to recursively update the nested dict state

src/ansys/fluent/core/data_model_cache.py
```diff
@@ -0,0 +1,143 @@
...
+                else:
+                    name = v.variant_map_state.item["_name_"].string_state
+                    key = f"{type}:{name}"
+                    source[key] = {"__iname__": iname}
+                    _update_cache_from_variant_state(source, key, v, dict.__setitem__)
+            else:
+                _update_cache_from_variant_state(source, k, v, dict.__setitem__)
+
+
+def update_cache(rules: str, state: Variant, deleted_paths: List[str]):
```
mkundu1, line 79: First version of update cache, tested through unittest so far

tests/test_data_model_cache.py
```diff
@@ -0,0 +1,70 @@
...
+            {"A": {"B": {"C": 5}}},
+            [],
+            {"r1": {"A": {"B": {"C": 5}}}},
+        ),
+        (
+            {"r1": {"A": {}}},
+            "r1",
+            {"A": {"B:B1": {"_name_": "B-1", "C": 5.0}}},
+            [],
+            {"r1": {"A": {"B:B-1": {"__iname__": "B1", "_name_": "B-1", "C": 5.0}}}},
```
mkundu1, line 53: I've introduced a hidden parameter `__iname__` in the cached state to store the internal name which is required to identify the named object within the cached state in the subsequent update. get_state from cache will need to strip all `__iname__` before returning the state to user.

seanpearsonuk: @mkundu the way I think it, the one driver for absolutely needing this is that the `deletedpaths` are (presumably) by internal name, such that if `__iname__` is missing, we would not know which elements to remove from the cache.

seanpearsonuk:
@mkundu1 this is looking very good - many thanks for taking it on! Do we still think there's an issue with BeginStreaming on the server-side for some datamodels (e.g., workflow)?

mkundu1:
> @mkundu1 this is looking very good - many thanks for taking it on! Do we still think there's an issue with BeginStreaming on the server-side for some datamodels (e.g., workflow)?

@seanpearsonuk I don't think there are any issues with BeginStreaming on the server-side for specific datamodel. We will probably face some issues due to usage of NoCommandsDifferencer for extracting the diff in server side.

mkundu1:
Rebased and squashed the branch

hpohekar (approved):

src/ansys/fluent/core/streaming_services/datamodel_streaming.py
```diff
@@ -0,0 +1,36 @@
...
+        """Processes datamodel events."""
+        responses = self._streaming_service.begin_streaming(started_evt)
+        while True:
+            try:
+                response: datamodel_se_pb2.DataModelResponse = next(responses)
+                with self._lock:
+                    self._streaming = True
+                    for _, cb_list in self._service_callbacks.items():
+                        state = (
+                            _convert_variant_to_value(response.state)
```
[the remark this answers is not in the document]
mkundu1: @seanpearsonuk It seems the dict state in meshing workflow is keyed with internal names. The cached state is currently keyed with display names which is causing inconsistency. Some unittests also assume that the internal names are the keys. Should I change the caching to have internal names as keys?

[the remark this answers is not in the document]
mkundu1: We can have both configurable in PyFluent similar to Fluent.

tests/test_data_model_cache.py
```diff
@@ -0,0 +1,198 @@
...
+                        "C": 5.0,
+                    }
+                }
+            },
+        ),
+        (
+            {"r1": {"B:B1": {"_name_": "B-1", "C": 5.0}}},
+            "r1",
+            {"B:B1": {"_name_": "B-2", "C": 5.0}},
+            ["B:B1"],
```
mkundu1, line 184: @seanpearsonuk Checked that this is streamed on renaming a named object in Fluent - deletion + creation

prmukherj (approved):

PR #1420, opened 2023-03-08
iter method for tasks

seanpearsonuk:


src/ansys/fluent/core/workflow.py
```diff
@@ -230,6 +202,46 @@ def _command(self):
         return self._cmd
```
seanpearsonuk, line 204: switched the order classes are defined in because `Task` is now mentioned int the type hints for `TaskContainer`.

src/ansys/fluent/core/workflow.py
```diff
@@ -230,6 +202,46 @@ def _command(self):
...
+    __getitem__(attr)
+    __getattr__(attr)
+    __dir__()
+    """
+
+    def __init__(self, command_source):
+        self._container = command_source
+        self._task_container = command_source._workflow.TaskObject
+
+    def __iter__(self) -> Iterator[Task]:
```
seanpearsonuk, line 220: only this method is added - no other changes

mkundu1 (approved):

tests/test_meshing_workflow.py
```diff
@@ -600,3 +600,13 @@ def downstream_names(task):
         "Enclose Fluid Regions (Capping)",
         "Create Regions",
     ]
+
+
+@pytest.mark.dev
```
seanpearsonuk, line 605: this is a failing test without the code changes

prmukherj (approved):

PR #1468, opened 2023-03-09
CI to build+deploy release doc

mkundu1:
Added a new CI to build and deploy the release doc. We have disabled that step from ci.yml as we now build only pyfluent source doc in ci.yml. The CI will be triggered when a non-dev tag is pushed.

### DIFF
--- a/.github/workflows/release-doc-build.yml
+++ b/.github/workflows/release-doc-build.yml
@@ -1,9 +1,11 @@
-name: Nightly Development Documentation Build
+name: Release Documentation Build
 
 on:
-  schedule:  # UTC at 0400
-    - cron:  '0 4 * * *'
   workflow_dispatch:
+  push:
+    tags:
+      - 'v*'
+      - '!*dev*'
 
 env:
   DOCUMENTATION_CNAME: 'fluent.docs.pyansys.com'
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly-dev-doc-build.yml
```diff
@@ -20,18 +20,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      # used for documentation deployment
```
dnwillia-work, line 23: This is no longer needed then?

dnwillia-work (approved):

.github/workflows/release-doc-build.yml
```diff
@@ -0,0 +1,81 @@
+name: Release Documentation Build
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
```
dnwillia-work, line 6: So, this means the workflow runs when a tag starting with *v* is pushed, but no tags including the character string *dev* will be built.

dnwillia-work (approved):

hpohekar (approved):

prmukherj (approved):

.github/workflows/nightly-dev-doc-build.yml
mkundu1: We are not using this kind of token in our workflows anymore.

.github/workflows/release-doc-build.yml
mkundu1: Yeah

PR #1473, opened 2023-03-14
Testing/scheduler testing improvments

Adam-Boutin:
Changes to load_machines.py:
---------------------------------
121-135: Checks if the host info is a file before parsing. I made this change because then I can pass the host info taken from a file into the same parsing procedure.

263-267: I was getting many "index out of range" errors and it doesn't make sense to send in default values. It makes sure there are at least the 2 required inputs and that the 2nd is of type int. I don't know if there would need to be a similar type constraint for a 4th argument.

190-302: Changed machineDict to a plain hostName: numberOfCores instead of having objects. At the end it passes it in once with all the required values.

341,380: Added a "." in the regular expression and it picks up a single number in brackets. (i.e. [3])

Changes to test_scheduler.py:
---------------------------------
Imported tempfile for testing the "with open" statements.

Through these new tests I found what I thought were issues in load_machines.py. I made changes but I am still learning and I am not sure they are better than the previous. As always I am open to any feedback that I can get. With these changes coverage of load_machines.py is up to 94%.

### DIFF
--- a/src/ansys/fluent/core/scheduler/load_machines.py
+++ b/src/ansys/fluent/core/scheduler/load_machines.py
@@ -118,11 +118,8 @@ def _parse_host_info(host_info):
         A list of dictionaries formatted as:
         {'machine-name' : ###, 'core-count' : ###}
     """
-    print(host_info)
-    if "\\" in host_info or "/" in host_info or "." in host_info:
-        # Filenames generally have '\\' or '/' or '.'
-        # so assume it's a file and parse accordingly
-        # Read from the file
+    if os.path.exists(host_info):
+        # Only opens a file if it exists
         with open(host_info, "r") as f:
             host_info = f.read()
 
--- a/src/ansys/fluent/core/scheduler/load_machines.py
+++ b/src/ansys/fluent/core/scheduler/load_machines.py
@@ -118,14 +118,8 @@ def _parse_host_info(host_info):
         A list of dictionaries formatted as:
         {'machine-name' : ###, 'core-count' : ###}
     """
-    print(host_info)
-    if ("\\" in host_info
-        or "/" in host_info
-        or "." in host_info
-    ):
-        # Filenames generally have '\\' or '/' or '.'
-        # so assume it's a file and parse accordingly
-        # Read from the file
+    if (os.path.exists(host_info)):
+        # only opens a file if it exists
         with open(host_info, "r") as f:
             host_info = f.read()
          
--- a/src/ansys/fluent/core/scheduler/load_machines.py
+++ b/src/ansys/fluent/core/scheduler/load_machines.py
@@ -119,10 +119,7 @@ def _parse_host_info(host_info):
         {'machine-name' : ###, 'core-count' : ###}
     """
     print(host_info)
-    if ("\\" in host_info
-        or "/" in host_info
-        or "." in host_info
-    ):
+    if "\\" in host_info or "/" in host_info or "." in host_info:
         # Filenames generally have '\\' or '/' or '.'
         # so assume it's a file and parse accordingly
         # Read from the file
--- a/src/ansys/fluent/core/scheduler/load_machines.py
+++ b/src/ansys/fluent/core/scheduler/load_machines.py
@@ -264,8 +264,7 @@ def _construct_machine_list_uge(host_filename):
             if len(row) == 1:
                 row.append(1)
             row[1] = int(row[1])
-            m = Machine(*row)
-            machineList.add(m)
+            machineList.add(Machine(*row))
     return machineList
 
 
@@ -289,15 +288,18 @@ def _construct_machine_list_pbs(host_filename):
     with open(host_filename, "r") as pbsFile:
         for hostname in pbsFile:
             hostname = hostname.rstrip("\r\n")
+            if len(hostname) == 0:
+                continue
             if hostname in machineDict:
-                machineDict[hostname].number_of_cores += 1
+                machineDict[hostname] += 1
             else:
-                machineDict[hostname] = Machine(hostname, 1)
+                machineDict[hostname] = 1
 
     # Convert accumulated dictionary to a MachineList
     machineList = MachineList()
-    for m in list(machineDict.values()):
-        machineList.add(m)
+    for m in list(machineDict):
+        machine = Machine(m, machineDict[m])
+        machineList.add(machine)
     return machineList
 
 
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/scheduler/load_machines.py
```diff
@@ -118,21 +118,18 @@ def _parse_host_info(host_info):
...
-    ):
-        # Filenames generally shouldn't have ':',
-        # so assume it's a string list and parse accordingly
-        sMod = 1 if host_info[0] == "[" else 0
-        sBeg = sMod
-        sEnd = len(host_info) - sMod
-        machine_data = host_info[sBeg:sEnd].split(",")
-    else:
+    print(host_info)
+    if "\\" in host_info or "/" in host_info or "." in host_info:
```
mkundu1, line 122: I think [Path.is_file()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_file) can be used here

src/ansys/fluent/core/scheduler/load_machines.py
```diff
@@ -118,21 +118,18 @@ def _parse_host_info(host_info):
...
-        and not "/" in host_info
-    ):
-        # Filenames generally shouldn't have ':',
-        # so assume it's a string list and parse accordingly
-        sMod = 1 if host_info[0] == "[" else 0
-        sBeg = sMod
-        sEnd = len(host_info) - sMod
-        machine_data = host_info[sBeg:sEnd].split(",")
-    else:
+    print(host_info)
```
mkundu1, line 121: Use `LOG.ino`

Adam-Boutin: Print line was a mistake I forgot to remove it. Thank you for catching it.

src/ansys/fluent/core/scheduler/load_machines.py
Adam-Boutin: I chose to use os.path.exists(host_info) as OS is already inported. Functioning in this way I removed the test_file_not_found. Would it make sense to add the test back or something similar?

src/ansys/fluent/core/scheduler/load_machines.py
```diff
@@ -118,21 +118,15 @@ def _parse_host_info(host_info):
...
+    if os.path.exists(host_info):
+        # Only opens a file if it exists
         with open(host_info, "r") as f:
-            machine_data = f.read().splitlines()
+            host_info = f.read()
+
+    sMod = 1 if host_info[0] == "[" else 0
+    sBeg = sMod
+    sEnd = len(host_info) - sMod
+    machine_data = host_info[sBeg:sEnd].split(",")
```
seanpearsonuk, line 129: @dnwillia-work the change to this function look like a solid improvement.

src/ansys/fluent/core/scheduler/load_machines.py
```diff
@@ -260,9 +254,11 @@ def _construct_machine_list_uge(host_filename):
         peReader = csv.reader(peFile, dialect="pemachines")
         for row in peReader:
             if len(row) == 0:
-                break
-            m = Machine(row[0], int(row[1]), row[2], None if len(row) == 4 else row[3])
-            machineList.add(m)
+                continue
+            if len(row) == 1:
+                row.append(1)
+            row[1] = int(row[1])
```
seanpearsonuk, line 260:
```suggestion
            else:
                row[1] = int(row[1])
```

src/ansys/fluent/core/scheduler/load_machines.py
seanpearsonuk: @Adam-Boutin : I agree with @mkundu1 that the more explicit query is preferable.

Adam-Boutin: @seanpearsonuk I am unsure of what you mean. Would you be able to give me some more info?

src/ansys/fluent/core/scheduler/load_machines.py
dnwillia-work: Yeah, if you can remove it that would be great.

src/ansys/fluent/core/scheduler/load_machines.py
dnwillia-work: It's the suggestion made by mkundu1.  You can replace the entire logic with a call to `Path.is_file()` because it's just trying to work out if `host_info` is a file or not.   I think you have to construct the `Path` object to do that.

tests/test_scheduler.py
```diff
@@ -249,6 +274,17 @@ def test_winhpc(self):
         self.assertEqual(fluentOpts, "-t32 -cnf=M0:8,M1:8,M2:16")
         del os.environ["CCP_NODES"]
 
+    def test_slurm_single_num(self):
+        os.environ["SLURM_JOB_NODELIST"] = "M[1-2],M[3]"
```
dnwillia-work, line 278: Interesting, was this use case not working before?  Or, was it working but returned the wrong result?

dnwillia-work (approved):

tests/test_scheduler.py
Adam-Boutin: There was not a test built for this case. It was in load_machine.py. I did have to make a small change in the regex (lines 336 and 375) to get the correct result.

src/ansys/fluent/core/scheduler/load_machines.py
Adam-Boutin: What would be the difference between using Path(host_info).is_file() vs. os.path.exists(host_info)?

Adam-Boutin: I switched it over to `Path(host_info).is_file()`. I saw in "test_scheduler.py" `os.path.isfile()` do each of these methods have a specific use case or would both `.is_file()` and `.isfile()` be interchangeable? 

Is there much of a downside to importing more modules?

mkundu1: Both `Path.is_file()` and `os.path.isfile()` are interchangeable ([doc](https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module)). Using the Path module is the modern usage, so that will be preferred in general.

dnwillia-work: > I switched it over to `Path(host_info).is_file()`. I saw in "test_scheduler.py" `os.path.isfile()` do each of these methods have a specific use case or would both `.is_file()` and `.isfile()` be interchangeable?
> 
> Is there much of a downside to importing more modules?

There's not really a downside to importing more modules.  As mkundu1 says though, the `PathLib` module is the preferred modern alternative to `os.path`

PR #1481, opened 2023-03-16
Bump version 0.12.5

mkundu1:
Added vpn network fix from main

### DIFF
--- a/src/ansys/fluent/core/_version.py
+++ b/src/ansys/fluent/core/_version.py
@@ -6,7 +6,7 @@ version_info = 0, 1, 'dev0'
 """
 
 # major, minor, patch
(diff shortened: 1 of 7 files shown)


src/ansys/fluent/core/_version.py
```diff
@@ -6,7 +6,7 @@
 """
 
 # major, minor, patch
-version_info = 0, 12, 3
+version_info = 0, 12, 5
```
mkundu1, line 9: This was not incremented during 0.12.4 release (which is in github but not in pypi)

hpohekar (approved):

prmukherj (approved):

Issue #1259, opened 2023-01-12
Assess logging framework


seanpearsonuk:
Implement module level logging using the native Python logging library. For each module, a new child logger will be initialized in the __init__.py file and its configuration will be loaded by a corresponding yaml file. Each child logger will have a "pyfluent-" suffix to prevent conflicts with other loggers created by other pyansys libraries. The logging for pyfluent will be initialized upon import.

For more information regarding the native python module and how it can be integrated in a python project, refer to: https://docs.python.org/3/howto/logging-cookbook.html

seanpearsonuk:
Completed

cj-hodgson:
@millerj97 Please could you check this.

millerj97:
@cj-hodgson opened issue [1879] (https://github.com/ansys/pyfluent/issues/1879)

millerj97:
@cj-hodgson testing carried out:
- set global logging level to all available int and str options as detailed in https://docs.python.org/3/library/logging.html#logging-levels
- set global level to default debug and individually set each logging type available in;
````
pyfluent.logging.list_loggers()
['pyfluent.general', 'pyfluent.launcher', 'pyfluent.networking', ...]
````
- results example;
````
###### Setting the pyfluent.networking level to 20

###### All levels are now
<Logger pyfluent (WARNING)>
<Logger pyfluent.networking (INFO)>
<Logger pyfluent.general (DEBUG)>
<Logger pyfluent.datamodel (DEBUG)>
<Logger pyfluent.tui (DEBUG)>
<Logger pyfluent.launcher (DEBUG)>
<Logger pyfluent.settings_api (DEBUG)>
<Logger pyfluent.post_objects (DEBUG)>
````
- Compare outputs between Linux and Windows and ensure correct settings applied
- for each global logging level run some fluent processes
    - inspect log file is created when required
    -  check logging level matches current settings and above only
- repeat the above process using the environment variable PYFLUENT_LOGGING 
- repeat above steps on Linux

PR #1491, opened 2023-03-22
Feat/logging overhaul

alikasim72:
Implementation of #1259. Replaced the old logging wrapper around the native python logging module.


seanpearsonuk:
@alikasim72 I haven't read every single line, but it looks good. I also haven't looked into the CI. Are these random issues again, or something triggered by the changes?

alikasim72:
@seanpearsonuk these were some random minor errors caused by me using absolute file location, fixed it by using the paths from the os library.

src/ansys/fluent/core/logging_config.yaml
mkundu1, line 1: Could you please check if this is packaged? You may need to add this in MANIFEST.in

alikasim72: Good Point!

seanpearsonuk (approved):

src/ansys/fluent/core/__init__.py
```diff
@@ -44,44 +44,16 @@ def version_info() -> str:
...
-
-def disable_logging_to_file() -> None:
-    """Disable logging to file."""
-    LOG.disable_logging_to_file()
+# Load the logging configuration from a YAML file
+with open(yaml_path, "rt") as f:
+    config = yaml.safe_load(f)
 
+# Configure the logging system
+logging.config.dictConfig(config)
```
seanpearsonuk, line 56: @alikasim72 I looked into the question of what happens if this method is called more than once (we are just one library):
https://stackoverflow.com/a/47496535

So, different configs cannot play together?

mkundu1 (approved):

PR #1492, opened 2023-03-23
Meshing Queries

hpohekar:
We are implementing Meshing Queries as rpcs.


seanpearsonuk:
@hpohekar 

Looks good so far. Are your next steps as follows:
- add queries object to the meshing sessions
- add a basic test
?

hpohekar:
> @hpohekar
> 
> Looks good so far. Are your next steps as follows:
> 
> * add queries object to the meshing sessions
> * add a basic test
>   ?

Yes. We will compare results of `SchemeEval` and queries object.

tests/test_meshing_queries.py
```diff
@@ -0,0 +1,123 @@
...
+    ) == meshing_session.scheme_eval.scheme_eval("(get-zones-of-group 'inlet)")
+
+    assert meshing_session.meshing_queries.GetFaceZonesOfFilter(
+        "*"
+    ) == meshing_session.scheme_eval.scheme_eval("(get-face-zones-of-filter '*)")
+
+    assert meshing_session.meshing_queries.GetCellZonesOfFilter("*") == [3460]
+    assert meshing_session.scheme_eval.scheme_eval("(get-cell-zones-of-filter '*)") == (
+        3460,
+    )
```
hpohekar, line 33: @seanpearsonuk @mkundu1 We can not compare all results as it is. We are getting correct results but with different return types. Hence we have compared above 2 same results separately.

tests/test_meshing_queries.py
```diff
@@ -0,0 +1,123 @@
...
+
+    assert meshing_session.meshing_queries.GetEdgeZonesOfFilter("*") == []
```
hpohekar, line 35: same is applicable here.

tests/test_meshing_queries.py
```diff
@@ -0,0 +1,123 @@
...
+    assert (
+        meshing_session.scheme_eval.scheme_eval("(get-edge-zones-of-filter '*)") == None
+    )
```
hpohekar, line 38: same is applicable here.

tests/test_meshing_queries.py
```diff
@@ -0,0 +1,123 @@
...
+
+    assert meshing_session.meshing_queries.GetNodeZonesOfFilter(
+        "*"
+    ) == meshing_session.scheme_eval.scheme_eval("(get-node-zones-of-filter '*)")
+
+    assert meshing_session.meshing_queries.GetObjectsOfType("mesh") == ["elbow-fluid"]
```
hpohekar, line 44: Here scheme returns `(elbow-fluid,)` hence it is not comparable directly.

tests/test_meshing_queries.py
```diff
@@ -0,0 +1,123 @@
...
+    # assert meshing_session.scheme_eval.scheme_eval("(get-objects-of-type 'mesh)") == (elbow-fluid,)
```
hpohekar, line 45: Here scheme returns (elbow-fluid,) hence it is not comparable directly.

mkundu1: @hpohekar Is `(elbow-fluid,)` a list of symbols? We can write some utility function to convert list/tuple to set and compare the set. If the individual members are symbols (of type `scheme_eval.Symbol`) we can replace with their `Symbol.str` attribute.

hpohekar: I see. Yes, it will save time for remaining rpcs.

hpohekar: I was not aware about use of `scheme_eval.Symbol`. Thank you.

tests/test_meshing_queries.py
mkundu1: In scheme_eval serve-side, scheme nil `()` is converted to empty gRPC message which is converted to `None` in Python. This is correct behaviour as scheme_eval is a generic interface. Whereas `GetEdgeZonesOfFilter` should always return a list.

I think what you've done here is very clear and fine.

tests/test_meshing_queries.py
seanpearsonuk: In fact, we do not *have to* incorporate Scheme code into these tests - bear in mind that we are really testing an API independently of its implementation. It's acceptable to get all the expected results up front and assert them in the test hard-coded.

hpohekar: Done. Thank you very much. It will save lot of time.

prmukherj (approved):

mkundu1 (approved):

PR #1513, opened 2023-03-30
Fix transcript behavior.

prmukherj:
transcript.start()  # write to stdout only

transcript.start(file_path="a.trn")  # write to file only

transcript.start(file_path="a.trn", write_to_stdout=True)  # write to both file and stdout


mkundu1 (approved):
Looks good, thanks. Please check if all 3 cases are covered in unittests before merging.

hpohekar (approved):

PR #1559, opened 2023-04-25
Fix bug in wildcard - flobject.py

prmukherj:
```
>>> [x for x in solver.setup.boundary_conditions.velocity_inlet]          
['inlet2', 'inlet1']
>>> [x for x in solver.setup.boundary_conditions.velocity_inlet['\*let*']]
TypeError: 'WildcardPath' object is not iterable
```

The above behavior is fixed.

Similar changes have been applied on the PyConsole side as well.

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -584,7 +584,7 @@ class WildcardPath(Group):
     def items(self):
         """Items."""
         for key, value in self._parent.items():
-            if fnmatch.fnmatch(key, self._path.split("/")[-1]):
+            if fnmatch.fnmatch(key, self._path.rsplit(sep="/", maxsplit=1)[-1]):
                 yield key, value
 
     def __iter__(self):
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -589,7 +589,7 @@ class WildcardPath(Group):
 
     def __iter__(self):
         for item in self._parent:
-            if fnmatch.fnmatch(item, self._path.split("/")[-1]):
+            if fnmatch.fnmatch(item, self._path.rsplit(sep="/", maxsplit=1)[-1]):
                 yield item
 
     def to_scheme_keys(self, value):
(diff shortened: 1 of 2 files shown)


hpohekar (approved):

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -572,13 +573,25 @@ def __getattr__(self, name: str):
...
         raise KeyError(
             allowed_name_error_message(
                 "Settings objects", name, self.get_object_names()
             )
         )
 
+    def items(self):
+        """Items."""
+        for key, value in self._parent.items():
+            if fnmatch.fnmatch(key, self._path.split("/")[-1]):
```
seanpearsonuk, line 587:
```suggestion
            if fnmatch.fnmatch(key, self._path.rsplit(sep="/", maxsplit=1)[-1]):
```

seanpearsonuk: (faster)

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -572,13 +573,25 @@ def __getattr__(self, name: str):
...
+                yield key, value
+
+    def __iter__(self):
+        for item in self._parent:
+            if fnmatch.fnmatch(item, self._path.split("/")[-1]):
```
seanpearsonuk, line 592: similar change as above

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -572,13 +573,25 @@ def __getattr__(self, name: str):
...
 
+    def items(self):
+        """Items."""
+        for key, value in self._parent.items():
+            if fnmatch.fnmatch(key, self._path.rsplit(sep="/", maxsplit=1)[-1]):
+                yield key, value
+
+    def __iter__(self):
+        for item in self._parent:
+            if fnmatch.fnmatch(item, self._path.split("/")[-1]):
```
seanpearsonuk, line 592:
```suggestion
            if fnmatch.fnmatch(item, self._path.rsplit(sep="/", maxsplit=1)[-1]):
```

seanpearsonuk (approved):

PR #1570, opened 2023-05-02
Case Reader to parse numerical values with units.

prmukherj:
Case Reader was not able to parse numerical values like: 2[m/s].

That behavior has been fixed.

### DIFF
--- a/src/ansys/fluent/core/filereader/case_file.py
+++ b/src/ansys/fluent/core/filereader/case_file.py
@@ -68,7 +68,12 @@ class InputParameter:
         float
             Numeric value of the Fluent input parameter.
         """
-        return float(self._component(0))
+        try:
+            num_val = float(self._component(0))
+        except ValueError:
+            num_val = float(self._component(0).split("[")[0])
+
+        return num_val
 
     def _component(self, idx: int):
         try:


src/ansys/fluent/core/filereader/case_file.py
```diff
@@ -68,7 +68,12 @@ def numeric_value(self):
         float
             Numeric value of the Fluent input parameter.
         """
-        return float(self._component(0))
+        try:
```
seanpearsonuk, line 71: @prmukherj this doesn't look like a complete fix because it only addresses the value part and not the units. Writing a unit test for this class will demonstrate. It is necessary to change the fundamental behaviour in the `_component` directly because it is there that we have made the assumptions that need to change due to the unforeseen input reported in the defect.

prmukherj: ok, sounds good. Will do it.

tests/test_casereader.py
```diff
@@ -224,6 +224,13 @@ def test_case_reader_input_parameter():
     assert momentum.numeric_value == 12.4
     assert momentum.value == "12.4 [kg m s^-1]"
 
+    velocity = InputParameter(raw_data=(("name", "v"), ("definition", "2[m / s]")))
+
+    assert velocity.name == "v"
+    assert velocity.units == "m / s"
```
seanpearsonuk, line 230: The main case to test is without the spaces, as reported. We don't necessarily anticipate this scenario but if it appeared, why would we retain these spaces?

src/ansys/fluent/core/filereader/case_file.py
```diff
@@ -72,7 +72,7 @@ def numeric_value(self):
 
     def _component(self, idx: int):
         try:
-            return self.value.split(maxsplit=1)[idx]
+            return self.value.split("[", maxsplit=1)[idx].strip()
```
seanpearsonuk, line 75: lstrip("[") is no longer relevant

tests/test_casereader.py
prmukherj: It is resolved as per our discussion.
Thank you.

seanpearsonuk (approved):

ajain-work (approved):

src/ansys/fluent/core/filereader/case_file.py
```diff
@@ -184,16 +190,16 @@ def __init__(self, case_filepath: str = None, project_filepath: str = None):
                     "Please provide a valid fluent project file path"
                 )
         try:
-            if "".join(Path(case_filepath).suffixes) == ".cas.h5":
+            if Path(case_filepath).match("*.cas.h5"):
```
seanpearsonuk, line 193: : )

tejalprabhu01 (approved):
Tested with the problematic case file. works! Thanks!

mkundu1 (approved):

PR #1576, opened 2023-05-04
Fix try 01

prmukherj:


raph-luc (approved):
Works fine locally here, and seems to fix the remote server env var issue as well

PR #1582, opened 2023-05-05
Fix/examples save path

raph-luc:
Related to this issue: https://github.com/pyansys/pyfluent/issues/1393 (see my comment there as well)

A few other changes:
- In downloads.py the variable `resp` was unused, also not used anywhere else that I could find in pyansys, so I removed it
- Changed `string.replace` to `re.sub("...$")` to force it to remove only the extension, and not potential `.zip` in the middle of the file name (like `test.zip.h` or something of the sort). `re` is also used elsewhere in pyansys, so it is not a completely new import

Notes/thoughts:
- urllib.request.urlretrieve is ported from python2 and may become deprecated in the future (see here https://docs.python.org/3/library/urllib.request.html#urllib.request.urlretrieve)
- maybe add output into the console explaining what is going on? Like " downloading file..." or "file already exists"
- as discussed with @mkundu1 this does not solve the underlying behavior that Fluent will by default save output files based on where it read input files from, but I think this makes sense and can be expected by the user. At least now pyfluent users can specify where to download example files to (and so Fluent will save its output there as well in this basic example workflow).

### DIFF
--- a/src/ansys/fluent/core/examples/downloads.py
+++ b/src/ansys/fluent/core/examples/downloads.py
@@ -52,7 +52,7 @@ def _retrieve_file(url: str, filename: str, save_path: Optional[str] = None) -> 
     # First check if file has already been downloaded
     if os.path.isfile(local_path_no_zip) or os.path.isdir(local_path_no_zip):
         logging.info("File already exists.")
-        logging.info(f"File path: {local_path_no_zip}")
+        logging.info(f"File path: {os.path.abspath(local_path_no_zip)}")
         return local_path_no_zip
 
     logging.info("Downloading specified file...")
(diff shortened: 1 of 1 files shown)


mkundu1 (approved):
Can you please add the logging calls for Note 2?

hpohekar (approved):

raph-luc:
> Can you please add the logging calls for Note 2?

Will do

raph-luc:
Added info logging calls, and removed get_ext as it was only used to verify the extension of the file name/path, which I believe is best done with a simple `string.endswith()` to avoid getting stuck with potential `zip` in the middle of the file path. 
This is what the logging looks like right now, input:
image
Output:
image

mkundu1 (approved):

PR #1597, opened 2023-05-15
allow allowed_values() for strings with has_allowed_values = True flag 

hpohekar:
Allow `allowed_values()` for strings and string-lists with `has_allowed_values = True` https://github.com/ansys/pyfluent/issues/1613

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -1121,7 +1121,7 @@ def get_cls(name, info, parent=None, version=None):
             bases = bases + (_CreatableNamedObjectMixin,)
         elif obj_type == "named-object":
             bases = bases + (_NonCreatableNamedObjectMixin,)
-        elif info.get("has-allowed-values") or version == "222":
+        elif info.get("has_allowed_values") or version == "222":
             bases += (_HasAllowedValuesMixin,)
 
         cls = type(pname, bases, dct)
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -1121,12 +1121,7 @@ def get_cls(name, info, parent=None, version=None):
             bases = bases + (_CreatableNamedObjectMixin,)
         elif obj_type == "named-object":
             bases = bases + (_NonCreatableNamedObjectMixin,)
-        elif info.get("has-allowed-values", False) and version in [
-            "222",
-            "231",
-            "232",
-            "241",
-        ]:
+        elif info.get("has-allowed-values") or version == "222":
             bases += (_HasAllowedValuesMixin,)
 
         cls = type(pname, bases, dct)
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -161,7 +161,7 @@ class Base:
                 attr_type_or_types = (attr_type_or_types,)
             if isinstance(val, attr_type_or_types):
                 return val
-            if val != [] and any(
+            if val is not None and any(
                 issubclass(x, bool) for x in attr_type_or_types
             ):  # cast to bool for boolean attributes
                 return bool(val)
@@ -215,12 +215,31 @@ class Numerical(Property):
 class Textual(Property):
     """Exposes attribute accessor on settings object - specific to string objects."""
 
+    _items_with_string_and_allowed_values = set()
+
+    def _filter_items_with_string_and_allowed_values(self, data):
+        for key, value in data.items():
+            if isinstance(value, dict):
+                if (
+                    "type" in value
+                    and "has_allowed_values" in value
+                    and value["type"] == "string"
+                    and value["has_allowed_values"] is True
+                ):
+                    self._items_with_string_and_allowed_values.add(key)
+                self._filter_items_with_string_and_allowed_values(value)
+
     def allowed_values(self):
         """Get the allowed values of the object."""
-        try:
-            return self.get_attr("allowed-values", (list, str))
-        except BaseException as ex:
-            return []
+
+        static_info = self.flproxy.get_static_info()
+        self._filter_items_with_string_and_allowed_values(static_info)
+
+        if self.obj_name in self._items_with_string_and_allowed_values:
+            try:
+                return self.get_attr("allowed-values", (list, str))
+            except BaseException as ex:
+                return []
 
 
 class SettingsBase(Base, Generic[StateT]):
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -165,7 +165,7 @@ class Base:
                 issubclass(x, bool) for x in attr_type_or_types
             ):  # cast to bool for boolean attributes
                 return bool(val)
-            return []
+            return None
         return val
 
     def is_active(self) -> bool:
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -1121,7 +1121,12 @@ def get_cls(name, info, parent=None, version=None):
             bases = bases + (_CreatableNamedObjectMixin,)
         elif obj_type == "named-object":
             bases = bases + (_NonCreatableNamedObjectMixin,)
-        elif info.get("has-allowed-values") and version in ["222", "231", "232", "241"]:
+        elif info.get("has-allowed-values", False) and version in [
+            "222",
+            "231",
+            "232",
+            "241",
+        ]:
             bases += (_HasAllowedValuesMixin,)
 
         cls = type(pname, bases, dct)
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -1121,7 +1121,7 @@ def get_cls(name, info, parent=None, version=None):
             bases = bases + (_CreatableNamedObjectMixin,)
         elif obj_type == "named-object":
             bases = bases + (_NonCreatableNamedObjectMixin,)
-        elif info.get("has_allowed_values") or version == "222":
+        elif info.get("has_allowed_values"):
             bases += (_HasAllowedValuesMixin,)
 
         cls = type(pname, bases, dct)
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -217,7 +217,10 @@ class Textual(Property):
 
     def allowed_values(self):
         """Get the allowed values of the object."""
-        return self.get_attr("allowed-values", (list, str))
+        try:
+            return self.get_attr("allowed-values", (list, str))
+        except BaseException as ex:
+            return []
 
 
 class SettingsBase(Base, Generic[StateT]):
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -215,32 +215,6 @@ class Numerical(Property):
 class Textual(Property):
     """Exposes attribute accessor on settings object - specific to string objects."""
 
-    _items_with_string_and_allowed_values = set()
-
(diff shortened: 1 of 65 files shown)


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -217,7 +217,10 @@ class Textual(Property):
 
     def allowed_values(self):
         """Get the allowed values of the object."""
-        return self.get_attr("allowed-values", (list, str))
+        try:
+            return self.get_attr("allowed-values", (list, str))
```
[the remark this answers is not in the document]
hpohekar: Right.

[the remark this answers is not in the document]
hpohekar: Done. Thank you.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -161,11 +161,11 @@ def get_attr(self, attr, attr_type_or_types=None) -> Any:
                 attr_type_or_types = (attr_type_or_types,)
             if isinstance(val, attr_type_or_types):
                 return val
-            if val is not None and any(
+            if val is not [] and any(
```
seanpearsonuk, line 164:
```
>>> [] is []
False
```

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -152,7 +152,7 @@ def get_attr(self, attr, attr_type_or_types=None) -> Any:
             attrs = attrs.get("attrs", attrs)
         if attr != "active?" and attrs and attrs.get("active?", True) is False:
             raise RuntimeError("Object is not active")
-        val = None
+        val = []
```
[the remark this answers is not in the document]
hpohekar: @seanpearsonuk 

It should be `None` not `[ ]`, otherwise it returns `[ ]` even for bool attribute type. It's corrected now. Thank you.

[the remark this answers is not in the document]
hpohekar: If `val = [ ]`

image

`-------------------------------------`

If `val = None`

image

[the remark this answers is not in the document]
hpohekar: Here we are getting expected behavior.

image

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -161,11 +161,11 @@ def get_attr(self, attr, attr_type_or_types=None) -> Any:
                 attr_type_or_types = (attr_type_or_types,)
             if isinstance(val, attr_type_or_types):
                 return val
-            if val is not None and any(
+            if val != [] and any(
                 issubclass(x, bool) for x in attr_type_or_types
             ):  # cast to bool for boolean attributes
                 return bool(val)
-            return None
+            return []
```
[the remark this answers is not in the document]
hpohekar: We want to return '[ ]` instead of 'None` in following case right ? 

image

tests/test_flobject.py
```diff
@@ -662,10 +662,7 @@ def test_accessor_methods_on_settings_object(load_static_mixer_case):
 
     existing = solver.file.read.file_type.get_attr("read-only?", bool)
     modified = solver.file.read.file_type.is_read_only()
-    if solver.get_fluent_version() < "23.2.0":
-        assert existing == modified
-    else:
-        assert existing == None and modified == False
+    assert existing == modified
```
hpohekar, line 665: @seanpearsonuk Here we have resolved 1 bug. Earlier it was returning `None` and `False` for same Boolean class type. Now both returns False.

image

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -215,9 +215,31 @@ def max(self):
...
-        return self.get_attr("allowed-values", (list, str))
+
+        static_info = self.flproxy.get_static_info()
+        self._filter_items_with_string_and_allowed_values(static_info)
+
+        if self.obj_name in self._items_with_string_and_allowed_values:
+            try:
+                return self.get_attr("allowed-values", (list, str))
+            except BaseException as ex:
+                return []
```
hpohekar, line 242: We check that if obj_name is in `items_with_string_and_allowed_values` then only we return allowed_values.

src/ansys/fluent/core/solver/flobject.py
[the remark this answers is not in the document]
h-krishnan: What about other attributes? For example, display-text?

src/ansys/fluent/core/solver/flobject.py
h-krishnan: I thought this would be implemented via mixin approach.

src/ansys/fluent/core/solver/flobject.py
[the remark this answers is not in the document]
hpohekar: We have restored original changes. We will discuss this soon.

src/ansys/fluent/core/solver/flobject.py

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1068,8 +1064,35 @@ def __setitem__(self, name: str, value):
...
+
+
+def _filter_items_with_string_and_allowed_values(static_info):
+    """Get strings with flag has_allowed_values=True"""
+    for key, value in static_info.items():
+        if isinstance(value, dict):
+            if (
+                "type" in value
+                and "has_allowed_values" in value
+                and value["type"] == "string"
```
seanpearsonuk, line 1077: @hpohekar are we also dealing with string lists?

hpohekar: No.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1115,6 +1138,8 @@ def get_cls(name, info, parent=None, version=None):
             bases = bases + (_CreatableNamedObjectMixin,)
         elif obj_type == "named-object":
             bases = bases + (_NonCreatableNamedObjectMixin,)
+        elif name in _items_with_string_and_allowed_values:
+            bases = bases + (_StringWithAllowedValuesMixin,)
```
h-krishnan, line 1143: Do you really require `_filter_items_with_string_and_allowed_values`? Can you not just do the following here:
``` python
if info.get('has-allowed-values'):
  bases += (_HasAllowedValuesMixin,)
```

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1068,8 +1064,35 @@ def __setitem__(self, name: str, value):
...
+                "type" in value
+                and "has_allowed_values" in value
+                and value["type"] == "string"
+                and value["has_allowed_values"] is True
+            ):
+                _items_with_string_and_allowed_values.add(key)
+            _filter_items_with_string_and_allowed_values(value)
+
+
+class _StringWithAllowedValuesMixin(Property):
```
h-krishnan, line 1084: Maybe this can be renamed as `_HasAllowedValuesMixin` and no need to derive from `Property`.

src/ansys/fluent/core/solver/flobject.py
hpohekar: Yes, it is required. If we modify according to above suggestion then following test fails.

image

h-krishnan: At least in PyConsole mode, this seems to be sufficient:
``` diff
diff --git a/fluent/cortex/pylib/flapi/flobject.py b/fluent/cortex/pylib/flapi/flobject.py
index eb2cb43b9a1..aec98d54045 100644
--- a/fluent/cortex/pylib/flapi/flobject.py
+++ b/fluent/cortex/pylib/flapi/flobject.py
@@ -215,6 +215,8 @@ class Numerical(Property):
 class Textual(Property):
     """Exposes attribute accessor on settings object - specific to string objects."""

+
+class _HasAllowedValuesMixin:
     def allowed_values(self):
         """Get the allowed values of the object."""
         return self.get_attr("allowed-values", (list, str))
@@ -1112,6 +1114,11 @@ def get_cls(name, info, parent=None, version=None):
         elif obj_type == "named-object":
             bases = bases + (_NonCreatableNamedObjectMixin,)

+        if obj_type in ['string'] and 'has-allowed-values' not in info:
+           print ('name =', name, 'info = ', info)
+        if info.get("has-allowed-values", False):
+            bases = bases + (_HasAllowedValuesMixin,)
+
         cls = type(pname, bases, dct)

         taboo = set(dir(cls))

Comment (h-krishnan) [reply src/ansys/fluent/core/solver/flobject.py:1143]: image

Comment (hpohekar) [reply src/ansys/fluent/core/solver/flobject.py:1125]: Done. Corrected above code. Thank you.

**Comment by hpohekar** [comment]
> > Allow `allowed_values()` only for strings with `has_allowed_values = True` #1613
> 
> Is the requirement to allow it for strings only?

Current modification will handle both `string` and `string-list`.

File: src/ansys/fluent/core/solver/flobject.py [hunk 10]
```diff
@@ -1115,6 +1121,8 @@ def get_cls(name, info, parent=None, version=None):
             bases = bases + (_CreatableNamedObjectMixin,)
         elif obj_type == "named-object":
             bases = bases + (_NonCreatableNamedObjectMixin,)
+        elif info.get("has-allowed-values") and version == "222":
```
h-krishnan, line 1124: Should it be `or version == "222"`?

h-krishnan (approved):

seanpearsonuk:
I'll hold of approving while you fox for the test failure

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1115,6 +1121,13 @@ def get_cls(name, info, parent=None, version=None):
             bases = bases + (_CreatableNamedObjectMixin,)
         elif obj_type == "named-object":
             bases = bases + (_NonCreatableNamedObjectMixin,)
+        elif info.get("has-allowed-values", False) and version in [
+            "222",
+            "231",
+            "232",
+            "241",
```
h-krishnan, line 1128: Sorry, I am totally confused now. Do we need to add every future release in this list from now on?

hpohekar: We tried this earlier, `elif info.get("has-allowed-values") and version in ["222", "231", "232", "241"]:` but this is also not working. It is failing again.

hpohekar: Do we need to add every future release in this list from now on? - Not sure, yet. It is not working for now.

hpohekar: @h-krishnan @seanpearsonuk 

We have observed following scenarios - 

1. `elif info.get("has-allowed-values") or version == "222":` 

`Build` check fails.

`--------------------------------------------------------------`

2. `elif info.get("has-allowed-values") and version in ["222", "231", "232", "241"]:` 

`Build` check passes but test fails in all versions.

image

`--------------------------------------------------------------`

2. `elif info.get("has-allowed-values", False) and version in ["222", "231", "232", "241"]:` 

`Build` check passes but test fails in all versions.

image

`--------------------------------------------------------------`

hpohekar: `elif name in _items_with_string_and_allowed_values:` is working fine for all versions.

tests/test_flobject.py
```diff
@@ -901,3 +901,19 @@ def get_child_nodes(node, nodes, type_list):
                 nodes[node_type] = node
                 if not type_list:
                     return
+
+
+@pytest.mark.fluent_241
```
h-krishnan, line 906: I don't think this needs to be marked as fluent_241 specific test. It should work for 23.2 (and may be 23.1) at least.

hpohekar:
> I'll hold of approving while you fox for the test failure

@seanpearsonuk  All tests are passing now.

tests/test_flobject.py
hpohekar: @h-krishnan Thanks very much for resolving all issues.

seanpearsonuk (approved):

PR #1606, opened 2023-05-16
Slight rewording change for version check

raph-luc:
Minor change. While testing different Fluent releases, I got this error message:
image

However, Fluent version 24.1 does exist in my machine, so I got confused for a second there. I believe what the original dev meant is that it is not yet supported by pyfluent.

With the proposed changes, the message now reads:
image

Additional thought probably for a different PR: do we need to block the execution of more recent versions than the latest fully supported Fluent release? Shouldn't more recent versions (always? usually?) work fine from now on? Perhaps just a warning that the recent/dev version is untested and subject to potential issues?

### DIFF
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -52,9 +52,9 @@ class FluentVersion(Enum):
                     return FluentVersion(version)
             else:
                 raise RuntimeError(
-                    f"The passed version '{version[:-2]}' does not exist."
-                    f" Available version strings are: "
-                    f"{[ver.value for ver in FluentVersion]} "
+                    f"The specified version '{version[:-2]}' is not supported."
+                    + f" Supported versions include: "
+                    + ", ".join([ver.value for ver in FluentVersion][::-1])
                 )
 
     def __str__(self):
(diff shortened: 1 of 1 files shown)


src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -52,9 +52,9 @@ def _missing_(cls, version):
                     return FluentVersion(version)
             else:
                 raise RuntimeError(
-                    f"The passed version '{version[:-2]}' does not exist."
-                    f" Available version strings are: "
-                    f"{[ver.value for ver in FluentVersion]} "
+                    f"The specified version '{version[:-2]}' is not supported."
+                    + f" Supported versions include: "
```
seanpearsonuk, line 56: "include" suggests that what follows might be a subset of all supported versions.

raph-luc: You are right, that was my intention, considering that pyfluent does work for example with Fluent v241 currently.

However, as the code blocks execution if the version is not on that list, I am going to change the word 'include' to 'are' as I agree it is more appropriate currently.

seanpearsonuk (approved):

Issue #1588, opened 2023-05-11
Loading pyfluent into memory immediately creates unexpected folders


raph-luc:
Two relatively minor issues: 
1) When the user loads or imports with `import ansys.fluent.core` into memory, a `%LocalAppData%\ansys_fluent_core\ansys_fluent_core\examples` directory on Windows, and `~/.local/share/ansys_fluent_core` on Linux, is immediately created, regardless of what the user intends to do with `pyfluent`. I think this is unexpected, and as far as I can tell unnecessary behavior. 

2) Note that on Windows it nests a `ansys_fluent_core` dir inside another `ansys_fluent_core`, which I think is not what the original developer intended. This is caused by the default behavior of appdirs as described in https://github.com/ActiveState/appdirs/blob/master/appdirs.py#L44

A better approach might be to, rather than create these directories when pyfluent is initialized, to create them only when the user starts working with the examples. I am not yet aware of other pyfluent functions that rely on these directories being preemptively created.

And, rather than saving to `ansys_fluent_core\ansys_fluent_core`, save to `Ansys\ansys_fluent_core` by default on Windows, which is how typical Windows apps would do this in my understanding.

mkundu1:
Shall we create a new pyansys package for this module? pymapdl is also using the same code.

raph-luc:
Not sure, I would not think that a separate package is required, as long as all pyansys packages like pyfluent and pymapdl use the same naming scheme i.e. `user_data_dir(appauthor='Ansys', appname=appname)` or something like that. Currently I believe they already don't use the same naming scheme for this, regardless of changes here

For example, 
- `pyprimemesh` uses `user_data_dir(appname='pyprimemesh', appauthor=False)`
- `pymapdl` uses `user_data_dir("ansys_mapdl_core")` which is going to result in the repeated dirs in Windows mentioned above `%LocalAppData%\ansys_mapdl_core\ansys_mapdl_core\examples`
- `ansys-tools-path` uses `user_data_dir("ansys_tools_path")` as well with the repeated dirs in Windows

Maybe we should all be using `ansys-tools-path` and update the naming scheme there to have a root local apps Ansys folder?

dnwillia-work:
It's open source?  Open a PR on all three of the repos and point back to this issue.

raph-luc:
Completed

PR #1615, opened 2023-05-17
Update to Windows data path name, and change to when examples folder is created

raph-luc:
Resolves issue https://github.com/ansys/pyfluent/issues/1588 for pyfluent

Changes the point of creation of the data and examples paths from when pyfluent is first imported, to when the example files are downloaded.

In addition, before these changes Windows was saving to `%LocalAppData%\ansys_fluent_core\ansys_fluent_core\example`, after these changes it will save to `%LocalAppData%\Ansys\ansys_fluent_core\example`. The root directory `%LocalAppData%\Ansys` is already created when installing Ansys Fluent releases on Windows.

These changes do not influence the default data paths on Linux, only changes them on Windows.

To make things easier for CI/testing or other purposes in the future, we could make `__init__.py` check for an existing env variable for the data paths, like I noticed `ansys/pyprimemesh` does. Let me know if anyone thinks this would be useful.

I am going to suggest the same naming change to other pyansys repositories so we all potentially use the same naming scheme and root Ansys folder, and avoid cluttering the user's local app data folder on Windows (and see if they care about folders being created on import regardless of whether they will actually be used or not).

### DIFF
--- a/src/ansys/fluent/core/__init__.py
+++ b/src/ansys/fluent/core/__init__.py
@@ -56,17 +56,8 @@ with open(yaml_path, "rt") as f:
 logging.config.dictConfig(config)
 
 # Setup data directory
-try:
-    USER_DATA_PATH = appdirs.user_data_dir("ansys_fluent_core")
-    if not os.path.exists(USER_DATA_PATH):
-        os.makedirs(USER_DATA_PATH)
-
-    EXAMPLES_PATH = os.path.join(USER_DATA_PATH, "examples")
-    if not os.path.exists(EXAMPLES_PATH):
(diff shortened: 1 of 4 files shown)


mkundu1 (approved):

raph-luc:
`start_fluent_container()` in `fluent_container.py` was not checking for existing directories before creating tempfile, and was failing with weird errors due to that when running the tests. Should now be fixed

hpohekar (approved):

prmukherj (approved):

seanpearsonuk (approved):

Issue #1553, opened 2023-04-20
Add interfaces to allow integration of PyFluent and PySystemCoupling


ochernuk:
### 📝 Description of the feature

The motivation is to integrate PyFluent and PySystemCoupling in such a way that PySystemCoupling can orchestrate a co-simulation involving one or more PyFluent solver instances, leveraging the existing connection between Fluent and System Coupling. In order to do that, changes are needed in both PyFluent and PySystemCoupling repositories. PySystemCoupling should be able to receive a handle to the PyFluent solver (session) object and to:

- issue queries via a generic interface to get the System Coupling-specific setup information related to this participant:
  - Regions involved in System Coupling analysis (and their properties, such as topology, etc.)
  - Variables involved in System Coupling analysis (and their properties, such as tensor type, etc.)
  - Whether the Fluent analysis is steady or transient
- invoke commands to establish connection between Fluent and System Coupling and to perform a coupled solve
  - Some of these commands will need to be called asynchronously (perhaps on a separate thread)

It would be preferrable to avoid introducing dependencies between PyFluent and PySystemCoupling and instead to just have PyFluent provide a generic interface ("duck-typing") to PySystemCoupling to perform the above operations.

A corresponding issue will be tracked within the PySystemCoupling repository.

### 💡 Steps for implementing the feature

The proposal is:

1. Add a member to the `ansys.fluent.core.Solver` class of type `ansys.fluent.core.SystemCoupling` and provide access to it, e.g `sc = solver.system_coupling`
2. The `SystemCoupling` class will have the following public interfaces:
  a. `get_variables() -> list`
  b. `get_regions() -> list`
  c. `get_analysis_type() -> str`
  d. `syc_connect(host : str, port : int, name : str)`
  e. `syc_solve()`
3. Internally, `SystemCoupling` class will have a handle to the `Solver` object to access to the following functionality:
  a. To get the list of regions and variables, SystemCoupling will call `solver.file.export.sc_def_file_settings.write_sc_file` to produce the .scp (XML) file, and will read the file and parse it to fill in the regions and variables data structures. For that, it will need to use `xml.etree.ElementTree` package.
  b. To connect to System Coupling, issue `sysc-connect-parallel` scheme command
  c. To solve the coupled analysis, issue `(sc-init-solve)` scheme command

### 🔗 Useful links and references

Corresponding issue in the PySystemCoupling repository: https://github.com/pyansys/pysystem-coupling/issues/148

[...]

PR #1620, opened 2023-05-18
Add system coupling object for integration with PySystemCoupling

ochernuk:
I am putting this up for a reference, related to the discussion here:
https://github.com/ansys/pyfluent/issues/1553


src/ansys/fluent/core/systemcoupling.py
```diff
@@ -0,0 +1,119 @@
...
+        minor_version = int(split_version[1])
+        if major_version >= 24 or (major_version == 23 and minor_version == 2):
+            self._solver.scheme_eval.exec(("(sc-init-solve)",))
+        else:
+            self._solver.scheme_eval.exec(
+                ('(ti-menu-load-string "/solve/initialize/initialize-flow")',)
+            )
+            self._solver.scheme_eval.exec(("(sc-solve)",))
+
+    def __get_syc_setup(self):
```
dnwillia-work, line 75: Do we even need the scp file anymore?  I would think you could do all the configuration directly by calling the API.

src/ansys/fluent/core/systemcoupling.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def get_analysis_type(self):
+        if self._solver.setup.general.get_state()["solver"]["time"] == "steady":
+            return "Steady"
+        else:
+            return "Transient"
+
+    def syc_connect(self, host: str, port: int, name: str):
+        connect_command = f'(%sysc-connect-parallel "{host}" {port} "{name}")'
+        self.scheme_eval.exec((connect_command,))
```
dnwillia-work, line 61: We would need to get the `scheme_eval` calls ripped out of here... Basically want to get it calling the API.

src/ansys/fluent/core/systemcoupling.py
```diff
@@ -0,0 +1,119 @@
...
+        output_variables: List[str]
+
+    def __init__(self, solver):
+        self._solver = solver
+        self.participant_type = "FLUENT"
+
+    # def __getattr__(self, attr):
+    #    return getattr(self.solver, attr)
+
+    def get_variables(self):
```
dnwillia-work, line 47: I would expect this function to be calling the API, not returning a dictionary that is constructed by writing the scp file.

src/ansys/fluent/core/systemcoupling.py
ochernuk: Ok `scheme_eval` is gone now.

src/ansys/fluent/core/systemcoupling.py
ochernuk: It would probably be good to refactor Fluent code so that `write_scp_file` function is implemented using settings APIs. If it's ok, I would like it if this was deferred until the corresponding changes are available in Fluent.

prmukherj (approved):

iboyd-ansys (approved):
Looks good to me.

mkundu1 (approved):

hpohekar (approved):

PR #1604, opened 2023-05-16
Fix product version in launch remote fluent.

prmukherj:
"launch_remote_fluent" does not use the "get_ansys_version()" as a corrected behavior. Instead, we use the product version provided as launch argument. If product version is not set, we use the version "latest".

Issue #1591, opened 2023-05-12
Starting a remote Fluent session is looking for a local Fluent version and ignores the given product version


plule-ansys:
### 🔍 Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### 🐞 Description of the bug

In an environment where PyFluent is starting a remote version of Fluent using PyPIM, PyFluent still looks for a local installation to determine which version to run. This check fails since this is precisely a scenario where there is no local installation.

In more details, the issue is located [here](https://github.com/pyansys/pyfluent/blob/main/src/ansys/fluent/core/launcher/launcher.py#L652), `product_version="".join(get_ansys_version().split("."))[:-1],`.

The correct behavior would be:
- Taking in account the `product_version` passed to `launch_fluent`, ensuring it is transformed in the 3-digit format ("231")
- If unset, there are two possible valid approaches:
  - Leaving to `None`, which is valid for [create_instance](https://pypim.docs.pyansys.com/api/_autosummary/ansys.platform.instancemanagement.Client.create_instance.html#ansys.platform.instancemanagement.Client.create_instance). In that case, the choice will be left up to the PIM server
  - Listing the available versions with [`pim.list_definitions("fluent-<mode>")`](https://pypim.docs.pyansys.com/api/_autosummary/ansys.platform.instancemanagement.Client.list_definitions.html), and pick a compatible one by crossing this information with the compatible Fluent version from `FluentVersion`. This is more complicated but allow for better error reporting, and is more flexible for a server supporting a wide range of versions

### 📝 Steps to reproduce

- Log in to an environment with PIM available such as Ansys Lab
- Execute `!pip install --user ansys-fluent-core==0.13.1`
- Restart the kernel
- Execute the standard launching code:
```python
import ansys.fluent.core as pyfluent
solver_session = pyfluent.launch_fluent(mode="solver")
solver_session.health_check_service.is_serving
```

The startup sequence fails with:

```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Input In [9], in <cell line: 5>()
      3 import ansys.fluent.core as pyfluent
      4 del os.environ["AWP_ROOT231"]
----> 5 solver_session = pyfluent.launch_fluent(mode="solver")
      6 solver_session.health_check_service.is_serving

File ~/.local/lib/python3.9/site-packages/ansys/fluent/core/launcher/launcher.py:652, in launch_fluent(***failed resolving arguments***)
[...]

PR #1639, opened 2023-05-25
Remote launcher: don't provide any product version by default

plule-ansys:
Follow-up to #1604 for the fix for #1591 This is just a small adjustment, I was out so I missed the review, but thank you for the quick fix @prmukherj

When the user does not explicitly request for a version, then `None` should be used. `latest` is not a standard name for a version so it will likely not be available. Leaving `None` will let the backend choose. I will add a note on this in the `pypim` documentation, at the moment it's only explain in the `.proto` I believe.

### DIFF
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -653,7 +653,7 @@ def launch_fluent(
             if product_version:
                 fluent_product_version = "".join(product_version.split("."))[:-1]
             else:
-                fluent_product_version = "latest"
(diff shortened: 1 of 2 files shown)


mkundu1 (approved):

hpohekar (approved):

raph-luc:
`test_transcript` has failed again after what are unrelated changes that I believe shouldn't have affected it, going to try re-running it

raph-luc (approved):

seanpearsonuk (approved):

Issue #1625, opened 2023-05-22
Set global log level in 0.13


ansmhoeij:
### Description of the modifications

With PR #1491 the logger seems to have changed. This is however not reflected in the documentation of [0.13 ](https://fluent.docs.pyansys.com/release/0.13/user_guide/index.html) . The `set_log_level` method does not exist anymore. 

```Python
import ansys.fluent.core as pyfluent
pyfluent.set_log_level("DEBUG") # by default, only errors are shown
```

With 0.13 do we still have control over the global log level? Is the following the intended way to control log levels? 
```Python
import logging
tui_logger = logging.getLogger("ansys.fluent.services.tui")
tui_logger.setLevel("DEBUG")
data_model_logger = logging.getLogger("ansys.fluent.services.datamodel")
data_model_logger.setLevel("DEBUG")
launcher_logger = logging.getLogger("ansys.fluent.services.launcher")
launcher_logger.setLevel("DEBUG")
```

mkundu1:
@ansmhoeij @raph-luc Note that, it is possible to configure the root logger which is set to output to console:
```
>>> import logging
>>> logging.warning("abc")
WARNING:root:abc
>>> logging.error("abc")
ERROR:root:abc
>>> assert logging.root.getEffectiveLevel() == logging.WARNING 
>>> logging.root.setLevel("ERROR")                              
>>> logging.warning("abc")                                      
>>> logging.error("abc")   
ERROR:root:abc
```

cj-hodgson:
@millerj97 Please could you check the doc is correct and that logging is behaving properly.

millerj97:
@cj-hodgson opened issue https://github.com/ansys/pyfluent/issues/1879

millerj97:
@cj-hodgson testing carried out:
- set global logging level to all available int and str options as detailed in https://docs.python.org/3/library/logging.html#logging-levels
- set global level to default debug and individually set each logging type available in;
````
pyfluent.logging.list_loggers()
['pyfluent.general', 'pyfluent.launcher', 'pyfluent.networking', ...]
````
- results example;
````
###### Setting the pyfluent.networking level to 20

###### All levels are now
<Logger pyfluent (WARNING)>
<Logger pyfluent.networking (INFO)>
<Logger pyfluent.general (DEBUG)>
<Logger pyfluent.datamodel (DEBUG)>
<Logger pyfluent.tui (DEBUG)>
<Logger pyfluent.launcher (DEBUG)>
[...]

Issue #9, opened 2022-01-17
Logging Fluent launch errors


mkundu1:
Consider piping stdout/stderr to a listener thread that writes to a logger. That way users can determine why fluent fails to launch. I've tried a similar implementation for MAPDL, but there's an issue with stdout/stderr piping otherwise we would have implemented it; I hope you don't encounter the same issue here.

_Originally posted by @akaszynski in https://github.com/pyansys/pyfluent/pull/7#discussion_r785035347_

PR #1651, opened 2023-05-29
Logging rework

raph-luc:
Closes https://github.com/ansys/pyfluent/issues/1259, https://github.com/ansys/pyfluent/issues/1625, https://github.com/ansys/pyfluent/issues/1532 and I believe also https://github.com/ansys/pyfluent/issues/9 (see discussion below)

Changes:

- Default error output level is now `WARNING` rather than `DEBUG`
- PyFluent loggers now have the `pyfluent_` prefix to avoid conflict with other loggers (also note that using `.` in the logger names is usually for logging propagation, which is not something we are using here)
- All logging output now goes to a single file, example: <details><summary> Logging file example </summary> image </details>
- Logging messages of level `ERROR` and above go to the python console (stdout) as well, example: <details><summary> Warning and error, Python console and file </summary>image </details>
- Added support for `PYFLUENT_LOGGING` environment variable, which we can use so that we don't have to change logging levels manually and so that logging is automatically set to `DEBUG` (or any other level you want) for development rather than the much more silent default `WARNING`
- Added convenience functions `list_loggers()` and `set_global_log_level()` (see documentation in the item below)
- Add documentation:  <details><summary> Screenshots</summary> User guide index: image API reference index (this isn't directly related to Fluent's API, but couldn't think of a better place in the current docs, please let me know of any suggestions): image  </details>

Recommendations:
- I think using root `logging` calls such as `logging.info()` or `logging.warning()` is something we should avoid and could conflict/cause issues with other packages or something that users have set up themselves,  using the custom defined loggers is better in every way that I can think of, so `logger=getLogger('pyfluent_...'); logger.info()`
- Use `print()` when the objective is to output non-error information to the Python console directly for the user to see or when it is an important non-error message and just logging to files is insufficient (we can set up additional logging for different combinations as well if necessary)

In regards to https://github.com/ansys/pyfluent/issues/9, the only scenario where no useful information or silent failing was being returned to the user that I could think of, and based on my experience as this was an issue for me too, was when running Fluent on containers, where it was failing completely silently with no logs or console output at all to help. My suggested fix (relevant lines below) is to set the docker image to run on the specified work directory that is also mounted on the container host system, so that the user, which has access to the host, at least has access to the Fluent automatic transcript output which is typically all that is needed to debug launch issues. I have also changed all the shorthand args for their actual names for the docker run command (check the other lines around the ones linked below), as I believe this makes them much clearer and easier to work with. Please let me know if I missed any scenarios where this isn't sufficient. https://github.com/ansys/pyfluent/blob/3ff4f0078bd66bcc6ccdddcd4b41c12bd9b6fc65/src/ansys/fluent/core/launcher/fluent_container.py#L60-L61

While this is fresh on my mind, please let me know if you have any questions, can think of any suggestions or ways in which logging would be more useful to you for development or any other purposes.

### DIFF
--- a/src/ansys/fluent/core/__init__.py
+++ b/src/ansys/fluent/core/__init__.py
@@ -44,16 +44,82 @@ def version_info() -> str:
     return _VERSION_INFO if _VERSION_INFO is not None else __version__
 
 
+def set_global_log_level(level: str | int) -> None:
+    """Method changing the levels of all PyFluent loggers.
+
+    Parameters
+    ----------
+    level : str or int
+        Specified logging level to set PyFluent loggers to. See logging levels in
+        https://docs.python.org/3/library/logging.html#logging-levels
+
+    Examples
+    --------
+    >>> import ansys.fluent.core as pyfluent
+    >>> pyfluent.set_global_log_level(10)
+
+    or
+
+    >>> pyfluent.set_global_log_level("DEBUG")
+
+    Notes
+    -------
+    By default loggers are set to WARNING.
+    """
+    if isinstance(level, str):
+        if level.isdigit():
+            level = int(level)
+    print(f"Setting PyFluent global logging level to {level}.")
+    pyfluent_loggers = list_loggers()
+    for name in pyfluent_loggers:
+        logging.getLogger(name).setLevel(level)
+
+
+def list_loggers():
+    """List with all PyFluent loggers.
+
+    Returns
+    -------
+    list of str
+        Each list element is a PyFluent logger name that can be controlled with logging.getLogger().
+
+    Examples
+    --------
+    >>> import ansys.fluent.core as pyfluent
+    >>> import logging
+    >>> pyfluent.list_loggers()
+    ['pyfluent_general', 'pyfluent_launcher', 'pyfluent_networking', ...]
+    >>> logger = logging.getLogger('pyfluent_networking')
+    >>> logger.setLevel("DEBUG")
+
+    Notes
+    -------
+    By default loggers are set to WARNING.
+    """
+    logger_dict = logging.root.manager.loggerDict
+    pyfluent_loggers = []
+    for name in logger_dict:
+        if name.startswith("pyfluent"):
+            pyfluent_loggers.append(name)
+    return pyfluent_loggers
+
+
+# Configure the logging system
 file_path = os.path.abspath(__file__)
 file_dir = os.path.dirname(file_path)
 yaml_path = os.path.join(file_dir, "logging_config.yaml")
 
-# Load the logging configuration from a YAML file
 with open(yaml_path, "rt") as f:
     config = yaml.safe_load(f)
 
-# Configure the logging system
 logging.config.dictConfig(config)
+print(f"PyFluent logging file in {os.path.join(os.getcwd(),'pyfluent.log')}")
+env_logging_level = os.getenv("PYFLUENT_LOGGING")
+if env_logging_level:
+    print(
+        "PYFLUENT_LOGGING environment variable found, setting global logging level..."
+    )
+    set_global_log_level(env_logging_level)
 
 # Setup data directory
 USER_DATA_PATH = appdirs.user_data_dir(appname="ansys_fluent_core", appauthor="Ansys")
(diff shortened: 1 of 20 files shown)


src/ansys/fluent/core/__init__.py
```diff
@@ -44,16 +44,82 @@ def version_info() -> str:
     return _VERSION_INFO if _VERSION_INFO is not None else __version__
 
 
+def set_global_log_level(level: str | int) -> None:
```
mkundu1, line 47: I think `|` in type-hint will not work in python 3.8

src/ansys/fluent/core/launcher/fluent_container.py
mkundu1, line 1: Nice improvement

src/ansys/fluent/core/__init__.py
raph-luc: You are right, going to change it to a Union[] which works in Python 3.8 according to my tests. Thank you for the heads-up!

mkundu1 (approved):
I think logging into a single file by default is a better idea as the order of the operations can easily be understood by looking into a single file.

raph-luc:
The changes exposed what I believe is a flaw in ansys/actions https://github.com/ansys/actions/issues/269, causing all of the tests that use `build-wheelhouse` to fail

Fixed some of the tests related to the broken github actions from PyFluent's side. For the ansys/actions side, the best way to fix this in the long term is going to require changes on their repo which I currently don't have permissions to create branches/PRs, so I just created the issue tracker for now

src/ansys/fluent/core/logging_config.yaml
```diff
@@ -1,84 +1,50 @@
...
+    formatter: console_fmt
     stream: ext://sys.stdout
-  fluent.file:
+  pyfluent_file:
     class: logging.handlers.RotatingFileHandler
-    level: DEBUG
-    formatter: simple
+    level: NOTSET #  no log level filtering on this handler, message log levels are managed by the loggers instead
+    formatter: logfile_fmt
     filename: pyfluent.log
```
mkundu1, line 25: Shall we append the timestamp to the log filename? Then the log file for each session will be preserved (like Fluent's automatic transcript). TimedRotatingFileHandler can probably be used here.

raph-luc:
Note this is currently on hold as I am reviewing how this is going to be set up, in light of some discussions with Sean, Mainak and the ansys/actions team, will push local changes to this branch once I am done

src/ansys/fluent/core/logging_config.yaml
raph-luc: It did not work very well, and logging package does not seem to support YAML very well for this. Using the `TimedRotatingFileHandler` instead of the `RotatingFileHandler` also disables rollover on file size (it instead rotates based on time interval only), which means that we end up with some absolutely massive log files some times. Ideally we'd use a custom handler that is a mix of these two handlers. 

Sticking to `RotatingFileHandler` for now, we can try adding timestamp again later on. Feel free to take a look at what I'm doing and see if you can find a good way to implement `TimedRotatingFileHandler`, I have pushed my local changes to this branch.

raph-luc:
Left an useful debug args for docker run commented here, may be useful in the future for others as well, let me know if you prefer that I remove it:
https://github.com/ansys/pyfluent/blob/c67798c0f99ad04ea8b53a662b0cb3b75660fa9f/src/ansys/fluent/core/launcher/fluent_container.py#L62-L63

mkundu1 (approved):

hpohekar (approved):

acarvalh-work (approved):

src/ansys/fluent/core/logging.py
```diff
@@ -0,0 +1,140 @@
+"""Module controlling PyFluent's logging functionality."""
```
seanpearsonuk, line 1: @raph-luc are you strongly advocating for this module or can the users live without it?

raph-luc: I would expect that most users should be able to live without debug logging. Print to console stdout for warning and error messages is automatically initialized in `__init__.py`, so users generally shouldn't need to worry about it either.

I am not yet aware of any use case where users necessarily need the debug output to be able to do anything.

I believe debugging is mostly for us and other developers or more advanced users.

PR #1657, opened 2023-05-31
doc: fix global logging

acarvalh-work:
Fixes global logging setup in user guide. 

old:
```
pyfluent.set_log_level("DEBUG") # by default, only errors are shown
```
new:
``` 
from ansys.fluent.core import logging

logger = logging.getLogger()
logger.setLevel(logging.DEBUG)
```

### DIFF
--- a/doc/source/user_guide/index.rst
+++ b/doc/source/user_guide/index.rst
@@ -208,6 +208,18 @@ You can control the global logging level at any time with:
 
 .. code:: python
 
-    import ansys.fluent.core as pyfluent
-    pyfluent.set_log_level("DEBUG") # by default, only errors are shown
+    from ansys.fluent.core import logging
 
+    logger = logging.getLogger()
+    logger.setLevel(logging.DEBUG)
+
+    logger.debug("DEBUG STATEMENT")
+
+Logging levels:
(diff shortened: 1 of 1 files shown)


doc/source/user_guide/index.rst
```diff
@@ -208,6 +208,18 @@ You can control the global logging level at any time with:
 
 .. code:: python
 
-    import ansys.fluent.core as pyfluent
-    pyfluent.set_log_level("DEBUG") # by default, only errors are shown
+    from ansys.fluent.core import logging
```
seanpearsonuk, line 211: @acarvalh-work this is just a standard Python module that can be imported directly. It happens to be available via ansys.fluent.core. @raph-luc is updating logging generally and I suspect that he will be completely removing this stuff.

raph-luc (commented):
This is only changing the root logger, which (even before my WIP changes, see https://github.com/ansys/pyfluent/pull/1651) is not being used by PyFluent except for one function in downloads.py which is also going to be reworked

raph-luc:
@acarvalh-work feel free to talk to me and contribute to https://github.com/ansys/pyfluent/pull/1651 if you want

PR #1659, opened 2023-06-02
Set container mount path from env var

mkundu1:
This is useful while launching the Fluent docker container from PyFluent in Windows host for development purpose. We cannot mount a Windows path within the Linux container.


raph-luc (approved):
Good idea and good change, removed some comments that are now outdated and not necessary anymore

hpohekar (approved):

mkundu1:
> Good idea and good change, removed some comments that are now outdated and not necessary anymore

Thanks

PR #1664, opened 2023-06-05
Maint/refactor quantity

### LINKED ISSUE #1619: requirement for quantity.py config data + comment on current data structures
quantity.py should be fed by config data separate from the quantity.py module. That can be yaml or json or even just python. The config file should contain minimal tables. 

The config data currently in quantity.py needs to be consolidated.

1.
```
    temperature_conversions = {
        # from : { to : formula }
        "C": {
            "K": lambda c: c + 273.15,
            "R": lambda c: (c + 273.15) * 1.8,
            "F": lambda c: c * 9 / 5 + 32,
            "C": lambda c: c,
        },
        "F": {
            "K": lambda f: (f - 32) * 5 / 9 + 273.15,
            "R": lambda f: f + 459.67,
            "F": lambda f: f,
            "C": lambda f: (f - 32) * 5 / 9,
        },
        "R": {
            "K": lambda r: r / 1.8,
            "R": lambda r: r,
            "F": lambda r: r - 459.67,
            "C": lambda r: (r - 491.67) / 1.8,
        },
        "K": {
            "K": lambda k: k,
            "R": lambda k: k * 1.8,
            "F": lambda k: k * 1.8 - 459.67,
            "C": lambda k: k - 273.15,
        },
    }
```
This should not be coded as a special case. *All* units have a conversion factor and an offset. It just happens that in this case the offsets are non-zero. We only need one comprehensive table that doesn't treat temperature as a special case. The conversion function would be generic and applicable to any of the data.

2.
```
    si_map = {
        "kg": "kg",
        "g": "kg",
        "lb": "kg",
        "lbm": "kg",
        "slug": "kg",
        "m": "m",
        "cm": "m",
        "ft": "m",
        "in": "m",
        "s": "s",
        "A": "A",
        "mol": "mol",
        "slugmol": "mol",
        "cd": "cd",
        "sr": "sr",
        "radian": "radian",
        "degree": "radian",
        "K": "K",
        "C": "K",
        "F": "K",
        "R": "K",
    }
``` 
The above dictionary is redundant - it does not need to be presented separately from the conversion table.

3.
```
    conversion_map = {
        "kg": 1,
        "g": 0.001,
        "lb": 0.45359237,
        "lbm": 0.45359237,
        "slug": 14.59390293720637,
        "m": 1,
        "cm": 0.01,
        "ft": 0.30479999999999996,
        "in": 0.0254,
        "s": 1,
        "A": 1,
        "mol": 1,
        "slugmol": 14593.9,
        "cd": 1,
        "sr": 1,
        "radian": 1,
        "degree": 0.017453292519943295,
        "K": 1,
        "C": 1,
        "F": 0.5555555555555556,
        "R": 0.5555555555555556,
        "delta_K": 1,
        "delta_C": 1,
        "delta_F": 0.5555555555555556,
        "delta_R": 0.5555555555555556,
    }
```
The deltas here are redundant and duplicate information already in the "temperature_conversions". To reiterate, the single table will suffice.

There are many more data structures here. I will not labour the point further, but only say that we can definitely reduce what is there. 

Refactoring steps could be to start by sensibly structuring config file(s) with simple tabular data without information duplication, and then start consuming them to populate the existing data structures. We can incrementally remove unneeded structures.

acarvalh-work:
Closes #1619

Changes: 

- All tables are stored within the `quantity_config.json` file within Utils. Tables are loaded into new `UnitsTable` class.
        - Removed duplicated tables.
- Created new classes for `dimensions` and `quantity_map` objects.
	- Both contain all properties and functions directly associated with respective class rather than scattered throughout `quantity.py`
- Condensed `remove_multipliers()` and `filter_multiplers()` into single function that does both.
	- `filter_unit_term()` : Filters a term within a unit string and returns the base, multiplier, and exponent as a tuple.
- Simplified unit system conversions to single method `convert()` within `Quantity` that creates a new Quantity object with desired unit system.
	- Removes needing to create a separate unit system object and passing it to quantity.
```     
Current                                 New
--------                                ----
unitSysSI = q.UnitSystem("SI")          v = q.Quantity(3.0, "ft s^-1")
v = q.Quantity(3.0, "ft s^-1")          siV = v.convert("SI")
siV = unitSysSI.convert(v)
```
- Removed `si_map` table. 
	- SI units are now mapped by the `_si_map()` function in `UnitsTable`.
- Removed `derived_units_with_conversion_factor` table. 
	- Incorporated `l`, `gal`, `BTU`, and `cal` into `derived_units` table within `quantity_config.json`.
- Incorporated all temperature conversion functions into a single general conversion function.
- Added offset and factor values to all units within `quantity_config.json`.
- Created `si_data()`  function within `UnitsTable` to get retrieve the SI unit string, SI offset, and SI factor from `quantity_config.json`.
- Created `condense()`  function within `UnitsTable` to combine like-terms in a unit string. ("m m^2" -> "m^3")
- Added types to all units. 
	- Currently only 'temperature' and 'temperature difference' types exist.
- Testing coverage increased to 99%.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,10 @@ env:
   DOCUMENTATION_CNAME: 'fluent.docs.pyansys.com'
   PACKAGE_NAME: 'ansys-fluent-core'
   PACKAGE_NAMESPACE: 'ansys.fluent.core'
+  ANSYSLMD_LICENSE_FILE: ${{ format('1055@{0}', secrets.LICENSE_SERVER) }}
+  PYFLUENT_START_INSTANCE: 0
+  PYFLUENT_TIMEOUT_FORCE_EXIT: 5
+  PYFLUENT_LAUNCH_CONTAINER: 1
 
 jobs:
 
@@ -102,8 +106,8 @@ jobs:
 
       - name: Retrieve PyFluent version
         run: |
-          echo "PYFLUENT_VERSION=$(python -c "from ansys.fluent.core import __version__; print(__version__)")" >> $GITHUB_OUTPUT
-          echo "PYFLUENT version is: $(python -c "from ansys.fluent.core import __version__; print(__version__)")"
+          echo "PYFLUENT_VERSION=$(python -c "from ansys.fluent.core import __version__; print(); print(__version__)" | tail -1)" >> $GITHUB_OUTPUT
+          echo "PYFLUENT version is: $(python -c "from ansys.fluent.core import __version__; print(); print(__version__)" | tail -1)"
         id: version
 
       - name: Cache API Code
@@ -141,9 +145,6 @@ jobs:
         if: steps.cache-api-code.outputs.cache-hit != 'true'
         run: make api-codegen
         env:
-          ANSYSLMD_LICENSE_FILE: ${{ format('1055@{0}', secrets.LICENSE_SERVER) }}
-          PYFLUENT_START_INSTANCE: 0
-          PYFLUENT_LAUNCH_CONTAINER: 1
           FLUENT_IMAGE_TAG: ${{ env.DOC_DEPLOYMENT_IMAGE_TAG }}
 
       - name: Install again after codegen
@@ -162,8 +163,6 @@ jobs:
       - name: Build Source Documentation
         run: make build-doc-source
         env:
-          ANSYSLMD_LICENSE_FILE: ${{ format('1055@{0}', secrets.LICENSE_SERVER) }}
-          PYFLUENT_START_INSTANCE: 0
           FLUENT_IMAGE_TAG: ${{ env.DOC_DEPLOYMENT_IMAGE_TAG }}
 
       - name: Zip HTML Documentation before upload
@@ -209,8 +208,8 @@ jobs:
 
       - name: Retrieve PyFluent version
         run: |
-          echo "PYFLUENT_VERSION=$(python -c "from ansys.fluent.core import __version__; print(__version__)")" >> $GITHUB_OUTPUT
-          echo "PYFLUENT version is: $(python -c "from ansys.fluent.core import __version__; print(__version__)")"
+          echo "PYFLUENT_VERSION=$(python -c "from ansys.fluent.core import __version__; print(); print(__version__)" | tail -1)" >> $GITHUB_OUTPUT
+          echo "PYFLUENT version is: $(python -c "from ansys.fluent.core import __version__; print(); print(__version__)" | tail -1)"
         id: version
 
       - name: Cache 22.2 API Code
@@ -247,9 +246,6 @@ jobs:
         if: steps.cache-222-api-code.outputs.cache-hit != 'true'
         run: make api-codegen
         env:
-          ANSYSLMD_LICENSE_FILE: ${{ format('1055@{0}', secrets.LICENSE_SERVER) }}
-          PYFLUENT_START_INSTANCE: 0
-          PYFLUENT_LAUNCH_CONTAINER: 1
           FLUENT_IMAGE_TAG: v22.2.0
 
       - name: Print 22.2 Fluent version info
@@ -284,9 +280,6 @@ jobs:
         if: steps.cache-231-api-code.outputs.cache-hit != 'true'
         run: make api-codegen
         env:
-          ANSYSLMD_LICENSE_FILE: ${{ format('1055@{0}', secrets.LICENSE_SERVER) }}
-          PYFLUENT_START_INSTANCE: 0
-          PYFLUENT_LAUNCH_CONTAINER: 1
           FLUENT_IMAGE_TAG: v23.1.0
 
       - name: Print 23.1 Fluent version info
@@ -316,7 +309,7 @@ jobs:
   test:
     name: Unit Testing
     needs: build
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, pyfluent]
     strategy:
       fail-fast: false
       matrix:
@@ -346,8 +339,8 @@ jobs:
 
       - name: Retrieve PyFluent version
         run: |
-          echo "PYFLUENT_VERSION=$(python -c "from ansys.fluent.core import __version__; print(__version__)")" >> $GITHUB_OUTPUT
-          echo "PYFLUENT version is: $(python -c "from ansys.fluent.core import __version__; print(__version__)")"
+          echo "PYFLUENT_VERSION=$(python -c "from ansys.fluent.core import __version__; print(); print(__version__)" | tail -1)" >> $GITHUB_OUTPUT
+          echo "PYFLUENT version is: $(python -c "from ansys.fluent.core import __version__; print(); print(__version__)" | tail -1)"
         id: version
 
       - name: Login to GitHub Container Registry
@@ -365,8 +358,6 @@ jobs:
       - name: Unit Testing
         run: make unittest-dev-${{ matrix.version }}
         env:
-          ANSYSLMD_LICENSE_FILE: ${{ format('1055@{0}', secrets.LICENSE_SERVER) }}
-          PYFLUENT_START_INSTANCE: 0
           FLUENT_IMAGE_TAG: ${{ matrix.image-tag }}
 
       - name: Upload 23.1 Coverage Results to Codecov
@@ -422,8 +413,8 @@ jobs:
 
       - name: Retrieve PyFluent version
         run: |
-          echo "PYFLUENT_VERSION=$(python -c "from ansys.fluent.core import __version__; print(__version__)")" >> $GITHUB_OUTPUT
-          echo "PYFLUENT version is: $(python -c "from ansys.fluent.core import __version__; print(__version__)")"
+          echo "PYFLUENT_VERSION=$(python -c "from ansys.fluent.core import __version__; print(); print(__version__)" | tail -1)" >> $GITHUB_OUTPUT
+          echo "PYFLUENT version is: $(python -c "from ansys.fluent.core import __version__; print(); print(__version__)" | tail -1)"
         id: version
 
       - name: Login to GitHub Container Registry
@@ -441,9 +432,6 @@ jobs:
       - name: Run API codegen
         run: make api-codegen
         env:
-          ANSYSLMD_LICENSE_FILE: ${{ format('1055@{0}', secrets.LICENSE_SERVER) }}
-          PYFLUENT_START_INSTANCE: 0
-          PYFLUENT_LAUNCH_CONTAINER: 1
           FLUENT_IMAGE_TAG: ${{ matrix.image-tag }}
 
       - name: Print Fluent version info
@@ -458,8 +446,6 @@ jobs:
       - name: Unit Testing
         run: make unittest-dev-${{ matrix.version }}
         env:
-          ANSYSLMD_LICENSE_FILE: ${{ format('1055@{0}', secrets.LICENSE_SERVER) }}
-          PYFLUENT_START_INSTANCE: 0
           FLUENT_IMAGE_TAG: ${{ matrix.image-tag }}
 
   release:
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -1,0 +1,20 @@
+# Dependency Review Action
+#
+# This Action will scan dependency manifest files that change as part of a Pull Request, surfacing known-vulnerable versions of the packages declared or updated in the PR. Once installed, if the workflow run is marked as required, PRs introducing known-vulnerable packages will be blocked from merging.
+#
+# Source repository: https://github.com/actions/dependency-review-action
+# Public documentation: https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-dependency-review#dependency-review-enforcement
+name: 'Dependency Review'
+on: [pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@v3
+      - name: 'Dependency Review'
+        uses: actions/dependency-review-action@v3
--- a/.github/workflows/nightly-dev-doc-build.yml
+++ b/.github/workflows/nightly-dev-doc-build.yml
@@ -8,6 +8,10 @@ on:
 env:
   DOCUMENTATION_CNAME: 'fluent.docs.pyansys.com'
   DOC_DEPLOYMENT_IMAGE_TAG: v23.1.0
+  ANSYSLMD_LICENSE_FILE: ${{ format('1055@{0}', secrets.LICENSE_SERVER) }}
+  PYFLUENT_START_INSTANCE: 0
+  PYFLUENT_TIMEOUT_FORCE_EXIT: 5
+  PYFLUENT_LAUNCH_CONTAINER: 1
 
 jobs:
   nightly_docs_build:
@@ -48,16 +52,11 @@ jobs:
       - name: Run API codegen
         run: make api-codegen
         env:
-          ANSYSLMD_LICENSE_FILE: ${{ format('1055@{0}', secrets.LICENSE_SERVER) }}
-          PYFLUENT_START_INSTANCE: 0
-          PYFLUENT_LAUNCH_CONTAINER: 1
           FLUENT_IMAGE_TAG: ${{ matrix.image-tag }}
 
       - name: Build All Documentation
         run: make build-all-docs
         env:
-          ANSYSLMD_LICENSE_FILE: ${{ format('1055@{0}', secrets.LICENSE_SERVER) }}
-          PYFLUENT_START_INSTANCE: 0
           FLUENT_IMAGE_TAG: ${{ matrix.image-tag }}
 
       - name: Zip HTML Documentation before upload
--- a/.github/workflows/nightly-test-run.yml
+++ b/.github/workflows/nightly-test-run.yml
@@ -5,6 +5,12 @@ on:
     - cron:  '0 4 * * *'
   workflow_dispatch:
 
+env:
+  ANSYSLMD_LICENSE_FILE: ${{ format('1055@{0}', secrets.LICENSE_SERVER) }}
+  PYFLUENT_START_INSTANCE: 0
+  PYFLUENT_TIMEOUT_FORCE_EXIT: 5
+  PYFLUENT_LAUNCH_CONTAINER: 1
+
 jobs:
   test:
     name: Unit Testing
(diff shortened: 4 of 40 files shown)


src/ansys/fluent/core/quantity.py
```diff
@@ -1,1369 +1,830 @@
...
+
+
+class UnitsTable(object):
+    """Initializes a UnitsTable object with all table values and unit string manipluation methods.
+
+    Properties
+    ----------
+    api_quantity_map : dict
+        Quantity map values from settings API.
+    fundamental_units : dict
```
dnwillia-work, line 343: Could you update the documentation for fundamental units.  Something like "Fundamental units are used for fundamental base units like length, mass, or time."

src/ansys/fluent/core/quantity.py
```diff
@@ -1,1369 +1,830 @@
...
+        Fundamental units and properties.
+    derived_units : dict
```
dnwillia-work, line 345: Same here for doc.  Something like "Common units that are derived from fundamental units".

src/ansys/fluent/core/quantity.py
```diff
@@ -1,1369 +1,830 @@
...
-                        if term.split("^")[0] == base
-                    ]
-                )
-            else:
-                return 1.0
 
-    def _compute_offset(self, unit_str: str) -> float:
-        """
-        Compute offset.
+        path = "src/ansys/fluent/core/utils/"
```
dnwillia-work, line 382: Does this work when installed as a package?

src/ansys/fluent/core/utils/quantity_config.json
```diff
@@ -0,0 +1,457 @@
...
+        },
+        "delta_C": {
+            "type": "Temperature Difference",
+            "factor": 1.0,
+            "offset": 0.0,
+            "dimension_order": 4
+        },
+        "delta_F": {
+            "type": "Temperature Difference",
+            "factor": 1.0,
```
dnwillia-work, line 144: Should that not be the factor to convert to `delta_K` or `delta_C`?

Maybe you can throw a comment at the top which covers what each entry is for.

acarvalh-work: https://github.com/ansys/pyfluent/blob/f3f6e135ef9804f9c563c2ae66eb4c9154ff8a2e/src/ansys/fluent/core/utils/quantity_config.yaml#L1-L9

Fixed the factors for delta_F, delta_C, and delta_R. Also switched from json to yaml to include comments in the config file explaining the properties within each unit.

dnwillia-work (approved):

dnwillia-work:
So, I approved, but I think this but I don't really agree with the design regarding unit systems.

Ideally I think you want to be able to do this:

```
mySys = UnitSystem(Name="Bob", BaseUnits =['ton', 'parsec', 'eon', 'R', ...])
q_ft_per_sec = Quanitity(3.0, 'ft s^-1')
q_parsec_per_eon = mySys.convert(q_ft_per_ec)
```

Right now the Quantity class does not allow it.

src/ansys/fluent/core/quantity.py
```diff
@@ -1,1369 +1,869 @@
...
+        return f"'{self.from_unit}' and '{self.to_unit}' have incompatible dimensions."
+
+
+class UnitsTable(object):
+    """Initializes a UnitsTable object with all table values and unit string manipluation methods.
+
+    Properties
+    ----------
+    api_quantity_map : dict
+        Quantity map values from settings API.
```
raph-luc, line 352: I believe `Properties` isn't recognized by our documentation set up, this section will just get ignored. Use `Parameters` instead. And private members will not show up in documentation either.

src/ansys/fluent/core/quantity.py
```diff
@@ -1,1369 +1,869 @@
...
+    """
+
+    def __init__(self):
+        self._data: dict = self._get_data()
+        self._api_quantity_map: dict = self._data["api_quantity_map"]
+        self._fundamental_units: dict = self._data["fundamental_units"]
+        self._derived_units: dict = self._data["derived_units"]
+        self._multipliers: dict = self._data["multipliers"]
+
+    def _get_data(self) -> dict:
```
raph-luc, line 385: `get_data` may as well be a function outside of the `UnitsTable` class, it is not doing anything specific to `UnitsTable` and makes no use of the object with `self`. Could be moved and renamed to `get_json_data` or something similar.

Alternatively, maybe just put the contents of `get_data` inside `__init__()` if it is only used there?

src/ansys/fluent/core/quantity.py
```diff
@@ -1,1369 +1,869 @@
...
+        """
+        self.from_unit = from_unit
+        self.to_unit = to_unit
+
+    def __str__(self):
+        return f"'{self.from_unit}' and '{self.to_unit}' have incompatible dimensions."
+
+
+class UnitsTable(object):
+    """Initializes a UnitsTable object with all table values and unit string manipluation methods.
```
raph-luc, line 347:
```suggestion
    """Initializes a UnitsTable object with all table values and unit string manipulation methods.
```

src/ansys/fluent/core/quantity.py
```diff
@@ -1,1369 +1,869 @@
...
+    multipliers : dict
+        Multiplier prefixes and respective factors.
+
+    Methods
+    -------
+    filter_unit_term()
+        Separate multiplier, base, and power from a unit term.
+    si_data()
+        Compute the SI unit string, SI multiplier, and SI offset from a unit string of any type.
+    si_converion()
```
raph-luc, line 366:
```suggestion
    si_conversion()
```

src/ansys/fluent/core/quantity.py
```diff
@@ -1,1369 +1,869 @@
...
-        if self._unit_system not in UnitSystem._supported_unit_sys:
-            raise ValueError(
-                "Unsupported unit system, only 'SI', 'CGS', 'BT' is allowed."
-            )
+        return self.condense(si_unit_str), si_multiplier, si_offset
 
-    def _get_unit_from_dim(self, dim_list: list) -> str:
-        """
-        Get unit from dimensions.
+    def si_conversion(
```
raph-luc, line 574: `si_conversion()` is also not using anything from the `UnitsTable` object. Maybe it would be better to rework it, or make it a module method rather than a `UnitsTable` method?

src/ansys/fluent/core/quantity.py
```diff
@@ -1,1369 +1,869 @@
...
-        """
-        Convert to a system.
+        # Perform conversion to SI value
+        return (
+            ((value + si_offset) * si_multiplier)
+            if type == "Temperature"
+            else (value * si_multiplier)
+        )
+
+    def condense(self, unit_str: str) -> str:
```
raph-luc, line 617: Used throughout the module: is the `_str` in `unit_str` necessary? Maybe just change it to `units` (as in "these are the __units__ of this number"), makes it easier to use (and it is apparently used a lot in the tests), and we could probably guess it is a string (what else would `units` be in this case?).

raph-luc (commented):
A few suggestions, and I think this is on the right path, just need some additional polishing and rethinking, particularly for the documentation

Additional ideas to consider: 

- adding a `UnitSystem` class as described by Dan, I think it is a good idea and should be possible to do, though could require significant work. Maybe rework `QuantityMap` to do this?

- I think Dan also mentioned splitting this nearly 1000 lines module up into smaller modules and possibly putting them inside a quantity folder. Might be the way to go if you can't think of a way to optimize and streamline the code. Maybe, to start with, `UnitsTable` and `Dimensions` could be in a separate module?

src/ansys/fluent/core/quantity.py
acarvalh-work: Update `UnitsTable` to get data from config file during `__init__`

https://github.com/ansys/pyfluent/blob/d785d42ec4c3fe35fd1c982c9b1c1d291f28eb6d/src/ansys/fluent/core/quantity/units_table.py#L27-L39

src/ansys/fluent/core/quantity.py
```diff
@@ -1,1369 +1,869 @@
-"""
-Module for creating quantities in terms of real values with units.
-"""
```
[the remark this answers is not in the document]
acarvalh-work: Updated quantity root documentation with current quantity implementation. 

https://github.com/ansys/pyfluent/blob/maint/refactor_quantity/doc/source/api/general/quantity.rst

src/ansys/fluent/core/quantity.py
```diff
@@ -1,1369 +1,869 @@
...
-        new_si_value = pow(self._si_value, exponent)
-        return Quantity(new_si_value, new_si_unit)
+    def __init__(self, quantity_map):
+        self._units_table = UnitsTable()
+        self._unit_str = self._map_to_unit_str(quantity_map)
 
-    def __str__(self):
-        return f'({self.value}, "{self.unit}")'
+    def _map_to_unit_str(self, quantity_map: dict) -> str:
+        """Convert a quantity map into a unit string.
```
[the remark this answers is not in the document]
acarvalh-work: I made docstrings for all functions and classes so that it's easier for developers to understand what a function or class does in the event that the code needs to be refactored in the future. If that's not required then I can go through and remove the docstrings on private members.

For `map_to_unit_str` and `unit_str_to_dim` I thought these should be private because they're only used by the quantity class and would not be called directly. Within the tests, getting a unit string from a quantity map look like this: 

```
v = q.Quantity(40.2, quantity_map={"Velocity": 1.0})
v.units
>>> "m s^-1"
``` 

and getting dimensions from a unit string would look like this:

```
v = q.Quantity(20.2, units="m s^-1")
v.dimensions
>>> [0.0, 1.0, -1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
```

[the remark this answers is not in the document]
raph-luc: My only concern was if this was documentation that you actually wanted to show up in the online documentation. If this was specifically for developers, then this is fine by me.

Regarding whether to make these private or not, if users are not expected to ever use it directly, then making it private also makes sense to me, thanks for the explanation.

raph-luc (commented):
I think this is a lot better, and another small suggestion: I believe you could consider using a `__init__.py` file inside the quantity folder, so that we can access the most common functions using e.g. `pyfluent.quantity.q` rather than repeating `pyfluent.quantity.quantity.q` 

or using `from ansys.fluent.core import quantity` rather than `from ansys.fluent.core.quantity import quantity`

src/ansys/fluent/core/quantity/quantity.py
```diff
@@ -0,0 +1,332 @@
...
+            or (quantity_map and dimensions)
+        ):
+            raise ValueError(
+                "Quantity only accepts 1 of the following: units, quantity_map, dimensions"
+            )
+
+        _units_table = q.UnitsTable()
+        _value = float(value)
+
+        if units or units == "":
```
seanpearsonuk, line 45: There is a subtle logical discrepancy between this assignment code and the above validation code. In both places, we should apply the simple condition `units is not None`.

src/ansys/fluent/core/quantity/quantity.py
```diff
@@ -0,0 +1,332 @@
...
+        if dimensions:
+            _dimensions = q.Dimensions(dimensions=dimensions)
+            _unit = _dimensions.units
+
+        _type = _units_table.get_type(_unit)
+        _, si_multiplier, si_offset = _units_table.si_data(units=_unit)
+        _si_value = (
+            ((_value + si_offset) * si_multiplier)
+            if _type == "Temperature"
+            else (_value * si_multiplier)
```
seanpearsonuk, line 61: The call to `_units_table.si_data(units=_unit)` should return `0` for `si_offset` for quantities other than temperature, so that the condition below is not needed - the value is always given by `(_value + si_offset) * si_multiplier`.

src/ansys/fluent/core/quantity/dimensions.py
```diff
@@ -0,0 +1,123 @@
...
+        """
+
+        power = power or 1.0
+        dimensions = dimensions or [0.0] * 9
+
+        # Split unit string into terms and parse data associated with individual terms
+        for term in units.split(" "):
+            _, unit_term, unit_term_power = self._units_table.filter_unit_term(term)
+
+            unit_term_power *= power
```
seanpearsonuk, line 94: I'm not sure about this `power` argument appearing in this method. If your output is a dimensions list then `power` can simply act as a multiplier mapped over the whole list. There really is no need to involve `power` prior to that point. It would be good to have a simple method only responsible for getting the `dimensions` for `units`.

src/ansys/fluent/core/quantity/units_table.py
```diff
@@ -0,0 +1,272 @@
...
+    def derived_units(self):
+        """Derived units and properties composed of fundamental units."""
+        return self._derived_units
+
+    @property
+    def multipliers(self):
+        """Multiplier prefixes and respective factors."""
+        return self._multipliers
+
+    def filter_unit_term(self, unit_term: str) -> tuple:
```
seanpearsonuk, line 99: It would be good to directly unit-test this code to ensure that it is always returning the three components correctly.

src/ansys/fluent/core/quantity/dimensions.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+
+    def __init__(
+        self, units: str = None, dimensions: list = None, unit_sys: str = None
+    ):
+        self._units_table = q.UnitsTable()
+        unit_sys = unit_sys or "SI"
+        units = units or " "
+
+        if units:
```
seanpearsonuk, line 26: This is always `True` given the line above.

src/ansys/fluent/core/quantity/dimensions.py
```diff
@@ -0,0 +1,123 @@
...
+    dimensions: list
+        List of dimensions.
+
+    Returns
+    -------
+    Dimensions instance.
+    """
+
+    def __init__(
+        self, units: str = None, dimensions: list = None, unit_sys: str = None
```
seanpearsonuk, line 20: Since it's supposed to be either `units` _or_ `dimensions`, the code should should check that both or not set.

src/ansys/fluent/core/quantity/dimensions.py
```diff
@@ -0,0 +1,123 @@
...
+        return dimensions
+
+    @property
+    def units(self):
+        """Unit string representation of dimensions"""
+        return self._unit
+
+    @property
+    def dimensions(self):
+        """Dimensions representation of unit string"""
```
seanpearsonuk, line 122: Comment is a bit inaccurate. The dimensions exist independently of units.

src/ansys/fluent/core/quantity/dimensions.py
```diff
@@ -0,0 +1,123 @@
...
+                dimensions = self._units_to_dim(
+                    units=self._units_table.derived_units[unit_term]["composition"],
+                    power=unit_term_power,
+                    dimensions=dimensions,
+                )
+
+        return dimensions
+
+    @property
+    def units(self):
```
seanpearsonuk, line 116: I'm not convinced that a dimensions object should have such an attribute. Are we currently depending to hold things together for now?

src/ansys/fluent/core/quantity/dimensions.py
```diff
@@ -0,0 +1,123 @@
...
+            Unit system of dimensions.
+
+        Returns
+        -------
+        units : str
+            Unit string representation of dimensions.
+        """
+        # Ensure dimensions list contains 9 terms
+        dimensions = [float(dim) for dim in dimensions + ((9 - len(dimensions)) * [0])]
+        sys_order = {
```
seanpearsonuk, line 53: ideally not embedded inside a method

src/ansys/fluent/core/quantity/quantity.py
acarvalh-work: Originally was having issues with evaluating the truthy value on empty vs non-empty strings. Fixed now checking `units is not None`.

https://github.com/ansys/pyfluent/blob/911b052f1e7162a45f2bc72965029fcf5045470b/src/ansys/fluent/core/quantity/quantity.py#L45-L46

https://github.com/ansys/pyfluent/blob/911b052f1e7162a45f2bc72965029fcf5045470b/src/ansys/fluent/core/quantity/quantity.py#L79-L81

src/ansys/fluent/core/quantity/dimensions.py
acarvalh-work: Fixed using `if units is not None:`

https://github.com/ansys/pyfluent/blob/911b052f1e7162a45f2bc72965029fcf5045470b/src/ansys/fluent/core/quantity/dimensions.py#L27-L32

src/ansys/fluent/core/quantity/dimensions.py
acarvalh-work: Added predefined unit systems to the yaml config file. If any new systems need to be embedded they can be added directly to the config file and work with the exiting quantity implementation. 

Not sure about adding a UnitSystems class at the moment. If one of the objectives of this quantity module is to allow users to define custom unit systems, then adding a UnitSystems class would definitely be effective. For now, the unit systems are just small dictionaries so I've added them to the config file.

https://github.com/ansys/pyfluent/blob/911b052f1e7162a45f2bc72965029fcf5045470b/src/ansys/fluent/core/quantity/quantity_config.yaml#L243-L279

src/ansys/fluent/core/quantity/dimensions.py
seanpearsonuk: OK, fair enough :)

src/ansys/fluent/core/quantity/dimensions.py
acarvalh-work: `power` is only applied to the terms that compose a derived unit string. 

For example: `m farad^-2`: 
- "m" is a fundamental unit and it's power in this case is 1 so nothing changes. 
- "farad^-2" becomes ("N m V^-2")^-2
- ("N V^-2")^-2 becomes ("kg m s^-2 m A ohm")^-2
- ("kg m s^-2 m A ohm")^-2 is simplified to "kg^2 m^4 s^-8 A^-4"
- the final SI unit string: `m^5 kg^2 s^-8 A^-4`

If power is mapped over the whole list then the `^-2` from farad in this case would be applied to the first `m` within the unit string even though the `^-2` is only supposed to apply to the terms derived from within `farad`.

src/ansys/fluent/core/quantity/dimensions.py
acarvalh-work: https://github.com/ansys/pyfluent/blob/9cbf840c81b3652466d7c61e22f63a5f0744907d/src/ansys/fluent/core/quantity/dimensions.py#L119-L122

src/ansys/fluent/core/quantity/quantity.py
acarvalh-work: Removed hard-coded check for temperature. Now if the unit string is a temperature it gets the offset, anything else the offset is set to 0.

https://github.com/ansys/pyfluent/blob/9cbf840c81b3652466d7c61e22f63a5f0744907d/src/ansys/fluent/core/quantity/quantity.py#L57
https://github.com/ansys/pyfluent/blob/9cbf840c81b3652466d7c61e22f63a5f0744907d/src/ansys/fluent/core/quantity/quantity.py#L92
https://github.com/ansys/pyfluent/blob/9cbf840c81b3652466d7c61e22f63a5f0744907d/src/ansys/fluent/core/quantity/quantity.py#L186
https://github.com/ansys/pyfluent/blob/9cbf840c81b3652466d7c61e22f63a5f0744907d/src/ansys/fluent/core/quantity/units_table.py#L173-L177

src/ansys/fluent/core/quantity/units_table.py
acarvalh-work: Added unit tests for `filter_unit_term`: 

https://github.com/ansys/pyfluent/blob/933480e02ff41edef549668b8a33b66019215349/tests/test_quantity.py#L961-L968

seanpearsonuk (approved):

raph-luc (approved):

PR #1671, opened 2023-06-08
Restructure Fluent Connection

prmukherj:

PR #1674, opened 2023-06-08
Adding functionality to force exit Fluent

raph-luc:
- Inside `fluent_connection.FluentConnection`: <details><summary>Changed exit() </summary>image</details><details><summary> Added force_exit() and force_exit_container() </summary>image</details>
- New frozen dataclass: <details><summary>connection_properties</summary>image</details>
- Simplified environment definition in GitHub actions YML files and made it so that GitHub runners always force exit in case normal exit fails. This is a potential fix for issues we were having with test runners randomly running out of memory due to the amount of zombie Fluent sessions - together with the WIP addition of the pyfluent watchdog process, and Fluent v241 auto exit on disconnect, should eliminate any possibility of these issues happening again in the future
- New `test_fluent_freeze_kill()` and `test_fluent_connection_properties()`


lagacep-ans:
Possibly rename kill_*() functions to terminate_*() , other suggestions for terminology?

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -85,6 +92,50 @@ def __call__(self):
...
+        self.cortex_pwd = cortex_pwd
+        self.cortex_pid = cortex_pid
+        self.cortex_host = cortex_host
+        self.inside_container = inside_container
+
+    def list_properties(self) -> list:
+        """List all property names."""
+        return [k for k, _ in vars(self).items()]
+
+    def list_properties_values(self) -> dict:
```
lagacep-ans, line 134: the pattern for this simple object seems a little verbose.    this could be even be only a dictionary.

raph-luc: True, that is also only a convenience function. I would suggest for now changing it to simply `list_values`, and the other method to just `list`. Going to see what other changes are suggested and then bundle them all up in a single commit

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -285,6 +361,95 @@ def id(self) -> str:
...
+            cleanup_line_end = ".sh\n"
+            cmd_list = ["bash"]
+        else:
+            logger.error(
+                f"Unrecognized or unsupported operating system, cancelling Fluent cleanup script execution."
+            )
+            return
+        list_dir = os.listdir(Path(pwd))
+        for name in list_dir:
+            if name.startswith("fluent-") and name.endswith(f"{pid}.trn"):
```
mkundu1, line 408: This assumes Fluent is writing out the automatic transcript, which can be turned off from Preferences.

Another approach is to look for the cleanup script filename pattern in Fluent's working directory.

We will add an API to get the cleanup script name to avoid all these complexities.

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -285,6 +361,95 @@ def id(self) -> str:
...
+        if not self.connection_properties.inside_container:
+            logger.error(
+                f"Session is not inside a container, cannot kill Fluent container. "
+                f"Use kill_fluent() instead."
+            )
+            return
+        if self._remote_instance is not None:
+            logger.error(f"Fluent is running remotely, cannot kill Fluent container.")
+            return
+        container_id = self.connection_properties.cortex_host
```
mkundu1, line 450: Where are we storing the container-id?

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -324,9 +489,57 @@ def get_fluent_version(self):
         """Gets and returns the fluent version."""
         return self.scheme_eval.version
 
-    def exit(self) -> None:
-        """Close the Fluent connection and exit Fluent."""
-        self._finalizer()
+    def exit(self, force: bool = None) -> None:
```
mkundu1, line 492: If we have the definition as `exit(self, timeout=None)`, `PYFLUENT_FORCE_EXIT` and `TIMEOUT_TO_FLUENT_KILL` can be combined to single env. `PYFLUENT_TIMEOUT_TO_FLUENT_KILL`.

.github/workflows/ci.yml
```diff
@@ -26,6 +26,10 @@ env:
   DOCUMENTATION_CNAME: 'fluent.docs.pyansys.com'
   PACKAGE_NAME: 'ansys-fluent-core'
   PACKAGE_NAMESPACE: 'ansys.fluent.core'
+  ANSYSLMD_LICENSE_FILE: ${{ format('1055@{0}', secrets.LICENSE_SERVER) }}
+  PYFLUENT_START_INSTANCE: 0
+  PYFLUENT_FORCE_EXIT: 1
+  PYFLUENT_LAUNCH_CONTAINER: 1
```
mkundu1, line 32: Thanks!

src/ansys/fluent/core/fluent_connection.py
lagacep-ans: would the transcript streaming be also a valid data source?

but yeah for all these methods we'll need to have
 - v232 compatible way of doing things
  - v241 cleaner approach, with native apis (as schemeeval is also deprecated)
  -

src/ansys/fluent/core/fluent_connection.py
raph-luc: I think that is a great idea, thanks Mainak

src/ansys/fluent/core/fluent_connection.py
raph-luc: > @mkundu1: This assumes Fluent is writing out the automatic transcript, which can be turned off from Preferences.

That is correct, I was assuming that typical PyFluent use would have the Fluent automatic transcripts enabled, as I believe we currently don't support turning them off through PyFluent (except maybe through tui/scheme-eval?) and auto transcripts are by default enabled on Fluent. 

> @mkundu1: Another approach is to look for the cleanup script filename pattern in Fluent's working directory.

The issue is that the user could and probably often will have multiple cleanup scripts on the same directory for different Fluent sessions, and the cleanup script name has the PID of the host Fluent process, not the Cortex PID, which is what we have access to. The Cortex PID is instead on the Fluent transcript file name. That is why the procedure I set up, to make sure we are killing the correct Fluent client, is: find the transcript for the correct Cortex PID -> find the correct cleanup file inside the transcript and check the hostname matches as well -> execute said cleanup file. 

An alternative would be to parse all cleanup script files in the work dir (as I am not aware of a way to tell which is the correct one, as we can with the transcripts, without going through all scripts in the dir) and then checking every individual PID being killed by the cleanup script, to find the one that matches the Cortex PID, but I believe this is ultimately more prone to failure). I am going to suggest an algorithm that combines both approaches. I am checking if there is a way to get the Fluent host-process PID directly rather than the Cortex PID. Please let me know if you can think of a better way of doing this.

> @mkundu1: We will add an API to get the cleanup script name to avoid all these complexities.

I think that would be the best solution in the long term, but would only be for next year's release of Fluent, right? Or can we do that for Fluent v232 as well? A Fluent settings API command that returns the session's cleanup script name (or specifically the Fluent host PID). That would make things significantly more straightforward. 

> @lagacep-ans: would the transcript streaming be also a valid data source?

Not in the current version. The initial portion of the transcript that has the cleanup script information is not streamed to PyFluent (it is printed out before the server has started and transcript streaming is set up).

src/ansys/fluent/core/fluent_connection.py
raph-luc: We are not storing the container id directly, only the cortex_host variable inside `connection_properties`. The hostname of the container is the same as the container id, which can be used for docker commands.

src/ansys/fluent/core/fluent_connection.py
raph-luc: Wasn't aware of frozen dataclasses, that is exactly what I wanted, thank you again Mainak

Putting together Patrick's suggestions and making FluentConnectionProperties a frozen dataclass

src/ansys/fluent/core/fluent_connection.py
raph-luc: That is great, tested it and `(cx-client-id)` returns the Fluent host PID that we need. So we can avoid having to find and check Fluent transcript output and simplifies things a bit, thank you for the heads-up

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -85,6 +93,42 @@ def __call__(self):
...
+    inside_container: bool
+
+    def list_names(self) -> list:
+        """List all property names."""
+        return [k for k, _ in vars(self).items()]
+
+    def list_values(self) -> dict:
+        """Dict with all property names and values."""
+        return vars(self)
+
```
mkundu1, line 130: I think the above 2 are same as `fields` and `asdict` methods of dataclass, we can mention them in the docstring.

raph-luc: I don't see those using dir() 
image
`__dataclass_fields__` isn't the same. `__dict__` returns the same as `list_values()`, I can make list_values a wrapper for `__dict__`, going to do that

Not finding anything particularly useful in dataclass documentation, just a separate `dataclasses.asdict()` https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict which seems to be very slow and does a lot more than we need here https://stackoverflow.com/questions/52229521/why-is-dataclasses-asdictobj-10x-slower-than-obj-dict

mkundu1: I didn't realize those are not instance methods. Current methods are fine, thanks for checking.

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -324,9 +483,74 @@ def get_fluent_version(self):
         """Gets and returns the fluent version."""
         return self.scheme_eval.version
 
-    def exit(self) -> None:
-        """Close the Fluent connection and exit Fluent."""
-        self._finalizer()
+    def exit(self, timeout: float = None, timeout_force: bool = True) -> None:
```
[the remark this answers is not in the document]
raph-luc: > What is the usage of `timeout = <some value>, timeout_force = False` combination?

It is for when the user wants timeout to happen but does not want to kill the Fluent client on exit. In other words, does not want the exit command to hang the Python interpreter, but also does not want kill the Fluent client when attempting exit). 

Does this make sense? In the above example, the user could set a timeout of 0 or above, and `timeout_force = False`.

> runs the cleanup script immediately after attempting to cleanly exit. I'm seeing the error

I believe that should only happen if you have the environment variable defined to timeout and kill, going to see if I can reproduce this issue here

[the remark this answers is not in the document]
raph-luc: > runs the cleanup script immediately after attempting to cleanly exit. I'm seeing the error

Found the issue, it was related to some local changes I was trying but didn't clean up properly before committing changes, so the resulting code was one tab space wrong (as you can see in the commit). Thanks for testing and the heads-up @mkundu1 !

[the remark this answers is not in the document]
mkundu1: Thanks for the explanation and fix.

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -1,6 +1,10 @@
 from ctypes import c_int, sizeof
+from dataclasses import dataclass
```
seanpearsonuk, line 2: @prmukherj n.b. you are also looking at updating this module currently. I don't know if you already started. If so, there will be conflicts.

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -690,6 +690,7 @@ def launch_fluent(
                     cleanup_on_exit=cleanup_on_exit,
                     start_transcript=start_transcript,
                     launcher_args=argvals,
+                    inside_container=True,
```
seanpearsonuk, line 693: In all other scenarios FluentConnection will have it that inside_container=False, which might not always be correct.

raph-luc: That is true. If the user for example manually starts up the Fluent docker container without using PyFluent, gets the proper connection information, and sets up the connection to PyFluent using `FluentConnection(..., inside_container=False)`, which is the default, the `FluentConnection` will work, but `force_exit` won't work. Ideally the user would correctly specify that `inside_container=True`. 

I am not aware of other scenarios where the current set up wouldn't work.

I don't currently know of a better way to tell whether Fluent is running inside a container or not. Going to investigate this. @mkundu1 perhaps any ideas that come to mind?

Regardless of whether we find a better way to check if inside container or not, for now I am going to change `inside_container` default from `False` to `None`, which I believe better passes the information that it hasn't been specified whether it is inside a container or not.

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -285,6 +362,87 @@ def id(self) -> str:
         """Return the session id."""
         return self._id
 
+    def force_exit(self):
```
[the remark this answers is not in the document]
raph-luc: I think that, regardless of server-side changes, it is a good idea to have this functionality in PyFluent, as it can also deal with situations where the server/container is unresponsive or frozen.

seanpearsonuk (approved):

mkundu1 (approved):

lagacep-ans (approved):

prmukherj (approved):

raph-luc:
Waiting for https://github.com/ansys/pyfluent/pull/1671

PR #1689, opened 2023-06-15
Feat/additional example

acarvalh-work:
Example added: 

- Modeling External Compressible Flow
https://ansyshelp.ansys.com/account/secured?returnurl=/Views/Secured/corp/v231/en/flu_tg/flu_tg_oneram6_wing.html


seanpearsonuk (approved):

mkundu1 (approved):

raph-luc (changes requested):
Good job, tested it here and it worked fine. I haven't tested the doc build.

Merging main to update this branch might help with the tests.

If possible (not sure due to the doc build) I would strongly suggest using `save_path` and specifying `cwd` for `launch_fluent` so that the example and output files don't end up in multiple different locations and are easier to clean. A good default path to save is `pyfluent.EXAMPLES_PATH`.

@seanpearsonuk quick question: how do we decide which examples go in `pyfluent-examples` and which ones go here?

raph-luc (commented):
Thanks, with the `cwd` change it is already much better as at least all Fluent input and output files end up in the same location.

examples/00-fluent/turbo_machinery.py
```diff
@@ -0,0 +1,815 @@
...
+#   :align: center
+
+#  - Total Mass Flow Rate at the Outlet
+
+###############################################################################
+# .. image:: /_static/turbo_machinery_018.png
+#   :width: 500pt
+#   :align: center
+
+solver_session.tui.solve.iterate(100)
```
raph-luc, line 797: Could we work with less iterations for this example so that it runs quicker? As long as it doesn't make the results completely misleading. I think (may be wrong) this is also going to influence how long the documentation build takes. 

I would try to use as few iterations as possible that still gets somewhat decent results. Maybe 20 iterations?

examples/00-fluent/turbo_machinery.py
```diff
@@ -0,0 +1,815 @@
...
+
+
+###############################################################################
+# Save data file
+# ~~~~~~~~~~~~~~
+# Save the data file as (``turbo_workflow.dat.h5``).
+
+solver_session.tui.file.write_data("turbo_workflow.dat.h5")
```
raph-luc, line 805: Currently fails on v241 here, with the error message:
`Error: Mesh has changed, cannot write the data file without writing the case file`

This should fix it:
```suggestion
solver_session.tui.file.write_case_data("turbo_workflow.cat.h5")
```

prmukherj (approved):

examples/00-fluent/turbo_machinery.py
```diff
@@ -0,0 +1,815 @@
...
+# Perform required imports
+# ~~~~~~~~~~~~~~~~~~~~~~~~
+# Perform required imports, which includes downloading and importing
+# the geometry files.
+
+import ansys.fluent.core as pyfluent
+from ansys.fluent.core import examples
+
+inlet_guide_vane_file, rotor_file, stator_file = [
+    examples.download_file(CAD_file, "pyfluent/turbo_workflow")
```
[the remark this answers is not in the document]
acarvalh-work: Agreed. It will be good to show the user they can define their own save paths within the example.

https://github.com/ansys/pyfluent/blob/1438197148adef64801dbbb1f936653a872bb725/examples/00-fluent/turbo_machinery.py#L52-L57

examples/00-fluent/turbo_machinery.py
acarvalh-work: Changes added.

https://github.com/ansys/pyfluent/blob/1438197148adef64801dbbb1f936653a872bb725/examples/00-fluent/turbo_machinery.py#L802-L807

examples/00-fluent/turbo_machinery.py
raph-luc: I think the tradeoff of having the iterations run ~4x faster is worth it (and improving doc build time). You could document that in these lines:
https://github.com/ansys/pyfluent/blob/1438197148adef64801dbbb1f936653a872bb725/examples/00-fluent/turbo_machinery.py#L767-L769

Saying that 100 iterations is recommended, but using 25 here as an example.

hpohekar (approved):

seanpearsonuk (approved):

examples/00-fluent/external_compressible_flow.py
```diff
@@ -255,89 +251,82 @@
...
+
+solver.setup.materials.fluid[
+    "air"
+].viscosity.sutherland.option = "three-coefficient-method"
+
+solver.setup.materials.fluid["air"].viscosity.sutherland.reference_viscosity = 1.716e-05
+
+solver.setup.materials.fluid["air"].viscosity.sutherland.reference_temperature = 273.11
+
+solver.setup.materials.fluid["air"].viscosity.sutherland.effective_temperature = 110.56
```
seanpearsonuk, line 266: We could create an alias for air e.g., 

```suggestion
air = solver.setup.materials.fluid["air"]

air.density.option = "ideal-gas"

air.viscosity.option = "sutherland"

viscosity_sutherland = air.viscosity.sutherland

viscosity_sutherland.option = "three-coefficient-method"

viscosity_sutherland.reference_viscosity = 1.716e-05

viscosity_sutherland.reference_temperature = 273.11

viscosity_sutherland.effective_temperature = 110.56
```
and so forth

seanpearsonuk (approved):

Issue #1616, opened 2023-05-17
Improvement in the docker run workflow


mkundu1:
1. Add a docker-compose file, instead of `docker run` call
2. Move all related env. vars in a .env file and document them.
3. Improve the exposure of docker container run in launch_fluent.

seanpearsonuk:
Re point 2, should this all be just env var configurable? Should we not allow it to be configured also via `launch_fluent` arguments?

Are we proposing to use the `docker` Python package?

N.b. during hackathon, I created the branch feat/hackathon as a basic way to get around hardcoded container repository (which we could not use) and mount path. Shall we address these in the same issue?

@mkundu1 @raph-luc

PR #1694, opened 2023-06-16
Watchdog to deal with Fluent stray processes

raph-luc:
Update: zero stray containers left on the GitHub CI runners after going through all tests multiple times (before would leave around 2 to 15 stray Fluent containers each run)

By default, Watchdog will only be started when Fluent is launched with `show_gui=False`, or with `start_watchdog=True`

For debugging, set `PYFLUENT_LOGGING=DEBUG` (for PyFluent) and `PYFLUENT_WATCHDOG_DEBUG=ON` (for the Watchdog)

Common situations that leave lots of stray Fluent processes behind include: 

- When running PyFluent tests that open and close many Fluent connections in quick succession (e.g. our CI runs on GitHub)
- Cancelling GitHub CI runs (automatically when for example a new commit is made while previous tests were running, or manually when cancelling workflows)
- When closing the window or terminal that is running Python while a launched Fluent process is active in the background
- When using the "restart" button on VSCode Jupyter or similar functions in other IDEs
- When sending multiple keyboard interrupts (Ctrl+C) to active Python sessions while a launched Fluent process is active in the background
- When either PyFluent or Fluent freeze/hang

Watchdog Pros:

- Can deal with Fluent stray processes in the most common situations listed above

Watchdog Cons:

- Slows down execution when launching Fluent, as the interpreter needs to wait a bit for Watchdog to properly hook to the Fluent client
- Shouldn't be necessary

To-do list (could be done in future PRs if we want to get this up asap to deal with the runner issues):

1. Done: ~Add some communication between watchdog and launcher, so that launcher does not need to wait more than necessary for watchdog, and to avoid situations where processes can still be left behind when Fluent is initialized and starts exiting (then gets stuck) before Watchdog can connect to Fluent and prepare to deal with the stray processes~ 
2. Lots of room to optimize and streamline the code (e.g., using wrappers for executing calls with timeout), review and improve logic
3. Done: ~Change how Fluent without GUI is launched, so that it is in a separate process group and can avoid bad VSCode Jupyter notebook termination attempts (and other similar situations) that leave `mpiexec` and `hydra_pmi_proxy` processes behind. Not an issue on v241 with @lagacep-ans Fluent error management fix~ 
4. Done: ~Figure out if there is a way to hide the Watchdog process window on Windows (this is an issue exclusive to Windows, on Linux no separate window is required)~ 
5. Tests
6. Future PR: update and restructure how Fluent is being launched by PyFluent, use Python Docker SDK for containers (?), use argument list rather than a single string for local non-container Fluent client, also use docker-composer as in #1616 raised by Mainak
7. Future PR: review Fluent launcher documentation and add watchdog documentation

### DIFF
--- a/src/ansys/fluent/core/fluent_connection.py
+++ b/src/ansys/fluent/core/fluent_connection.py
@@ -253,6 +253,7 @@ class FluentConnection:
                 inside_container = True
             else:
                 logger.debug("Assuming Fluent is not running inside a container.")
+                inside_container = False
 
         self.connection_properties = FluentConnectionProperties(
             ip,
@@ -437,6 +438,7 @@ class FluentConnection:
                     _ = async_result.get(timeout=timeout)
                     logger.debug("session.exit() successful")
                     pool.close()
+                    return
                 except TimeoutError:
                     logger.debug("session.exit() timeout")
                     pool.terminate()
--- a/src/ansys/fluent/core/fluent_connection.py
+++ b/src/ansys/fluent/core/fluent_connection.py
@@ -95,7 +95,7 @@ class _IsDataValid:
         return self._scheme_eval.scheme_eval("(data-valid?)")
 
 
-def get_container_ids() -> list[str]:
+def get_container_ids() -> List[str]:
     try:
         logger.debug("Attempting to get docker container IDs...")
         proc = subprocess.Popen(
--- a/src/ansys/fluent/core/fluent_connection.py
+++ b/src/ansys/fluent/core/fluent_connection.py
@@ -2,6 +2,8 @@ from ctypes import c_int, sizeof
 from dataclasses import dataclass
 import itertools
 import logging
+import multiprocessing
+from multiprocessing.context import TimeoutError
 import os
 from pathlib import Path
 import socket
@@ -91,6 +93,21 @@ class _IsDataValid:
 
     def __call__(self):
         return self._scheme_eval.scheme_eval("(data-valid?)")
+
+
+def get_container_ids() -> list[str]:
+    try:
+        logger.debug("Attempting to get docker container IDs...")
+        proc = subprocess.Popen(
+            ["docker", "ps", "-q"], stdout=subprocess.PIPE, stderr=subprocess.DEVNULL
+        )
+        output_bytes = proc.stdout.read()
+        output_str = output_bytes.decode()
+        ids = output_str.strip().split()
+    except:
+        logger.warning("Failed to get docker container IDs.")
+        ids = []
+    return ids
 
 
 @dataclass(frozen=True)
@@ -322,6 +339,14 @@ class FluentConnection:
             cortex_pwd = None
             fluent_host_pid = None
 
+        if inside_container is None and not remote_instance and cortex_host is not None:
+            logger.debug("Checking if Fluent is running inside a container.")
+            if cortex_host in get_container_ids():
+                logger.debug("Fluent is running inside a container.")
+                inside_container = True
+            else:
+                logger.debug("Assuming Fluent is not running inside a container.")
+
         self.connection_properties = FluentConnectionProperties(
             ip,
             port,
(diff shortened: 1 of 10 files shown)


src/ansys/fluent/core/launcher/watchdog.py
```diff
@@ -0,0 +1,302 @@
...
+    ] = f"pyfluent_watchdog_{watchdog_id}.log"
+
+    logger = pyfluent.logging.get_logger("pyfluent.watchdog")
+
+    if os.getenv("PYFLUENT_WATCHDOG_DEBUG", "off").upper() in [1, "1", "ON"]:
+        pyfluent.logging.enable(custom_config=log_config)
+        logger.setLevel("DEBUG")
+        logger.handlers = pyfluent.logging.get_logger(
+            "pyfluent.general"
+        ).handlers  # using same handlers as already defined
```
raph-luc, line 149: This is an example of how the current logging setup can be customized when e.g. running scripts (in this case being used for Watchdog as it runs outside and separate from main PyFluent)

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -55,6 +57,21 @@ def run(self) -> None:
             cb()
 
 
+def get_container_ids() -> List[str]:
+    try:
+        logger.debug("Attempting to get docker container IDs...")
+        proc = subprocess.Popen(
```
seanpearsonuk, line 63: @raph-luc 
Some time subsequent to this PR, would you be interested in switching the code to use the Python docker package?

src/ansys/fluent/core/launcher/watchdog.py
```diff
@@ -0,0 +1,302 @@
...
+
+def launch(main_pid: int, sv_port: int, sv_password: str, sv_ip: str = None):
+    watchdog_id = "".join(
+        random.choices(
+            string.ascii_uppercase + string.ascii_lowercase + string.digits, k=6
+        )
+    )
+
+    env_watchdog_debug = os.getenv("PYFLUENT_WATCHDOG_DEBUG", "off").upper()
+    if env_watchdog_debug in [1, "1", "ON"]:
```
seanpearsonuk, line 24:
```suggestion
    if env_watchdog_debug in ("1", "ON"):
```
1 is not possible.

src/ansys/fluent/core/launcher/watchdog.py
```diff
@@ -0,0 +1,302 @@
...
+        print(
+            f"PYFLUENT_WATCHDOG_DEBUG environment variable found, "
+            f"enabling debugging for watchdog id {watchdog_id}..."
+        )
+
+    logger = pyfluent.logging.get_logger("pyfluent.launcher")
+
+    watchdog_env = os.environ.copy()
+
+    # No auto PyFluent logging to file on the watchdog
```
seanpearsonuk, line 37: @raph-luc is the following code producing the required behaviour?

If the var is not in env, it will still not be in env
If it is in env and falsy, it will still be in env and falsy
If it is in env and truthy, it will not be in env

src/ansys/fluent/core/launcher/watchdog.py
```diff
@@ -0,0 +1,302 @@
...
+    if watchdog_env.get("PYFLUENT_LOGGING"):
```
seanpearsonuk, line 38:
```suggestion
    if "PYFLUENT_LOGGING" in watchdog_env:
```

src/ansys/fluent/core/launcher/watchdog.py
```diff
@@ -0,0 +1,302 @@
...
+
+    # Configure logger for Watchdog process
+    log_config = pyfluent.logging.get_default_config()
+    log_config["handlers"]["pyfluent_file"][
+        "filename"
+    ] = f"pyfluent_watchdog_{watchdog_id}.log"
+
+    logger = pyfluent.logging.get_logger("pyfluent.watchdog")
+
+    if os.getenv("PYFLUENT_WATCHDOG_DEBUG", "off").upper() in [1, "1", "ON"]:
```
seanpearsonuk, line 144:
```suggestion
    if os.getenv("PYFLUENT_WATCHDOG_DEBUG", "OFF").upper() in ("1", "ON"):
```

src/ansys/fluent/core/launcher/watchdog.py
raph-luc: Good catch, thanks!

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -608,6 +622,9 @@ def launch_fluent(
             if topy:
                 launch_string += scm_to_py(topy)
 
+            if _is_windows():
+                launch_string = 'start "" ' + launch_string
```
mkundu1, line 626: Why is it required to start a command prompt?

mkundu1 (approved):

seanpearsonuk (approved):

PR #1695, opened 2023-06-19
Update streaming service and SVARs test.

prmukherj:


raph-luc:
@mkundu1 @prmukherj I think once we fix `test_transcripts`, and possibly after we finish and merge the watchdog PR to avoid zombie containers, we could try reverting this change: https://github.com/ansys/pyfluent/pull/1677, so that we have more runners available and so that test queues run quicker

hpohekar (approved):

seanpearsonuk (approved):

PR #1701, opened 2023-06-20
Stop all containers before each test to fix test_transcript

mkundu1:


mkundu1:
There seems to be some issue due to the running containers from previous tests (probably due to resource limit). Also, The test [runs](https://github.com/ansys/pyfluent/actions/runs/5329140067) singly without any issues.

.github/workflows/run-custom-tests.yml
```diff
@@ -0,0 +1,88 @@
+
+name: Run Custom Tests
```
mkundu1, line 2: Useful for debugging specific test(s) in github

mkundu1:
> There seems to be some issue due to the running containers from previous tests (probably due to resource limit). Also, The test [runs](https://github.com/ansys/pyfluent/actions/runs/5329140067) singly without any issues.

But the test has failed in the latest run. Server-side issue probably?

.github/workflows/run-custom-tests.yml
prmukherj: this will be useful, Thank you.

raph-luc:
> @mkundu1: There seems to be some issue due to the running containers from previous tests (probably due to resource limit). Also, The test [runs](https://github.com/ansys/pyfluent/actions/runs/5329140067) singly without any issues.

If that is part of the issue, then it will also be addressed by the watchdog

> @mkundu1: But the test has failed in the latest run. Server-side issue probably?

Does look like it is an issue exclusive to the CI runners, I am currently attempting to reproduce it locally using this branch's version of `test_transcript`, with and without containers, so far unable to reproduce. Will report if I get any failures

PR #1706, opened 2023-06-22
Enable nightly doc build warnings

hpohekar:
Enable nightly doc build warnings.

https://github.com/ansys/pyfluent/actions/runs/5352121166

image


raph-luc (approved):

hpohekar:
Nightly doc build - https://github.com/ansys/pyfluent/actions/runs/5352121166

prmukherj (approved):

dnwillia:
> If I understood it correctly, this is going to cause nightly doc builds to fail if there were any warnings at all.
> 
> My first thought is that it may be a good idea to disable some warnings such as `GL08: The object does not have a docstring` as this is not always necessarily an issue in my opinion. Adding a docstring to every single object/method, even when it is obvious and arguably unnecessary (e.g. just repeating the method or class name), just because of that warning, looks like a mistake to me.
> 
> Regardless, if others prefer to always have docstrings on all objects and go ahead with this, that is fine with me as well.
> 
> Also, wouldn't it be a good idea to have this same behavior on branches, before they get merged into main as well? Otherwise I expect we will forget about this and make small doc changes that result in warnings that no one notices on the PR, and have nightly doc builds failing constantly.

Yes, that's what it would do.  When I originally did the doc infrastructure work last year I activated it with warnings as errors but a few months ago this got deactivated and now there are a number of undocumented classes and modules in the code.

The problem with this IMO is that you have no context about why the module is there, why the class is there, what it's methods are for and you are left to reverse engineer the context by staring at the code or writing tests for it to see what it does.

The bottom line is everyone should be documenting their code if they want anyone to use it.

dnwillia:
Regarding this:

> Also, wouldn't it be a good idea to have this same behavior on branches, before they get merged into main as well? Otherwise I expect we will forget about this and make small doc changes that result in warnings that no one notices on the PR, and have nightly doc builds failing constantly.

The answer is yes.  It used to be active for branches as well... but that was also deactivated.

hpohekar:
We will merge this after fixing all existing warnings.

hpohekar:
Mainak has already fixed `noindex` warnings and right now we are adding missing docstrings.

PR #1733, opened 2023-06-28
Package 23.2 API files

mkundu1:
Just keeping the branch ready, will finalize and have a public release around 23.2 FCA.


.github/workflows/nightly-dev-doc-build.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        image-tag: [v23.1.0, v23.2.0, v24.1.0]
+        image-tag: [v22.2.0, v23.1.0, v23.2.0, v24.1.0]
```
mkundu1, line 22: I removed v22.2.0 in a previous PR. As we are packaging 22.2 files, it is better to have the doc tested - so reinstating it.

examples/00-fluent/mixing_elbow_settings_api.py
```diff
@@ -180,7 +180,7 @@
 # Solve for 150 iterations.
 
 solver.solution.run_calculation.iterate.get_attr("arguments")
-if solver.get_fluent_version() >= "23.1.0":
+if solver.get_fluent_version() >= "23.2.0":
```
raph-luc, line 183: I think we should not change this one? 

We could otherwise consider just removing this version check and backward compatibility from the example, assuming that examples are always run with the latest released version (e.g. doc builds).

raph-luc (approved):

seanpearsonuk (approved):

PR #1735, opened 2023-06-28
Maint/quantity unit systems

acarvalh-work:
Fixes - https://github.com/ansys/pyfluent/pull/1664#issuecomment-1581523362

- Added a unit system class (and tests)
   - Users can create their own custom units and  unit systems and use them with the existing per-defined units and unit systems. 
- Added error handling for common errors (all modules)

### DIFF
--- a/src/ansys/fluent/core/quantity/dimensions.py
+++ b/src/ansys/fluent/core/quantity/dimensions.py
@@ -127,7 +127,7 @@ class Dimensions(object):
         return self._dimensions
 
 
-class DimensionsError(TypeError):
+class DimensionsError(ValueError):
     """Custom dimensions errors."""
 
     def __init__(self, err):
@@ -135,8 +135,12 @@ class DimensionsError(TypeError):
 
     @classmethod
     def EXCESSIVE_PARAMETERS(cls):
-        cls("Dimensions only accepts 1 of the following: (units) or (dimensions).")
+        return cls(
+            "Dimensions only accepts 1 of the following: (units) or (dimensions)."
+        )
 
     @classmethod
     def EXCESSIVE_DIMENSIONS(cls, len):
-        cls(f"`dimensions` must contain 9 values or less, currently there are {len}.")
+        return cls(
+            f"`dimensions` must contain 9 values or less, currently there are {len}."
+        )
--- a/src/ansys/fluent/core/quantity/dimensions.py
+++ b/src/ansys/fluent/core/quantity/dimensions.py
@@ -22,9 +22,7 @@ class Dimensions(object):
         self, units: str = None, dimensions: list = None, unit_sys: str = None
     ):
         if units and dimensions:
-            raise ValueError(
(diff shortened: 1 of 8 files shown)


hpohekar (approved):

src/ansys/fluent/core/quantity/dimensions.py
```diff
@@ -120,3 +125,22 @@ def units(self):
...
+    @classmethod
+    def EXCESSIVE_PARAMETERS(cls):
+        return cls(
+            "Dimensions only accepts 1 of the following: (units) or (dimensions)."
+        )
+
+    @classmethod
+    def EXCESSIVE_DIMENSIONS(cls, len):
+        return cls(
+            f"`dimensions` must contain 9 values or less, currently there are {len}."
```
seanpearsonuk, line 145: Shall we write **D**imensions in both cases

src/ansys/fluent/core/quantity/dimensions.py
```diff
@@ -20,31 +22,32 @@ def __init__(
...
         self._units_table = q.UnitsTable()
-        unit_sys = unit_sys or "SI"
+        unit_sys = unit_sys or self._units_table.unit_systems["SI"]
 
         if units is not None:
             self._unit = units
             self._dimensions = self._units_to_dim(units=units)
 
         if dimensions:
+            if len(dimensions) > 9:
```
seanpearsonuk, line 35: Let's expose this dimensions list size (=9) from this dimensions module. Then we can reference it here, in unit_system, and wherever else needed.

seanpearsonuk:
@acarvalh-work looks very clean, thanks. I only made a couple of minor comments.

prmukherj (approved):

src/ansys/fluent/core/quantity/dimensions.py
acarvalh-work: Added more clarification to error messages, Dimensions was referring to the class while `dimensions` is the parameter that accepts a list of dimensions.

https://github.com/ansys/pyfluent/blob/16bfd1f51b57625263ffefce7cf960c6691ef305/src/ansys/fluent/core/quantity/dimensions.py#L144-L154

src/ansys/fluent/core/quantity/dimensions.py
acarvalh-work: Added static method to Dimensions to access max dimensions list length (9). exp -> `Dimensions.max_dim_len()`

https://github.com/ansys/pyfluent/blob/16bfd1f51b57625263ffefce7cf960c6691ef305/src/ansys/fluent/core/quantity/dimensions.py#L132-L135

src/ansys/fluent/core/quantity/dimensions.py
seanpearsonuk: thanks for clarifying

seanpearsonuk (approved):

Issue #1747, opened 2023-06-30
grpc error during initial connection


mkundu1:
Currently, any grpc error during initial connection results in a timeout with the runtime error "The connection to the Fluent server could not be established...".  It is better to fail early in those situations by exposing the internal grpc error which may provide useful triaging information from the user's system. 

Relevant is the test test_create_session_from_server_info_file_with_wrong_password which is checking the failure due to the timeout (with the 100 s wait) instead of an immediate failure due to password mismatch.

@raph-luc (originally found the issue)

raph-luc:
Related example test failures:
https://github.com/ansys/pyfluent/actions/runs/5457287691/jobs/9931174803?pr=1759
https://github.com/ansys/pyfluent/actions/runs/5357964161/jobs/9719803655?pr=1715
https://github.com/ansys/pyfluent/actions/runs/5355190417/jobs/9713258474?pr=1712

raph-luc:
The catched gRPC exceptions when these random failures occur are:

First this exception once: 
`HealthCheckService.status() caught RuntimeError: End of TCP stream`

and then repeatedly every second on every gRPC attempt until timeout: 
`HealthCheckService.status() caught RuntimeError: failed to connect to all addresses; last error: UNKNOWN: ipv4:127.0.0.1:63599: connection attempt timed out before receiving SETTINGS frame`

Any ideas what may be causing these seemingly random failures? The Fluent client remains open after this, but haven't checked if it is responsive or not as running it without GUI.

This happens not only when launching Fluent, but also when connecting to an already existing Fluent server (e.g. also happens in the middle of a `tests/test_streaming_services.py::test_transcript` run)

There is no particular test where this seems to happen more often than others, and nothing is written to the `.trn` file when this happens

Also, we probably should consider raising different exceptions for these different gRPC errors rather than `RuntimeError` for all/most of them, maybe we could parse the RpcError text and raise an appropriate exception if we can identify it based on the text?

raph-luc:
First one to be logged on GitHub: https://github.com/ansys/pyfluent/actions/runs/5695317917/job/15438219836

Same issue as I had one time locally, as described in my previous message:
`INFO     pyfluent.general:health_check.py:88 HealthCheckService.status() caught RuntimeError: failed to connect to all addresses; last error: UNKNOWN: ipv4:127.0.0.1:34431: connection attempt timed out before receiving SETTINGS frame`

Until timeout

PR #1748, opened 2023-06-30
Docker API support and container launch exposure

raph-luc:
WIP


src/ansys/fluent/core/utils/execution.py
raph-luc, line 1: Added `timeout_exec()` and `timeout_loop()` functions to this file (previously async_execution.py). 

These can be used to minimize wait times and avoid lockouts, for example rather than `time.sleep(5)` we can `timeout_loop(func, 5)` instead, which will exit earlier if `func()` returns the expected value earlier. This can quickly add up, particularly in our CI runs with many `time.sleep()`s which are always waiting for significantly longer than necessary. I have replaced the longest time.sleep calls, but haven't worked through all of them.

src/ansys/fluent/core/services/health_check.py
```diff
@@ -81,6 +84,10 @@ def status(self) -> str:
             try:
                 return self.check_health()
             except Exception:
+                ex_type, ex_value, _ = sys.exc_info()
+                logger.info(
+                    f"HealthCheckService.status() caught {ex_type.__name__}: {ex_value}"
+                )
```
raph-luc, line 90: Added a logging report whenever a general exception is caught here, as this was hiding relevant exceptions (see https://github.com/ansys/pyfluent/issues/1747)

tests/test_session.py
```diff
@@ -169,7 +169,9 @@ def test_create_session_from_server_info_file_with_wrong_password(
     server_info_file.write_text(f"{ip}:{port}\n1234")
     with pytest.raises(RuntimeError):
         session = BaseSession.create_from_server_info_file(
-            server_info_filepath=str(server_info_file), cleanup_on_exit=False
+            server_info_filepath=str(server_info_file),
+            cleanup_on_exit=False,
+            start_timeout=2,
         )
```
raph-luc, line 175: Reduced the timeout from this test, as it was always "succeeding" when timing out after 100 seconds every time. This is related to the general exception catcher that was hiding an expected wrong password exception, see changes to `health_check.py` file in this commit.

As no actual Fluent session or client is being launched, this test should ideally fail instantly due to the wrong password.

dnwillia:
> See comments below

Raphael is there an issue that documents what this is for?

raph-luc:
> @dnwillia  Raphael is there an issue that documents what this is for?

Hi Dan, yes, there is an issue associated to this draft PR, it should appear under the "Development" heading in the right sidebar: https://github.com/ansys/pyfluent/issues/1616 (see Sean's reply in particular)

src/ansys/fluent/core/launcher/fluent_container.py
raph-luc, line 1: Most important changes are in this file

src/ansys/fluent/core/utils/execution.py
raph-luc, line 1: Added `timeout_exec()` and `timeout_loop()` functions to this file (previously async_execution.py). 

These can be used to minimize wait times and avoid lockouts, for example rather than `time.sleep(5)` we can `timeout_loop(func, 5)` instead, which will exit earlier if `func()` returns the expected value earlier. This can quickly add up, particularly in our CI runs with many `time.sleep()`s which are always waiting for significantly longer than necessary. I have replaced the longest time.sleep calls, but haven't worked through all of them.

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -58,20 +59,33 @@ def run(self) -> None:
...
-    except:
-        logger.warning("Failed to get docker container IDs.")
-        ids = []
-    return ids
+        docker_client = docker.from_env()
+        container = docker_client.containers.get(container_id_or_name)
+    except docker.errors.NotFound:  # NotFound is a child from DockerException
+        return False
+    except docker.errors.DockerException:
+        return None
```
mkundu1, line 87: probably better to propagate the exception than to differentiate between `None` and `False`

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -671,19 +718,19 @@ def launch_fluent(
         if meshing_mode:
             args.append(" -meshing")
 
-        import ansys.fluent.core as pyfluent
-
-        host_mount_path = pyfluent.EXAMPLES_PATH
-        if not os.path.exists(host_mount_path):
-            os.makedirs(host_mount_path)
+        if dry_run:
+            image_name, config_dict, _, _, _, _ = configure_container_dict(
```
mkundu1, line 722:
```suggestion
            image_name, config_dict, *_ = configure_container_dict(
```

setup.py
```diff
@@ -31,6 +31,8 @@
     "h5py>=3.8.0",
     "lxml>=4.9.2",
     "pyyaml",
+    "docker",
```
mkundu1, line 34: It is generally better to specify dependency versions to avoid any incompatibility due to user's installed version.

src/ansys/fluent/core/launcher/fluent_container.py
mkundu1: Looks very good!

mkundu1 (approved):

hpohekar (approved):

prmukherj (approved):

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -355,17 +375,32 @@ def force_exit_container(self):
...
+                f"Fluent is running remotely, cannot terminate Fluent container."
             )
             return
-        if self._remote_instance is not None:
-            logger.error(f"Fluent is running remotely, cannot kill Fluent container.")
+        container = self.connection_properties.inside_container
+        if not container:
+            logger.error(
+                f"Session is not inside a container, cannot terminate Fluent container. "
+                f"Try force_exit() instead."
```
seanpearsonuk, line 387: there are a few redundant f-strings in this module (already here)

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -564,31 +594,48 @@ def launch_fluent(
...
         fluent_launch_mode = LaunchMode.PIM
-    elif os.getenv("PYFLUENT_LAUNCH_CONTAINER") == "1":
-        fluent_launch_mode = LaunchMode.CONTAINER
+    elif start_container is True or (
+        start_container is None
+        and container_dict
+        or os.getenv("PYFLUENT_LAUNCH_CONTAINER") == "1"
+    ):
+        if check_docker_support():
+            start_container = True
```
seanpearsonuk, line 608: rather than assigning `start_container` at this point, we could just `del` it and refer only to `fluent_launch_mode` hereafter?

seanpearsonuk:
We might want to refactor the launcher code again. The functions here have a tendency to grow. It would be good to keep the functions small so that whole is modular -  we want to promote reuse for PyCFX etc. I will create a separate issue.

src/ansys/fluent/core/fluent_connection.py
raph-luc: @seanpearsonuk @mkundu1 My objective here, as in the function docstring, is: 

If the system is not correctly set up to run Docker containers, returns ``None``.
If the container was not found, returns ``False``.

Raising the exception would stop execution, while that is excessive when just checking if a container exists. This is used in a few places where we do not necessarily know whether Fluent is inside a container or not (e.g. when starting up FluentConnection without specifying when it is inside a container or not, or when checking from the watchdog whether Fluent is inside a container).

src/ansys/fluent/core/fluent_connection.py
raph-luc: Thank you for the heads-up, going to double-check. Not sure why they ended up with a f-format specifier, probably due to changes over time and I forgot to remove them.

src/ansys/fluent/core/fluent_connection.py
raph-luc: On second thought, I think it is a good idea to at least log the exception, going to do that.

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -58,20 +59,34 @@ def run(self) -> None:
...
-        logger.warning("Failed to get docker container IDs.")
-        ids = []
-    return ids
+        docker_client = docker.from_env()
+        container = docker_client.containers.get(container_id_or_name)
+    except docker.errors.NotFound:  # NotFound is a child from DockerException
+        return False
+    except docker.errors.DockerException as exc:
+        logger.info("%s: %s" % (type(exc).__name__, exc))
+        return None
```
raph-luc, line 88: Added a logging report when catching Docker exceptions (I would expect these can usually be safely ignored when getting a container object, but still probably best to at least log them)

mkundu1 (approved):

raph-luc:
By the way, with these changes (mainly I believe due to `time.sleep` -> `timeout_loop()`), the time taken for all tests to complete appears to have been reduced by approximately 20%, e.g., unit test v232 from 25 min to 20 min, unit test v241 from 14 min to 11 min, disregarding other factors
(see https://github.com/ansys/pyfluent/actions/runs/5467911359/jobs/9955261111 vs https://github.com/ansys/pyfluent/actions/runs/5467665522/jobs/9954548633)

PR #1760, opened 2023-07-05
Remove settings API arguments method

hpohekar:
Removed settings API arguments method.

image


seanpearsonuk (approved):
Thanks for looking into this so quickly @hpohekar.

prmukherj (approved):

Issue #1769, opened 2023-07-07
launch_fluent server_info_filepath


seanpearsonuk:
- server_info_filepath is undocuemented - did we intend to remove it?
- it is reassigned before being used

seanpearsonuk:
@raph-luc would you like to look at this one

raph-luc:
Will do

PR #1780, opened 2023-07-07
Revising fluent launch arguments

raph-luc:
Fixes https://github.com/ansys/pyfluent/issues/1769

Removing support for `launch_fluent(server_info_filepath=...)` as it does not make much sense to support it considering that the file will be deleted by Fluent when it exits, regardless of what we do in PyFluent.

Raising `ValueError` exceptions if arguments are provided in unsupported combinations. Might've missed a few but I think I at least got the most important ones, please let me know (or directly commit) if you notice any missed combinations.

### DIFF
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -611,6 +611,14 @@ def launch_fluent(
 
     del start_container
 
+    if additional_arguments is None:
+        additional_arguments = ""
+    elif fluent_launch_mode == LaunchMode.PIM:
+        raise ValueError(
+            "'additional_arguments' option for 'launch_fluent' is currently not supported "
+            "when starting a remote Fluent PyPIM client."
+        )
+
     if fluent_launch_mode == LaunchMode.PIM and start_watchdog:
         raise ValueError(
             "'start_watchdog' argument for 'launch_fluent' is currently not supported "
@@ -637,7 +645,6 @@ def launch_fluent(
 
     if fluent_launch_mode != LaunchMode.STANDALONE:
         arg_names = [
-            "additional_arguments",
             "env",
             "cwd",
             "topy",
@@ -661,11 +668,9 @@ def launch_fluent(
 
     if fluent_launch_mode == LaunchMode.STANDALONE:
         if lightweight_mode is None:
+            # note argvals is no longer locals() here due to _get_session_info() pass
+            argvals.pop("lightweight_mode")
             lightweight_mode = False
-        if additional_arguments is None:
-            additional_arguments = ""
-
-        argvals = locals()
 
         _raise_exception_g_gu_in_windows_os(additional_arguments)
 
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -628,8 +628,7 @@ def launch_fluent(
     if (
         start_watchdog is None
         and cleanup_on_exit
-        and (fluent_launch_mode in (LaunchMode.CONTAINER, LaunchMode.STANDALONE)
-        )
+        and (fluent_launch_mode in (LaunchMode.CONTAINER, LaunchMode.STANDALONE))
     ):
         start_watchdog = True
 
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -461,7 +461,7 @@ def launch_fluent(
     processor_count: int = None,
     journal_filepath: str = None,
     start_timeout: int = 100,
-    additional_arguments: str = "",
+    additional_arguments: str = None,
     env: Dict[str, Any] = None,
     start_container: bool = None,
     container_dict: dict = None,
@@ -471,9 +471,8 @@ def launch_fluent(
     show_gui: bool = None,
     case_filepath: str = None,
     case_data_filepath: str = None,
-    lightweight_mode: bool = False,
+    lightweight_mode: bool = None,
     mode: Union[FluentMode, str, None] = None,
-    server_info_filepath: str = None,
     py: bool = None,
     gpu: bool = None,
     cwd: str = None,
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -615,20 +611,69 @@ def launch_fluent(
...
+            "topy",
+            "case_filepath",
+            "lightweight_mode",
+            "journal_filepath",
+            "case_data_filepath",
+        ]
+        invalid_arg_names = []
+        for arg_name in arg_names:
+            if argvals[arg_name] is not None:
+                invalid_arg_names.append(arg_name)
```
seanpearsonuk, line 659:
```suggestion
        invalid_arg_names=filter(lambda arg_name:argvals[arg_name] is not None, arg_names)
```

(for example)

raph-luc: I knew there had to be a better way to do that. Wasn't familiar with `filter()`, that is going to be useful, thanks Sean

mkundu1: `invalid_arg_names = [arg_name for arg_name in arg_names if argvals[arg_name] is not None]`

if you prefer list-comprehension

raph-luc: Using list comprehension is also better than what I suggested initially, thanks Mainak. 

Between filter and list comprehension, either one works fine I think, going to stick with filter for this one as it is an interesting built-in function :)

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -615,20 +611,68 @@ def launch_fluent(
...
+            "'start_watchdog' argument for 'launch_fluent' is currently not supported "
+            "when starting a remote Fluent PyPIM client."
+        )
+
+    if (
+        start_watchdog is None
+        and cleanup_on_exit
+        and (
+            fluent_launch_mode == LaunchMode.CONTAINER
+            or fluent_launch_mode == LaunchMode.STANDALONE
```
seanpearsonuk, line 633:
```suggestion
        and (fluent_launch_mode in (LaunchMode.CONTAINER, LaunchMode.STANDALONE)
```

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -615,20 +611,68 @@ def launch_fluent(
 
     del start_container
 
-    if start_watchdog is None and cleanup_on_exit:
+    if additional_arguments is None:
+        additional_arguments = ""
+    elif fluent_launch_mode == LaunchMode.PIM:
```
seanpearsonuk, line 616: In later changes, fluent_launch_mode could be a key `launcher_factory.create(key)` then each launcher could be responsible for constraining/validating its own launch configuration in `launcher.launch(<args>)`. E.g., `pim_launcher.launch(<args>)` would not have an `additional_arguments` argument.

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -615,20 +611,68 @@ def launch_fluent(
...
+        raise ValueError(
```
seanpearsonuk, line 617: these are normally `TypeError`. 
```
>>> def foo(): pass
...
>>>
>>> foo(1)
TypeError: foo() takes 0 positional arguments but 1 was given
>>> foo(x=1)
TypeError: foo() got an unexpected keyword argument 'x'
>>>
```

src/ansys/fluent/core/launcher/launcher.py
seanpearsonuk: Bear in mind that there is a performance difference. One of these is almost twice as quick a the other:
```
>>> xs = range(100000000)
>>> ys = list(filter(lambda x: x%2, xs))
>>> zs = [x for x in xs if x%2]
```
but it doesn't really matter here.

src/ansys/fluent/core/launcher/launcher.py
raph-luc: In this case the "type" (`NoneType` or `str`) is correct as the function does accept and expect that variable with those types. However, with this combination of arguments, the "value" is incorrect as it needs to be `None`. I think `ValueError` makes a bit more sense than `TypeError`.

Googling this, most people seem to suggest `ValueError` as well, first Google links that show up here:
https://stackoverflow.com/questions/256222/which-exception-should-i-raise-on-bad-illegal-argument-combinations-in-python

https://www.adamsmith.haus/python/answers/how-to-raise-an-exception-for-illegal-argument-combinations-in-python

seanpearsonuk: Okay, that makes sense, yes.

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -615,20 +611,68 @@ def launch_fluent(
...
-        if not fluent_launch_mode == LaunchMode.CONTAINER:
+    if dry_run and fluent_launch_mode != LaunchMode.CONTAINER:
+        raise ValueError(
+            "'dry_run' argument for 'launch_fluent' currently is only "
+            "supported when starting containers."
+        )
+
+    argvals = locals()
+
+    if fluent_launch_mode != LaunchMode.STANDALONE:
```
seanpearsonuk, line 646: We can really see how the launcher factory is going to smash this! I will create a detailed gh issue.

raph-luc:
@seanpearsonuk should I close or wait on this PR due to the incoming launch fluent rework which is probably going to make most of this obsolete? Or should we have that rework perhaps based on this branch?

seanpearsonuk:
> @seanpearsonuk should I close or wait on this PR due to the incoming launch fluent rework which is probably going to make most of this obsolete? Or should we have that rework perhaps based on this branch?

I would be very happy for you to merge this, assuming there are no outstanding points.

hpohekar (approved):

mkundu1 (approved):

seanpearsonuk (approved):

PR #1805, opened 2023-07-14
Remove dunders from auto completer

hpohekar:


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -743,6 +743,27 @@ def get_object_names(self):
...
+        for command_name in self.command_names:
+            if command_name.startswith(prefix):
+                command = getattr(self, command_name)
+                if command.is_active():
+                    ret.append([command_name, Command.__name__, command.__doc__])
+        for query_name in self.query_names:
+            if query_name.startswith(prefix):
+                query = getattr(self, query_name)
+                if query.is_active():
+                    ret.append([query_name, Query.__name__, query.__doc__])
```
seanpearsonuk, line 764: I think we should eliminate the duplication of these two almost identical blocks.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -359,6 +359,36 @@ class BooleanList(SettingsBase[BoolListType], Property):
...
+                        ]
+                    )
+                else:
+                    ret.append(
+                        [
+                            command_or_query_name,
+                            Query.__name__,
+                            command_or_query_name.__doc__,
+                        ]
+                    )
```
seanpearsonuk, line 388: @hpohekar 
- it's only the selection between `Command` and `Query` that depends on the value of `command`
- `command` could be renamed `is_command` or similar because `command` reads as if it is a command object
- `command_or_query_names` and `command_or_query_name` are a bit long-winded. `name` is fine for the latter

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -359,33 +359,21 @@ class BooleanList(SettingsBase[BoolListType], Property):
...
+                    ret.append([name, Command.__name__, name.__doc__])
                 else:
-                    ret.append(
-                        [
-                            command_or_query_name,
-                            Query.__name__,
-                            command_or_query_name.__doc__,
-                        ]
-                    )
+                    ret.append([name, Query.__name__, name.__doc__])
```
seanpearsonuk, line 376: @hpohekar @mkundu1 I had in mind `ret.append([name, (Command if is_command else Query).__name__, name.__doc__])`. But that might not be considered good style. Otherwise the conditional could be pulled out of the `append` call - the key is to avoid all duplication either way.

src/ansys/fluent/core/solver/flobject.py

seanpearsonuk (approved):

raph-luc (approved):

PR #1807, opened 2023-07-14
Enable parametric test for 24.1

mkundu1:
Test run result - https://github.com/ansys/pyfluent/actions/runs/5559483041

PR #1809, opened 2023-07-17
Fluent container and testing improvements

raph-luc:
Originally these were suggested changes for https://github.com/ansys/pyfluent/pull/1807

Improving how `fluent_container.py` parses mount path specifications for container run

Re-working  the nightly `test_parametric_workflow` so that it works with containers on both Windows and Linux hosts (test results for this test here: https://github.com/ansys/pyfluent/actions/runs/5592225363), also note this test currently did not work using standalone Fluent on my end: https://github.com/ansys/pyfluent/issues/1820

Activating previously skipped `test_batch_ops_create_mesh_and_access_fails` 

Improving `test_get_all_rp_vars` so it does not need a unique `new_solver_session_no_transcript_examples_path` setup, and so it works on Linux and Windows hosts as well

For `test_batch_ops_create_mesh` , improving skip condition coverage, and connecting test skip message to the issue tracker on GitHub so we can more easily track why this test was skipped

### DIFF
--- a/tests/parametric/test_parametric_workflow.py
+++ b/tests/parametric/test_parametric_workflow.py
@@ -1,5 +1,6 @@
 import os
 from pathlib import Path
+import shutil
 import tempfile
 
 import pytest
@@ -11,7 +12,7 @@ from ansys.fluent.core import examples
 @pytest.mark.nightly
 @pytest.mark.fluent_version(">=23.2")
 def test_parametric_workflow(monkeypatch: pytest.MonkeyPatch):
-    # path needs to exist for mkdtemp
+    # parent path needs to exist for mkdtemp
     Path(pyfluent.EXAMPLES_PATH).mkdir(parents=True, exist_ok=True)
     tmp_save_path = tempfile.mkdtemp(dir=pyfluent.EXAMPLES_PATH)
     import_filename = examples.download_file(
@@ -171,3 +172,5 @@ def test_parametric_workflow(monkeypatch: pytest.MonkeyPatch):
     solver_session.file.parametric_project.archive(archive_name=archive_name)
     assert (Path(tmp_save_path) / new_study_path / archive_name).exists()
     solver_session.exit()
+
+    shutil.rmtree(tmp_save_path)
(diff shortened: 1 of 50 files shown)


tests/parametric/test_parametric_workflow.py
```diff
@@ -130,28 +138,36 @@ def test_parametric_workflow(monkeypatch: pytest.MonkeyPatch):
     assert "New Study" in solver_session.parametric_studies
     del solver_session.parametric_studies[study1_name]
     assert len(solver_session.parametric_studies) == 1
-    project_filename = Path(save_path) / "static_mixer_study.flprj"
-    solver_session.file.parametric_project.save_as(
-        project_filename=str(project_filename)
-    )
-    assert project_filename.exists()
+    project_filename = "static_mixer_study.flprj"
+    solver_session.file.parametric_project.save_as(project_filename=project_filename)
```
mkundu1, line 142: Does it save at the mount path dir if we pass something like `/tmpdir/static_mixer_study.flprj` here?

raph-luc: It should work, yes. 

I tried to minimize the amount of times that we have to check whether a container run was specified, do you think it would be better to, rather than use relative paths with `.flprj.cffdb/...`, check if running in a container and then using absolute paths with the known container mount path?

mkundu1: When we are using `.flprj.cffdb/...`, we are basing the test on a defective Fluent behaiour. So passing an absolute path is better in this case IMHO. The container mount path can be stored in a session property, then we can remove the hardcoding of `/tmpdir` too.

raph-luc: That makes sense, and I agree. I did not realize the Fluent `.flprj.cffdb/...` pathing was considered a mistake, but I can see where you are coming from, that is definitely very unintuitive behavior (same as in the other instances where Fluent changes its own working directory without the user necessarily knowing this).

Going to suggest some changes in light of this. 

By the way, even if we store the container mount path in a session connection property, we'd still need to specify this path every time we are sending commands to Fluent from PyFluent, as Fluent itself is not aware of the container mount path difference, only PyFluent is (unless we rework how Fluent works). Also, we can currently use the `DEFAULT_CONTAINER_MOUNT_PATH` const as it is a little less hardcoded than specifying `/tmpdir` directly.

mkundu1 (approved):

hpohekar (approved):

prmukherj (approved):

seanpearsonuk (approved):

Issue #1800, opened 2023-07-13
Due to recent Fluent v241 changes, flobject.get_root() is returning ~2 MB of data through gRPC


raph-luc:
Call: `GRPC_TRACE: rpc = /ansys.api.fluent.v0.settings.Settings/GetStaticInfo, request = {'root': 'fluent'}`
Response: [response.txt](https://github.com/ansys/pyfluent/files/12042906/response.txt)

seanpearsonuk:
@raph-luc I see that you merged a PR. Is this issue closable?

raph-luc:
@seanpearsonuk The "bigger picture" issue is that we are requesting the full flobject static info often (probably more often than we need to, would require researching whether we can optimize/reduce the number of requests).

In my understanding, as more objects are added to the settings API, the larger these full static info calls will become. Even though the gRPC messages are compressed (2 MB was the uncompressed value, considering compression it was around 1 MB if I recall correctly), this could/will become a performance issue at some point, particularly if Fluent is running remotely from PyFluent.

The issue that actually prompted me to look into this, was that these very large messages were causing logging issues, and that has been resolved by the mentioned PR.

mkundu1:
@raph-luc @seanpearsonuk Currently, the settings API static info is requested even though we have the latest generated files from codegen, because we are comparing the hashes to check for a potential mismatch in the API. We can make the hash comparison a dev feature, i.e. we skip the hash comparison if the connected Fluent is from a released version and there are generated files from codegen for that Fluent version.

Regarding the size of the static info response, we can change it to a streaming response if needed, let's discuss that on Fluent server side.

seanpearsonuk:
@mkundu1 With the current state of play, do you envisage that improvements can be made in this area?

mkundu1:
@seanpearsonuk #3399 has removed the static-info query during runtime.

PR #1825, opened 2023-07-20
Reworking gRPC and TUI debug logging

raph-luc:
Minor suggestion to have debug logs that are much easier to work with, and not overwhelmed by gRPC calls that pass millions of characters or with multiple line breaks each time. Partially addresses https://github.com/ansys/pyfluent/issues/1800 from the logging point of view (but I don't know if passing such gigantic gRPC calls so often is intentional or desired). 

<details><summary> Before </summary>

image

Fluent root settings API call
```
2023-07-19 16:39:20,210 pyfluent.networking   DEBUG    GRPC_TRACE: rpc = /ansys.api.fluent.v0.settings.Settings/GetStaticInfo, request = {'root': 'fluent'}
2023-07-19 16:39:22,301 pyfluent.networking   DEBUG    GRPC_TRACE: response = {'info': ... <2.2 MILLION CHARACTERS> ... }
```

TUI call
```
2023-07-19 16:39:10,286 pyfluent.tui          DEBUG    TUI Command: path: "/file/read_case"
args {
  fields {
    key: "tui_args"
    value {
      list_value {
        values {
          string_value: "ablation.msh.h5"
        }
      }
    }
  }
}

2023-07-19 16:39:10,287 pyfluent.networking   DEBUG    GRPC_TRACE: rpc = /ansys.api.fluent.v0.DataModel/ExecuteCommand, request = {'path': '/file/read_case', 'args': {'tui_args': ['ablation.msh.h5']}}
2023-07-19 16:39:20,164 pyfluent.networking   DEBUG    GRPC_TRACE: response = {}

```

</details>

<details><summary> After </summary>

image

Fluent root settings API call
```
2023-07-20 12:34:31,332 pyfluent.networking   DEBUG    GRPC_TRACE: rpc = /ansys.api.fluent.v0.settings.Settings/GetStaticInfo, request = {'root': 'fluent'}
2023-07-20 12:34:32,882 pyfluent.networking   DEBUG    GRPC_TRACE: response hidden, 2221700 bytes > 1000 bytes limit.

```

TUI call
```
2023-07-20 12:34:21,939 pyfluent.tui          DEBUG    TUI Command: path: "/file/read_case" args { fields { key: "tui_args" value { list_value { values { string_value: "ablation.msh.h5" } } } } }
2023-07-20 12:34:21,941 pyfluent.networking   DEBUG    GRPC_TRACE: rpc = /ansys.api.fluent.v0.DataModel/ExecuteCommand, request = {'path': '/file/read_case', 'args': {'tui_args': ['ablation.msh.h5']}}
2023-07-20 12:34:31,304 pyfluent.networking   DEBUG    GRPC_TRACE: response = {}
```

</details>

### DIFF
--- a/src/ansys/fluent/core/services/datamodel_tui.py
+++ b/src/ansys/fluent/core/services/datamodel_tui.py
@@ -192,7 +192,8 @@ class PyMenu:
         if self._path.startswith("/query/"):
             return self._execute_query(request)
         else:
-            logger.debug(f"TUI Command: {request}")
+            request_str = " ".join(str(request).split())
+            logger.debug(f"TUI Command: {request_str}")
             return self._execute_command(request)
 
     def get_doc_string(self, include_unavailable: bool = False) -> str:
--- a/src/ansys/fluent/core/services/interceptors.py
+++ b/src/ansys/fluent/core/services/interceptors.py
@@ -10,7 +10,7 @@ import grpc
 from ansys.fluent.core.services.batch_ops import BatchOps
 
 network_logger = logging.getLogger("pyfluent.networking")
-log_bytes_limit = int(int(os.getenv("PYFLUENT_GRPC_LOG_BYTES_LIMIT", 1000)))
+log_bytes_limit = int(os.getenv("PYFLUENT_GRPC_LOG_BYTES_LIMIT", 1000))
 
 
 class TracingInterceptor(grpc.UnaryUnaryClientInterceptor):
--- a/src/ansys/fluent/core/services/interceptors.py
+++ b/src/ansys/fluent/core/services/interceptors.py
@@ -1,6 +1,7 @@
 """Interceptor classes to use with gRPC services."""
 
 import logging
+import sys
 from typing import Any
 
 from google.protobuf.json_format import MessageToDict
@@ -9,6 +10,7 @@ import grpc
 from ansys.fluent.core.services.batch_ops import BatchOps
 
 network_logger = logging.getLogger("pyfluent.networking")
+log_bytes_limit = 1000
 
 
 class TracingInterceptor(grpc.UnaryUnaryClientInterceptor):
@@ -29,9 +31,15 @@ class TracingInterceptor(grpc.UnaryUnaryClientInterceptor):
         )
         response = continuation(client_call_details, request)
         if not response.exception():
-            network_logger.debug(
-                f"GRPC_TRACE: response = {MessageToDict(response.result())}"
-            )
+            response_str = str(MessageToDict(response.result()))
+            response_bytes = sys.getsizeof(response_str)
+            if response_bytes < log_bytes_limit:
+                network_logger.debug(f"GRPC_TRACE: response = {response_str}")
+            else:
+                network_logger.debug(
+                    f"GRPC_TRACE: response hidden, {response_bytes} bytes > "
+                    f"{log_bytes_limit} bytes limit."
+                )
         return response
 
     def intercept_unary_unary(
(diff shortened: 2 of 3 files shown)


src/ansys/fluent/core/services/datamodel_tui.py
```diff
@@ -192,7 +192,8 @@ def execute(self, *args, **kwargs) -> Any:
         if self._path.startswith("/query/"):
             return self._execute_query(request)
         else:
-            logger.debug(f"TUI Command: {request}")
+            request_str = " ".join(str(request).split())
```
seanpearsonuk, line 195: What's going on here? Is it just to minimise whitespace?

raph-luc: `.split()` removes whitespaces and line breaks

src/ansys/fluent/core/services/interceptors.py
```diff
@@ -29,9 +31,15 @@ def _intercept_call(
...
         if not response.exception():
-            network_logger.debug(
-                f"GRPC_TRACE: response = {MessageToDict(response.result())}"
-            )
+            response_str = str(MessageToDict(response.result()))
+            response_bytes = sys.getsizeof(response_str)
+            if response_bytes < log_bytes_limit:
+                network_logger.debug(f"GRPC_TRACE: response = {response_str}")
+            else:
+                network_logger.debug(
```
seanpearsonuk, line 39: should we hide it or truncate it?

raph-luc: Good question.

We could truncate to show the beginning of the response, but I think that is rarely going to be useful unless we are specifically trying to debug gRPC; and in that case we want to see the full response string not a truncated portion anyway, so setting `pyfluent.services.interceptors.log_bytes_limit` to a higher value, as needed, would make more sense to me

We could add an env var to make this easier to control in different environments

We could also make it so `log_bytes_limit` can be manually set to `None` (perhaps through env var definitions) so it does not hide any messages when desired

src/ansys/fluent/core/services/datamodel_tui.py
raph-luc: See "before" and "after" for "TUI calls" in the original post above to see what that does

src/ansys/fluent/core/services/interceptors.py
```diff
@@ -29,9 +31,15 @@ def _intercept_call(
         )
         response = continuation(client_call_details, request)
         if not response.exception():
-            network_logger.debug(
-                f"GRPC_TRACE: response = {MessageToDict(response.result())}"
-            )
+            response_str = str(MessageToDict(response.result()))
```
mkundu1, line 34: I think using [ByteSize()](https://googleapis.dev/python/protobuf/latest/google/protobuf/message.html#google.protobuf.message.Message.ByteSize) method of protobuf message is better than converting it to string.

raph-luc: Thanks for letting me know about that method, I wasn't aware of it. 

The main reason why I did not look for an alternate method to get the size, is because we have to serialize the message result into a string regardless, so just getting the size of the Python string directly made sense to me. 

I will try using ByteSize anyway to see what it behaves like.

src/ansys/fluent/core/services/interceptors.py
raph-luc: Added optional `PYFLUENT_GRPC_LOG_BYTES_LIMIT` env var to make this easier to control, and made it so setting it to 0 disables message hiding and shows the full response, if required

src/ansys/fluent/core/services/interceptors.py
raph-luc: Using ByteSize is indeed better, avoids working with huge strings unless we actually want to log them, using it now, also optimized the logic a little bit

mkundu1 (approved):

hpohekar (approved):

PR #1833, opened 2023-07-21
Interrupt PyFluent when Fluent server sends error event

raph-luc:


mkundu1:
What is the behaviour when multiple solvers are connected to PyFluent? I'm thinking if `solver1` returns an error-event while the main thread is processing something on `solver2`, should we interrupt the main thread? Perhaps we can make sure any future remote call on `solver1` throws exception.

raph-luc:
> @mkundu1: What is the behaviour when multiple solvers are connected to PyFluent? I'm thinking if `solver1` returns an error-event while the main thread is processing something on `solver2`, should we interrupt the main thread? Perhaps we can make sure any future remote call on `solver1` throws exception.

Those are good questions. That is correct, it will interrupt the main thread regardless of which solver raised the error. My initial thoughts are that if there is an error event (currently only fatal events will be caught like this) on `solver1` while the user may be doing something else on `solver2` while using the same Python interpreter, we should still interrupt the main thread and let the user know immediately. 

If the user is doing two completely unrelated things with different solvers on the same Python interpreter, I think it will be expected that significant errors in one of the solvers will stop whatever the user is doing momentarily to let them know. This should not be a common occurrence anyway.

The major downside of waiting for future remote calls on `solver1` before raising the error is that, assuming the user needs both solvers working correctly simultaneously, and that is why they are executing both on the same Python interpreter, they may never get the interruption and realize that something went wrong on `solver1`, or only realize it too late due to unrelated errors on `solver2` which will only make figuring out what happened to `solver1` harder.

seanpearsonuk:
@raph-luc @mkundu1 it will be useful to discuss this and determine the required behaviour, considering that pyfluent can run multiple remote sessions asynchronously and that it can be a single component of a broader application.

PR #1845, opened 2023-07-26
Add initialize step in Field Data docs

hpohekar:
Added `solver.solution.initialization.hybrid_initialize()` step.

otherwise we get,

```
>>> field_data = solver.field_data
>>> field_data.is_data_valid()                                                                            
False
```

### DIFF
--- a/doc/source/api/solver/fielddata.rst
+++ b/doc/source/api/solver/fielddata.rst
@@ -13,8 +13,6 @@ In order to access field data, launch the fluent solver, and make field data
 available (for example, either by reading a case file and then initializing as in the following code, or
 by reading case and data files).
 
-1. Reading case file and initializing
-
 .. code-block:: python
 
   >>> import ansys.fluent.core as pyfluent
--- a/doc/source/api/solver/fielddata.rst
+++ b/doc/source/api/solver/fielddata.rst
@@ -10,8 +10,8 @@ Accessing field data objects
 ----------------------------
 
 In order to access field data, launch the fluent solver, and make field data
-available (for example, either by reading a case file and then initializing as in the following code,
-or by reading case and data files)".
+available (for example, either by reading a case file and then initializing as in the following code, or
+by reading case and data files).
 
 1. Reading case file and initializing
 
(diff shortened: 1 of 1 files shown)


mkundu1 (approved):

doc/source/api/solver/fielddata.rst
```diff
@@ -19,6 +19,7 @@ available (for example, by reading case and data files):
   >>> import_filename = examples.download_file("mixing_elbow.msh.h5", "pyfluent/mixing_elbow")
   >>> solver = pyfluent.launch_fluent(mode="solver")
   >>> solver.file.read(file_type="case", file_name=import_filename)
+  >>> solver.solution.initialization.hybrid_initialize()
```
seanpearsonuk, line 22: We should now also adjust the above comment for consistency with the example code: "(for example, either by reading a case file and then initializing as in the following code, or by reading case and data files)".

hpohekar: We are not reading data files separately here.

We are using `solver.solution.initialization.hybrid_initialize()` in following example.

https://fluent.docs.pyansys.com/version/dev/examples/00-fluent/mixing_elbow_settings_api.html#sphx-glr-examples-00-fluent-mixing-elbow-settings-api-py

hpohekar: @seanpearsonuk  Earlier I thought we should be consistent with this example - https://fluent.docs.pyansys.com/version/dev/examples/00-fluent/mixing_elbow_settings_api.html#sphx-glr-examples-00-fluent-mixing-elbow-settings-api-py

Therefore I added above line only.  

but we are mentioning using both case and data files here.

image

Therefore we will add both initialize and data file reading steps.

seanpearsonuk: The code and comment should correlate better.  The comment is saying for example do this, but them the code is doing something different. Experts don't need the comment but non-experts just need clear comments that correlate with the code.

hpohekar: Done. Thank you.

doc/source/api/solver/fielddata.rst
```diff
@@ -21,6 +24,17 @@ available (for example, by reading case and data files):
...
+
+.. code-block:: python
+
+  >>> import ansys.fluent.core as pyfluent
+  >>> from ansys.fluent.core import examples
+  >>> import_filename = examples.download_file("mixing_elbow.msh.h5", "pyfluent/mixing_elbow")
+  >>> data_filename = examples.download_file("mixing_elbow.dat.h5", "pyfluent/mixing_elbow")
+  >>> solver = pyfluent.launch_fluent(mode="solver")
+  >>> solver.file.read(file_type="case", file_name=import_filename)
+  >>> solver.file.read(file_type="data", file_name=data_filename)
```
seanpearsonuk, line 37: @hpohekar Now that this code has changed, it has become inconsistent with the changed comments.

hpohekar: @seanpearsonuk This is not clear to me.

seanpearsonuk: @hpohekar the above comment now reads "...(for example, either by reading a case file and then initializing as in the following code, ...)" which contradicts the "following code", which reading data instead.

hpohekar: Do we just want "(for example, either by reading a case file and then initializing as in the following code, or by reading case and data files)"  with previous version ? No need to add reading data file code ?

hpohekar: Done.

seanpearsonuk (approved):

raph-luc (approved):

PR #1849, opened 2023-07-27
PyFluent Example File Path

hpohekar:
Added `path` function to check existence of valid file name.

```
>>> from ansys.fluent.core import examples
>>> filepath = examples.path("elbow_param.cas.h5")
  File "D:\Repos\pyfluent\src\ansys\fluent\core\examples\downloads.py", line 141, in path
    raise FileNotFoundError(f"{filename} does not exist.")                               
FileNotFoundError: elbow_param.cas.h5 does not exist.                                    
>>> case_filename = examples.download_file("elbow_param.cas.h5", "pyfluent/mixing_elbow")
Checking if specified file already exists...      
File does not exist. Downloading specified file...
Download successful. File path:                                                    
C:\Users\hpohekar\AppData\Local\Ansys\ansys_fluent_core\examples\elbow_param.cas.h5
>>> filepath = examples.path("elbow_param.cas.h5")
>>> filepath
'C:\\Users\\<user_name>\\AppData\\Local\\Ansys\\ansys_fluent_core\\examples\\elbow_param.cas.h5'
>>>
```

### DIFF
--- a/src/ansys/fluent/core/examples/__init__.py
+++ b/src/ansys/fluent/core/examples/__init__.py
@@ -1,1 +1,2 @@
 from .downloads import download_file  # noqa: F401
+from .downloads import path  # noqa: F401
--- a/src/ansys/fluent/core/examples/downloads.py
+++ b/src/ansys/fluent/core/examples/downloads.py
@@ -1,5 +1,6 @@
 """Functions to download sample datasets from the Ansys example data repository."""
 import os
+from pathlib import Path
 import re
 import shutil
 from typing import Optional
@@ -131,3 +132,11 @@ def download_file(
 
     url = _get_file_url(filename, directory)
     return _retrieve_file(url, filename, save_path, return_only_filename)
+
+
+def path(filename: str):
+    file_path = str(Path(pyfluent.EXAMPLES_PATH) / filename)
+    if os.path.isfile(file_path):
+        return file_path
+    else:
+        raise FileNotFoundError(f"{filename} does not exist.")
--- a/tests/test_rp_vars.py
+++ b/tests/test_rp_vars.py
@@ -1,10 +1,7 @@
-from pathlib import Path
-
 import pytest
 from util.solver_workflow import new_solver_session_no_transcript  # noqa: F401
 
-import ansys.fluent.core as pyfluent
-from ansys.fluent.core.examples import download_file
+from ansys.fluent.core.examples import download_file, path
 from ansys.fluent.core.filereader.casereader import CaseReader
 
 
@@ -47,7 +44,7 @@ def test_get_all_rp_vars(new_solver_session_no_transcript) -> None:
     assert len(all_vars) == pytest.approx(9000, 20)
 
     # CaseFile comparison, note that the PyFluent work dir is not necessarily the same as the Fluent work dir
-    case = CaseReader(case_filepath=str(Path(pyfluent.EXAMPLES_PATH) / case_path))
+    case = CaseReader(case_filepath=path(case_path))
     case_vars = case.rp_vars()
     assert len(case_vars) == pytest.approx(9000, 450)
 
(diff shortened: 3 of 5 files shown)


seanpearsonuk:
looks cleaner

src/ansys/fluent/core/examples/downloads.py
```diff
@@ -131,3 +132,11 @@ def download_file(
...
     url = _get_file_url(filename, directory)
     return _retrieve_file(url, filename, save_path, return_only_filename)
+
+
+def path(filename: str):
+    file_path = str(Path(pyfluent.EXAMPLES_PATH) / filename)
+    if os.path.isfile(file_path):
+        return file_path
+    else:
+        raise FileNotFoundError(f"{filename} does not exist.")
```
raph-luc, line 142:
```suggestion
        raise FileNotFoundError(f"{file_path} does not exist.")
```
Small suggestion to make this exception statement more accurate (and informative)

src/ansys/fluent/core/examples/downloads.py
```diff
@@ -131,3 +132,11 @@ def download_file(
 
     url = _get_file_url(filename, directory)
     return _retrieve_file(url, filename, save_path, return_only_filename)
+
+
+def path(filename: str):
+    file_path = str(Path(pyfluent.EXAMPLES_PATH) / filename)
+    if os.path.isfile(file_path):
```
mkundu1, line 139: If you want to use the `Path` object from the previous line it also has an equivalent method.

mkundu1 (approved):

src/ansys/fluent/core/examples/downloads.py
```diff
@@ -131,3 +132,11 @@ def download_file(
...
+        return file_path
```
raph-luc, line 140: I believe this could cause some inconsistencies when running tests on containers vs standalone, as `download_file` returns only a filename on container tests but a fully qualified path when standalone.

A suggestion would be:
```suggestion
def path(filename: str):
    if os.path.isabs(filename):
        return filename
    file_path = str(Path(pyfluent.EXAMPLES_PATH) / filename)
    if os.path.isfile(file_path):
        return file_path
```

raph-luc (approved):

src/ansys/fluent/core/examples/__init__.py
```diff
@@ -1 +1,2 @@
 from .downloads import download_file  # noqa: F401
+from .downloads import path  # noqa: F401
```
raph-luc, line 2:
```suggestion
from .downloads import download_file, path  # noqa: F401
```

tests/test_rp_vars.py
```diff
@@ -47,7 +44,7 @@ def test_get_all_rp_vars(new_solver_session_no_transcript) -> None:
     assert len(all_vars) == pytest.approx(9000, 20)
 
     # CaseFile comparison, note that the PyFluent work dir is not necessarily the same as the Fluent work dir
-    case = CaseReader(case_filepath=str(Path(pyfluent.EXAMPLES_PATH) / case_path))
+    case = CaseReader(case_filepath=path(case_path))
```
raph-luc, line 47: This `path(case_path)` is an example where it would break when running this test with standalone Fluent without the changes I suggested, as `case_path` is already going to be an absolute path to begin with

src/ansys/fluent/core/examples/downloads.py
hpohekar: image

src/ansys/fluent/core/examples/downloads.py
hpohekar: The message with `filename` is concise than `file_path`.

src/ansys/fluent/core/examples/downloads.py
hpohekar: Thank you.

tests/test_rp_vars.py
hpohekar: We get same output with both file name and file path.

image

src/ansys/fluent/core/examples/__init__.py
hpohekar: Thank you.

PR #1851, opened 2023-07-27
Move the Fluent API tests to run with 24.1 Fluent version

mkundu1:
This will fix 23.1 nightly failure - https://github.com/ansys/pyfluent/actions/runs/5676374753/job/15383014718#step:24:1229

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,8 +15,6 @@ addopts = """
 --cov-report=html
 --durations=0
 --show-capture=all
---ignore-glob=tests/test_pro/*
---ignore-glob=tests/test_datamodel_server/*
 """
 markers = [
     "integration: Short-running tests which read a simple input file and check the exposure and values of various setting objects",
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -5,6 +5,10 @@ from packaging.specifiers import SpecifierSet
 from packaging.version import Version
 import pytest
 
+from ansys.fluent.core.launcher.launcher import FluentVersion
+
+_fluent_dev_version = next(iter(FluentVersion))
+
 
 def pytest_addoption(parser):
     parser.addoption(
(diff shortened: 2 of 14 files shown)


seanpearsonuk (approved):

raph-luc (approved):

mkundu1:
> **FOR SEPARATE CONSIDERATION** For a directory like `test_solvermode`, where we are purely exercising a Fluent API, the testing of released Fluent versions can be considered complete, and anyway any issues cannot be resolved any more. Therefore, I would propose that this directory is only applicable to the latest version - the current unreleased version under development. If we can associate a whole directory with a release version without additional need for marking individual tests, that would be ideal. Other tests, that aren't purely concerned with API testing should try to minimise and simplify their API use and try to use stable parts of APIs.

Agree, these tests should test the Fluent API under development as that is the only thing which may fail these tests. We already have documentation examples and pyfluent-examples repo to demonstrate Fluent API usage of the released Fluent version. I'll mark these tests with 24.1 which will also reduce the test running time.

seanpearsonuk:
> > **FOR SEPARATE CONSIDERATION** For a directory like `test_solvermode`, where we are purely exercising a Fluent API, the testing of released Fluent versions can be considered complete, and anyway any issues cannot be resolved any more. Therefore, I would propose that this directory is only applicable to the latest version - the current unreleased version under development. If we can associate a whole directory with a release version without additional need for marking individual tests, that would be ideal. Other tests, that aren't purely concerned with API testing should try to minimise and simplify their API use and try to use stable parts of APIs.
> 
> Agree, these tests should test the Fluent API under development as that is the only thing which may fail these tests. We already have documentation examples and pyfluent-examples repo to demonstrate Fluent API usage of the released Fluent version. I'll mark these tests with 24.1 which will also reduce the test running time.

Thanks a lot @mkundu1. 

@cj-hodgson @millerj97 letting you know what we're up to here

PR #1864, opened 2023-08-01
Mark some tests to nightly

mkundu1:
I have identified some tests to move to nightly based on the fact that the client-side code covered by these tests is very minimal and stable (exercising settings API, flobject.py or stable APIs like the batch operation, reduction services). These are basically testing the server-side APIs.

Additional thoughts:
1. The tests for fielddata and svars services are also calling settings API. But as these services have been developed recently and contain some wrapper code on the client side, I haven't moved them to the nightly, but let me know whether we should move them too.
2. Another idea is to run the settings API tests with Fluent <=23.2 during PRs but run them with Fluent 24.1 nightly. But I think settings API tests with Fluent 23.2 won't be impacted by PyFluent changes except for changes in flobject.py. But we should consider flobject.py belonging to the Fluent package, so any changes in flobject.py should first be tested with Fluent's PyConsole which should cover the PyFluent behaviour in most scenarios. Edit: This is implemented now using the "latest" marker.
3. I think we need to run some tests with Fluent 24.1 which are testing 24.1 server-side changes for any PyFluent feature.

We are also considering running some of these "API tests" in PyFluent during the certification process of the Fluent docker image, so a new revision of the docker image won't break these tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,8 @@ name: GitHub Actions
 
 on:
   pull_request:
+      paths-ignore:
+      - '.github/workflows/run-custom-tests.yml'
   workflow_dispatch:
   push:
     tags:
--- a/tests/test_flobject.py
+++ b/tests/test_flobject.py
@@ -653,6 +653,7 @@ class root(Group):
     )  # noqa: W293
 
 
+@pytest.mark.nightly
 def test_accessor_methods_on_settings_object(load_static_mixer_case):
     solver = load_static_mixer_case
 
@@ -700,6 +701,7 @@ def test_accessor_methods_on_settings_object(load_static_mixer_case):
     )
 
 
+@pytest.mark.nightly
 @pytest.mark.fluent_version(">=23.1")
 def test_accessor_methods_on_settings_object_types(load_static_mixer_case):
     solver = load_static_mixer_case
@@ -751,6 +753,7 @@ def test_find_children_from_settings_root_231(load_static_mixer_case):
     }
 
 
+@pytest.mark.nightly
 @pytest.mark.fluent_version(">=23.2")
 @pytest.mark.codegen_required
 def test_find_children_from_settings_root_232(load_static_mixer_case):
@@ -776,6 +779,7 @@ def test_find_children_from_settings_root_232(load_static_mixer_case):
     }
 
 
+@pytest.mark.nightly
 @pytest.mark.fluent_version(">=23.1")
 def test_find_children_from_fluent_solver_session(load_static_mixer_case):
     setup_children = find_children(load_static_mixer_case.setup)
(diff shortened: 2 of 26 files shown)


.github/workflows/ci.yml
```diff
@@ -3,6 +3,8 @@ name: GitHub Actions
 
 on:
   pull_request:
+      paths-ignore:
```
mkundu1, line 6: The intention is not to run the full CI when the custom tests CI is edited in an ongoing PR, but it is not currently working due to https://github.com/actions/runner/issues/2324.

hpohekar (approved):

prmukherj (approved):

tests/test_flobject.py
```diff
@@ -653,6 +653,7 @@ class a_2(Boolean):
     )  # noqa: W293
 
 
+@pytest.mark.nightly
 def test_accessor_methods_on_settings_object(load_static_mixer_case):
```
seanpearsonuk, line 657: @mkundu 
Tests like this are heavily exercising the API, that's true, but here we most interested interested in the behaviour of the client-side flobject.py. In fact, for CI, we want to check that we haven't broken flobject.py, and we can do that with a stable API version. So, shall we mark test such as this one as using the latest released fluent version, currently 23R2? In terms of marking tests, it would be useful to have labels to indicate _latest released version_ and _current development version_. Then those labels can be maintained and one place and incremented as Fluent is released. I don't believe we have something like that(?)

seanpearsonuk (approved):

tests/test_flobject.py
mkundu1: `pytest.mark.fluent_version("latest")` means it will run

1. with latest dev version of Fluent in PRs, and
2. with latest release and dev versions of Fluent in nightly

tests/test_fluent_version_marker.py
```diff
@@ -5,6 +5,11 @@ def test_dev_fluent_any():
     pass
 
 
+@pytest.mark.fluent_version("latest")
+def test_dev_fluent_latest():
+    pass
```
raph-luc, line 10: Are these tests just to check if the "latest" and other markers exist? Think I missed these before

raph-luc (commented):
Typos:

https://github.com/ansys/pyfluent/blob/eb8e0b2bc28fe02218715b05319f7aaee5619a6f/tests/test_settings_api.py#L41


https://github.com/ansys/pyfluent/blob/eb8e0b2bc28fe02218715b05319f7aaee5619a6f/tests/test_settings_api.py#L91


https://github.com/ansys/pyfluent/blob/eb8e0b2bc28fe02218715b05319f7aaee5619a6f/tests/test_settings_api.py#L119

tests/test_rp_vars.py
```diff
@@ -29,15 +29,15 @@ def test_get_and_set_rp_vars(new_solver_session_no_transcript) -> None:
...
     rp_vars = solver.rp_vars
     # all vars
     all_vars = rp_vars()
     assert len(all_vars) == pytest.approx(9000, 10)
 
     # refresh
-    solver.file.write(file_type="case", file_name=case_path)
-    solver.file.read(file_type="case", file_name=case_path)
+    solver.tui.file.write_case(case_path)
+    solver.tui.file.read_case(case_path)
```
raph-luc, line 40: I'm not sure if this is related to the test that is currently failing, but is it a good idea to use the `tui` here?

raph-luc (approved):

PR #1866, opened 2023-08-02
Instances of BaseException to be replaced with Exception.

prmukherj:
Resolved several instances of except BaseException

### DIFF
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -393,8 +393,8 @@ def _get_running_session_mode(
                 if fluent_connection.scheme_eval.scheme_eval("(cx-solver-mode?)")
                 else "meshing"
             )
-        except BaseException:
-            raise RuntimeError("Fluent session password mismatch")
+        except Exception as ex:
+            raise RuntimeError("Fluent session password mismatch") from ex
     return session_mode.value[1]
 
 
--- a/tests/test_casereader.py
+++ b/tests/test_casereader.py
@@ -202,7 +202,7 @@ def test_case_reader_get_rp_and_config_vars():
     assert reader.rp_var.pressure.output_dpdt__q() is True
     assert len(reader.rp_var.context.map_r17__plus()) == 53
     assert reader.rp_var.defaults.pre_r19__dot0_early__q() is False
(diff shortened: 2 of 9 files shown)


hpohekar (approved):

seanpearsonuk (approved):

tests/test_casereader.py
```diff
@@ -202,7 +202,7 @@ def test_case_reader_get_rp_and_config_vars():
     assert reader.rp_var.pressure.output_dpdt__q() is True
     assert len(reader.rp_var.context.map_r17__plus()) == 53
     assert reader.rp_var.defaults.pre_r19__dot0_early__q() is False
-    with pytest.raises(BaseException):
+    with pytest.raises(Exception):
```
mkundu1, line 205: generally better to test the specific exception which is thrown in this case

mkundu1 (approved):

tests/test_casereader.py
raph-luc: Not only in tests, but ideally all of the exception catches should be as specific as possible

Regardless, not criticizing as this is still a big improvement, thanks @prmukherj

raph-luc (approved):

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -393,8 +393,8 @@ def _get_running_session_mode(
                 if fluent_connection.scheme_eval.scheme_eval("(cx-solver-mode?)")
                 else "meshing"
             )
-        except BaseException:
-            raise RuntimeError("Fluent session password mismatch")
+        except Exception as ex:
+            raise RuntimeError("Fluent session password mismatch") from ex
```
raph-luc, line 397: Good catch

tests/test_casereader.py
prmukherj: Done, Thank you @mkundu1.

Thank you @raph-luc, I totally agree with the point you raised. However, covering all the specific exceptions for the cases where these changes have been made seems a bit tricky. That's why for now I think we can re-raise the Exception using the from keyword which provides user the information about the original error as well. Seems fine for now, right?

raph-luc: That is right, I fully agree

PR #1871, opened 2023-08-03
Handling Fluent server-side errors

raph-luc:


mkundu1:
Shall we design this to throw exception on any future remote call (because the remote access is errored, not the attribute access within client-side session)? We can pass the error-state object to individual services and a grpc-interceptor will check the error-state before making the remote call and throw instead if necessary.

prmukherj (approved):

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -89,6 +89,31 @@ def get_container(container_id_or_name: str) -> Union[bool, Container, None]:
...
+    @property
+    def details(self):
+        return self._details
+
+    def __init__(self, name: str = "", details: str = ""):
+        self._name = name
+        self._details = details
+
+    def __eq__(self, other):
+        return self._name == other
```
seanpearsonuk, line 106: This version of `__eq__` should not be in the public interface. This version requires that RHS is required to be a string and not another object of the same type. `details` which is a public property is not involved in the comparison.

src/ansys/fluent/core/services/interceptors.py
```diff
@@ -61,6 +61,38 @@ def intercept_unary_unary(
...
+
+    def _intercept_call(
+        self,
+        continuation: Any,
+        client_call_details: grpc.ClientCallDetails,
+        request: Any,
+    ):
+        if self._fluent_error_state == "fatal":
+            details = self._fluent_error_state.details
+            self._fluent_error_state.clear()
```
seanpearsonuk, line 80: I think we should not clear this.

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -407,24 +437,22 @@ def register_finalizer_cb(self, cb):
         """Register a callback to run with the finalizer."""
         self.finalizer_cbs.append(cb)
 
-    def create_service(self, service, add_arg=None):
+    def create_service(self, service, args=()):
```
mkundu1, line 440: For additional optional arguments, the signature would be `def create_service(self, service, *args):`

src/ansys/fluent/core/services/interceptors.py
mkundu1, line 80: Is it better to clear the state in events_streaming.py in an else block of `if event_name == "errorevent"`?

mkundu1 (approved):

PR #1882, opened 2023-08-09
replace code

seanpearsonuk:
This temporarily replaces the units code with a slightly modified copy of PyUnits (modified at the highest level)


hpohekar (approved):

prmukherj (approved):

src/ansys/fluent/core/quantity/systems.py
```diff
@@ -114,7 +115,7 @@ def BASE_UNITS_LENGTH(cls, len):
     @classmethod
     def UNIT_UNDEFINED(cls, unit):
         return cls(
-            f"`{unit}` is an undefined unit. To use `{unit}` add it to the `fundamental_units` table within quantity_config.yaml."
+            f"`{unit}` is an undefined unit. To use `{unit}` add it to the `fundamental_units` table within cfg.yaml."
```
hpohekar, line 118: @seanpearsonuk 

Why are we suggesting to add an undefined unit to fundamental  units in  .yml file ? We have already defined all fundamental units.

hpohekar: It should be added to derived units with correct composition and factor.

seanpearsonuk: This is just being copied back from PyUnits. Yes, I think the guidance is innaccurate here. But it might depend what is meant by "fundamental".

mkundu1 (approved):

PR #1883, opened 2023-08-09
Deploy documentation from github runner

mkundu1:
As discussed with Roberto and Jorge, this will avoid generating the actions_github_pages_* folders on self-hosted runners which is causing disk space to run out.

I'll make this change in other repos after testing.

### DIFF
--- a/.github/workflows/nightly-dev-doc-build.yml
+++ b/.github/workflows/nightly-dev-doc-build.yml
@@ -16,7 +16,7 @@ env:
   PYFLUENT_HIDE_LOG_SECRETS: 1
 
 jobs:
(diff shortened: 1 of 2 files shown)


hpohekar (approved):

prmukherj (approved):

raph-luc (approved):
I had manually cleaned the oldest of those folders in the past after talking about this with Roberto as well, and I just realized I think I forgot to discuss this with you and the others. 

Regardless, good change. If I understood this correctly, this will move those actions_github_pages_* folders from the self-hosted runners to the GitHub runners, which automatically clean up after themselves

PR #1886, opened 2023-08-09
Remove old files

mkundu1:
Looking at gh-pages structure of other repos

https://github.com/ansys/pygeometry/tree/gh-pages
https://github.com/ansys/pymapdl/tree/gh-pages

I'm removing these files to match the folder structure. Let's see if this helps.

Old structure: https://github.com/ansys/pyfluent/tree/gh-pages
New structure: https://github.com/ansys/pyfluent/tree/maint/gh-pages-files


raph-luc (approved):
LGTM, and a question you may already have the answer to: why does our current docs only show 0.14 and newer PyFluent versions, while we have older versions back to 0.11 in the doc version files? Is this a PyAnsys-wide decision? I noticed the same thing in PyMAPDL's repo. 

If we were to keep older doc versions up, I think it would make sense to also expose the actual older versions (starting from 0.11), and not only the 3 most recent ones.

I noticed you are also deleting the older doc versions from the files, and that makes sense to me too. They are not going to be accessible anyway as things currently stand.

raph-luc:
I guess keeping the older files in the versions folder would be useful for people who have older bookmarks, as those links would still work (even if they can't get back to those pages from the newest versions). 

This may be a reason to not delete the older files in the versions folder

mkundu1:
> I guess keeping the older files in the versions folder would be useful for people who have older bookmarks, as those links would still work (even if they can't get back to those pages from the newest versions, they can still get to the newest version from the older versions), and they will get a banner on the top of the page letting them know they are looking at deprecated documentation.
> 
> This may be a reason to not delete the older files in the versions folder

That is a good point. Let's keep the version folder as before for now.

PR #1887, opened 2023-08-09
doc updates

seanpearsonuk:


src/ansys/fluent/core/filereader/case_file.py
```diff
@@ -175,9 +198,21 @@ class CaseFile:
...
+        """Initialize a CaseFile object. Exactly one file path argument must be specified.
+
+        Parameters
+        ----------
+        case_filepath : Optional[str]
+            The path of a case file.
+        project_filepath : Optional[str]
+            The path of a project file from which the case file is selected.
+        """
+        if (not case_filepath) == (not project_filepath):
```
seanpearsonuk, line 215: This is the only code change in the PR. Added no additional tests and not tested interactively.

mkundu1 (approved):

src/ansys/fluent/core/filereader/case_file.py
raph-luc: Curious why this change was made. Isn't `(case_filepath and project_filepath)` significantly clearer than  `(not case_filepath) == (not project_filepath)` when checking whether both were specified? Is this to cover for a potential situation where neither one is specified?

raph-luc (approved):

hpohekar (approved):

prmukherj (approved):

src/ansys/fluent/core/filereader/case_file.py
seanpearsonuk: _Is this to cover for a potential situation where neither one is specified?_

@raph-luc Yes. Also, I've written the docstring to state _exactly one_.

PR #1906, opened 2023-08-18
Global search method

mkundu1:
Added a global search method to search statically through all API paths. This will be useful to discover object paths while writing scripts without running a Fluent session. E.g., all queries in #1890 can be addressed using this method (outputs are truncated):
```
>>> pyfluent.search("periodic")
...
<solver_session>.setup.boundary_conditions.periodic["<name>"] (Object)
<solver_session>.setup.mesh_interfaces.interface["<name>"].periodic (Parameter)
...
>>> pyfluent.search("compute") 
...
<solver_session>.setup.reference_values.compute (Command)
...
>>> pyfluent.search("console")      
...
<solver_session>.tui.define.parameters.output_parameters.print_to_console (Command)
...
>>> pyfluent.search("report_file") 
<solver_session>.tui.solution.monitor.report_files (Object)
<solver_session>.tui.solve.report_files (Object)
<solver_session>.solution.monitor.report_files["<name>"] (Object)
```
The documentation search (https://fluent.docs.pyansys.com/version/stable/search.html?q=compute) also outputs these objects/commands, but that does not show the runtime hierarchy.

Please see the unittests for more usage.

For the implementation, a nested dictionary of all API paths is stored during codegen which is parsed during the search.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,4 +2,5 @@ include src/ansys/fluent/core/launcher/fluent_launcher_options.json
 include src/ansys/fluent/core/docs/README.rst
 include src/ansys/fluent/core/logging_config.yaml
 include src/ansys/fluent/core/quantity/cfg.yaml
-recursive-include src *pyi
+recursive-include src/ansys/fluent/core *pyi
+recursive-include src/ansys/fluent/core/data api_tree_*.pickle
--- a/codegen/allapigen.py
+++ b/codegen/allapigen.py
@@ -1,14 +1,11 @@
 import argparse
 import os
 from pathlib import Path
-
-# import pickle
-from pprint import pprint
+import pickle
 
 import datamodelgen
 import print_fluent_version
-
-# import settingsgen
+import settingsgen
 import tuigen
 
 from ansys.fluent.core.launcher.launcher import FluentVersion, get_ansys_version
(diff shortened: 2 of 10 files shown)


hpohekar (approved):

raph-luc (approved):
Played around with it a bit on my end, and it works great and will certainly be very useful.

Note it did not automatically work with the latest Fluent installed on my end, and tried to read a `ansys/fluent/core/data/api_tree_232.pickle` that did not exist. I had `ansys/fluent/core/data/api_tree_241.pickle`. A minor suggestion would be to check if the file exists; and if it doesn't, then output a logging warning and default to whatever the latest `api_tree_*.pickle` available is.

Should we add this to the documentation in `user_guide/index.html` as I would expect a lot of users are going to want to use this?

mkundu1:
> Played around with it a bit on my end, and it works great and will certainly be very useful.
> 
> Note it did not automatically work with the latest Fluent installed on my end, and tried to read a `ansys/fluent/core/data/api_tree_232.pickle` that did not exist. I had `ansys/fluent/core/data/api_tree_241.pickle`. A minor suggestion would be to check if the file exists; and if it doesn't, then output a logging warning and default to whatever the latest `api_tree_*.pickle` available is.
> 
> Should we add this to the documentation in `user_guide/index.html` as I would expect a lot of users are going to want to use this?

Thanks for testing this, I've pushed a change to dynamically find the available api_tree file. I'll add a section in the user-guide.

Also, in a separate discussion with Krishnan, we plan to move the implementation to server-side, so the data/code can be re-used in both Fluent console and PyFluent (won't change the usage).

mkundu1:
Merging this for now, will update the user-guide separately

PR #1945, opened 2023-09-01
Fix for menus which are python keyword names

mkundu1:


src/ansys/fluent/core/services/datamodel_tui.py
```diff
@@ -359,4 +342,4 @@ def convert_path_to_grpc_path(path: Path) -> str:
     str
         gRPC path.
     """
-    return "/" + "/".join(convert_func_name_to_tui_menu(x) for x in path)
```
mkundu1, line 362: This conversion is currently done on the server side.

prmukherj (approved):

PR #1951, opened 2023-09-05
Fix bug in journal read for settings API in launcher.

prmukherj:


mkundu1 (approved):
There are 2 inconsistencies between Fluent and PyFluent:
1. We should be able to pass multiple journal files through launch_fluent
2. -topy is Fluent converts the journal files passed through -i

I'll open an issue in PyFluent.

seanpearsonuk (approved):

raph-luc (approved):
Tested with v241, works fine

Issue #1955, opened 2023-09-05
KeyError: 'named-expressions' in casereader


tejalprabhu01:
### 🔍 Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### 🐞 Description of the bug

Following error is seen while reading a case file:

  File "D:\ANSYSInstallation\ANSYS Inc\v241\optiSLang\lib\python3.10\lib\site-packages\ansys\fluent\core\filereader\case_file.py", line 230, in input_parameters
    exprs = self._named_expressions()
  File "D:\ANSYSInstallation\ANSYS Inc\v241\optiSLang\lib\python3.10\lib\site-packages\ansys\fluent\core\filereader\case_file.py", line 280, in _named_expressions
    return self._find_rp_var("named-expressions")
  File "D:\ANSYSInstallation\ANSYS Inc\v241\optiSLang\lib\python3.10\lib\site-packages\ansys\fluent\core\filereader\case_file.py", line 286, in _find_rp_var
    return self._rp_vars[name]
KeyError: 'named-expressions'

### 📝 Steps to reproduce

[keyerror.zip](https://github.com/ansys/pyfluent/files/12525541/keyerror.zip)

from ansys.fluent.core.filereader.casereader import CaseReader
reader = CaseReader(case_filepath = casefilepath)

### 💻 Which operating system are you using?

Windows

### 📀 Which ANSYS version are you using?

24R1 (Dev)

### 🐍 Which Python version are you using?

3.7

### 📦 Installed packages

```shell
ansys-fluent-core
ansys-fluent-parametric
```
@seanpearsonuk @mkundu1

PR #1956, opened 2023-09-06
Settings Reader

prmukherj:


hpohekar (approved):

src/ansys/fluent/core/filereader/case_file.py
```diff
@@ -282,19 +282,19 @@ class CaseFile:
 
     def __init__(
         self,
-        case_filepath: Optional[str] = None,
+        settings_filepath: Optional[str] = None,
```
seanpearsonuk, line 285: docstring below is out of date

src/ansys/fluent/core/filereader/case_file.py
```diff
@@ -282,19 +282,19 @@ class CaseFile:
...
         ----------
-        case_filepath : Optional[str]
-            The path of a case file.
+        settings_filepath : Optional[str]
+            The path of a settings file.
         project_filepath : Optional[str]
             The path of a project file from which the case file is selected.
         """
-        if (not case_filepath) == (not project_filepath):
+        if (not settings_filepath) == (not project_filepath):
```
seanpearsonuk, line 297: Is `project_filepath` valid to use here?

src/ansys/fluent/core/filereader/case_file.py
```diff
@@ -304,7 +304,7 @@ def __init__(
                     dirname(project_filepath),
                     Path(project_filepath).name.split(".")[0] + ".cffdb",
                 )
-                case_filepath = Path(
+                settings_filepath = Path(
```
seanpearsonuk, line 307: case mentioned in runtime error message above

seanpearsonuk, line 307: regarding the call to _get_case_filepath_from_flprj, that's literally just getting the Case. So for Settings file, we should leave this out.

src/ansys/fluent/core/filereader/case_file.py
```diff
@@ -313,41 +313,48 @@ def __init__(
                 )
 
         try:
-            if Path(case_filepath).match("*.cas.h5"):
-                file = h5py.File(case_filepath)
+            if Path(settings_filepath).match("*.cas.h5"):
+                file = h5py.File(settings_filepath)
```
seanpearsonuk, line 317: This code is specific to case files. We have classes to separate out the different functionality so that we don't need conditional code.

src/ansys/fluent/core/filereader/casereader.py
```diff
@@ -1,7 +1,5 @@
 # Compatibility aliases
-from .case_file import (  # noqa: F401
-    InputParameter,
-    OutputParameter,
-    _get_processed_string,
-)
 from .case_file import CaseFile as CaseReader  # noqa: F401
+from .case_file import InputParameter, OutputParameter  # noqa: F401
+from .case_file import SettingsFile as SettingsReader  # noqa: F401
```
seanpearsonuk, line 4: This is not needed as the other only exists for compatibility

tests/test_settings_reader.py
```diff
@@ -0,0 +1,90 @@
+import pytest
+
+from ansys.fluent.core import examples
+from ansys.fluent.core.filereader.casereader import SettingsReader
```
seanpearsonuk, line 4: this is the old import location

src/ansys/fluent/core/filereader/casereader.py
prmukherj: Keeping it now till OptiSlang uses the new files.

prmukherj:
@tejalprabhu01, the bug reported in #1955 has been addressed in this PR. A test case added gives some insights on how it can be parsed. Please let me know if it suits your use case. This is only used for old case files though.
cc @mkundu1

tejalprabhu01 (approved):
I have tested the changes and the issue is fixed! Thank you.

src/ansys/fluent/core/filereader/casereader.py
```diff
@@ -1,7 +1,4 @@
 # Compatibility aliases
-from .case_file import (  # noqa: F401
-    InputParameter,
-    OutputParameter,
-    _get_processed_string,
-)
 from .case_file import CaseFile as CaseReader  # noqa: F401
+from .case_file import InputParameter, OutputParameter  # noqa: F401
```
seanpearsonuk, line 3: Why any new lines needed here?

prmukherj: Removed. Thank you.

seanpearsonuk (approved):

PR #1961, opened 2023-09-07
Meshing queries using fdl

hpohekar:
- [x] Add PyQuery class
- [x] Add meshing queries to static info and pure meshing session.
- [x] Add tests for meshing utilities


src/ansys/fluent/core/session_pure_meshing.py
```diff
@@ -7,10 +7,11 @@
 import ansys.fluent.core as pyfluent
 from ansys.fluent.core.data_model_cache import DataModelCache
 from ansys.fluent.core.fluent_connection import FluentConnection
-from ansys.fluent.core.services.meshing_queries import (
-    MeshingQueries,
-    MeshingQueriesService,
-)
+
+# from ansys.fluent.core.services.meshing_queries import (
```
hpohekar, line 11: It's our previous implementation using C++.

src/ansys/fluent/core/session_pure_meshing.py
```diff
@@ -43,10 +50,10 @@ def __init__(self, fluent_connection: FluentConnection):
...
         )
 
-        self.meshing_queries_service = fluent_connection.create_service(
-            MeshingQueriesService, self.error_state
-        )
-        self.meshing_queries = MeshingQueries(self.meshing_queries_service)
+        # self.meshing_queries_service = fluent_connection.create_service(
+        #     MeshingQueriesService, self.error_state
+        # )
+        # self.meshing_queries = MeshingQueries(self.meshing_queries_service)
```
hpohekar, line 56: It's our previous implementation using C++.

src/ansys/fluent/core/session_pure_meshing.py
```diff
@@ -7,10 +7,11 @@
...
 from ansys.fluent.core.fluent_connection import FluentConnection
-from ansys.fluent.core.services.meshing_queries import (
-    MeshingQueries,
-    MeshingQueriesService,
-)
+
+# from ansys.fluent.core.services.meshing_queries import (
+#     MeshingQueries,
+#     MeshingQueriesService,
+# )
```
hpohekar, line 14: It's our previous implementation using C++.

mkundu1 (approved):

hpohekar:
> I wonder if we should also keep the 23R2-specific meshing_queries object for 23R2 users? @hpohekar @mkundu1

@seanpearsonuk @mkundu1 

If we want to keep both old and new, `meshing_queries` name is already used  for C++ implementation. What should be replacement from meshing_queries using fdl ?

hpohekar:
> > meshing_queries
> 
> The new one should be `meshing_queries` and is instantiated only if the server >=24.1. The old one can also be `meshing_queries` and is instantiated only if the server == 23.2. 23.2 users already have access to `meshing_queries` in pyfluent up to now and we have no strong reason to remove that.

Yes, sure. Thank you.

seanpearsonuk:
> > > meshing_queries
> > 
> > 
> > The new one should be `meshing_queries` and is instantiated only if the server >=24.1. The old one can also be `meshing_queries` and is instantiated only if the server == 23.2. 23.2 users already have access to `meshing_queries` in pyfluent up to now and we have no strong reason to remove that.
> 
> Yes, sure. Thank you.

I think same name is OK because they are quite similar apart from implementation.

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -135,6 +138,16 @@ def execute_command(
         logger.debug(f"Command: {request.command}")
         return self._stub.executeCommand(request, metadata=self._metadata)
 
+    if ansys_version == "24.1.0":
```
seanpearsonuk, line 141: It should be safe to write `>=` here. I would prefer it if we had a function that we could call for all such queries. Something like:
if `ansys_versions.current >= ansys_versions.version_24_1`:
```suggestion
    if ansys_version >= "24.1.0":
```

seanpearsonuk (approved):

prmukherj (approved):

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1329,9 +1394,15 @@ def _get_child_names(self):
                         singleton_names.append(member)
                 creatable_type_names = struct_field.creatabletypes
                 command_names = [x.name for x in struct_field.commands]
+                if hasattr(struct_field, "queries"):
+                    query_names = [x.name for x in struct_field.queries]
+        if hasattr(struct_field, "queries"):
+            return singleton_names, creatable_type_names, command_names, query_names
         return singleton_names, creatable_type_names, command_names
```
mkundu1, line 1401: I think it is fine to return an empty list for query_names in this case, the check at lines 1404, 1415 won't be needed

codegen/datamodelgen.py
```diff
@@ -223,6 +239,9 @@ def _write_static_info(self, name: str, info: Any, f: FileIO, level: int = 0):
         singletons = sorted(info.singletons)
         parameters = sorted(info.parameters)
         commands = sorted(info.commands)
+
+        if int(self.version) >= 241:
+            queries = sorted(info.queries)
```
mkundu1, line 244: Here also the queries can be an empty list otherwise, the version checks won't be required at other places of this method.

src/ansys/fluent/core/services/datamodel_se.py

codegen/datamodelgen.py
hpohekar: Done. Thank you.

codegen/datamodelgen.py
```diff
@@ -79,6 +78,19 @@ def _build_command_docstring(name: str, info: Any, indent: str):
     return doc
 
 
+def _build_query_docstring(name: str, info: Any, indent: str):
```
seanpearsonuk, line 81: This seems to be the same function as the one above aside from the first line.

hpohekar: Yes, added check for command and query to handle this. Thanks.

PR #1979, opened 2023-09-14
Example of creating new materials using settings API

ansjsia:
Added an example to the docs showing how to create new materials using the settings API, as an alternative to using the TUI.

### DIFF
--- a/doc/source/user_guide/materials.rst
+++ b/doc/source/user_guide/materials.rst
@@ -70,17 +70,17 @@ Create new material
 
 .. code:: python
 
-    solver.setup.materials.solid["mysolid"] = {}
-    solver.setup.materials.solid["mysolid"].chemical_formula = "SiO2"
-    solver.setup.materials.solid["mysolid"].density.value = 2650
-    solver.setup.materials.solid["mysolid"].specific_heat.value = 1887
-    solver.setup.materials.solid["mysolid"].thermal_conductivity.value = 7.6
+    mysolid = solver.setup.materials.solid.create("mysolid")
+    mysolid.chemical_formula = "SiO2"
+    mysolid.density.value = 2650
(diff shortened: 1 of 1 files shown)


hpohekar (approved):

doc/source/user_guide/materials.rst
```diff
@@ -53,9 +53,34 @@ Use settings objects
...
     solver.setup.cell_zone_conditions.fluid["elbow-fluid"].material = "water-liquid"
+
+Create new material
+~~~~~~~~~~~~~~~~~~~
+
+**Python code**
+
+.. code:: python
+
+    solver.setup.materials.solid["mysolid"] = {}
```
seanpearsonuk, line 73: @ansjsia 
```
>>> mysolid = solver.setup.materials.solid.create("mysolid")
>>> mysolid.chemical_formula = "SiO2"
```
Note the above create syntax. Regardless of the means of creation, I prefer not to repeat the trainwreck over multiple lines. I am referring to duplication of solver.setup.materials.solid["mysolid"]. I know that this is what a lot of authors do, but this creates a negative impression, and we should change a lot of existing test and example code.

I tried also your dict style access. It only works for already existing materials. It does not create. It raises an exception if the material does not exist.

ansjsia: Thanks for the comments, I have revised the syntax in the following commit. Please let me know if further changes are required.

seanpearsonuk (approved):
Thanks @ansjsia that looks good now

mkundu1 (approved):

h-krishnan:
@ansjsia @mkundu1 @seanpearsonuk Note that setting each property separately is a bit slower than setting them all at once via a dictionary.

PR #1994, opened 2023-09-20
Fix nightly tests

mkundu1:
The tests within tests/test_solvermode directory will now be run weekly through the new `weekly-solvermode-test-run.yml`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,6 +15,7 @@ addopts = """
 --cov-report=html
 --durations=0
 --show-capture=all
+--ignore=tests/test_solvermode
 """
 markers = [
     "integration: Short-running tests which read a simple input file and check the exposure and values of various setting objects",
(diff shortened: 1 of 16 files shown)


pyproject.toml
```diff
@@ -15,6 +15,7 @@ addopts = """
 --cov-report=html
 --durations=0
 --show-capture=all
+--ignore=tests/test_solvermode
```
mkundu1, line 18: These tests are not testing any pyfluent-specific code, thus disabling.

These tests are very sensitive to settings API change and frequently fail nightly test run.

raph-luc (approved):
Would it be more appropriate in the long term to add a pytest marker specific for `tests/test_solvermode` and then enable it in the suggested `unittest-solvermode-241`? 

Regardless, this works fine I think, and this minor refactoring I suggest can easily be done at another time

prmukherj (approved):

seanpearsonuk (approved):

PR #1997, opened 2023-09-21
Update installation instructions

hpohekar:
Updated installation section.

image

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -72,13 +72,6 @@ Install the latest release from `PyPI
 
    pip install ansys-fluent-core
 
-Alternatively, install the latest from
-`PyFluent GitHub <https://github.com/ansys/pyfluent>`_ via:
-
-.. code:: console
-
-   pip install git+https://github.com/ansys/pyfluent.git
-
 For developers
 --------------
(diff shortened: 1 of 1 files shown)


prmukherj (approved):

raph-luc:
Wouldn't this additional/specific information be more appropriate in https://fluent.docs.pyansys.com/version/stable/getting_started/installation.html ?

I think the main GitHub readme should probably tend towards being leaner and only containing necessary or basic information

In fact, should we also remove the whole "for developers" section from the main GitHub readme, as it is rather specific? We already have a "for developers" section in the installation page I mentioned above, that is probably a more appropriate location for this information.

mkundu1:
`pip install git+https://github.com/ansys/pyfluent.git` will pull the files from git repo and compile the package from them. The installed package won't have the generated files from codegen, so we should not recommend it for non-developers.

I'd avoid recommending `pip install git+https://github.com/ansys/pyfluent.git` in general. For developers, starting from a `git clone` is quite standard.

mkundu1 (approved):

raph-luc (approved):

dnwillia-work (approved):

PR #2005, opened 2023-09-22
Maint/update topy arg

prmukherj:


mkundu1 (approved):

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -676,8 +676,18 @@ def launch_fluent(
         kwargs = _get_subprocess_kwargs_for_fluent(env)
         if cwd:
             kwargs.update(cwd=cwd)
+        if journal_filepaths:
+            if not isinstance(journal_filepaths, (str, list)):
+                raise TypeError("Journal name should be a list of strings.")
+            if isinstance(journal_filepaths, str):
+                warnings.warn("Journal name should be a list of strings.")
```
mkundu1, line 683: I think we should not show the warning as the `journal_filepaths` parameter is documented to accept a single file.

hpohekar (approved):

src/ansys/fluent/core/launcher/launcher.py
prmukherj: Thank you @mkundu1. Done.

PR #2014, opened 2023-09-22
Make sure Fluent is shutdown on exit

mkundu1:
`session.exit()` is handled in a daemon thread (MonitorThread) which ensures shutdown of all streaming (non-daemon) threads. A daemon thread is terminated abruptly during Python process exit, after all non-daemon threads are exited). In the current code, the streaming threads are shut down after the server is exited within `session.exit()`. The MonitorThread shuts down immediately after shutting down the streaming threads without exiting the last server.

In this PR, a long-running thread is added which is exited at the end of `session.exit()` to ensure everything within `session.exit()` gets executed during exit.

Need to see how to unittest this (need a Python process exit).

### DIFF
--- a/src/ansys/fluent/core/fluent_connection.py
+++ b/src/ansys/fluent/core/fluent_connection.py
@@ -348,6 +348,13 @@ class FluentConnection:
         self.launcher_args = launcher_args
 
         self._exit_evt = threading.Event()
+
+        # session.exit() is handled in the daemon thread (MonitorThread) which ensures
+        # shutdown of non-daemon threads. A daemon thread is terminated abruptly
+        # during interpreter exit (after all non-daemon threads are exited).
+        # self._waiting_thread is a long-running thread which is exited
+        # at the end of session.exit() to ensure everything within session.exit()
+        # gets executed during exit.
         self._waiting_thread = threading.Thread(target=self._exit_evt.wait)
         self._waiting_thread.start()
 
--- a/src/ansys/fluent/core/fluent_connection.py
+++ b/src/ansys/fluent/core/fluent_connection.py
@@ -347,6 +347,10 @@ class FluentConnection:
         self._remote_instance = remote_instance
         self.launcher_args = launcher_args
 
+        self._exit_evt = threading.Event()
+        self._waiting_thread = threading.Thread(target=self._exit_evt.wait)
+        self._waiting_thread.start()
+
         self._finalizer = weakref.finalize(
             self,
             FluentConnection._exit,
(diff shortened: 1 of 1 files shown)


seanpearsonuk (approved):

prmukherj (approved):

hpohekar (approved):

tests/test_fluent_session.py
```diff
@@ -220,3 +220,22 @@ def test_interrupt(load_static_mixer_case):
...
+
+def test_fluent_exit():
+    inside_container = os.getenv("PYFLUENT_LAUNCH_CONTAINER")
+    script = (
+        "import ansys.fluent.core as pyfluent;"
+        "solver = pyfluent.launch_fluent(start_watchdog=False);"
+        f'{"print(solver.connection_properties.cortex_host);" if inside_container else "print(solver.connection_properties.cortex_pid);"}'
+        "exit()"
+    )
+    output = subprocess.check_output(f'python -c "{script}"')
```
raph-luc, line 233: On my end need to have `shell = True` for this script to work:
```suggestion
    output = subprocess.check_output(f'python -c "{script}"', shell=True)
```

tests/test_fluent_session.py
```diff
@@ -220,3 +220,22 @@ def test_interrupt(load_static_mixer_case):
...
+
+
+def test_fluent_exit():
+    inside_container = os.getenv("PYFLUENT_LAUNCH_CONTAINER")
+    script = (
+        "import ansys.fluent.core as pyfluent;"
+        "solver = pyfluent.launch_fluent(start_watchdog=False);"
+        f'{"print(solver.connection_properties.cortex_host);" if inside_container else "print(solver.connection_properties.cortex_pid);"}'
+        "exit()"
+    )
```
raph-luc, line 232: And I think will need to disable logging, in case it is active:
```suggestion
    script = (
        "import ansys.fluent.core as pyfluent;"
        "import os;"
        "os.environ.pop('PYFLUENT_LOGGING', None);"
        "os.environ.pop('PYFLUENT_WATCHDOG_DEBUG', None);"
        "solver = pyfluent.launch_fluent(start_watchdog=False);"
        f'{"print(solver.connection_properties.cortex_host);" if inside_container else "print(solver.connection_properties.cortex_pid);"}'
        "exit()"
    )
```

raph-luc (approved):
LGTM, and nice that you are adding a test that can work both with and without containers as well

tests/test_fluent_session.py
raph-luc: Note I edited my suggestion above, `import ansys.fluent.core as pyfluent` needs to come after disabling logging for this command

mkundu1:
> LGTM, and nice that you are adding a test that can work both with and without containers as well
> 
> Very minor suggestion would be to rename `_exit_evt` to `_exit_event` just to make this var name clearer/obvious

Thanks for the suggestions, they are very helpful.

PR #2017, opened 2023-09-25
Make parent public.

prmukherj:


src/ansys/fluent/core/meta.py
```diff
@@ -149,10 +149,10 @@ def __create_get_ancestors_by_type(cls):
         def wrapper(self, obj_type, obj=None):
             obj = self if obj is None else obj
             parent = None
-            if getattr(obj, "_parent", None):
-                if isinstance(obj._parent, obj_type):
-                    return obj._parent
-                parent = self.get_ancestors_by_type(obj_type, obj._parent)
+            if getattr(obj, "parent", None):
```
seanpearsonuk, line 152: We also want a method to avoid writing `parent.parent.parent.parent.parent`. Do we have that here?

seanpearsonuk (approved):

src/ansys/fluent/core/meta.py
prmukherj: yes, this is that method. @seanpearsonuk

PR #2028, opened 2023-09-26
Added environmental variable PYFLUENT_ADDITIONAL_ARGUMENTS

schrummy14:
Added an environmental variable to allow a user to pass additional arguments to the launcher without having to change their python file.

Launcher checks if the environment variable is set and that the user did not launch with the additional_arguments option. If this is the case, the user additional_arguments are used instead.

### DIFF
--- a/doc/source/user_guide/launching_ansys_fluent.rst
+++ b/doc/source/user_guide/launching_ansys_fluent.rst
@@ -111,7 +111,7 @@ scheduler environments are Univa Grid Engine (UGE), Load Sharing Facility (LSF),
 Portable Batch System (PBS), and Slurm.
 
 This example shows a bash shell script that can be submitted to a Slurm
-scheduler using the ```sbatch``` command:  
+scheduler using the ```sbatch``` command:
 
 .. code:: bash
 
@@ -194,3 +194,29 @@ For distributed parallel processing, you usually pass both parameters:
       mode="solver",
       additional_arguments="-t16 -cnf=m1:8,m2:8",
    )
+
+Using the PYFLUENT_ADDITIONAL_ARGUMENTS environment variable
+------------------------------------------------------------
+You can set additional arguments also using this environment variable.
+This can be useful for running on multiple platforms and allows you to keep
+a the same python script to use on the platforms.
+
+.. code:: bash
+
+   #!/bin/bash
+   #
+   # Activate your favorite Python environment
+   #
+   export AWP_ROOT232=/apps/ansys_inc/v232
+   . ./venv/bin/activate
+   #
+   # Set additional arguments
+   #
+   export PYFLUENT_ADDITIONAL_ARGUMENTS="-cnf=m1:8,m2:8 -t16 -peth.efa"
+   #
+   # Run a PyFluent script
+   #
+   python run.py
+
+Note: You should omit the additional_arguments option in the launch_fluent function as
+that variable is taken over the environmental variable.
--- a/doc/source/user_guide/launching_ansys_fluent.rst
+++ b/doc/source/user_guide/launching_ansys_fluent.rst
@@ -197,7 +197,7 @@ For distributed parallel processing, you usually pass both parameters:
 
 Using the PYFLUENT_ADDITIONAL_ARGUMENTS environment variable
 ------------------------------------------------------------
-You can set additional arguments also using this environment variable.
+You can also set additional arguments using this environment variable.
 This can be useful when running on multiple platforms, avoiding the need to change the
 Python scripts being run.
 
@@ -217,6 +217,3 @@ Python scripts being run.
    # Run a PyFluent script
    #
    python run.py
-
-Note: You should omit the additional_arguments option in the launch_fluent function as
-that variable is ignored when using the environmental variable.
--- a/doc/source/user_guide/launching_ansys_fluent.rst
+++ b/doc/source/user_guide/launching_ansys_fluent.rst
@@ -198,8 +198,8 @@ For distributed parallel processing, you usually pass both parameters:
 Using the PYFLUENT_ADDITIONAL_ARGUMENTS environment variable
 ------------------------------------------------------------
 You can set additional arguments also using this environment variable.
-This can be useful for running on multiple platforms and allows you to keep
-a the same python script to use on the platforms.
+This can be useful when running on multiple platforms, avoiding the need to change the
+Python scripts being run.
 
 .. code:: bash
 
--- a/doc/source/user_guide/launching_ansys_fluent.rst
+++ b/doc/source/user_guide/launching_ansys_fluent.rst
@@ -219,4 +219,4 @@ a the same python script to use on the platforms.
    python run.py
 
 Note: You should omit the additional_arguments option in the launch_fluent function as
-that variable is taken over the environmental variable.
+that variable is ignored when using the environmental variable.
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -600,6 +600,17 @@ def launch_fluent(
 
     del start_container
 
+    # Check if user set PYFLUENT_ADDITIONAL_ARGUMENTS var
+    if os.environ.get("PYFLUENT_ADDITIONAL_ARGUMENTS",None) is not None:
+        # Now check if user used the additional_arguments option
+        if additional_arguments is None:
+            additional_arguments = os.environ.get("PYFLUENT_ADDITIONAL_ARGUMENTS")
+        else:
+            logger.warning(
+                "Both the 'PYFLUENT_ADDITIONAL_ARGUMENTS' variable and 'additional_arguments' "
+                "options cannot be used together. Using the 'additional_arguments' option."
+            )
+
     if additional_arguments is None:
         additional_arguments = ""
     elif fluent_launch_mode == LaunchMode.PIM:
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -499,7 +499,7 @@ def launch_fluent(
         format they are normally passed to Fluent on the command line.
         As an alternative, a user may set the PYFLUENT_ADDITIONAL_ARGUMENTS environment
         variable instead of setting this option. If both the environment variable is set,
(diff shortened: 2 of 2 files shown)


src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -600,6 +600,17 @@ def launch_fluent(
 
     del start_container
 
+    # Check if user set PYFLUENT_ADDITIONAL_ARGUMENTS var
+    if os.environ.get("PYFLUENT_ADDITIONAL_ARGUMENTS", None) is not None:
```
mkundu1, line 604:
```suggestion
    if os.getenv("PYFLUENT_ADDITIONAL_ARGUMENTS"):
```

mkundu1: Please also update the function docstring

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -600,6 +600,17 @@ def launch_fluent(
...
+        # Now check if user used the additional_arguments option
+        if additional_arguments is None:
+            additional_arguments = os.environ.get("PYFLUENT_ADDITIONAL_ARGUMENTS")
+        else:
+            logger.warning(
+                "Both the 'PYFLUENT_ADDITIONAL_ARGUMENTS' variable and 'additional_arguments' "
+                "options cannot be used together. Using the 'additional_arguments' option."
```
mkundu1, line 611: The purpose of the environment variable is to override the script behaviour. So, if both are specified, the environment variable will take precedence.

doc/source/user_guide/launching_ansys_fluent.rst
```diff
@@ -193,4 +193,30 @@ For distributed parallel processing, you usually pass both parameters:
...
+   # Set additional arguments
+   #
+   export PYFLUENT_ADDITIONAL_ARGUMENTS="-cnf=m1:8,m2:8 -t16 -peth.efa"
+   #
+   # Run a PyFluent script
+   #
+   python run.py
+
+Note: You should omit the additional_arguments option in the launch_fluent function as
+that variable is taken over the environmental variable.
```
raph-luc, line 222:
```suggestion
Note: You should omit the ``additional_arguments`` option in the ``launch_fluent`` function as
that option overrides what is set in the environment variable.
```

raph-luc: Should it be the other way around? Env var overriding what is set in the `launch_fluent` arguments?

src/ansys/fluent/core/launcher/launcher.py
raph-luc: Agreed, makes more sense to me

doc/source/user_guide/launching_ansys_fluent.rst
schrummy14: I have updated the documentation

src/ansys/fluent/core/launcher/launcher.py
schrummy14: I has been updated

mkundu1 (approved):

doc/source/user_guide/launching_ansys_fluent.rst
```diff
@@ -193,4 +193,30 @@ For distributed parallel processing, you usually pass both parameters:
...
+   # Set additional arguments
+   #
+   export PYFLUENT_ADDITIONAL_ARGUMENTS="-cnf=m1:8,m2:8 -t16 -peth.efa"
+   #
+   # Run a PyFluent script
+   #
+   python run.py
+
+Note: You should omit the additional_arguments option in the launch_fluent function as
+that variable is ignored when using the environmental variable.
```
raph-luc, line 222:
```suggestion
Note: You should omit the ``additional_arguments`` option in the ``launch_fluent`` function as
that variable is ignored when using the environment variable.
```
Can also remove this point as suggested by Mainak

raph-luc (approved):

doc/source/user_guide/launching_ansys_fluent.rst
schrummy14: The note has been removed

prmukherj (approved):

dnwillia-work (commented):
@schrummy14  

There seems to be no issue linked to this PR where the specifics about why this is needed are discussed.

The thing is that you could easily implement this into your Python journal:
```
import os
import ansys.fluent.core as pf
add_args = os.environ.get("PYFLUENT_ADDITIONAL_ARGUMENTS", None)
pf.launch_fluent(... additional_arguments=add_args ...)
```

So I guess I don't really understand the use case to make it worth cluttering up the launch fluent implementation logic with this detail.

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -600,6 +603,16 @@ def launch_fluent(
...
 
+    # Check if user set PYFLUENT_ADDITIONAL_ARGUMENTS var
+    if os.getenv("PYFLUENT_ADDITIONAL_ARGUMENTS") is not None:
+        # Now check if user used the additional_arguments option
+        if additional_arguments is not None:
+            logger.warning(
+                "Both the 'PYFLUENT_ADDITIONAL_ARGUMENTS' variable and 'additional_arguments' "
+                "options cannot be used together. Using the 'PYFLUENT_ADDITIONAL_ARGUMENTS' option."
+            )
+        additional_arguments = os.environ.get("PYFLUENT_ADDITIONAL_ARGUMENTS")
```
dnwillia, line 614: I don't agree with this behaviour.  The arguments I explicitly pass to the launch_fluent call should always take precedence over the configured environment.

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -600,6 +603,16 @@ def launch_fluent(
 
     del start_container
 
+    # Check if user set PYFLUENT_ADDITIONAL_ARGUMENTS var
```
dnwillia-work, line 606: I suppose that I don't agree with this behaviour. The arguments I explicitly pass to the launch_fluent call should always take precedence over the configured environment.

doc/source/user_guide/launching_ansys_fluent.rst
```diff
@@ -193,4 +193,27 @@ For distributed parallel processing, you usually pass both parameters:
       version="3d",
       mode="solver",
       additional_arguments="-t16 -cnf=m1:8,m2:8",
-   )
\ No newline at end of file
+   )
+
+Using the PYFLUENT_ADDITIONAL_ARGUMENTS environment variable
```
dnwillia-work, line 198: I don't think you need this doc here at all unless you want to go through the library and document all the other environment variables that are used for various purposes.

Generally we've not really been documenting them by design, because we don't want to encourage library users to be using them.  i.e. they are kind of meant to be hidden.

seanpearsonuk (changes requested):
I believe we can do without this.

seanpearsonuk:
I am not able to see the code changes. Does anyone have a measure of cyclomatic complexities before and after the change?

raph-luc:
> I am not able to see the code changes. Does anyone have a measure of cyclomatic complexities before and after the change?

@seanpearsonuk the only non-documentation change is adding 
```
# Check if user set PYFLUENT_ADDITIONAL_ARGUMENTS var
if os.getenv("PYFLUENT_ADDITIONAL_ARGUMENTS") is not None:
    # Now check if user used the additional_arguments option
    if additional_arguments is not None:
        logger.warning(
            "Both the 'PYFLUENT_ADDITIONAL_ARGUMENTS' variable and 'additional_arguments' "
            "options cannot be used together. Using the 'PYFLUENT_ADDITIONAL_ARGUMENTS' option."
        )
    additional_arguments = os.environ.get("PYFLUENT_ADDITIONAL_ARGUMENTS")
```
After this line: 

https://github.com/ansys/pyfluent/blob/fa41f347745740c41601aea45ca8c9209fa98e0d/src/ansys/fluent/core/launcher/launcher.py#L601

src/ansys/fluent/core/launcher/launcher.py
raph-luc: In my understanding, the purpose of this environment variable was specifically to override whatever is defined in scripts. 

However, I can see now that this could cause some unexpected/unintuitive behavior: defining this env var this way will also cause Python interpreter calls to `launch_fluent` with `additional_arguments` to be ignored. This is not necessarily an issue, but certainly an undesirable side-effect.

In this case, I think it makes the most sense to revert to the behavior that had been initially suggested by @schrummy14, where the actual arguments specified for `launch_fluent` take precedence over what is defined in the env var. Thoughts @schrummy14 @mkundu1 ?

doc/source/user_guide/launching_ansys_fluent.rst
dnwillia-work: In some cases they have a specific uses such as:

- point the library at a local development version of Fluent
- control some specific detail of the CI/CD pipeline that is part of building and testing the library, but not using it

Neither of these things is needed to _use_ the library right. Fundamentally I don't think there should be any behaviour of the _customer facing_ API calls that depend on any environment variable setting.  Environment variables should not be part of an API IMHO.

dnwillia-work:
> > @schrummy14
> > There seems to be no issue linked to this PR where the specifics about why this is needed are discussed.
> > The thing is that you could easily implement this into your Python journal:
> > ```
> > import os
> > import ansys.fluent.core as pf
> > add_args = os.environ.get("PYFLUENT_ADDITIONAL_ARGUMENTS", None)
> > pf.launch_fluent(... additional_arguments=add_args ...)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So I guess I don't really understand the use case that makes it worth cluttering up the launch fluent implementation logic with this detail.
> 
> @dnwillia-work @seanpearsonuk I have linked the issue (with a link to the [discussion](https://discuss.ansys.com/discussion/2535/pyfluent-environment-variable/p1?new=1) in the developer forum) with this PR. Shall we close the forum discussion with this suggestion?

It seems like a valid way for the user to handle it that does not require changing the library so I don't see why not.  I've never looked at the developer forum so thanks for pointing that out.

mkundu1:
Closing as user-side solution exists.

PR #2035, opened 2023-09-27
General doc updates, refactoring

hpohekar:
1. Field data and Field info moved to General section.

2. type = None -> Optional[type] = None.

3. Updated asynchronous doc.

4. Updated scheduler info in `launch_fluent`.


seanpearsonuk (approved):
Great effort, @hpohekar, thanks!

raph-luc:
In my local doc build of this branch, toggle-prompt doesn't seem to be working. Is it perhaps because we have the copy button in the same location? Maybe I am looking for it in the wrong place?
image

requirements/requirements_doc.txt
```diff
@@ -12,4 +12,5 @@ sphinx-gallery==0.13.0
 sphinx-notfound-page==0.8.3
 sphinxcontrib-websupport==1.2.5
 sphinxemoji==0.2.0
+sphinx-toggleprompt
```
raph-luc, line 15: We should specify the version, as in the other dependencies

The automated install used version 0.0.5 but that is a very old version from 3 years ago, maybe that is why it isn't working?

prmukherj (approved):

mkundu1:
> In my local doc build of this branch, toggle-prompt doesn't seem to be working. Is it perhaps because we have the copy button in the same location? Maybe I am looking for it in the wrong place? image

Yeah, it should appear like [this](https://sphinx-toggleprompt.readthedocs.io/en/latest/#using-with-sphinx-copybutton)

seanpearsonuk:
We have many code snippets that are just calls without prompts or return values, like:
```
x.foo()
x.bar()
```
Now we are able to toggle between
```
>>> x.foo()
42
>>> x.bar()
0
```
and
```
x.foo()
x.bar()
```

requirements/requirements_doc.txt
hpohekar: Done. Thank you.

hpohekar:
sphinx_toggleprompt==0.4.0 needs sphinx <7 and >4.5.

We need to wait until sphinx_toggleprompt supports sphinx >= 7.1.2

PR #2045, opened 2023-09-28
Advise TUI -> settings API upgrade

mkundu1:
The following advice will be printed if an equivalent settings API exists:
```
>>> solver.tui.define.models.energy("yes", "no", "no", "no", "yes")
The following settings API could also be used to execute the above command:
<solver_session>.setup.models.energy = {"enabled" : True}
```
The feature will work with Fluent version 23.1 onward.


mkundu1:
We may need to disable the feature in the documentation examples run.

src/ansys/fluent/core/services/api_upgrade.py
```diff
@@ -0,0 +1,41 @@
...
+            journal_str = self._scheme_eval(
+                "(close-output-port pyfluent-journal-str-port)"
+            ).strip()
+            if (
+                journal_str.startswith("solver.")
+                and not journal_str.startswith("solver.tui")
+                and not journal_str.startswith("solver.execute_tui")
+            ):
+                print(
+                    "The following settings API could also be used to execute the above command:"
```
seanpearsonuk, line 39:
```suggestion
                    "The following solver settings object method could also be used to execute the above command:"
```
?

seanpearsonuk (approved):

prmukherj (approved):

PR #2052, opened 2023-09-29
Removing most of the unhelpful console output in examples

raph-luc:


examples/00-fluent/exhaust_system.py
```diff
@@ -46,6 +44,7 @@
 # Perform required imports, which includes downloading and importing
 # the geometry file.
 
+# sphinx_gallery_thumbnail_path = '_static/exhaust_system.png'
```
raph-luc, line 47: Thumbnail specification needs to be in a code block (also mentioned in the [sphinx-gallery documentation](https://sphinx-gallery.github.io/stable/configuration.html#providing-thumbnail)), otherwise the command will show up in the webpage to users, as it was doing before

doc/source/conf.py
```diff
@@ -156,6 +156,9 @@ def _stop_fluent_container(gallery_conf, fname):
     "thumbnail_size": (350, 350),
     "reset_modules_order": "after",
     "reset_modules": (_stop_fluent_container),
+    "capture_repr": (),
+    "remove_config_comments": True,
+    "abort_on_example_error": True,
```
raph-luc, line 161: Unrelated to the issue, but a command that a noticed among the options that could be useful: `abort_on_example_error`, to stop building the examples and fail when an error occurs, which I think is what we expected it to do by default. Let me know if the behavior of continuing despite failures is preferred and I will remove this.

hpohekar (approved):

PR #2063, opened 2023-10-03
pre-commit CI should catch removeprefix like issues in code - testing

mkundu1:


mkundu1:
It seems linters cannot be run for a different python version than what is already installed. This change will cause more issues in local development.

PR #2068, opened 2023-10-04
Update examples doc

hpohekar:
* Changed debug message into warning.

* Added clarification about precedence of arguments / environment variables. 

* Updated information of specifying file paths 

image

### DIFF
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -367,8 +367,8 @@ def _get_server_info(
 ):
     """Get server connection information of an already running session."""
     if ip and port:
-        logger.debug(
-            "The server-info file was not parsed because ip and port were provided explicitly."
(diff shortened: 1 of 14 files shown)


src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -367,8 +367,8 @@ def _get_server_info(
 ):
     """Get server connection information of an already running session."""
     if ip and port:
-        logger.debug(
-            "The server-info file was not parsed because ip and port were provided explicitly."
+        logger.warning(
+            "Could not parsed server-info file because ip and port were provided explicitly."
```
seanpearsonuk, line 371:
```suggestion
            "Could not parse server-info file because ip and port were provided explicitly."
```

seanpearsonuk (approved):

prmukherj (approved):

PR #2071, opened 2023-10-04
Seamless file handling

hpohekar:
Support upload and download to PyFluent file read and write methods in Ansys Lab respectively.

* Added an existing `client.py` file from `third_party/mayth/simple_upload_server/client.py` to support upload and download of files. 

*  Removed dependency on separate whl file, PyFluent-core package is self-contained now.

* No need to use `simple_upload_server_client-0.0.1-py3-none-any.whl` in Ansys Lab now onwards.

* Verified that pyfluent-core working fine without installation of `simple_upload_server_client-0.0.1-py3-none-any.whl` in Ansys Lab.

* Added `read_case` and `write_case` methods on session objects.

* Added `_ServerFileManager` class for Meshing and Solver session.

`-----------------------------------------------------`

image

image

image

image

image

image


prmukherj:
@hpohekar, it is not possible to add local tests, right?

hpohekar:
> @hpohekar, it is not possible to add local tests, right?

No.

src/ansys/fluent/core/session_solver.py
```diff
@@ -228,3 +235,120 @@ def read_case(self, file_name: str, lightweight_mode: bool = False):
             fut.add_done_callback(functools.partial(Solver._sync_from_future, self))
         else:
             self.file.read(file_type="case", file_name=file_name)
+
```
seanpearsonuk, line 238: @hpohekar It will be a good exercise for you to extract and consolidate all the duplicated code.

src/ansys/fluent/core/utils/client.py
```diff
@@ -0,0 +1,106 @@
+# ©2023, ANSYS Inc. Unauthorized use, distribution or duplication is prohibited.
```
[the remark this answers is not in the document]
hpohekar: We will remove this copyright notice. Thanks.

src/ansys/fluent/core/session_solver.py
hpohekar: Right, I tried to add them in `BaseSession` to avoid all duplication but `tui` for meshing and `file` for solver is not available there.

src/ansys/fluent/core/session_meshing.py
```diff
@@ -66,3 +72,120 @@ def PMFileManagement(self):
...
+        ----------
+        file_name : str
+            Case file name
+        """
+        if pypim.is_configured():
+            if os.path.isfile(file_name):
+                self.upload(os.path.basename(file_name))
+            else:
+                raise FileNotFoundError(f"{file_name} does not exist.")
+            self._wait_for_file(file_name, self.file_service)
```
seanpearsonuk, line 169: A slightly different strategy that we discussed is checking if the file exists remotely.

I think it is reasonable and correct to first check local existence, as you are doing. But if it doesn't exist locally, then we could check remote existence directly. This could cover a valid use case.

Can `upload()` fail if the file already exists remotely, and if so, does it raise an exception. If so, that is fine, we should try-except that code, and simply proceed with the `read_case()`.

It would be useful to have a notebook avaialbel to run in the lab and test all these scenarios.

src/ansys/fluent/core/session_solver.py
hpohekar: Yes sure. Thank you.

src/ansys/fluent/core/session_meshing.py
hpohekar: Yes, we will implement this check. Thank you.

PR #2094, opened 2023-10-05
Add doc for exceptions

mkundu1:
1. Added [missing-raises-doc](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/missing-raises-doc.html) check from pylint as a part of pre-commit checks.
2. Added the missing documentation for raised exceptions.
3. Using `# pylint: disable=missing-raises-doc` for some methods which users won't directly use.


src/ansys/fluent/core/session.py
```diff
@@ -314,14 +314,35 @@ def __init__(self, pim_instance):
...
 
     def download(self, file_name: str, local_file_path: Optional[str] = None):
-        """Downloads a file from the server."""
+        """Downloads a file from the server.
+
+        Parameters
+        ----------
+        file_name : str
+            filename
+        local_file_path : Optional[str], optional
```
raph-luc, line 338:
```suggestion
        local_file_path : str, optional
```

src/ansys/fluent/core/session.py
```diff
@@ -314,14 +314,35 @@ def __init__(self, pim_instance):
...
 
     def upload(self, file_path: str, remote_file_name: Optional[str] = None):
-        """Uploads a file on the server."""
+        """Uploads a file on the server.
+
+        Parameters
+        ----------
+        file_path : str
+            filepath
+        remote_file_name : Optional[str], optional
```
raph-luc, line 323:
```suggestion
        remote_file_name : str, optional
```

src/ansys/fluent/core/launcher/fluent_container.py
```diff
@@ -118,6 +118,15 @@ def configure_container_dict(
...
     remove_server_info_file: bool
 
+    Raises
+    ------
+    KeyError
+        If license server is not specified through environment variable or ``container_dict``.
+    ValueError
+        If server info file is specified through both command-line argument and ``container_server_info_file``.
+    ValueError
+        If ``fluent_image`` or ``image_tag`` and ``image_name`` is not specified.
```
raph-luc, line 128:
```suggestion
        If ``fluent_image`` or ``image_tag`` and ``image_name`` are not specified.
```

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -295,7 +296,13 @@ def __init__(
         self._service.events[self.tag] = self
 
     def unsubscribe(self):
-        """Unsubscribe the datamodel event."""
+        """Unsubscribe the datamodel event.
+
+        Raises
+        ------
+        RuntimeError
+            If server fails to unsubscribe event.
```
raph-luc, line 304:
```suggestion
            If server fails to unsubscribe from event.
```

src/ansys/fluent/core/file_session.py
```diff
@@ -537,6 +586,11 @@ def get_pathlines_field_data(
         Dict
             Dictionary containing a map of surface IDs to the pathline data.
             For example, pathlines connectivity, vertices, and field.
+
+        Raises
+        ------
+        RuntimeError
+            Always, as pathlines are not supported.
```
raph-luc, line 593: Why do we even have this function if the only thing it does is raise an exception? 🤔 Do we plan to support this later on and this is a placeholder for now?

src/ansys/fluent/core/launcher/fluent_container.py
```diff
@@ -118,6 +118,15 @@ def configure_container_dict(
     host_server_info_file : Path
     remove_server_info_file: bool
 
+    Raises
+    ------
+    KeyError
+        If license server is not specified through environment variable or ``container_dict``.
+    ValueError
+        If server info file is specified through both command-line argument and ``container_server_info_file``.
```
raph-luc, line 126:
```suggestion
        If server info file is specified through both a command-line argument inside ``container_dict`` and the  ``container_server_info_file`` parameter.
```

src/ansys/fluent/core/file_session.py
```diff
@@ -105,6 +105,13 @@ def add_scalar_fields_request(
         Returns
         -------
         None
+
+        Raises
+        ------
+        RuntimeError
+            If surface names or surface ids are not provided.
```
raph-luc, line 112: I believe "id" should be uppercase as "ID", but I think we already use it in too many places as "id" to change this in this PR, better do this in a separate PR if anyone considers this an actual issue in the future

src/ansys/fluent/core/launcher/fluent_container.py
```diff
@@ -118,6 +118,15 @@ def configure_container_dict(
     host_server_info_file : Path
     remove_server_info_file: bool
 
+    Raises
+    ------
+    KeyError
+        If license server is not specified through environment variable or ``container_dict``.
```
raph-luc, line 124:
```suggestion
        If license server is not specified through an environment variable or in ``container_dict``.
```

raph-luc (approved):
Looks great, few minor suggestions/thoughts

src/ansys/fluent/core/file_session.py
seanpearsonuk: I agree with @raph-luc, especially as id (plural ids) is a noun.

src/ansys/fluent/core/file_session.py
seanpearsonuk: @raph-luc it's not obvious, but the reason is that here we have (re)implemented an established interface. An implementation already existed for live data (in the application). This newer implementation is for file-based data, where the particular data is unavailable.

src/ansys/fluent/core/launcher/fluent_container.py
```diff
@@ -118,6 +118,15 @@ def configure_container_dict(
...
     remove_server_info_file: bool
 
+    Raises
+    ------
+    KeyError
+        If license server is not specified through an environment variable or in ``container_dict``.
+    ValueError
+        If server info file is specified through both a command-line argument inside ``container_dict`` and the  ``container_server_info_file`` parameter.
+    ValueError
+        If ``fluent_image`` or ``image_tag`` and ``image_name`` are not specified.
```
seanpearsonuk, line 128: The bonus value here is that developers can see where we need to design higher-level exception types.

See #2098

@mkundu1 @raph-luc

seanpearsonuk (approved):

hpohekar (approved):

prmukherj (approved):

src/ansys/fluent/core/launcher/fluent_container.py
raph-luc: I think that adding higher-level custom exception classes is something that may not end up being helpful, particularly if we aren't catching those custom exceptions and also don't expect users to catch and handle these exceptions in different ways necessarily, if that makes sense. Will comment on the issue tracker

PR #2095, opened 2023-10-05
Remove reader code, support Python version 3.12, drop 3.8

mkundu1:
1. Move reader code to separate repo https://github.com/ansys/pyfluent-reader.
2. Support Python version 3.12, drop support for 3.8.


tests/test_rp_vars.py
```diff
@@ -44,9 +43,10 @@ def test_get_all_rp_vars(new_solver_session_no_transcript) -> None:
     assert len(all_vars) == pytest.approx(9000, 20)
 
     # CaseFile comparison, note that the PyFluent work dir is not necessarily the same as the Fluent work dir
-    case = CaseReader(case_filepath=path(case_path))
-    case_vars = case.rp_vars()
-    assert len(case_vars) == pytest.approx(9000, 450)
+    # To be moved to reader repo
```
mkundu1, line 46: https://github.com/ansys/pyfluent-reader/issues/9

hpohekar (approved):

.github/workflows/ci.yml
```diff
@@ -365,7 +365,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: 3.9
```
RobPasMue, line 368: @mkundu1 - have a look at other repos. We typically define a MAIN_PYTHON_VERSION. See https://github.com/ansys/pyprimemesh/blob/main/.github/workflows/ci_cd.yml#L19

That way it is easier to maintain and to see what is the main Python version you are doing operations against.

seanpearsonuk:
@mkundu1 do we definitely want to break up the package at this point? I'm not against it per se, but if it's to support 3.12, should we wait, given that we have 6 mo soft deadline?

.github/workflows/ci.yml
prmukherj: Done, Thank you.

mkundu1:
> @mkundu1 do we definitely want to break up the package at this point? I'm not against it per se, but if it's to support 3.12, should we wait, given that we have 6 mo soft deadline?

Yeah, we shall wait at least till pyfluent-reader package is released.

mkundu1:
closing this, will reactivate once we are ready to merge

PR #2102, opened 2023-10-06
Cavitation Model Example And Example Warning Fix

ansaboutin:
Created a new PyFluent example following the Chapter 22: Modeling Cavitation Fluent example. 

A few tui commands remain. Equivalent settings commands do not exist.


examples/00-fluent/modeling_cavitation.py
```diff
@@ -0,0 +1,385 @@
...
+**Problem description**
+
+The problem considers the cavitation caused by the flow separation
+after a sharp-edged orifice. The flow is pressure driven, with an
+inlet pressure of 5 x 10^5 Pa and an outlet pressure of 9.5 x 10^4 Pa.
+The orifice diameter is 4 x 10^-3 m, and the geometrical parameters
+of the orifice are D/d = 2.88 and L/d = 4, where D, d, and L are the
+inlet diameter, orifice diameter, and orifice length respectively.
+"""
+# sphinx_gallery_thumbnail_path = '_static\cavitation_model.png'
```
raph-luc, line 32:
```suggestion
```
This sphinx gallery command needs to be in a code block, otherwise it will show up in the rendered webpage (see my other suggestion)

examples/00-fluent/modeling_cavitation.py
```diff
@@ -0,0 +1,385 @@
...
+# Example Setup
+# -------------
+# Before you can begin, you must set up the example and initialize this workflow.
+#
+# Perform required imports
+# ~~~~~~~~~~~~~~~~~~~~~~~~
+# Perform required imports, which includes downloading and importing
+# the geometry file.
+
+import ansys.fluent.core as pyfluent
```
raph-luc, line 44:
```suggestion

# sphinx_gallery_thumbnail_path = '_static\cavitation_model.png'
import ansys.fluent.core as pyfluent
```

examples/00-fluent/modeling_cavitation.py
```diff
@@ -0,0 +1,385 @@
...
+# Launch Fluent as a service in 2d solution mode with double precision running
+# on one processor.
+
+solver = pyfluent.launch_fluent(
+    precision="double",
+    processor_count=1,
+    mode="solver",
+    version="2d",
+    cwd=pyfluent.EXAMPLES_PATH,
+    show_gui=True,
```
raph-luc, line 63:
```suggestion
```

examples/00-fluent/modeling_cavitation.py
ansaboutin: Thank you, the changes have been added.

seanpearsonuk:
Hi @ansaboutin Re the question about replacing TUI commands with settings API commands, please see this merged PR by @mkundu1 : https://github.com/ansys/pyfluent/pull/2045, which automates the discovery of such translations: "advice will be printed if an equivalent settings API exists".

examples/00-fluent/modeling_cavitation.py
```diff
@@ -0,0 +1,377 @@
...
+# ~~~~~~~~~~~~~~~~~~~~~~~
+solver.read_case("cav.msh")
+
+solver.mesh.check()
+
+###############################################################################
+# Set units for length
+# ~~~~~~~~~~~~~~~~~~~~
+
+solver.tui.define.units("length", "m")
```
seanpearsonuk, line 76: @ansaboutin We need to understand whether this call is really required for this example to work, and if so, what the purpose/effect of calling it is.

examples/00-fluent/modeling_cavitation.py
```diff
@@ -0,0 +1,377 @@
+"""
+
+Modeling Cavitation
+-------------------
+This example examines the pressure-driven cavitating flow of water
```
dnwillia-work, line 5: If you are editing this stuff in Visual Studio Code you should install this extension:

https://marketplace.visualstudio.com/items?itemName=stkb.rewrap

and rewrap to 80 columns.

examples/00-fluent/modeling_cavitation.py
dnwillia-work: yeah, if the mesh is already in meters then I don't think we need it.  Adam, did you just copy this from one of the other examples?

examples/00-fluent/modeling_cavitation.py
```diff
@@ -0,0 +1,377 @@
...
+
+# Enable the cavitation model
+
+# Set the number of mass transfer mechanisms to 1, set liquid as the from phase,
+# set vapor as the to phase and cavitation as the mechanism.
+
+solver.tui.define.phases.set_domain_properties.interaction_domain.heat_mass_reactions.mass_transfer(
+    1, "liquid", "vapor", "cavitation", "1", "no", "no", "no"
+)
+"""Set the vaporization pressure to 3540 Pa and bubble number density to 1e+11."""
```
dnwillia-work, line 156: What is this here for?

examples/00-fluent/modeling_cavitation.py
```diff
@@ -0,0 +1,377 @@
...
+    "cav.msh", "pyfluent/cavitation", save_path=pyfluent.EXAMPLES_PATH
+)
+
+###############################################################################
+# Launch Fluent
+# ~~~~~~~~~~~~~
+# Launch Fluent as a service in 2d solution mode with double precision running
+# on one processor.
+
+solver = pyfluent.launch_fluent(
```
dnwillia-work, line 57: Staring at the documentation build error from the last pipeline execution I think you might need to force use of 24.1 here.

examples/00-fluent/modeling_cavitation.py
```diff
@@ -0,0 +1,377 @@
...
+)
+
+# Initialize the solution with hybrid initialization.
+
+solver.solution.initialization.hybrid_initialize()
+
+###############################################################################
+# Save and Run
+# ~~~~~~~~~~~~
+# Solve the case file (``cav.cas.h5``).
```
dnwillia-work, line 290: `Save` the case file

examples/00-fluent/modeling_cavitation.py
```diff
@@ -0,0 +1,377 @@
...
+contour_static_pressure = {
+    "coloring": {"option": "banded", "smooth": False},
+    "contour_lines": False,
+    "field": "pressure",
+    "filled": True,
+}
+
+solver.results.graphics.contour["contour_static_pressure"] = contour_static_pressure
+
+# Mirror the display.
```
dnwillia-work, line 322: Maybe write here something more descriptive like `Mirror the symmetry zone so that the full model is shown in the graphical objects`

examples/00-fluent/modeling_cavitation.py
raph-luc: You are correct, already mentioned this to Adam: we are using the last publicly released Fluent version to run all the examples. It might be possible to find equivalent commands that work in 23.2 so the example can be released now, or can wait until 24.1 is publicly released

dnwillia-work: Yeah, I'm kind of guessing that's what might be the issue.  There is also this bit in the documentation build output:
```
pyfluent.launcher WARNING: These specified arguments are only supported when starting local standalone Fluent clients: cwd.
```

raph-luc: That is a somewhat expected warning, in the sense that contributors usually test using Windows and local standalone Fluent clients, whereas in GitHub the tests run using Linux and inside containers. 

Simply removing `cwd` specification here will make the warning go away and otherwise make no difference in the GitHub examples gallery run, but will make a difference when running the examples locally using Windows/standalone Fluent.

We made changes (mainly myself and Mainak) in past PRs to try to abstract these differences away and make PyFluent work as similarly as possible and independent of the system and Fluent setup, as far as GitHub tests and examples are concerned. But there are still remaining differences, and this is one of them as you can tell by the warning message. 

Alternatively, we could add support for the `launch_fluent(cwd=...)` argument in Fluent container launches, perhaps using a `pyfluent.examples_path` object that behaves differently depending on how Fluent is being launched; or redirecting `cwd` to specify the host path that is being mounted into the guest container (though I am not sure if this is very intuitive, and will also muddy the `cwd` argument a bit)

examples/00-fluent/modeling_cavitation.py
ansaboutin: It can be removed. I had it in to match the fluent Chapter 22 as closely as possible.

examples/00-fluent/modeling_cavitation.py
ansaboutin: This was a command I could not find. These are default values which will not need to be set. This line has been removed.

examples/00-fluent/modeling_cavitation.py
```diff
@@ -0,0 +1,377 @@
...
+
+###############################################################################
+# Launch Fluent
+# ~~~~~~~~~~~~~
+# Launch Fluent as a service in 2d solution mode with double precision running
+# on one processor.
+
+solver = pyfluent.launch_fluent(
+    precision="double",
+    processor_count=1,
```
raph-luc, line 59: To improve GitHub run speed, I would recommend using at least 2, preferably 4, threads:
```suggestion
    processor_count=4,
```

raph-luc: Ah just realized this is 2D, so depending on the mesh, this might not make much of a difference at all, still probably a good idea

ansaboutin: @raph-luc I used 1 to follow the fluent tutorial. (https://ansyshelp.ansys.com/account/secured?returnurl=/Views/Secured/corp/v232/en/flu_tg/flu_tg_cavitation.html) Will this effect anything else in the example?

raph-luc: That is fair, but I don't think we need to follow the examples that closely. 

I believe the number of threads being used should not have any side effects in the example, please try increasing the number of threads and see if you notice any issues or significant speed improvements on your end.

src/ansys/fluent/core/streaming_services/monitor_streaming.py
```diff
@@ -176,7 +176,7 @@ def _process_streaming(self, id, stream_begin_method, started_evt, *args, **kwar
                         if monitor_data:
                             new_df = pd.DataFrame([monitor_data], columns=monitors)
                             new_df.set_index("xvalues", inplace=True)
-                            df_data["df"] = pd.concat([df, new_df])
+                            df_data["df"] = new_df
```
raph-luc, line 179: Why was this change made?

ansaboutin: I am trying to resolve this warning:
```
/home/ansys/actions-runner/_work/pyfluent/pyfluent/src/ansys/fluent/core/streaming_services/monitor_streaming.py:179: FutureWarning: The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.
  df_data["df"] = pd.concat([df, new_df])
```
from issue #2117. Is this the incorrect way to change this?

raph-luc: Ah yes, for future reference I think it would've probably been more appropriate to tackle the issues in that tracker in a separate PR, rather than bundling them up into this one, as the PR title and description do not really match what you are doing here.

Regarding that warning, if I'm understanding this correctly, would it make sense to check if `df` is indeed empty with `df.empty`, rather than just immediately discarding it? I don't think `df` is always going to be empty (if it is, then that part of the code should probably be revised). Would something like `df = df.append(new_df)` work?

@ajain-work @seanpearsonuk @mkundu1 any advice?

ansaboutin: Ok I will do that in the future, thank you for the heads up.
When I was looking at it I thought line 167 was creating an empty section to be filled later. I am unsure of how this works as a whole. 
image

examples/00-fluent/modeling_cavitation.py
dnwillia-work: OK I think we should leave it out and have no warnings in the build pipeline.  The `cwd` arg is not necessarily required in local standalone runs either right and someone could always add it locally to get things working.

src/ansys/fluent/core/streaming_services/monitor_streaming.py
dnwillia-work: Yeah, in this case it probably is better for a separate PR, but it's fine this time.  You have linked the issue to this PR which is good.

examples/00-fluent/modeling_cavitation.py
raph-luc: That is correct. I believe the only reason to keep `cwd` in the examples would be to demonstrate a very commonly used/useful command to folks learning PyFluent.

Regardless, leaving it out is fine by me.

examples/00-fluent/modeling_cavitation.py
```diff
@@ -0,0 +1,359 @@
...
+# ~~~~~~~~~~~~~
+# Launch Fluent as a service in 2d solution mode with double precision running
+# on one processor.
+
+solver = pyfluent.launch_fluent(
+    precision="double",
+    processor_count=4,
+    mode="solver",
+    version="2d",
+    product_version="23.2.0",
```
raph-luc, line 64:
```suggestion
```
This specification is unnecessary for the documentation pipeline, the version is specified in the GitHub workflow file

examples/00-fluent/modeling_cavitation.py
```diff
@@ -0,0 +1,359 @@
...
+# workflow.
+#
+# Perform required imports
+# ~~~~~~~~~~~~~~~~~~~~~~~~
+# Perform required imports, which includes downloading and importing
+# the geometry file.
+
+# sphinx_gallery_thumbnail_path = '_static/cavitation_model.png'
+
+import ansys.fluent.core as pyfluent
```
raph-luc, line 46:
```suggestion
# sphinx_gallery_thumbnail_path = '_static/cavitation_model.png'
import ansys.fluent.core as pyfluent
```
These two lines need to be in the same "block" with no lines in between, otherwise Sphinx will consider them two different blocks and display the `# sphinx_gallery_thumbnail_path = '_static/cavitation_model.png'` config command in the webpage output as described [here](https://sphinx-gallery.github.io/stable/configuration.html#providing-thumbnail).

examples/00-fluent/modeling_cavitation.py
ansaboutin: Yes, sorry forgot to check this.

examples/00-fluent/modeling_cavitation.py
ansaboutin: Ok I will keep them together.

examples/00-fluent/modeling_cavitation.py
```diff
@@ -0,0 +1,355 @@
...
+}
+
+solver.setup.materials.fluid["water"] = water
+
+# Copy water vapor from the materials database and modify the properties of your
+# local copy.
+
+solver.setup.materials.database.copy_by_name(type="fluid", name="water-vapor")
+
+# Set 0.02558 kg/m3 for Density and 1.26e-06 kg/m–s for Viscosity.
```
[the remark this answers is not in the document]
raph-luc: Same for code comments in other sections below

raph-luc (commented):
There are too many code comments (compare this [with other examples in the repo](https://fluent.docs.pyansys.com/version/stable/examples/index.html)): 
<details><summary>Screenshot example</summary>

image

</details>

I would suggest transforming almost all of the code comments into text (everywhere in this example, not only the screenshot above), and moving them to appropriate headings (or alternating between code and text as you see fit). I mentioned a few instances below.

See this for more details on how to switch between text and code: https://sphinx-gallery.github.io/stable/tutorials/plot_parse.html (download python source code at the bottom)

examples/00-fluent/modeling_cavitation.py
```diff
@@ -104,56 +103,48 @@
...
-# set vapor as the to phase and cavitation as the mechanism.
-
 solver.tui.define.phases.set_domain_properties.interaction_domain.heat_mass_reactions.mass_transfer(
     1, "liquid", "vapor", "cavitation", "1", "no", "no", "no"
 )
 
 ###############################################################################
 # Boundary Conditions
 # ~~~~~~~~~~~~~~~~~~~
-# Set momentum and turbulence boundary conditions for the first inlet
```
raph-luc, line 137: This extra empty line is causing the text to become a code block:
image

examples/00-fluent/modeling_cavitation.py
```diff
@@ -79,23 +79,22 @@
...
 ###############################################################################
 # Define materials
 # ~~~~~~~~~~~~~~~~
-# Create the material ``water``.
-
-# name: water density : constant reference density : 1000 [kg/m^3] viscosity :
-# constant viscosity method : constant reference viscosity : 0.001 [kg/(m s)]
+# Create the material named water. Use 1000 kg/m3 for density and 0.001 kg/m–s
+# for viscosity. Copy water vapor from the database and modify the copy. Change
+# the density to 0.02558 kg/m3 and viscosity to 1.26e-06 kg/m–s.
```
raph-luc, line 97:
```suggestion
# Create a material named water, using 1000 kg/m3 for density and 0.001 kg/m–s
# for viscosity. Then, copy water vapor properties from the database and modify the copy by changing
# the density to 0.02558 kg/m3 and the viscosity to 1.26e-06 kg/m–s.
```

examples/00-fluent/modeling_cavitation.py
```diff
@@ -104,56 +103,48 @@
...
 solver.setup.materials.fluid["water-vapor"] = water_vapor
 
 ###############################################################################
 # Phases
 # ~~~~~~
-# Change phase names to "liquid" and "water-vapor".
+# Change phase names to "liquid" for the primary phase and "water-vapor" for the
+# secondary phase. Enable the cavitation model and set the number of mass
+# transfer mechanisms to 1, liquid as the from phase, vapor as the to phase and
+# cavitation as the mechanism.
```
raph-luc, line 118:
```suggestion
# Change the name of the primary phase to "liquid" and the secondary phase to "water-vapor".
# Then, enable the cavitation model and set the number of mass
# transfer mechanisms to 1. Finally, specify cavitation as a mass transfer mechanism 
# occurring from the liquid to the vapor.
```

raph-luc (commented):
If the description for each section is going to be in a single block of text, then the text needs to be revised to be more readable (I suggested a few changes as examples). Breaking the longest text blocks into paragraphs may help as well.

In addition, if you think it would be more readable, feel free to consider breaking the larger code blocks into smaller parts, and spreading the text in between code blocks. For reference, see the link I sent above with information on how to more easily switch between text and code blocks.

raph-luc (commented):
Looks much better, thanks @ansaboutin, some additional suggestions below

dnwillia (approved):

examples/00-fluent/modeling_cavitation.py
```diff
@@ -0,0 +1,309 @@
...
+
+###############################################################################
+# Mirror the display around the symmetry plane to show the full model.
+
+solver.tui.display.set.mirror_zones(["symm_2", "symm_1"])
+
+###############################################################################
+# .. image:: /_static/cavitation_model_012.png
+#   :width: 500pt
+#   :align: center
```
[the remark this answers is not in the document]
dnwillia: We need them as static images only for the purpose of the documentation.  Yes we could add those commands but I don’t think we could generate the images dynamically during the build and use them for the doc.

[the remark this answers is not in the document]
ansaboutin: I added the save image lines. They come out in this format: 
image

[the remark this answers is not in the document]
raph-luc: @dnwillia I did not mean to generate the images dynamically and use them for the documentation build, my point was to just add the image output commands so users have an idea of how they might be able to obtain these images through PyFluent. Otherwise, and as things currently stand, when running the example they won't see any images at all, despite all the `graphics` and `contours` commands that are effectively not doing anything if these images aren't written to file or visible anywhere through PyFluent.

I guess users could see those contours if they opened a Fluent session with its own GUI and loaded the saved case file then displayed the contours manually.

And you're right, the other examples are also using graphics commands that don't output anything that is visible to the PyFluent user.

Moving this to a separate issue tracker sounds ok to me, going to do that.

@ansaboutin that is correct, that is what I meant by "the image output isn't going to look perfect".

examples/00-fluent/modeling_cavitation.py
```diff
@@ -0,0 +1,319 @@
...
+
+solver = pyfluent.launch_fluent(
+    precision="double",
+    processor_count=4,
+    mode="solver",
+    version="2d",
+)
+
+###############################################################################
+# Read the Mesh and Specify Models
```
raph-luc, line 61: The intention of my previous suggestion was to turn this into a heading, like so:
```suggestion
# Read the Mesh and Specify Models
# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
If you prefer this as text rather than a heading, then you should use sentence case and not title case.

raph-luc (approved):

PR #2108, opened 2023-10-07
Add type annotations for some modules under services

mkundu1:
Found [pytype](https://google.github.io/pytype/) is quite useful to generate the stubs for missing type annotations.

### DIFF
--- a/src/ansys/fluent/core/services/batch_ops.py
+++ b/src/ansys/fluent/core/services/batch_ops.py
@@ -2,14 +2,18 @@
 
 import inspect
 import logging
-from typing import List, Tuple
+from types import ModuleType
+from typing import List, Optional, Tuple, TypeVar
 import weakref
 
+from google.protobuf.message import Message
 import grpc
 
 import ansys.api.fluent.v0 as api
 from ansys.api.fluent.v0 import batch_ops_pb2, batch_ops_pb2_grpc
 from ansys.fluent.core.services.error_handler import catch_grpc_error
+
+_TBatchOps = TypeVar("_TBatchOps", bound="BatchOps")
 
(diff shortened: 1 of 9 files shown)


src/ansys/fluent/core/services/batch_ops.py
```diff
@@ -2,28 +2,34 @@
...
+_TBatchOps = TypeVar("_TBatchOps", bound="BatchOps")
+
+network_logger: logging.Logger = logging.getLogger("pyfluent.networking")
 
 
 class BatchOpsService:
     """Class wrapping methods in batch RPC service."""
 
-    def __init__(self, channel: grpc.Channel, metadata: List[Tuple[str, str]]):
+    def __init__(self, channel: grpc.Channel, metadata: list[tuple[str, str]]) -> None:
```
raph-luc, line 24: Don't know if this already has been mentioned somewhere else and I missed it, but these changes from `typing.List` to `list` shouldn't cause any issues, as we dropped support for Python 3.8, and these are a standard part of Python 3.9, right?

In this case we could clean up a lot of deprecated `typing` imports that we had been using to support Python 3.8 (there are many instances)

raph-luc (approved):

raph-luc:
Also, do you plan on adding this type-checking to the automated CI testing? Seems like a nice addition indeed

mkundu1:
> Also, do you plan on adding this type-checking to the automated CI testing? Seems like a nice addition indeed

As we cannot yet use the newer typing features added after Python 3.9, I feel it is best not to enforce this right now

raph-luc (approved):

seanpearsonuk (approved):

PR #2129, opened 2023-10-12
Seamless File Handling

hpohekar:
* Added `_ServerFileManager` class in Meshing and Solver session.

Prior to this, we were uploading a file using `session.upload()` method to read it using `session.api.file.read_case()` method.

We were downloading a file using `session.download()` method after execution of `session.api.file.write_case()` method.

Now we are handling upload and download part internally without need of specifying it explicitly.

We have demonstrated it in following screenshots.

image

`----------------------------------------------------`

image

image

image

image

### DIFF
--- a/src/ansys/fluent/core/session_meshing.py
+++ b/src/ansys/fluent/core/session_meshing.py
@@ -1,12 +1,9 @@
 """Module containing class encapsulating Fluent connection."""
-import os
-import time
 from typing import Any
 
 from ansys.fluent.core.fluent_connection import FluentConnection
 from ansys.fluent.core.session_pure_meshing import PureMeshing
 from ansys.fluent.core.session_solver import Solver
-import ansys.platform.instancemanagement as pypim
 
 
 class Meshing(PureMeshing):
@@ -29,10 +26,8 @@ class Meshing(PureMeshing):
             fluent_connection (:ref:`ref_fluent_connection`): Encapsulates a Fluent connection.
         """
         super(Meshing, self).__init__(fluent_connection=fluent_connection)
-        self._fluent_connection = fluent_connection
         self.switch_to_solver = lambda: self._switch_to_solver()
         self.switched = False
-        self._server_file_manager = None
 
     def _switch_to_solver(self) -> Any:
         self.tui.switch_to_solution_mode("yes")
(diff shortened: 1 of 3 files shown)


prmukherj (approved):

seanpearsonuk:
@hpohekar 
I don't know how I can review this if it doesn't explain the way forward. Is there a plan for getting rid of the duplication? I'm not sure that this is the best structure for that because the new code mixes all the concerns together in a way that makes it difficult to parametrize for the specific session types. But if you have a particular design in mind, you should explain it at this stage rather than leaving the reviewer to try to piece everything together. If you have no plan then you'll probably just end up having to undo all the changes later and implement something new again. In the meantime, all the reviewers have to wade through two copies of the new code without understanding any of the rationale behind the design. 

Note that all the code is still completely duplicated in the two classes, just as in the original PR. The code has been restructured but none of the duplication has been eliminated. On opening the PR, it was instantly obvious that none of the duplication had been eliminated before looking at any of the code, because the only changed files are the two session modules. For duplication to be eliminated there would be new code in a third, shared module. Why is there no shared code? If you look at the two changed modules side-by-side, you can see the same big code block - 130 lines in each file with a couple of incidental differences. You must have been aware of that because you pasted the code.

It's important to state properly in the description what the point of the PR is. One key comment from the original PR that was reiterated subsquently in the issue was the emilniation of duplication. I need to know that this point is properly acknowledged and how this PR fits into that plan.

@prmukherj: you approved the PR. What was your view on the large-scale duplication here? What did you think of this design? Every approval counts, so we must not approve code where we can see issues and just ignore them. The issues here are plainly obvious - it's impossible to glance at the code without realizing it.

hpohekar:
> @hpohekar I don't know how I can review this if it doesn't explain the way forward. Is there a plan for getting rid of the duplication? I'm not sure that this is the best structure for that because the new code mixes all the concerns together in a way that makes it difficult to parametrize for the specific session types. But if you have a particular design in mind, you should explain it at this stage rather than leaving the reviewer to try to piece everything together. If you have no plan then you'll probably just end up having to undo all the changes later and implement something new again. In the meantime, all the reviewers have to wade through two copies of the new code without understanding any of the rationale behind the design.
> 
> Note that all the code is still completely duplicated in the two classes, just as in the original PR. The code has been restructured but none of the duplication has been eliminated. On opening the PR, it was instantly obvious that none of the duplication had been eliminated before looking at any of the code, because the only changed files are the two session modules. For duplication to be eliminated there would be new code in a third, shared module. Why is there no shared code? If you look at the two changed modules side-by-side, you can see the same big code block - 130 lines in each file with a couple of incidental differences. You must have been aware of that because you pasted the code.
> 
> It's important to state properly in the description what the point of the PR is. One key comment from the original PR that was reiterated subsquently in the issue was the elimination of duplication. I need to know that this point is properly acknowledged and how this PR fits into that plan.
> 
> @prmukherj: you approved the PR. What was your view on the large-scale duplication here? What did you think of this design? Every approval counts, so we must not approve code where we can see issues and just ignore them. The issues here are plainly obvious - it's impossible to glance at the code without realizing it.

The description has been updated.

@seanpearsonuk I just followed your suggestion to do refactoring in a separate PR after merging this PR. Therefore I have created a new issue to address this.

image

We just removed `client.py` due to copyright issues in this PR. If we want to remove duplication before merging this PR then we   can make this PR as draft.

I already tested it in the Ansys Lab, if anyone is interested I can share the notebook.

hpohekar:
We will merge this PR after refactoring existing code.

src/ansys/fluent/core/session_meshing.py
```diff
@@ -66,3 +71,130 @@ def PMFileManagement(self):
     def preferences(self):
         """Datamodel root of preferences."""
         return super(Meshing, self).preferences if not self.switched else None
+
+    def read_case(
```
seanpearsonuk, line 75: If you add it at this level, then it won't be available in `PureMeshing`.

src/ansys/fluent/core/session_meshing.py
```diff
@@ -66,3 +71,130 @@ def PMFileManagement(self):
...
+
+        Raises
+        ------
+        FileNotFoundError
+            If a case file does not exist.
+        """
+        if pypim.is_configured():
+            if os.path.isfile(file_name):
+                if not self.file_service.file_exist(os.path.basename(file_name)):
+                    self.upload(os.path.basename(file_name))
```
seanpearsonuk, line 173: I'm curious about how this line works. If I have some `file_name` that specifies a file that can be at any path on the local machine, and I just strip it down to its base name and ask for the file of that name to be uploaded, what's actually happening there?

seanpearsonuk:
> We will merge this PR after refactoring existing code.

@hpohekar 

the simplesr way forward is to separate out the upload/download decision logic (**SHARED**) from the Fluent API calls (**NOT SHARED**).

```
def wrapped_solver_api_thing():
      shared_before_stuff()
      solver_api_thing()
      shared_after_stuff()

def wrapped_meshing_api_thing():
      shared_before_stuff()
      meshing_api_thing()
      shared_after_stuff()
```

There are more elaborate approaches, but the above is all you need here.

src/ansys/fluent/core/session_meshing.py
hpohekar: It is required due to following logic used in `client.py`.

image

If we pass `file_name` as complete file path to `self.file_service.file_exist()` , then it shows that file does not exist on the server.

Therefore we are reducing it to base name.

src/ansys/fluent/core/session_meshing.py
hpohekar: Right, I will update this change. Thank you.

hpohekar:
Closing due to 3.10.9 doc formatter issue.

PR #2138, opened 2023-10-13
Seamless File Handling - Ansys Lab

hpohekar:
- [x]  Design primary interface.
- [x] Test in the Ansys Lab.
- [x] Refactor code - removed duplication.
- [x] Again tested in the Ansys Lab.
- [x] Create a separate module `FileTransferService`.
- [x] Separated pypim upload and download methods. 
- [x] Fix a regression for local sessions.
- [x] Separate upload and download methods. 
- [x] Integrate remote file handler.
- [x] Test in the Ansys Lab 
- [x] Test local sessions.

We were doing following steps to read and write a case file in the Ansys Lab

```
meshing.upload(file_path="read/elbow.cas.h5", remote_file_name="uploaded_elbow.cas.h5")
meshing.tui.file.read_case("uploaded_elbow.cas.h5")
```

```
meshing.tui.file.write_case("written_sample_elbow_tui.cas.h5")
meshing.download(file_name="written_sample_elbow_tui.cas.h5", local_file_path=".")
```

Now we are handling upload and download part internally. So, we just need to execute following steps to read and write a case file.

```
meshing.read_case("read/elbow.cas.h5")
meshing.write_case("write_meshing_sample_elbow.cas.h5")
```
 
 *  `upload` to upload files if pypim is configured.
 *  `download` to download files if pypim is configured.

image

`----------------------------------------------------`

image

image

image

image


src/ansys/fluent/core/session.py
```diff
@@ -343,8 +306,93 @@ def download(self, file_name: str, local_file_path: Optional[str] = None):
...
+            either session.tui or session.file
+        Raises
+        ------
+        FileNotFoundError
+            If a case file does not exist.
+        """
+        file_service = self._file_service()
+        if os.path.isfile(file_name):
+            if not file_service.file_exist(os.path.basename(file_name)):
+                self.upload(os.path.basename(file_name))
```
seanpearsonuk, line 358: Hi @hpohekar I just want to double-check this logic.

We check if file with given path exists locally 
Then (after the remote check), we upload by base name only, i.e. stripping off the path information. How can `upload()` know where the file resides locally?

Am I understanding correctly?

hpohekar: @seanpearsonuk Yes, got it. In the Ansys Lab I have `elbow.cas.h5` in both current directory and in `read/elbow.cas.h5` directory. So every time it is taking `elbow.cas.h5` from current directory after stripping off the path information.

It should be `self.upload(file_name)` and `file_name` can be absolute or relative path.

Thank you very much, for pointing this out.

I will update this change.

hpohekar:
No idea, why following tests are failing ? (for 24R1)

This PR does not involve changes related to these tests.

image

src/ansys/fluent/core/session.py
hpohekar: Done and verified. Thank you.

src/ansys/fluent/core/session.py
```diff
@@ -258,60 +259,21 @@ def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any):
...
-        Download a file from the server.
-    """
-
-    def __init__(self, pim_instance):
-        self.pim_instance = pim_instance
-        self.file_service = None
+    def _file_service(self):
+        """File upload and download service."""
+        pim_instance = self.fluent_connection._remote_instance
+        file_service = None
```
seanpearsonuk, line 265: @hpohekar Should we be caching this file service or executing all this code each time?

src/ansys/fluent/core/session.py
```diff
@@ -343,8 +306,93 @@ def download(self, file_name: str, local_file_path: Optional[str] = None):
...
+                self.upload(file_name)
+                self._wait_for_file(file_name)
+        elif file_service.file_exist(os.path.basename(file_name)):
+            pass
+        else:
+            raise FileNotFoundError(f"{file_name} does not exist.")
+        if is_meshing:
+            api(os.path.basename(file_name))
+        else:
+            api(file_name=os.path.basename(file_name))
```
seanpearsonuk, line 367: @hpohekar 

This part of the design prevents this utility from being reusable. 

Any differences between clients of the utility need to be placed _behind the abstraction_.
So, it means that we need each client to pass in a simple callable, and if one client says _I need to use a named argument_ then he must achieve that inside his own implementation. You are almost there - you just need this one final adjustment!

src/ansys/fluent/core/session.py
```diff
@@ -343,8 +306,93 @@ def download(self, file_name: str, local_file_path: Optional[str] = None):
...
+        file_name : str
+            Case file name
+        Raises
+        ------
+        FileNotFoundError
+            If a case file does not exist.
+        """
+        file_service = self._file_service()
+        start_time = time.time()
+        max_wait_time = 100
```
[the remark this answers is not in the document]
hpohekar: I think another work around for this is to remove time check and just check for `file_service.file_exist()` until case file gets ready for reading or writing.

src/ansys/fluent/core/session.py
hpohekar: No. Right, `_file_service` is called twice. (in upload and download separately). We can combine it in one method and we can handle upload and download by passing an argument then we will call this single method in `session.upload()` and `session.download()` methods. I got your point, I will update this soon. Thank you.

src/ansys/fluent/core/session.py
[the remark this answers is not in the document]
seanpearsonuk: Yes, we shall get rid of the time limit. The "user" will have to fix any "problems" if the function does not return.

src/ansys/fluent/core/session.py
hpohekar: Ok. I will try to handle this. Thank you.

src/ansys/fluent/core/session.py
hpohekar: Implemented `_upload_download_helper`. We are going to do similar implementation for `_pypim_upload_download_helper`.

src/ansys/fluent/core/session.py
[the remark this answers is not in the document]
hpohekar: only check for `file_service.file_exist()` is not working. We can use `sys.maxsize` instead of `max_wait_time = 100`.

image

src/ansys/fluent/core/session.py
hpohekar: moved `_file_service` to `__init__` and we are creating it only once and using in all required methods.

src/ansys/fluent/core/session.py
hpohekar: ok.

src/ansys/fluent/core/session.py
```diff
@@ -343,8 +346,82 @@ def download(self, file_name: str, local_file_name: Optional[str] = None):
...
+            max_wait_time -= 1
+            time.sleep(3)
+        else:
+            raise FileNotFoundError(f"{file_name} does not exist.")
+
+    def _pypim_upload_download_helper(
+        self,
+        is_upload: bool,
+        file_name: str,
+        api: Optional[Any] = None,
```
seanpearsonuk, line 379: 👍 for making this optional. Since it can be None, we should check it before calling it.

Can we separate out the upload and download helpers:
```
def _pypim_upload(
        self,
        file_name: str,
        on_uploaded: Optional[Any] = None):
    ...

Comment (hpohekar) [reply src/ansys/fluent/core/session.py]: * Added check for `on_uploaded`.
* Separated pypim upload and download.

Thank you.

File: src/ansys/fluent/core/session.py [hunk 6]
```diff
@@ -343,8 +346,82 @@ def download(self, file_name: str, local_file_name: Optional[str] = None):
...
+        """Uploads a file if not available on the server.
+
+        Parameters
+        ----------
+        is_upload: bool
+            True if pypim is configured, False otherwise
+        file_name : str
+            File name
+        api: Session object property
+            either ``session.tui`` or ``session.file``
```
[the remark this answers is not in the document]
hpohekar: Done. Thank you.

src/ansys/fluent/core/session_pure_meshing.py
```diff
@@ -147,3 +147,45 @@ def transfer_mesh_to_solvers(
...
+
+        Parameters
+        ----------
+        file_name : str
+            Case file name
+        """
+        if pypim.is_configured():
+            self._pypim_upload(
+                file_name=file_name,
+                on_uploaded=self.tui.file.read_case,
```
hpohekar, line 165: Updated `on_uploaded = (lambda file_names: self.tui.file.read_case(file_names) -> self.tui.file.read_case` (for meshing)

seanpearsonuk: This looks good 👍

src/ansys/fluent/core/session_solver.py
```diff
@@ -228,3 +230,47 @@ def read_case(self, file_name: str, lightweight_mode: bool = False):
...
+        """Writes a case file.
+
+        Parameters
+        ----------
+        file_name : str
+            Case file name
+        """
+        self._no_pypim_helper(
+            file_name,
+            api=(lambda file_names: self.file.write_case(file_name=file_names)),
```
seanpearsonuk, line 271: Interesting that you made this plural - does this support multiple files?

hpohekar: if we use singular then it shadows `file_name` from outer scope, hence we have made it singular. Another option was to use `filename` but we are not using it now.

hpohekar: Any other suggestion will be helpful for this. I also don't want to use plural.

src/ansys/fluent/core/session.py
hpohekar: * Called callback before the download.

src/ansys/fluent/core/session.py
```diff
@@ -343,8 +346,87 @@ def download(self, file_name: str, local_file_name: Optional[str] = None):
...
+        """
+        start_time = time.time()
+        max_wait_time = sys.maxsize
+        while (time.time() - start_time) < max_wait_time:
+            if self.file_service.file_exist(os.path.basename(file_name)):
+                break
+            max_wait_time -= 1
+            time.sleep(3)
+        else:
+            raise FileNotFoundError(f"{file_name} does not exist.")
```
raph-luc, line 373: `max_wait_time = sys.maxsize` is basically the same as an infinite amount of time, is that correct? 

In that case, the `raise FileNotFoundError(f"{file_name} does not exist.")` below is realistically never going to be raised.

Assuming that waiting for up to one minute should be more than enough, my suggestion would be something like:
```suggestion
        from ansys.fluent.core.utils.execution import timeout_loop
        file_exists = timeout_loop(
            self.file_service.file_exist, 
            timeout=60, 
            args=(os.path.basename(file_name),),
            idle_period=3
        )
        if not file_exists:
            raise FileNotFoundError(f"{file_name} does not exist.")
```
Which will check if the file exists every three seconds until `file_exist` returns true (or fail if timeout of 60 seconds is reached).

hpohekar: It is the same case whether we use current implementation or `timeout_loop`. We don't know what is the ideal time limit to upload or download file due to following uncertainties. Sean has already mentioned this. 

1. Internet Speed
2. Size of file

We will also use `sys.maxsize` in `timeout_loop` due to above mentioned uncertainties. (60 seconds is not suitable, 100 was prior to this, but we changed it to `sys.maxsize`)

raph-luc: Do we plan on changing the max wait time from `sys.maxsize` to something that can realistically ever be reached? Otherwise there is no point to including `raise FileNotFoundError(f"{file_name} does not exist.")` here, as it is effectively unreachable.

And yes, `timeout_loop` is just a convenience function that we already have that achieves what you are adding here. If you want to use this code block rather than that function, that is fair, but why `time.sleep(3)` and then `max_wait_time -= 1`? Wouldn't it make more sense to reduce the wait time by the same amount of time that you set it to sleep for?

src/ansys/fluent/core/session.py
```diff
@@ -343,8 +346,87 @@ def download(self, file_name: str, local_file_name: Optional[str] = None):
...
+            Read a file.
+        Raises
+        ------
+        FileNotFoundError
+            If a file does not exist.
+        """
+        if os.path.isfile(file_name):
+            if not self.file_service.file_exist(os.path.basename(file_name)):
+                self.upload(file_name)
+                self._wait_for_file(file_name)
```
mkundu1, line 392: I'm wondering why the `_wait_for_file` call is required. Is the `upload` function asynchronous, i.e. it immediately returns before the file-upload is completed?

src/ansys/fluent/core/session.py
seanpearsonuk: @raph-luc we decided not to bother with a finite upper limit since we cannot quantify it.

seanpearsonuk: but you're correct about the unreachable exception

src/ansys/fluent/core/session.py
```diff
@@ -343,8 +346,87 @@ def download(self, file_name: str, local_file_name: Optional[str] = None):
...
+        """Used if pypim is not configured.
+
+        Parameters
+        ----------
+        file_name : str
+            File name
+        api: Callable[str]
+            Read or write a file.
+        """
+        api(file_name)
```
seanpearsonuk, line 432: - We don't seem to need any _no_pypim code as it's simply calling the callback
- As discussed, please update the callback names (not `api` but `on_uploaded` etc)

src/ansys/fluent/core/session.py
hpohekar: @raph-luc yes we can update this code according to above suggestion. So it will avoid dependency on time limit. Thank you.

src/ansys/fluent/core/session_solver.py
```diff
@@ -204,7 +206,7 @@ def _sync_from_future(self, fut: Future):
             except Exception:
                 fut_session.exit()
 
-    def read_case(self, file_name: str, lightweight_mode: bool = False):
+    def read_case_lightweight(self, file_name: str, lightweight_mode: bool = False):
```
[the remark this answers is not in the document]
hpohekar: We can set `lightweight_mode` to `True`, but we are not going to integrate this function into `read_case`.

src/ansys/fluent/core/session.py
hpohekar: @raph-luc  `but why time.sleep(3) and then max_wait_time -= 1?`

It's ok to have `time.sleep(1)` and `max_wait_time -= 1` for small amount of time like 60 seconds, but in current case it will be much more than 60 seconds. In Ansys Lab we don't want to query status of file for every second therefore we have made it delay of 3 seconds. There is no standard relationship between `time.sleep()` and decrement operator, we are using it according to our purpose.

src/ansys/fluent/core/session.py
```diff
@@ -343,8 +346,87 @@ def download(self, file_name: str, local_file_name: Optional[str] = None):
...
+        Parameters
+        ----------
+        file_name : str
+            File name
+        api: Callable[str]
+            Write a file.
+        """
+        if api:
+            api(os.path.basename(file_name))
+        self._wait_for_file(file_name)
```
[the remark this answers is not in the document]
hpohekar: WE / pyfluent developers are uploading/downloading the file on the server internally, because it's dependency for Ansys Lab. File needs to be uploaded before reading and downloaded after writing. We can't do read/download immediately, we need to wait until file gets ready.

src/ansys/fluent/core/session.py
hpohekar: * Removed `_no_pypim` code.
*  Renamed `api` to `on_uploaded` 

Thank you.

src/ansys/fluent/core/session.py
```diff
@@ -343,8 +346,64 @@ def download(self, file_name: str, local_file_name: Optional[str] = None):
...
+
+        Parameters
+        ----------
+        file_name : str
+            File name
+        """
+        while True:
+            if self.file_service.file_exist(os.path.basename(file_name)):
+                break
+            time.sleep(3)
```
hpohekar, line 364: @raph-luc This approach eliminating time constraint. Thank you.

src/ansys/fluent/core/session.py
hpohekar: @mkundu1 @seanpearsonuk  Removed `_wait_for_file`. Thank you.

I tested both scenario with and without `_wait_for_file`, both are working fine. Last time we have seen file does not exist error raised by `simple_upload_server` therefore we had added this code block, but it has been removed now.

src/ansys/fluent/core/session.py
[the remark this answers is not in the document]
hpohekar: It's removed now. Thank you.

src/ansys/fluent/core/session.py
```diff
@@ -343,8 +345,49 @@ def download(self, file_name: str, local_file_name: Optional[str] = None):
...
+            If a file does not exist.
+        """
+        if pypim.is_configured():
+            if os.path.isfile(file_name):
+                if not self.file_service.file_exist(os.path.basename(file_name)):
+                    self.upload(file_name)
+            elif self.file_service.file_exist(os.path.basename(file_name)):
+                pass
+            else:
+                raise FileNotFoundError(f"{file_name} does not exist.")
```
seanpearsonuk, line 373: rather than 
```
if ...:
  ...
elif ...:
  pass
else:
  ...
```
prefer:
```
if ...:
  ...
elif not ...:
  ...
```

src/ansys/fluent/core/session.py
```diff
@@ -258,75 +272,63 @@ def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any):
...
-    def __init__(self, pim_instance):
-        self.pim_instance = pim_instance
-        self.file_service = None
-        try:
-            upload_server = self.pim_instance.services["http-simple-upload-server"]
-        except (AttributeError, KeyError):
-            pass
-        else:
-            from simple_upload_server.client import Client
+    def _upload_download_helper(
```
seanpearsonuk, line 275: @hpohekar 
There are too many responsibilities being managed in this module.

Rather than merging the uploader code with new code, it would be better to keep them separate. The Uploader doesn't even need to be in this module. It could have been retained as it was and moved into a new module. Then your new functionality could have been built on that - separately. This session class would simply be a user of that.

hpohekar: Ok. We can create a separate file `uploader.py` will move `_Uploader` class to it.

src/ansys/fluent/core/session.py
hpohekar: It's not clear to me about `elif not ...`;

hpohekar: We don't have `elif not` scenario in above code block.

seanpearsonuk: I see it. Its equivalent.

hpohekar: Current code implemented according to following discussion.

image

hpohekar: We can do following change. 

```
if self.file_service.file_exist(os.path.basename(file_name)):
	pass
elif not self.file_service.file_exist(os.path.basename(file_name)):
	try:
		if os.path.isfile(file_name):
			self.upload(file_name)
	except FileNotFoundError as e:
		print(e.message)
else:
	raise FileNotFoundError(f"{file_name} does not exist.")
```
 OR
 
```
 if self.file_service.file_exist(os.path.basename(file_name)):
	pass
elif not self.file_service.file_exist(os.path.basename(file_name)):
	if os.path.isfile(file_name):
		self.upload(file_name)
	else:
		raise FileNotFoundError(f"{file_name} does not exist.")
else:
	raise FileNotFoundError(f"{file_name} does not exist.")
```

We need to raise an error or exception 2 times if we follow your suggestion and I think existing code handles this already.

@seanpearsonuk Could you please explain how your suggestion improves current code if it is equivalent, if I missing any point in existing code ?

src/ansys/fluent/core/session.py
hpohekar: Done. Thank you.

raph-luc:
I used the latest version of this branch and `meshing.read_case("read/elbow.cas.h5")`, but it seems that it isn't working in Ansys Lab, I get this traceback:
```
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
Input In [9], in <cell line: 1>()
----> 1 meshing.read_case("read/elbow.cas.h5")

File /opt/conda/lib/python3.9/site-packages/ansys/fluent/core/session_pure_meshing.py:162, in PureMeshing.read_case(self, file_name)
    150 def read_case(
    151     self,
    152     file_name: str,
    153 ):
    154     """Reads a case file.
    155 
    156     Parameters
   (...)
    159         Case file name
    160     """
--> 162     self._pypim_upload(
    163         file_name=file_name,
    164         on_uploaded=self.tui.file.read_case,
    165     )

File /opt/conda/lib/python3.9/site-packages/ansys/fluent/core/session.py:271, in BaseSession._pypim_upload(self, file_name, on_uploaded)
    269 def _pypim_upload(self, file_name: str, on_uploaded: Optional[Any] = None):
    270     """If pypim is configured, uploads a file on the server."""
--> 271     return self._uploader._pypim_upload(file_name, on_uploaded)

File /opt/conda/lib/python3.9/site-packages/ansys/fluent/core/utils/uploader.py:147, in _Uploader._pypim_upload(self, file_name, on_uploaded)
    145         pass
    146     else:
--> 147         raise FileNotFoundError(f"{file_name} does not exist.")
    148 if on_uploaded:
    149     on_uploaded(os.path.basename(file_name))

FileNotFoundError: read/elbow.cas.h5 does not exist.
```

raph-luc (commented):
These `<solver/meshing>.read_case()` and `.write_case()` methods are only going to work in Ansys Lab or when PyPIM is configured, if I'm understanding this correctly. 

Do we want these commands to work for local PyFluent sessions as well? For example, as a user, I would expect that the `solver.read_case(file_name)` command would do the same that a `solver.file.read(file_type="case", file_name=file_name)` command would do.

They are named as general commands that are not specific to Ansys Lab, and the docstrings don't mention this limitation either.

src/ansys/fluent/core/utils/uploader.py
```diff
@@ -0,0 +1,167 @@
+import os
+from typing import Any, Callable, Optional  # noqa: F401
+
+import ansys.platform.instancemanagement as pypim
+
+
+class _Uploader:
```
seanpearsonuk, line 7: Since this does upload and download, we can call it something like `FileTransferService` and rename the filename accordingly.

hpohekar:
> I used the latest version of this branch and `meshing.read_case("read/elbow.cas.h5")` with the other files you provided, but it seems that it isn't working in Ansys Lab, I get this traceback:
> 
> ```
> ---------------------------------------------------------------------------
> FileNotFoundError                         Traceback (most recent call last)
> Input In [9], in <cell line: 1>()
> ----> 1 meshing.read_case("read/elbow.cas.h5")
> 
> File /opt/conda/lib/python3.9/site-packages/ansys/fluent/core/session_pure_meshing.py:162, in PureMeshing.read_case(self, file_name)
>     150 def read_case(
>     151     self,
>     152     file_name: str,
>     153 ):
>     154     """Reads a case file.
>     155 
>     156     Parameters
>    (...)
>     159         Case file name
>     160     """
> --> 162     self._pypim_upload(
>     163         file_name=file_name,
>     164         on_uploaded=self.tui.file.read_case,
>     165     )
> 
> File /opt/conda/lib/python3.9/site-packages/ansys/fluent/core/session.py:271, in BaseSession._pypim_upload(self, file_name, on_uploaded)
>     269 def _pypim_upload(self, file_name: str, on_uploaded: Optional[Any] = None):
>     270     """If pypim is configured, uploads a file on the server."""
> --> 271     return self._uploader._pypim_upload(file_name, on_uploaded)
> 
> File /opt/conda/lib/python3.9/site-packages/ansys/fluent/core/utils/uploader.py:147, in _Uploader._pypim_upload(self, file_name, on_uploaded)
>     145         pass
>     146     else:
> --> 147         raise FileNotFoundError(f"{file_name} does not exist.")
>     148 if on_uploaded:
>     149     on_uploaded(os.path.basename(file_name))
> 
> FileNotFoundError: read/elbow.cas.h5 does not exist.
> ```
> 
> Edit: never mind this, I had misunderstood what this was supposed to be doing, the error was just because the file was not in the correct directory in Ansys Lab. `meshing.read_case("elbow.cas.h5")` works. Thanks for providing the files to be able to test this too

@raph-luc Thanks for testing it.

hpohekar:
> These `<solver/meshing>.read_case()` and `.write_case()` methods are only going to work in Ansys Lab or when PyPIM is configured, if I'm understanding this correctly.
> 
> Do we want these commands to work for local PyFluent sessions as well? For example, as a user, I would expect that the `solver.read_case(file_name)` command would do the same that a `solver.file.read(file_type="case", file_name=file_name)` command would do.
> 
> They are named as general commands that are not specific to Ansys Lab, and the docstrings don't mention this limitation either.

These methods will work for both local PyFluent sessions and Ansys Lab. We are handling it on basis of whether PyPIM is configured or not.

src/ansys/fluent/core/utils/uploader.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/session.py
```diff
@@ -111,7 +111,9 @@ def build_from_fluent_connection(self, fluent_connection: FluentConnection):
         self.error_state = self.fluent_connection.error_state
         self.scheme_eval = self.fluent_connection.scheme_eval
         self.rp_vars = RPVars(self.scheme_eval.string_eval)
-        self._uploader = None
+        self._file_transfer_service = FileTransferService(
```
seanpearsonuk, line 114: It would be really nice if this service were actually passed in to the constructor. We can leave that as a separate piece of work because we can look at the other dependencies too.

src/ansys/fluent/core/utils/file_transfer_service.py
```diff
@@ -0,0 +1,167 @@
...
+                        expanded_file_path
+                    )
+                    self.file_service.upload_file(expanded_file_path, upload_file_name)
+                else:
+                    raise FileNotFoundError(f"{file_name} does not exist.")
+            else:
+                if self.file_service.file_exist(file_name):
+                    self.file_service.download_file(file_name, local_file_name)
+                else:
+                    raise FileNotFoundError("Remote file does not exist.")
```
seanpearsonuk, line 87: rather than having a helper with conditional logic, implement as two methods

raph-luc:
> These methods will work for both local PyFluent sessions and Ansys Lab. We are handling it on basis of whether PyPIM is configured or not.

@hpohekar You are right, sorry for my confusion: I had misunderstood what `_pypim_upload()` does. The docstring and function name are misleading in my opinion, it only uploads using PyPIM if it is configured, if it is not then it just executes the callback defined (going to make a separate suggestion regarding this)

src/ansys/fluent/core/utils/file_transfer_service.py
```diff
@@ -0,0 +1,165 @@
...
+        Raises
+        ------
+        FileNotFoundError
+            If the remote file does not exist.
+        """
+        self._upload_download_helper(
+            is_upload=False, file_name=file_name, local_file_name=local_file_name
+        )
+
+    def _pypim_upload(self, file_name: str, on_uploaded: Optional[Any] = None):
```
raph-luc, line 126: I think the function name `"_pypim_upload"`, the argument name `"on_uploaded"` and the docstring are misleading: this function does not necessarily upload files using the PIM API, `on_uploaded` callback is executed even when the upload does not occur, and the docstring does not properly describe the function behavior or the arguments. As set up here, `on_uploaded` does not necessarily read a file, it is just a callback that will do something, usually read or write, with the file name, depending on how `_pypim_upload` is called. 

Quick name suggestions for you to consider would be, for example:
- `_pypim_upload` -> `_upload_if_configured`
- `on_uploaded` -> `callback` 

Also a quick question and suggestion: is using a callback a good idea here? The callback isn't very helpful as it can be replaced in less characters and lines than you have currently in the function callback definition and where the function is called, for example under `read_case` it could be simply:
```
self._pypim_upload(file_name)
self.file.read_case(file_name=file_name)
```
rather than what you suggest currently:
```
self._pypim_upload(
    file_name=file_name,
    on_uploaded=self.file.read_case,
)
```

hpohekar:
> > These methods will work for both local PyFluent sessions and Ansys Lab. We are handling it on basis of whether PyPIM is configured or not.
> 
> @hpohekar You are right, sorry for my confusion: I had misunderstood what `_pypim_upload()` does. The docstring and function name are misleading in my opinion, it only uploads using PyPIM if it is configured, if it is not then it just executes the callback defined (going to make a separate suggestion regarding this)

@raph-luc No worries at all. Your suggestion will be helpful. Thank you.

src/ansys/fluent/core/utils/file_transfer_service.py
raph-luc: @seanpearsonuk That is a good point, in that case to reproduce the current behavior I believe it would be:
```
self._pypim_upload(file_name)
self.file.read_case(file_name=Path(file_name).name)
```
if using `pathlib`, else, as is currently done here
```
self._pypim_upload(file_name)
self.file.read_case(file_name=os.path.basename(file_name))
```

(Unless other file path/name processing is being done that I missed)

This is not a strong opinion from my side either, just a thought that occurred to me and I think could be considered, if it wasn't already

raph-luc (commented):
Related to my previous points: the discussion with @seanpearsonuk made me realize that this `read_case()` implementation won't work with local/container Fluent sessions when specifying files not in the current working dir, i.e. inside folders or when providing an absolute path, as only the `basename` is being passed to the actual `read` command

seanpearsonuk:
> Related to my previous points: the discussion with @seanpearsonuk made me realize that this `read_case()` implementation won't work with local/container Fluent sessions when specifying files not in the current working dir, i.e. inside folders or when providing an absolute path, as only the `basename` is being passed to the actual `read` command

correct

src/ansys/fluent/core/utils/file_transfer_service.py
hpohekar: `on_uploaded` was suggested by @seanpearsonuk. Earlier it was `api` and `api` may be `tui` or `file`.

hpohekar: The class docstring already mentions about behavior but if it is not sufficient then may be we can update docstrings.

hpohekar: is using a callback a good idea here? - I have added it considering point of reusability in other methods like `read_mesh` etc..

So for `read_mesh` we just need to change callback from `self.file.read_case` to `self.file.read_mesh`

hpohekar:
> Related to my previous points: the discussion with @seanpearsonuk made me realize that this `read_case()` implementation won't work with local/container Fluent sessions when specifying files not in the current working dir, i.e. inside folders or when providing an absolute path, as only the `basename` is being passed to the actual `read` command

Yes, I verified and your point is correct. I just realized that after refactoring I need to update file_name passed to `on_uploaded` callback.

src/ansys/fluent/core/utils/file_transfer_service.py
hpohekar: `Pathlib` approach looks good to me, we can integrate it current code but I just want to avoid separate line like `self.file.read_case(file_name=Path(file_name).name)` in each method.

I'm thinking we can do following change right ?

```
        if on_uploaded:
            on_uploaded(Path(file_name).name)
```

src/ansys/fluent/core/session.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/utils/file_transfer_service.py

src/ansys/fluent/core/session.py
```diff
@@ -103,15 +106,24 @@ def __init__(self, fluent_connection: FluentConnection):
         Args:
             fluent_connection (:ref:`ref_fluent_connection`): Encapsulates a Fluent connection.
         """
-        BaseSession.build_from_fluent_connection(self, fluent_connection)
+        remote_file_handler = RemoteFileHandler(
+            transfer_service=PimFileTransferService(fluent_connection._remote_instance)
```
seanpearsonuk, line 110: As commented previously, any dependency on a file transfer services will be passed in. Follow-up work, no action required.

src/ansys/fluent/core/session.py
```diff
@@ -103,15 +106,24 @@ def __init__(self, fluent_connection: FluentConnection):
...
+        )
+        BaseSession.build_from_fluent_connection(
+            self, fluent_connection, remote_file_handler
+        )
```
seanpearsonuk, line 114: This is mysterious

src/ansys/fluent/core/session.py
```diff
@@ -103,15 +106,24 @@ def __init__(self, fluent_connection: FluentConnection):
...
 
-    def build_from_fluent_connection(self, fluent_connection: FluentConnection):
+    def build_from_fluent_connection(
```
seanpearsonuk, line 116: This is not commented. I don't understand the rationale without some explanation.

hpohekar:
<Local solver session testing>

```
>>> from ansys.fluent.core import examples                                                    
>>> import_file_name = examples.download_file("mixing_elbow.msh.h5", "pyfluent/mixing_elbow")
Checking if specified file already exists...
File already exists. File path:
C:\Users\hpohekar\AppData\Local\Ansys\ansys_fluent_core\examples\mixing_elbow.msh.h5
>>> solver = launch_fluent()
>>> solver.read_case(import_file_name)
pyfluent.settings_api WARNING: Mismatch between generated file and server object info. Dynamically created settings classes will be used.
Fast-loading "D:\ANSYSDev\vNNN\fluent\fluent24.1\\addons\afd\lib\hdfio.bin"
Done.

Reading from AAPQ13pUvWxQtxN:"C:\Users\hpohekar\AppData\Local\Ansys\ansys_fluent_core\examples\mixing_elbow.msh.h5" in NODE0 mode ...
  Reading mesh ...
       66417 nodes,     3 node zones ...

Building...
     mesh
     materials,
     interface,
     domains,
     zones,
        Skipping thread 20 of domain 1 (not referenced by grid).
        Skipping thread 21 of domain 1 (not referenced by grid).
        Skipping thread 22 of domain 1 (not referenced by grid).
        Skipping thread 23 of domain 1 (not referenced by grid).
        Skipping thread 24 of domain 1 (not referenced by grid).
        Skipping thread 25 of domain 1 (not referenced by grid).
        Skipping thread 26 of domain 1 (not referenced by grid).
        Skipping thread 27 of domain 1 (not referenced by grid).
        Skipping thread 28 of domain 1 (not referenced by grid).
        wall-elbow
        wall-inlet
        outlet
        cold-inlet
        hot-inlet
        symmetry-xyplane
        interior--elbow-fluid
        elbow-fluid
     parallel,
Done.
Mesh is now scaled to meters.
>>> solver.read_case("mixing_elbow.msh.h5")
Warning: reading 2 partition grid onto 1 compute node machine.
         Combining every 2 partitions.
  Done.Multicore SMT processors detected. Processor affinity set!

       Clearing partially read grid.

Reading from AAPQ13pUvWxQtxN:"D:\hackathon\pyfluent\mixing_elbow.msh.h5" in NODE0 mode ...
  Reading mesh ...
       66417 nodes,     3 node zones ...

Building...
     mesh
     materials,
     interface,
     domains,
     zones,
        Skipping thread 20 of domain 1 (not referenced by grid).
        Skipping thread 21 of domain 1 (not referenced by grid).
        Skipping thread 22 of domain 1 (not referenced by grid).
        Skipping thread 23 of domain 1 (not referenced by grid).
        Skipping thread 24 of domain 1 (not referenced by grid).
        Skipping thread 25 of domain 1 (not referenced by grid).
        Skipping thread 26 of domain 1 (not referenced by grid).
        Skipping thread 27 of domain 1 (not referenced by grid).
        Skipping thread 28 of domain 1 (not referenced by grid).
        wall-elbow
        wall-inlet
        outlet
        cold-inlet
        hot-inlet
        symmetry-xyplane
        interior--elbow-fluid
        elbow-fluid
     parallel,
Done.
Mesh is now scaled to meters.
>>> solver.write_case("latest_zzz.cas.h5")

Writing to AAPQ13pUvWxQtxN:"D:\hackathon\pyfluent\latest_zzz.cas.h5" in NODE0 mode and compression level 1 ...
Grouping cells for Laplace smoothing ...
Warning: reading 2 partition grid onto 1 compute node machine.
         Combining every 2 partitions.
  Done.       17822 cells,     1 zone  ...
       91581 faces,     7 zones ...
       66417 nodes,     1 zone  ...
  Done.
Done.
>>> solver.write_case("latest_zzz.cas.h5")

Writing to AAPQ13pUvWxQtxN:"D:\hackathon\pyfluent\latest_zzz.cas.h5" in NODE0 mode and compression level 1 ...
Grouping cells for Laplace smoothing ...
       17822 cells,     1 zone  ...
       91581 faces,     7 zones ...
       66417 nodes,     1 zone  ...
  Done.
Done.
>>> solver.write_case("D:\hackathon\pyfluent\examples\latest_zzz.cas.h5")

Writing to AAPQ13pUvWxQtxN:"D:\hackathon\pyfluent\examples\latest_zzz.cas.h5" in NODE0 mode and compression level 1 ...
Grouping cells for Laplace smoothing ...
       17822 cells,     1 zone  ...
       91581 faces,     7 zones ...
       66417 nodes,     1 zone  ...
  Done.
Done.
```

</Local solver session testing>

hpohekar:
Local Meshing session testing.

<details>


```
>>> from ansys.fluent.core import examples                                                    
>>> import_file_name = examples.download_file("mixing_elbow.msh.h5", "pyfluent/mixing_elbow")
Checking if specified file already exists...
File already exists. File path:
C:\Users\hpohekar\AppData\Local\Ansys\ansys_fluent_core\examples\mixing_elbow.msh.h5
>>> meshing = launch_fluent(mode="meshing")                               
>>> meshing.read_case(import_file_name)
Reading "C:\Users\hpohekar\AppData\Local\Ansys\ansys_fluent_core\examples\mixing_elbow.msh.h5" ...
3D mesh
nodes: 66417
edges: 300
faces: 91581
cells: 17822
reading 3 node zones
reading 9 edge zones
reading 7 face zones
reading 1 cell zones
updating sub-domains...
done.
generating pointers...done.
extracting boundary entities...
 13264 boundary nodes.
 4799 boundary faces.
 6 boundary face zones.
done.
generating cells...done.
analyzing boundary connectivity...done.
>>> meshing.read_case("mixing_elbow.msh.h5")
Reading "mixing_elbow.msh.h5" ...
copy file from "mixing_elbow.msh.h5" to "C:\\Users\\hpohekar\\\\mixing_elbow.msh.h5" 
file copy done
3D mesh
nodes: 66417
edges: 300
faces: 91581
cells: 17822
reading 3 node zones
reading 9 edge zones
reading 7 face zones
reading 1 cell zones
updating sub-domains...
done.
generating pointers...done.
extracting boundary entities...
 13264 boundary nodes.
 4799 boundary faces.
 6 boundary face zones.
done.
generating cells...done.
analyzing boundary connectivity...done.
>>> meshing.write_case("latest_yyy.cas.h5")

The case file is being written in the mesher mode.
Previously saved data files are incompatible and should not be read with this case file.

Writing "latest_yyy.cas.h5" ...
writing 3 node zones
writing 7 face zones 
writing 1 cell zones 
writing boundary layer flags
writing node curvature data...
>>> meshing.write_case("D:\hackathon\pyfluent\examples\latest_yyy.cas.h5")
done.
The case file is being written in the mesher mode.
Previously saved data files are incompatible and should not be read with this case file.

Writing "D:\hackathon\pyfluent\examples\latest_yyy.cas.h5" ...
writing 3 node zones
writing 7 face zones 
writing 1 cell zones 
writing boundary layer flags
writing node curvature data...
>>> meshing.exit()                                                        
```


</details>

src/ansys/fluent/core/session.py
hpohekar: @seanpearsonuk 

I thought adding it here will avoid following duplication in 2 files. 

`session_pure_meshing.py` and `session_solver.py`

image

`----------------------------------------------`

image

src/ansys/fluent/core/session.py
hpohekar: As per above explanation.

src/ansys/fluent/core/session.py
hpohekar: It's moved to `__init__` of `BaseSession`.

src/ansys/fluent/core/session.py
seanpearsonuk: I agree that it's good to avoid duplication. I was just trying to understand the particular design changes here.

seanpearsonuk (approved):

raph-luc (approved):

PR #2139, opened 2023-10-13
Fix vale warnings

mkundu1:
These warnings started appearing in multiple PRs - https://github.com/ansys/pyfluent/actions/runs/6505185628/job/17668355789?pr=2136#step:6:63

Some of these warnings seem false-positive. Why does `Import geometry` raise a warning, but `Describe geometry` does not?

### DIFF
--- a/.vale.ini
+++ b/.vale.ini
@@ -31,3 +31,4 @@ BasedOnStyles = Vale, Google
 Google.WordList = NO
 Google.Colons = NO
 Google.OxfordComma = NO
+Google.Headings = NO
--- a/.github/styles/Vocab/ANSYS/accept.txt
+++ b/.github/styles/Vocab/ANSYS/accept.txt
@@ -1,9 +1,11 @@
 ANSYS
 Ansys
 ansys
+API
 batch_ops
 CaseFile
 DataFile
+CAD
 CommandArguments
 error_handler
 events
@@ -63,6 +65,7 @@ Streaming
 Session
 Slurm
 transcript
+TUI
 UDFs
 Univa
 venv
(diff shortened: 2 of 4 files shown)


raph-luc (approved):

mkundu1:
> Could it be that it wants us to write something like `Import the geometry`?

The rule in the google link mentioned in https://github.com/errata-ai/Google/blob/master/Google/Headings.yml. Funnily it shows the warning also for `Capitalization in titles and headings` from the google page.

mkundu1:
> > Could it be that it wants us to write something like `Import the geometry`?
> 
> The rule in the google link mentioned in https://github.com/errata-ai/Google/blob/master/Google/Headings.yml. Funnily it shows the warning also for `Capitalization in titles and headings` from the google page.

raph-luc:
@hpohekar That is a good point, not all of the warnings are false-positives, but there are way too many false-positives currently for us to keep it active. Even stuff like ` 'Journaling' should use sentence-style capitalization.`, what does it even expect us to do? I have no idea.

hpohekar:
> @hpohekar That is a good point, not all of the warnings are false-positives, but there are way too many false-positives currently for us to keep it active. Even stuff like ` 'Journaling' should use sentence-style capitalization.`, what does it even expect us to do? I have no idea.

Yes, I agree. No idea.

hpohekar (approved):

prmukherj (approved):

.vale.ini
```diff
@@ -30,4 +30,5 @@ BasedOnStyles = Vale, Google
 # Removing Google-specific rule - Not applicable under some circumstances
 Google.WordList = NO
 Google.Colons = NO
-Google.OxfordComma = NO
\ No newline at end of file
+Google.OxfordComma = NO
+Google.Headings = NO
```
prmukherj, line 34: Applied the same to UI repo, Thank you.

PR #2147, opened 2023-10-16
Fix set_state implementation for command argument instance.

prmukherj:


src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1132,6 +1132,17 @@ def get_state(self) -> Any:
 
     getState = get_state
 
+    def set_state(self, state) -> Any:
+        """Get state of the command argument."""
```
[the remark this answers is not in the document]
prmukherj: Thank you @raph-luc, good catch.

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1132,6 +1132,17 @@ def get_state(self) -> Any:
...
+        parent_state = self.parent.get_state()
+        try:
+            parent_state[self.name] = state
+            self.parent.set_state(parent_state)
+        except KeyError:
+            pass
```
raph-luc, line 1142: Question unrelated to this PR: why are we ignoring `KeyError` exceptions here? Should we at least log these?

prmukherj: I think we should.

cc. @mkundu1, @seanpearsonuk

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1132,6 +1132,17 @@ def get_state(self) -> Any:
 
     getState = get_state
 
+    def set_state(self, state) -> Any:
+        """Set state of the command argument."""
+        parent_state = self.parent.get_state()
+        try:
+            parent_state[self.name] = state
+            self.parent.set_state(parent_state)
+        except KeyError:
```
mkundu1, line 1141: In which scenario will we get a KeyError?

If the KeyError happens due to some user error, we should throw the appropriate exception. If it never happens (guaranteed by the implementation), we should remove the try-except.

src/ansys/fluent/core/services/datamodel_se.py
prmukherj: Done.

seanpearsonuk (approved):

hpohekar (approved):

raph-luc (approved):

mkundu1 (approved):

PR #2149, opened 2023-10-17
Expose settings root like in pyconsole.

prmukherj:
```
>>> solver = pf.launch_fluent(...)
>>> solver()
```
Solver object is callable now.

### DIFF
--- a/src/ansys/fluent/core/session_solver.py
+++ b/src/ansys/fluent/core/session_solver.py
@@ -228,3 +228,9 @@ class Solver(BaseSession):
             fut.add_done_callback(functools.partial(Solver._sync_from_future, self))
         else:
             self.file.read(file_type="case", file_name=file_name)
+
+    def __call__(self):
+        return self._root.get_state()
+
+    def __getattr__(self, attr):
+        return getattr(self._root, attr)
--- a/src/ansys/fluent/core/session_solver.py
+++ b/src/ansys/fluent/core/session_solver.py
@@ -231,6 +231,3 @@ class Solver(BaseSession):
 
     def __call__(self):
         return self._root.get_state()
-
-    def __getattr__(self, attr):
-        return getattr(self._root, attr)


src/ansys/fluent/core/session_solver.py
```diff
@@ -228,3 +228,9 @@ def read_case(self, file_name: str, lightweight_mode: bool = False):
             fut.add_done_callback(functools.partial(Solver._sync_from_future, self))
         else:
             self.file.read(file_type="case", file_name=file_name)
+
+    def __call__(self):
+        return self._root.get_state()
+
+    def __getattr__(self, attr):
```
prmukherj, line 235: @seanpearsonuk, @mkundu1, would this implementation cause any related issues issues?

seanpearsonuk: @prmukherj This stuff is always tricky. Bear in mind that we would also want the objects to show up in the `dir()` results. Is that the case. An alternative is to proactively add the attributes to this object, which might be most appropriate here anyway.

@mkundu1

prmukherj: Right @seanpearsonuk. Thank you. I think it is best to remove the get_attr for now.

prmukherj: @seanpearsonuk, shall we keep the get_attr as it is just as a back up so that nothing breaks if some new api is used?

mkundu1: I think it is better to remove `__getattr__`, `solver_sesssion.xyz` now goes to the server and hangs instead of throwing attribute error.

Note that, we want to load the settings root on demand. As the top-level child names cannot be known without querying the settings root, we cannot add the top-level children as attributes or properties dynamically. So, the explicit properties as currently implemented here are OK. We also have the docstrings and type-hints (session_solver.pyi) based on those properties.

prmukherj: ok, @mkundu1. Thank you. I also noticed the hang, actually it throws the attribute error after a long pause.

raph-luc (approved):

hpohekar (approved):

Issue #2098, opened 2023-10-06
Higher-level exception types


seanpearsonuk:
Design and use higher-level exceptions that specify the actual issue rather than emitting bland, lower-level built-in exception types.

E.g., fluent_container.py
```
Raises
    ------
    KeyError
        If license server is not specified through an environment variable or in ``container_dict``.
    ValueError
        If server info file is specified through both a command-line argument inside ``container_dict`` and the  ``container_server_info_file`` parameter.
    ValueError
        If ``fluent_image`` or ``image_tag`` and ``image_name`` are not specified.
```

raph-luc:
I think some nuance and discussion is required on which exceptions actually need custom types, and which ones will just end up being unnecessary additional lines of code that no one actually ever catches and handles. 

I agree that we overuse some exception types (like `RuntimeError`) and those should be changed to more appropriate types (not necessarily custom types). 

In general, sticking to the [built-in concrete exceptions](https://docs.python.org/3/library/exceptions.html) as much as possible seems to be the most efficient way to go about this, and makes the most sense, to me.

In the example above and based on the built-in exceptions documentation: 

- `KeyError` is raised "when a mapping (dictionary) key is not found in the set of existing keys", which is exactly the situation there. 

- `ValueError` is raised "when an operation or function receives an argument that has the right type but an inappropriate value", which IMO also accurately describes both situations there, as arguments have the right types but inappropriate values. Types can be `NoneType`, so it is not a `TypeError` when they aren't specified, but in these cases the values or value combinations are inappropriate. 

I don't think custom exceptions for these examples would be helpful. With that said, I do agree there may be exceptions types/classes that need to be re-evaluated in the code, and I think we should need an actual reason/use to add custom exceptions (i.e., say if we need to differentiate between exception types and catch a specific one that may be raised simultaneously to others, or the built-in exceptions just aren't appropriate at all).

seanpearsonuk:
So
```
try:
  # call high-level operation that defines no exception types
  some_high_level_operation()
# I know what low-level methods are called in the implementation and what they raise, and thneither will ever change
except ValiueError as ex:
  # I know all the various possibilities - and the details of the messages won't ever change.
  # Also, there's surely only one method that can raise a ValueError.
  if error_indicator_1 in str(ex):
     handle_error_indicator_1()
  if error_indicator_2 in str(ex):
     handle_error_indicator_2()
except KeyError:
  # etc
```

raph-luc:
[...]

PR #2159, opened 2023-10-19
Higher Level Exceptions

hpohekar:
* Adding higher level custom exceptions in `src\ansys\fluent\core\solver\error_message.py`.

* Updating respective tests for added exceptions.

### DIFF
--- a/src/ansys/fluent/core/file_session.py
+++ b/src/ansys/fluent/core/file_session.py
@@ -108,9 +108,9 @@ class Transaction:
 
         Raises
         ------
-        Transaction.SURFACE_NAMES_IDs_NOT_PROVIDED
+        TransactionError.SURFACE_NAMES_IDs_NOT_PROVIDED
             If surface names or surface ids are not provided.
-        Transaction.INVALID_FIELD_NAME_PREFIX
+        TransactionError.INVALID_FIELD_NAME_PREFIX
             If field name does not have prefix ``phase-`` for multi-phase cases.
         """
         if surface_ids is None:
@@ -160,9 +160,9 @@ class Transaction:
 
         Raises
         ------
-        Transaction.SURFACE_NAMES_IDs_NOT_PROVIDED
+        TransactionError.SURFACE_NAMES_IDs_NOT_PROVIDED
             If surface names or surface ids are not provided.
-        Transaction.INVALID_FIELD_NAME_PREFIX
+        TransactionError.INVALID_FIELD_NAME_PREFIX
             If field name does not have prefix ``phase-`` for multi-phase cases.
         """
         if surface_ids is None:
@@ -223,7 +223,7 @@ class Transaction:
 
         Raises
         ------
-        Transaction.INVALID_FIELD_NAME
+        TransactionError.INVALID_FIELD_NAME
             If any field other than ``"velocity"`` is provided.
         """
         mesh = self._file_session._case_file.get_mesh()
@@ -319,11 +319,11 @@ class FileFieldData:
 
         Raises
         ------
-        RuntimeError
+        TransactionError.SURFACE_NAMES_IDs_NOT_PROVIDED
             If surface names or surface ids are not provided.
         """
         if surface_ids and surface_name:
-            raise RuntimeError("Please provide either surface name or surface ids.")
+            raise TransactionError.SURFACE_NAMES_IDs_NOT_PROVIDED()
 
         if data_type == SurfaceDataType.Vertices:
             if surface_name:
@@ -403,15 +403,13 @@ class FileFieldData:
 
         Raises
         ------
-        RuntimeError
+        TransactionError.SURFACE_NAMES_IDs_NOT_PROVIDED
             If surface names or surface ids are not provided.
-        RuntimeError
-            If field name does not have prefix ``phase-`` for multi-phase cases.
-        RuntimeError
+        TransactionError.INVALID_FIELD_NAME_PREFIX
             If field name does not have prefix ``phase-`` for multi-phase cases.
         """
         if surface_ids and surface_name:
-            raise RuntimeError("Please provide either surface name or surface ids.")
+            raise TransactionError.SURFACE_NAMES_IDs_NOT_PROVIDED()
 
         if surface_name:
             surface_ids = self._field_info.get_surfaces_info()[surface_name][
@@ -419,9 +417,7 @@ class FileFieldData:
             ]
             if len(self._file_session._data_file.get_phases()) > 1:
                 if not field_name.startswith("phase-"):
-                    raise RuntimeError(
-                        "For multi-phase cases field name should have a prefix of phase name."
-                    )
+                    raise TransactionError.INVALID_FIELD_NAME_PREFIX()
                 return ScalarFieldData(
                     surface_ids[0],
                     self._file_session._data_file.get_face_scalar_field_data(
@@ -440,9 +436,7 @@ class FileFieldData:
         else:
             if len(self._file_session._data_file.get_phases()) > 1:
                 if not field_name.startswith("phase-"):
-                    raise RuntimeError(
-                        "For multi-phase cases field name should have a prefix of phase name."
-                    )
+                    raise TransactionError.INVALID_FIELD_NAME_PREFIX()
                 return {
                     surface_id: ScalarFieldData(
                         surface_id,
@@ -491,23 +485,21 @@ class FileFieldData:
 
         Raises
         ------
-        RuntimeError
+        TransactionError.SURFACE_NAMES_IDs_NOT_PROVIDED
             If surface names or surface ids are not provided.
-        RuntimeError
+        TransactionError.INVALID_FIELD_NAME
             If any field other than ``"velocity"`` is provided.
-        RuntimeError
-            If field name does not have prefix ``phase-`` for multi-phase cases.
-        RuntimeError
+        TransactionError.INVALID_FIELD_NAME_PREFIX
             If field name does not have prefix ``phase-`` for multi-phase cases.
         """
         if surface_ids and surface_name:
-            raise RuntimeError("Please provide either surface name or surface ids.")
+            raise TransactionError.SURFACE_NAMES_IDs_NOT_PROVIDED()
 
         if (
             field_name.lower() != "velocity"
             and field_name.split(":")[1].lower() != "velocity"
         ):
-            raise RuntimeError("Only 'velocity' is allowed field.")
+            raise TransactionError.INVALID_FIELD_NAME()
 
         if surface_name:
             surface_ids = self._field_info.get_surfaces_info()[surface_name][
@@ -515,9 +507,7 @@ class FileFieldData:
             ]
             if len(self._file_session._data_file.get_phases()) > 1:
                 if not field_name.startswith("phase-"):
-                    raise RuntimeError(
-                        "For multi-phase cases field name should have a prefix of phase name."
-                    )
+                    raise TransactionError.INVALID_FIELD_NAME_PREFIX()
                 vector_data = self._file_session._data_file.get_face_vector_field_data(
                     field_name.split(":")[0], surface_ids[0]
                 )
@@ -530,9 +520,7 @@ class FileFieldData:
         else:
             if len(self._file_session._data_file.get_phases()) > 1:
                 if not field_name.startswith("phase-"):
-                    raise RuntimeError(
-                        "For multi-phase cases field name should have a prefix of phase name."
-                    )
+                    raise TransactionError.INVALID_FIELD_NAME_PREFIX()
                 return {
                     surface_id: VectorFieldData(
                         surface_id,
--- a/src/ansys/fluent/core/file_session.py
+++ b/src/ansys/fluent/core/file_session.py
@@ -12,6 +12,7 @@ from ansys.fluent.core.services.field_data import (
     VectorFieldData,
     Vertices,
 )
+from ansys.fluent.core.solver.error_message import TransactionError
 
 
 class Transaction:
@@ -108,15 +109,17 @@ class Transaction:
 
         Raises
         ------
-        TransactionError.SURFACE_NAMES_IDs_NOT_PROVIDED
+        TransactionError.SurfaceNamesIDsNotProvidedError
             If surface names or surface ids are not provided.
-        TransactionError.INVALID_FIELD_NAME_PREFIX
+        TransactionError.InvalidFieldNamePrefixError
             If field name does not have prefix ``phase-`` for multi-phase cases.
         """
         if surface_ids is None:
             surface_ids = []
         if surface_ids and surface_names:
-            raise TransactionError.SURFACE_NAMES_IDs_NOT_PROVIDED()
+            raise TransactionError.SurfaceNamesIDsNotProvidedError(
+                "Please provide either surface names or surface ids."
+            )
 
         if surface_names:
             for surface_name in surface_names:
@@ -126,7 +129,9 @@ class Transaction:
 
         if len(self._file_session._data_file.get_phases()) > 1:
             if not field_name.startswith("phase-"):
-                raise TransactionError.INVALID_FIELD_NAME_PREFIX()
+                raise TransactionError.InvalidFieldNamePrefixError(
+                    "For multi-phase cases field name should have a prefix of phase name."
+                )
             self._scalar_field_transactions.append(
                 Transaction._ScalarFieldTransaction(
                     field_name, surface_ids, field_name.split(":")[0]
@@ -160,15 +165,17 @@ class Transaction:
 
         Raises
         ------
-        TransactionError.SURFACE_NAMES_IDs_NOT_PROVIDED
+        TransactionError.SurfaceNamesIDsNotProvidedError
             If surface names or surface ids are not provided.
-        TransactionError.INVALID_FIELD_NAME_PREFIX
+        TransactionError.InvalidFieldNamePrefixError
             If field name does not have prefix ``phase-`` for multi-phase cases.
         """
         if surface_ids is None:
             surface_ids = []
         if surface_ids and surface_names:
-            raise TransactionError.SURFACE_NAMES_IDs_NOT_PROVIDED()
+            raise TransactionError.SurfaceNamesIDsNotProvidedError(
+                "Please provide either surface names or surface ids."
+            )
 
         if surface_names:
             for surface_name in surface_names:
@@ -178,7 +185,9 @@ class Transaction:
 
         if len(self._file_session._data_file.get_phases()) > 1:
             if not field_name.startswith("phase-"):
-                raise TransactionError.INVALID_FIELD_NAME_PREFIX()
+                raise TransactionError.InvalidFieldNamePrefixError(
+                    "For multi-phase cases field name should have a prefix of phase name."
+                )
             self._vector_field_transactions.append(
                 Transaction._VectorFieldTransaction(
                     field_name, surface_ids, field_name.split(":")[0]
@@ -223,7 +232,7 @@ class Transaction:
 
         Raises
         ------
-        TransactionError.INVALID_FIELD_NAME
+        TransactionError.InvalidFieldNameError
             If any field other than ``"velocity"`` is provided.
         """
         mesh = self._file_session._case_file.get_mesh()
@@ -254,7 +263,9 @@ class Transaction:
 
         for transaction in self._vector_field_transactions:
             if "velocity" not in transaction.field_name:
-                raise TransactionError.INVALID_FIELD_NAME()
+                raise TransactionError.InvalidFieldNameError(
+                    "Only 'velocity' is allowed field."
+                )
             if vector_field_tag not in field_data:
                 field_data[vector_field_tag] = {}
             field_data_surface = field_data[vector_field_tag]
@@ -319,11 +330,13 @@ class FileFieldData:
 
         Raises
         ------
-        TransactionError.SURFACE_NAMES_IDs_NOT_PROVIDED
+        TransactionError.SurfaceNamesIDsNotProvidedError
             If surface names or surface ids are not provided.
         """
         if surface_ids and surface_name:
-            raise TransactionError.SURFACE_NAMES_IDs_NOT_PROVIDED()
+            raise TransactionError.SurfaceNamesIDsNotProvidedError(
+                "Please provide either surface names or surface ids."
+            )
 
         if data_type == SurfaceDataType.Vertices:
             if surface_name:
@@ -403,13 +416,15 @@ class FileFieldData:
 
         Raises
         ------
-        TransactionError.SURFACE_NAMES_IDs_NOT_PROVIDED
+        TransactionError.SurfaceNamesIDsNotProvidedError
             If surface names or surface ids are not provided.
-        TransactionError.INVALID_FIELD_NAME_PREFIX
+        TransactionError.InvalidFieldNamePrefixError
             If field name does not have prefix ``phase-`` for multi-phase cases.
         """
         if surface_ids and surface_name:
-            raise TransactionError.SURFACE_NAMES_IDs_NOT_PROVIDED()
+            raise TransactionError.SurfaceNamesIDsNotProvidedError(
+                "Please provide either surface names or surface ids."
+            )
 
         if surface_name:
             surface_ids = self._field_info.get_surfaces_info()[surface_name][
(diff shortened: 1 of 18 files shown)


src/ansys/fluent/core/fluent_connection.py
```diff
@@ -670,3 +666,20 @@ def _exit(
             remote_instance.delete()
 
         exit_event.set()
+
+
+class FluentConnectionError(ValueError):
```
seanpearsonuk, line 671: :+1: for the inheritance.

seanpearsonuk (approved):
:+1: for starting with a single example. This is exactly what we wanted.

src/ansys/fluent/core/fluent_connection.py
hpohekar: Right, Thank you.

src/ansys/fluent/core/file_session.py
```diff
@@ -724,3 +708,24 @@ def read_case(self, case_file_name):
...
+
+
+class TransactionError(RuntimeError):
+    """Custom transaction errors."""
+
+    def __init__(self, error):
+        super().__init__(error)
+
+    @classmethod
+    def SURFACE_NAMES_IDs_NOT_PROVIDED(cls):
```
mkundu1, line 720: At the consumer level, these exception objects are still of type TransactionError and the way to identify a SURFACE_NAMES_IDs_NOT_PROVIDED exception is to check for the error message. Do we want to refine the exception hierarchy further by deriving a SurfaceNotSpecified from TransactionError?

All the custom exception classes can be moved to a separate file, so user can find them in one place.

src/ansys/fluent/core/file_session.py
```diff
@@ -108,15 +108,15 @@ def add_scalar_fields_request(
 
         Raises
         ------
-        RuntimeError
+        TransactionError.SURFACE_NAMES_IDs_NOT_PROVIDED
             If surface names or surface ids are not provided.
```
raph-luc, line 112: `TransactionError.SURFACE_NAMES_IDs_NOT_PROVIDED` is not a exception type, it can't be caught using `except`. 

I believe it makes more sense to list here just `TransactionError` as that is the exception type that is actually being raised and that can be caught.

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -670,3 +666,20 @@ def _exit(
...
+    """Custom Fluent connection errors."""
+
+    def __init__(self, error):
+        super().__init__(error)
+
+    @classmethod
+    def PORT_NOT_PROVIDED(cls):
```
raph-luc, line 678: I'm unsure about the usefulness of creating new exception class methods for exception messages that are only raised in one location in the code, as seems to be the case with `PORT_NOT_PROVIDED`.

Regardless, if this is how we would like to structure things moving forward, I understand.

src/ansys/fluent/core/file_session.py
hpohekar: Until now, we are not catching exceptions, therefore we have not derived `TransactionError` from exception class.

src/ansys/fluent/core/file_session.py
hpohekar: Discussed with @mkundu1, we will refine the exception hierarchy. Thank you.

src/ansys/fluent/core/file_session.py
hpohekar: @raph-luc I just came across following test code.

image

src/ansys/fluent/core/fluent_connection.py
seanpearsonuk: @raph-luc Please allow me to explain the idea here via an example.  
`launch_fluent(...)` is called. Maybe I call it from my Python interpreter, or it's called from some other application (could be a cloud application). Fluent launches successfully but PyFluent fails to connect to Fluent. The calling application will catch exceptions raised by `launch_fluent`. It will use the exception information for observability and potentially for recovery. The exception type is the main source of information. Programmatically, that type is the best thing for deciding how to proceed. The message is useless in this respect if we are trying to do things rigorously. What are the useful exception types that can potentially help us to determine the recovery strategy? Does a FluentConnection error tell us anything more than a ValueError (or any other low-level exception)? Regarding the one location argument: if we can reuse exception types, that's good, but it's not a requirement.

src/ansys/fluent/core/file_session.py
raph-luc: Thanks @hpohekar, that is an example of exactly what I am saying: you can't catch `TransactionError.SURFACE_NAMES_IDs_NOT_PROVIDED`, but you can catch `TransactionError`. 

Change your except block to `except Error` and it will work. In this example, it is therefore the `Error` type that should be listed in the numpy-style docstring, in my understanding.

src/ansys/fluent/core/fluent_connection.py
raph-luc: Thanks @seanpearsonuk, I understand that part, in this example `FluentConnectionError` can be caught, and that can therefore be a useful addition.

However, `FluentConnectionError.PORT_NOT_PROVIDED` is of little use in my understanding. It cannot be caught directly, it is just wrapping around the `"The port to connect to Fluent session is not provided."` message.

In this case, if `FluentConnectionError.PORT_NOT_PROVIDED` is only used in one location, then we are replacing what could be just `raise FluentConnectionError("The port to connect to Fluent session is not provided.")` in that location, by:
```
raise FluentConnectionError.PORT_NOT_PROVIDED()
```
plus the definition
```
    @classmethod
    def PROCESS_NOT_FINISHED(cls):
        return cls(
            "Fluent remote instance not supported by FluentConnection.wait_process_finished()."
        )
```
Which doesn't seem very efficient, to me at least. 

Am I misunderstanding or missing something?

src/ansys/fluent/core/file_session.py
seanpearsonuk: @raph-luc judging by @hpohekar 's other comments below, this issue should resolve itself shortly.

src/ansys/fluent/core/file_session.py
seanpearsonuk: @hpohekar you could also look into refining exception hierarchy in the units repo if it interests you - I believe that what we are doing there is similar to what we currently see here.

src/ansys/fluent/core/fluent_connection.py
seanpearsonuk: @raph-luc No, you are not misunderstanding. There are just some mistakes in the implementation which @hpohekar is working through. Thanks for notifying him about that.

hpohekar: @raph-luc The suggestion given by @mkundu1 to separate out class methods as hierarchy will resolve your concern mentioned above.

src/ansys/fluent/core/file_session.py
hpohekar: Yes, after completing this task I will look into it.

seanpearsonuk:
@hpohekar You might have seen my early comment stating that the fact this PR represents a small change is a good thing. 

Let's keep these changes focussed in a small area. Let's focus on `launch_fluent` first. It's a great example of a function that's composed of multiple steps that can go wrong.

Write tests that demonstrate that we can catch distinct custom exceptions for each failure type, including (in the first instance) _failed to start fluent_ and _failed to connect to fluent_.

src/ansys/fluent/core/file_session.py
mkundu1: @hpohekar @seanpearsonuk Another usage pattern could be:
```
try:
    <pyfluent code>
except TransactionError as ex:
    if ex.code == TransactionError.ErrorCode.SURFACE_NAMES_IDS_NOT_PROVIDED:
        <handle SURFACE_NAMES_IDS_NOT_PROVIDED exception>
    elif ex.code == TransactionError.ErrorCode.INVALID_FIELD_NAME_PREFIX:
        <handle INVALID_FIELD_NAME_PREFIX exception>
    else:
        <handle other TransactionError>
```
`ErrorCode` is an enum. This way we won't have too many custom exception classes, but users will be able to identify the fine-grained errors.

raph-luc: I personally would vote for using and catching specific exception types if possible, like `SurfaceNotSpecifiedError` as you said earlier, otherwise that structure you just shared is going back to being very close to just parsing the Exception error message as in this comment and my comment after that: https://github.com/ansys/pyfluent/issues/2098#issuecomment-1750759704

Which doesn't require any custom exception classes.

raph-luc: They do have this:
image
Which is closer to what I believe we should be aiming for if we are going to add more custom exception classes:

mkundu1: @raph-luc There is a lack of official documentation here. [This](https://github.com/ansys/pyfluent-reader/pull/14#pullrequestreview-1688178548) is the general usage pattern for handling exceptions from a Python client.

Right, internally they have implemented a full custom class hierarchy, but that implementation is private.

raph-luc: I see, thanks for the explanation. As they have those status codes which are used repeatedly and relevant everywhere in their modules (see https://grpc.github.io/grpc/core/md_doc_statuscodes.html), it makes sense that they would include the status codes as part of their custom exception classes.

hpohekar: @mkundu1 This looks good to me. Thank you.

image

tests/test_launcher.py
```diff
@@ -84,9 +85,9 @@ def test_gpu_launch_arg_additional_arg(monkeypatch):
 
 
 def test_kwargs():
-    with pytest.raises(RuntimeError):
+    with pytest.raises(LauncherError.MeshingModeError):
         pyfluent.launch_fluent(abc=1, meshing_mode=True)
-    with pytest.raises(TypeError):
+    with pytest.raises(LauncherError.UnexpectedKeywordArgumentError):
         pyfluent.launch_fluent(abc=1, xyz=2)
```
seanpearsonuk, line 91: @hpohekar It's kind of remarkable that the exception types are different. That's the great thing about this kind of testing. Please continue to add more scenarios, and as you go, you can review if the raised exceptions match the error scenarios (and update the the tests and implementations as necessary). Thanks.

hpohekar: Yes sure.

src/ansys/fluent/core/meta.py
```diff
@@ -738,3 +735,29 @@ def Create(self, name=None):
...
+
+
+class MetaClassError(ValueError):
+    """Custom meta class errors."""
+
+    def __init__(self, error):
+        super().__init__(error)
+
+    @classmethod
+    def DISALLOWED_VALUES(
```
raph-luc, line 747: Are we going to upgrade this `DISALLOWED_VALUES` to an exception class as well so that it can be caught directly (similarly to what was done for `FluentConnectionError.PortNotProvidedError` and the others)?

src/ansys/fluent/core/solver/error_message.py
raph-luc, line 1: Would a file name like `exceptions.py` rather than `error_message.py` work better?

So that we can, for example,
`from ansys.fluent.core.solver.exceptions import TransactionError`
rather than
`from ansys.fluent.core.solver.error_message import TransactionError`

raph-luc (commented):
Looks great, thanks @hpohekar, only a couple minor nitpicks below

Will wait to review again and approve when you set this as ready for review again

src/ansys/fluent/core/solver/error_message.py
hpohekar: Let's see what others think. Personally I just wanted to incorporate this change without disturbing current use and structure of `error_message.py`, therefore I haven't renamed it yet.

src/ansys/fluent/core/meta.py
hpohekar: We haven't updated it with latest change yet. We will do it soon.

src/ansys/fluent/core/solver/error_message.py
seanpearsonuk: If the name is no longer accurate, it shall be changed.

tests/test_launcher.py
seanpearsonuk: @hpohekar we should keep the comment unresolved while the points are still being addressed. Let me specify what they are. I was referring to the arbitrary exception type differences. The original `launch_fluent` code here was wrong. Both exceptions should have been `TypeError`, and that has led you to create distinct types, which looks very odd in this test, right? We should ask about the value of adding new exception types in each case. These `TypeError`s will crop up do to incorrect usage of the interface. `TypeError` is already adequate for that, and the message contains all the context. In creating new exception types, we should focus on the more sophisticated failure scenarios.

src/ansys/fluent/core/solver/error_message.py
```diff
@@ -25,3 +25,55 @@ def allowed_values_error(
     context: str, trial_name: str, allowed_values: List[str]
 ) -> ValueError:
     return ValueError(allowed_name_error_message(context, trial_name, allowed_values))
+
+
+class FluentConnectionError(ValueError):
```
seanpearsonuk, line 30: Why does this inherit from `ValueError`?

src/ansys/fluent/core/solver/error_message.py
```diff
@@ -25,3 +25,55 @@ def allowed_values_error(
...
+    class RemoteNotSupportedError(ValueError):
+        def __init__(self, error):
+            super().__init__(error)
+
+    class WaitTypeError(TypeError):
+        def __init__(self, error):
+            super().__init__(error)
+
+
+class LauncherError(RuntimeError):
```
seanpearsonuk, line 50: Why does this inherit from `RuntimeError`?

src/ansys/fluent/core/solver/error_message.py
```diff
@@ -25,3 +25,55 @@ def allowed_values_error(
...
+
+    class DockerContainerLaunchNotSupportedError(SystemError):
+        def __init__(self, error):
+            super().__init__(error)
+
+
+class TransactionError(RuntimeError):
+    """Custom transaction errors."""
+
+    class SurfaceNamesIDsNotProvidedError(RuntimeError):
```
seanpearsonuk, line 69: Something not provided sounds like a `TypeError`. CHeck if the new type is really needed. If you focus on testing then we can see what is truly needed.

src/ansys/fluent/core/solver/error_message.py
seanpearsonuk: ...although if we have XError directly associated with class x.X then XError should probably be in module x as well. That would also mean that we can eliminate all this extra namespacing which looks heavyweight.

src/ansys/fluent/core/solver/error_message.py
hpohekar: image

https://stackoverflow.com/questions/39077623/can-a-python-inner-class-be-a-subclass-of-its-own-outer-class#:~:text=not%20a%20subclass.-,No%2C%20an%20inner%20class%20can't%20inherit%20(not%20extend,while%20defining%20the%20inner%20class.

src/ansys/fluent/core/solver/error_message.py
hpohekar: Yes, there are many places where incorrect exceptions has been raised. I'm not focusing on correcting each of them right now, I will first ensure that all of them have been covered as per new exception format then after passing all tests, I will correct those exceptions one by one.

tests/test_launcher.py
hpohekar: @seanpearsonuk 

`The original launch_fluent code here was wrong. Both exceptions should have been TypeError, and that has led you to create distinct types, which looks very odd in this test, right?` 

Even if we had corrected both of them to `TypeError` still we need to create and inherit distinctly as we have done in `TransactionError`.

`We should ask about the value of adding new exception types in each case.`

The issue description says that we don't want lower-level built-in exception types, therefore we are updating all existing exceptions. We have not decided any criteria to add new exception types (places where we want to add new exceptions types and places where we want to use built-in exception types).

src/ansys/fluent/core/solver/error_message.py
hpohekar: shall we move all new exceptions to respective modules ? It will also keep module documentation consistent with the new exception type usage.

src/ansys/fluent/core/solver/error_message.py
hpohekar: Right, we followed `TransactionError` format which includes all `RuntimeError`. it's removed now. Thank you.

src/ansys/fluent/core/solver/error_message.py
hpohekar: It's removed now. Thank you.

src/ansys/fluent/core/file_session.py
```diff
@@ -108,15 +109,17 @@ def add_scalar_fields_request(
 
         Raises
         ------
-        RuntimeError
+        TransactionError.SurfaceNamesIDsNotProvidedError
             If surface names or surface ids are not provided.
-        RuntimeError
```
hpohekar, line 113: Both of them should be `TypeError`

src/ansys/fluent/core/file_session.py
```diff
@@ -108,15 +109,17 @@ def add_scalar_fields_request(
 
         Raises
         ------
-        RuntimeError
```
hpohekar, line 111: Should be TypeError

src/ansys/fluent/core/file_session.py
```diff
@@ -227,7 +232,7 @@ def get_fields(self):
...
```
hpohekar, line 230: Should be ValueError

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -603,12 +613,12 @@ def launch_fluent(
     """
     if kwargs:
         if "meshing_mode" in kwargs:
-            raise RuntimeError(
```
hpohekar, line 606: Should be ValueError

src/ansys/fluent/core/solver/error_message.py
raph-luc: My vote would be for leaving `error_message.py` as it is (before these changes), and either:
- Adding a new `exceptions.py` module for all the PyFluent exception classes (assuming they are shared by different modules and each exception class isn't used only in a single module) or 
- As Sean suggested, adding each Exception class to the module where it is used (if each exception class is used in one location)

src/ansys/fluent/core/meta.py
hpohekar: Done.

tests/test_launcher.py
hpohekar: We can replace `LauncherError.MeshingModeError` with `LauncherError.UnexpectedKeywordArgumentError`.

We can also remove `LauncherError.MeshingModeError`.

hpohekar: Replaced `LauncherError.MeshingModeError` with `LauncherError.UnexpectedKeywordArgumentError`

src/ansys/fluent/core/solver/error_message.py
hpohekar: @seanpearsonuk This should be `ValueError` not `TypeError`.

"`TypeError` is raised when an operation or function is applied to an object of inappropriate type. The associated value is a string giving details about the type mismatch."

hpohekar: Corrected error type.

src/ansys/fluent/core/solver/error_message.py
hpohekar: @seanpearsonuk @raph-luc 

We have `error_message.py` at 2 places.

1. src/ansys/fluent/core/solver/error_message.py
2. fluent/cortex/pylib/flapi/error_message.py

Therefore if we are going to add new `exceptions.py` then we will have to add it at above 2 mentioned places.

I think we should follow only 1 approach out of following - 

1. Keep all new exceptions in `error_message.py` and we can rename it to `exceptions.py`. 
(It is useful as `DisallowedValuesError` is defined in single file and we are using it in various other files) 

OR

2.  Keep all new exceptions in respective modules. 
(We will have to import various common errors from various files e.g. `DisallowedValuesError`)

I will go with 1st approach.

hpohekar: @raph-luc 

"Edit: also this exceptions.py module probably should be under ansys.fluent.core and not ansys.fluent.core.solver as it is not solver-specific and is for PyFluent in general, right?"

In Fluent we have `error_message.py` if we leave it as it is, then we will have to add and maintain `exceptions.py` there additionally and I think it's not useful. 

We have renamed `error_message.py` to `exceptions.py` and we are keeping it with `flobject.py` to be consistent with Fluent.

seanpearsonuk: Re:1 error_message.py contains functions that build similar error messages (i.e. currently messages, not exception types) for different contexts. That was done simply to avoid code duplication. It has had to be a shared module because of the wide range of usage contexts. If something is associated with one module, be it a class, function or their associated exception types, then it belongs in that module, unless a strong counterargument can be provided.

seanpearsonuk: > @raph-luc
> 
> "Edit: also this exceptions.py module probably should be under ansys.fluent.core and not ansys.fluent.core.solver as it is not solver-specific and is for PyFluent in general, right?"
> 
> In Fluent we have `error_message.py` if we leave it as it is, then we will have to add and maintain `exceptions.py` there additionally and I think it's not useful.
> 
> We have renamed `error_message.py` to `exceptions.py` and we are keeping it with `flobject.py` to be consistent with Fluent.

You need to start resisting the urge to bundle unrelated things together.

src/ansys/fluent/core/file_session.py
```diff
@@ -108,15 +109,17 @@ def add_scalar_fields_request(
...
+        TransactionError.SurfaceNamesIDsNotProvidedError
```
seanpearsonuk, line 112: `TransactionError` is being used to provide a namespace, but the module already provides a namespace. Also, `file_session` would addsome meaning, whereas `TransactionError` is generic.

src/ansys/fluent/core/solver/error_message.py
hpohekar: > > @raph-luc
> > "Edit: also this exceptions.py module probably should be under ansys.fluent.core and not ansys.fluent.core.solver as it is not solver-specific and is for PyFluent in general, right?"
> > In Fluent we have `error_message.py` if we leave it as it is, then we will have to add and maintain `exceptions.py` there additionally and I think it's not useful.
> > We have renamed `error_message.py` to `exceptions.py` and we are keeping it with `flobject.py` to be consistent with Fluent.
> 
> You need to start resisting the urge to bundle unrelated things together.

@seanpearsonuk 

I don't think that above point is unrelated. let me clarify myself.

1. First we decided to rename `error_message.py` to `exceptions.py` because we are adding exceptions to it now.

image

2. Then we have decided to have XError which is directly associated to x.X in the module as well to avoid extra namespacing.

image

Related to this I asked shall we move all new exception types to respective module ? - unanswered yet.

3. I found that we have added a new exception type in flobject.py.

image

So there are following scenarios right now.

1. Add `InactiveObjectError` in flobject.py itself. (Haven't decided yet)

2. Add `InactiveObjectError` in `error_message.py` which is renamed to `exceptions.py`. (Done in current PR)

3. Add new `exceptions.py` in PyFluent core and in Fluent. (Not done yet)

I just thought adding `InactiveObjectError` in `error_message.py` which is renamed to `exceptions.py` is good option from point of code duplication and maintenance.

Could you please explain how above viewpoint is unrelated to the change I made, regarding renaming `error_message.py` to `exception.py` and adding `InactiveObjectError` to it  to be consistent with Fluent ?

Thank you.

hpohekar: > Re:1 error_message.py contains functions that build similar error messages (i.e. currently messages, not exception types) for different contexts. That was done simply to avoid code duplication. It has had to be a shared module because of the wide range of usage contexts. If something is associated with one module OTOH, be it a class, function or their associated exception types, then it belongs in that module, unless a strong counterargument can be provided otherwise.

@seanpearsonuk 

Yes I agree. Therefore I have added all of them to respective modules in first commit right ? Then we got suggestions that we should maintain all exceptions in a single file therefore I started adding them in `error_message.py`.

seanpearsonuk: OK, by your description, we are converging on the same thing. Sounds good.

hpohekar: So from discussion regarding this, we are going to do following changes.

1. Add exception.py at 
    a) ansys/fluent/core/exception.py
    b) fluent/cortex/pylib/flapi/exception.py
    
 2. Keep `error_message.py` as it is in both PyFluent and Fluent.
 
 3. Keep common exceptions in `exceptions.py`
 
 4. Move distinct exception types belonging to single class into respective modules.

hpohekar:
https://github.com/ansys/pyfluent/pull/2176

PR #2163, opened 2023-10-20
Extract connection-info from server's stdout - not ready for merge

mkundu1:
This doesn't work with `show_gui`, I also need to test in linux.

### DIFF
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -11,7 +11,6 @@ from pathlib import Path
 import platform
 import subprocess
 import tempfile
-import time
 from typing import Any, Dict, List, Optional, Union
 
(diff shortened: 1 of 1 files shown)


src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -201,6 +202,7 @@ def _get_subprocess_kwargs_for_fluent(env: Dict[str, Any]) -> Dict[str, Any]:
     fluent_env = os.environ.copy()
     fluent_env.update({k: str(v) for k, v in env.items()})
     fluent_env["REMOTING_THROW_LAST_TUI_ERROR"] = "1"
+    fluent_env["FLUENT_LAUNCHED_FROM_PYFLUENT"] = "1"
```
mkundu1, line 205: Need to skip the password as currently there is no way to print the password to stdout

mkundu1:
Closing till we plan further on this

PR #2166, opened 2023-10-20
Enable some skipped tests

mkundu1:
Enabling the skipped tests which are passing in the latest version.

### DIFF
--- a/tests/test_session.py
+++ b/tests/test_session.py
@@ -17,6 +17,7 @@ from ansys.fluent.core import connect_to_fluent, examples, session
 from ansys.fluent.core.fluent_connection import FluentConnection
 from ansys.fluent.core.launcher.launcher import LaunchFluentError
 from ansys.fluent.core.session import BaseSession
+from ansys.fluent.core.utils.execution import timeout_loop
 from ansys.fluent.core.utils.networking import get_free_port
 
 
@@ -296,7 +297,6 @@ def test_solverworkflow_not_in_solver_session(new_solver_session):
 
 
 @pytest.mark.fluent_version(">=23.2")
-@pytest.mark.skip("failing because set_state at settings root is not working")
 def test_read_case_using_lightweight_mode():
     import_file_name = examples.download_file(
         "mixing_elbow.cas.h5", "pyfluent/mixing_elbow"
@@ -306,10 +306,16 @@ def test_read_case_using_lightweight_mode():
     )
     solver.setup.models.energy.enabled = False
     old_fluent_connection_id = id(solver.fluent_connection)
-    while id(solver.fluent_connection) == old_fluent_connection_id:
-        time.sleep(1)
-    time.sleep(5)
-    assert solver.setup.models.energy.enabled() == False
+    timeout_loop(
+        id(solver.fluent_connection) != old_fluent_connection_id,
+        timeout=60,
+        idle_period=1,
+    )
+    timeout_loop(
+        solver.setup.models.energy.enabled() == False,
+        timeout=60,
+        idle_period=1,
+    )
     solver.exit()
 
 
--- a/tests/test_session.py
+++ b/tests/test_session.py
@@ -8,7 +8,7 @@ import grpc
 from grpc_health.v1 import health_pb2, health_pb2_grpc
 import pytest
 from util.meshing_workflow import new_mesh_session  # noqa: F401
-from util.solver_workflow import new_solver_session  # noqa: F401
+from util.solver_workflow import make_new_session, new_solver_session  # noqa: F401
 
 from ansys.api.fluent.v0 import scheme_eval_pb2, scheme_eval_pb2_grpc
 from ansys.api.fluent.v0.scheme_pointer_pb2 import SchemePointer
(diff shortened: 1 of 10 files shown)


tests/test_session.py
```diff
@@ -257,15 +256,6 @@ def test_journal_creation(file_format, new_mesh_session):
     assert new_stat.st_mtime > prev_mtime or new_stat.st_size > prev_size
 
 
-@pytest.mark.skip("Failing in GitHub CI")
-def test_old_style_session():
```
mkundu1, line 261: no longer relevant

tests/test_settings_api.py
```diff
@@ -21,77 +21,61 @@ def test_setup_models_viscous_model_settings(new_solver_session) -> None:
     assert viscous_model.model() == "inviscid"
 
 
-@pytest.mark.skip(reason="failing if run with other tests")
-def test_results_graphics_mesh_settings(new_solver_session) -> None:
```
mkundu1, line 25: removed this as we have sufficient coverage for settings API under nightly

tests/test_session.py
```diff
@@ -304,7 +294,7 @@ def test_solverworkflow_not_in_solver_session(new_solver_session):
 
 
 @pytest.mark.fluent_version(">=23.2")
-@pytest.mark.skip("Failing in github")
+@pytest.mark.skip("seems to be broken")
```
mkundu1, line 297: looking into

tests/test_session.py
```diff
@@ -304,7 +294,7 @@ def test_solverworkflow_not_in_solver_session(new_solver_session):
 
 
 @pytest.mark.fluent_version(">=23.2")
-@pytest.mark.skip("Failing in github")
+@pytest.mark.skip("failing because set_state at settings root is not working")
```
mkundu1, line 297: noted in the issue - will investigate

tests/test_tui_api.py
```diff
@@ -5,12 +5,9 @@
 from ansys.fluent.core.services.datamodel_tui import TUIMenu
 
 
-@pytest.mark.skip("randomly failing due to missing transcript capture")
+@pytest.mark.skip("Failing in github")
```
mkundu1, line 8: This one is passing locally in both standalone and container mode, not sure why it is failing only on github

tests/test_settings_api.py
```diff
@@ -21,77 +21,61 @@ def test_setup_models_viscous_model_settings(new_solver_session) -> None:
...
         "fluid": {
             "source_terms": {
-                "y-momentum": [{"option": "value", "value": 2}],
-                "x-momentum": [{"option": "value", "value": 1}],
-                "z-momentum": [{"option": "value", "value": 3}],
+                "source_terms": {
+                    "x-momentum": [{"option": "value", "value": 1}],
+                    "y-momentum": [{"option": "value", "value": 2}],
+                    "z-momentum": [{"option": "value", "value": 3}],
+                }
```
raph-luc, line 59: This repetition seems like it could be improved, in `cell_zone_conditions` of `source_terms.source_terms` and `["source_terms"]["source_terms"]` for the dict 

Though I don't know how much trouble it would be to change this

tests/test_tui_api.py
raph-luc: Also works fine here in Windows and Linux inside/outside containers, very odd

Few ideas below, don't know if you've already tried these or not (we can also try to fix this later in a separate PR)

tests/test_tui_api.py
```diff
@@ -5,12 +5,9 @@
...
 def test_report_system_proc_stats_tui(new_solver_session, capsys) -> None:
-    new_solver_session.transcript.start()
-    # Issue: Transcript missing for the first TUI command
-    new_solver_session.solver.tui.report.system.sys_stats()
-    new_solver_session.solver.tui.report.system.sys_stats()
+    new_solver_session.tui.report.system.sys_stats()
```
raph-luc, line 10: Would using the settings API equivalent make any difference? Both TUI and settings API work on my end (v232 and above)
```suggestion
    new_solver_session.results.report.system.sys_statistics()
```

tests/test_tui_api.py
```diff
@@ -5,12 +5,9 @@
...
     captured = capsys.readouterr()
     assert "CPU" in captured.out
```
raph-luc, line 12: Is it possible the output ended up in stderr rather than stdout for some reason? This will check both of them:
```suggestion
    assert "CPU" in str(captured)
```

tests/test_tui_api.py
raph-luc, line 10: Would perhaps a small delay after the command and before reading the output help?
```suggestion
    new_solver_session.tui.report.system.sys_stats()
    time.sleep(0.5)
```

raph-luc (approved):
Looks good to me, a minor comment and few ideas for `test_report_system_proc_stats_tui` below

tests/test_tui_api.py
mkundu1: I have tried the delay in the past, but it didn't help. The transcript output should be available in stdout. I think there is some issue in the streaming service which is manifested only in a github runner (in an environment with limited resource perhaps). We'll try to investigate it separately.

hpohekar (approved):

PR #2167, opened 2023-10-20
More robust Windows launch command for Watchdog

raph-luc:
Attempting to make the command as explicit as possible, to see if this error, that only occurs in some systems, and that I have been unable to reproduce, is finally resolved without requiring the user to change anything about their system

Changing from command list, which seems to have behavior that can be system dependent on Windows, to a command string, which should be more robust

### DIFF
--- a/src/ansys/fluent/core/launcher/watchdog.py
+++ b/src/ansys/fluent/core/launcher/watchdog.py
@@ -78,16 +78,15 @@ def launch(
 
     python_executable = Path(python_executable)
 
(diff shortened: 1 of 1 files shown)


ansaboutin (approved):
Running `launch.fluent` in the current main branch will result in a pop up asking for me to choose the app to run `watchdog_exec`. In this branch it initializes the watchdog properly.

mkundu1 (approved):

prmukherj (approved):

Issue #2179, opened 2023-10-26
get_scp_string() throws error if SCP file is not in same directory as Python console


ansjsia:
### 🔍 Before submitting the issue

- [X] I have searched among the existing issues
- [x] I am using a Python virtual environment

### 🐞 Description of the bug

Any time the `get_scp_string()` method is called from a directory other than the Fluent working directory, PyFluent throws `AssertionError: ERROR: Could not create System Coupling .scp file`.

### 📝 Steps to reproduce

1. Call `solver = pyfluent.launch_fluent(mode='solver', cwd=..., case_file_name=...)` from the Python console. Set the `cwd` parameter to a directory other than the one the console is currently in. Load a case file in the `case_file_name` parameter.
2. Attempt to call a System Coupling-related method, e.g. `solver.system_coupling.get_analysis_type()`
3. PyFluent throws an `AssertionError: ERROR: Could not create System Coupling .scp file`.

### 💻 Which operating system are you using?

Windows

### 📀 Which ANSYS version are you using?

24R1

### 🐍 Which Python version are you using?

3.11

### 📦 Installed packages

```shell
alabaster==0.7.13
ansys-api-fluent==0.3.18
ansys-api-platform-instancemanagement==1.0.0
ansys-api-systemcoupling==0.1.0
-e git+https://github.com/ansys/pyfluent.git@00a5b4f928c41c743d766af069eea8e66293188b#egg=ansys_fluent_core
ansys-platform-instancemanagement==1.1.2
-e git+https://github.com/ansys/pysystem-coupling.git@82e50fe7c81a1d59c888df28a99760e4b64bb614#egg=ansys_systemcoupling_core
anyio==4.0.0
appdirs==1.4.4
argon2-cffi==23.1.0
argon2-cffi-bindings==21.2.0
arrow==1.3.0
asttokens==2.4.1
async-lru==2.0.4
attrs==23.1.0
Babel==2.13.1
backcall==0.2.0
beautifulsoup4==4.12.2
bleach==6.1.0
cattrs==23.1.2
certifi==2023.7.22
cffi==1.16.0
charset-normalizer==3.3.1
colorama==0.4.6
comm==0.1.4
debugpy==1.8.0
decorator==5.1.1
defusedxml==0.7.1
docker==6.1.3
docutils==0.20.1
esbonio==0.16.2
executing==2.0.0
fastjsonschema==2.18.1
fqdn==1.5.1
googleapis-common-protos==1.61.0
[...]

PR #2180, opened 2023-10-26
Prepend working dir to .scp file name

ansjsia:
To correct issue https://github.com/ansys/pyfluent/issues/2179

Obtains Fluent's working directory using the `cortex_pwd` variable and prepends that to the `.scp` file name in `get_scp_string()`.

### DIFF
--- a/src/ansys/fluent/core/systemcoupling.py
+++ b/src/ansys/fluent/core/systemcoupling.py
@@ -72,7 +72,8 @@ class SystemCoupling:
     def __get_syc_setup(self) -> dict:
         def get_scp_string() -> str:
             """Get SCP file contents in the form of the XML string."""
-            scp_file_name = "fluent.scp"
+            fluent_cwd = self._solver.connection_properties.cortex_pwd
+            scp_file_name = os.path.join(fluent_cwd, "fluent.scp")
             self._solver.setup.models.system_coupling.write_scp_file(
                 file_name=scp_file_name
             )


ochernuk (approved):

raph-luc (approved):
Unrelated to this PR: should tests be added for `systemcoupling.py`? Doesn't look like we have any from PyFluent's side.

raph-luc:
Note also that this won't work when using Fluent containers, as the file system inside the container (where Fluent is) and in the host system (where Python is) are not the same.

ansjsia:
> Note also that this won't work when using Fluent containers, as the file system inside the container (where Fluent is) and in the host system (where Python is) are not the same.

Would this PR break existing behavior in that case?

raph-luc:
@ansjsia Depends, is this being used with Fluent containers or remote Fluent clients? If you are only using it with conventional local Fluent installs, then it shouldn't break anything. 

However, if we, for example, implement GitHub tests or examples for `systemcoupling.py` that run using containers, then this won't work by default.

ansjsia:
Is there a better way to get Fluent's working directory rather than relying on the `cortex_pwd` variable?

ochernuk:
@raph-luc you're right. Eventually, I am hoping we do not have to write out the .scp file at all, which would address the containers issue, but I think that requires Fluent changes to provide the same information directly, e.g. via settings APIs.

raph-luc:
@ansjsia No, I believe `connection_properties.cortex_pwd` as you used here is the best way and it is the correct value for Fluent. The issue is the fact that the working directory path may be different (or inaccessible) for Python (and PyFluent) when Fluent is remote/containerized.

@ochernuk I see, thanks for the additional context. If required, we can also make this work in both situations from the Python side, as if you are using PyFluent to launch the Fluent process, then we will usually have enough information to properly handle any path differences from Python.

PR #2214, opened 2023-11-06
refactor: Upgrade examples to 24.1

ansjsia:
Upgrades:
- TUI calls replaced with Settings API calls where possible (except for Exhaust System example, where a new example using the Settings API has been created to avoid overwriting the TUI example)
- Tested with Fluent 24R1
- Tested with PyFluent 0.18.1

This should be merged only once the switchover to 24R1 is complete.


raph-luc:
Converting this to a draft: to be reopened after 24.1 is released and the repo is updated to support it

examples/00-fluent/external_compressible_flow.py
```diff
@@ -301,9 +302,9 @@
 
 pressure_farfield.flow_direction[2] = 0.053382
 
-pressure_farfield.turb_intensity = 0.05
+pressure_farfield.turbulent_intensity = 0.05
 
-pressure_farfield.turb_viscosity_ratio = 10
+pressure_farfield.turbulent_viscosity_ratio_real = 10
```
dnwillia-work, line 307: @mkundu1 @seanpearsonuk Any idea why this has the strange `_real` suffix on it now.  It was fine before.

dnwillia-work: See #2251

examples/00-fluent/mixing_elbow_settings_api.py
```diff
@@ -61,7 +61,10 @@
 # in this example. If you want working units for length to be other than inches
 # (for example, millimeters), make the appropriate change.
 
-solver.tui.define.units("length", "in")
+solver.setup.general.units.set_units(
+    quantity="length",
+    units_name="mm",
+)
```
seanpearsonuk, line 67: I did not realise that this was in the settings API. Let me comment on the API here, rather than on this change, which is fine.

This is purely a user interface setting. Not only is it irrelevant to a remote user, it's also misleading in a way that could have a negative side effect. After setting the length unit (for instance), the user can then get (or set) a length parameter in the belief that the values are in their chosen unit. Avoidance of this kind of muddled API design is something that has been raised several times. 

Furthermore, the help string is inaccurate: "Set unit conversion factors."  The help   string has to say precisiely what this method does. Additionally, the method name should be updated to reflect that this is purely a Fluent UI setting: e.g., `set_fluent_ui_units`.

@dnwillia-work @h-krishnan

ansjsia: Thanks for calling this out, Sean. Since this purely a GUI change and the GUI is not shown in this example, would it be prudent for me to delete this section entirely from all the examples that use it?

raph-luc: Yes @ansjsia, great point. If those commands are only relevant for the GUI, I would say they must be removed from all the PyFluent examples, otherwise they will be very misleading to users

ansjsia: I've removed all calls to `root.setup.general.units.set_units()` in all the PyFluent examples.

ansjsia:
The last commit is dependent on this PR: https://github.com/ansys/example-data/pull/214

ansjsia:
The CI/CD pipeline is throwing `RuntimeError: menu not found` on line 629 of `exhaust_system.py`, when it makes the call `solver.tui.display.set.picture.use_window_resolution("no")`. I can't replicate this problem locally nor directly in the TUI. Any insights on why this might be happening? @dnwillia-work @raph-luc

mkundu1:
@ansjsia This can be merged after #2349 which updates `fluent_release_version`.

We need to update the Fluent version in the .yml files for documentation CI.

cc: @prmukherj

mkundu1 (approved):

prmukherj (approved):

hpohekar (approved):

Issue #2187, opened 2023-11-01
Fluent launch process is very slow giving warnings.


giannoko:
### 🔍 Before submitting the issue

- [x] I have searched among the existing issues
- [X] I am using a Python virtual environment

### 🐞 Description of the bug

When try to launch Fluent session, it takes a long time to get launched getting warnings messages regarding run-time issues.

Input:

```
import ansys.fluent.core as pyfluent
session = pyfluent.launch_fluent(show_gui=False)

```
Output:

```
pyfluent.launcher WARNING: Exception caught - RuntimeError: The launch process has been timed out.
pyfluent.launcher WARNING: Retrying Fluent launch with less robust command: C:\Program Files\ANSYS Inc\v241\fluent\ntbin\win64\fluent.exe 3ddp   -sifile="C:\Users\kgiannok\AppData\Local\Temp\serverinfo-zy71n9vz.txt" -nm -hidden
```

The session starts but the launching time is quite high. 

I get the same behavior for pyFluent 0.16.dev0 - 0.19.dev1 versions and for Fluent 23R2 and 24R1. 
I have no issues with v0.15.0.

### 📝 Steps to reproduce

```
import ansys.fluent.core as pyfluent
session = pyfluent.launch_fluent(show_gui=False)
```

### 💻 Which operating system are you using?

Windows

### 📀 Which ANSYS version are you using?

tested pyfluent versions  : 0.16.dev0 - 0.19.dev1
tested Fluent versions     : 23R2, 24R1

### 🐍 Which Python version are you using?

3.10

### 📦 Installed packages

```shell
accessible-pygments==0.0.4
alabaster==0.7.13
ansys-api-fluent==0.3.18
ansys-api-platform-instancemanagement==1.0.0
ansys-api-systemcoupling==0.1.0
ansys-dpf-core==0.9.0
ansys-dpf-gate==0.4.1
ansys-dpf-gatebin==0.4.1
ansys-fluent-core==0.17.0
ansys-grpc-dpf==0.8.1
ansys-platform-instancemanagement==1.1.2
ansys-sphinx-theme==0.12.3
-e git+https://github.com/pyansys/pysystem-coupling.git@82e50fe7c81a1d59c888df28a99760e4b64bb614#egg=ansys_systemcoupling_core
appdirs==1.4.4
asttokens==2.4.0
attrs==23.1.0
Babel==2.13.0
backcall==0.2.0
beautifulsoup4==4.12.2
black==23.10.0
[...]

PR #2217, opened 2023-11-08
Fixing launch behavior for various Windows shell environments

raph-luc:
Closes https://github.com/ansys/pyfluent/issues/2187

Two things were causing issues for custom shell environments on Windows and are fixed here:
- Unnecessary empty spaces between arguments
- Using quotes when they are not needed

Tested with Command Prompt, PowerShell and Cmder, this should not cause any regressions (famous last words)

Just to make sure no additional issues introduced, it would be helpful if some reviewers could test this branch on their own Windows systems: just `launch_fluent` and then see if any obvious issues occur, and if PyFluent and the Watchdog were launched correctly.


mkundu1 (approved):
works fine as per my testing on windows

raph-luc:
Confirmed with @giannoko the issues he had are resolved when using this branch

dnwillia-work (approved):

Issue #2221, opened 2023-11-08
Unused arguments and copy-pasted unused method


raph-luc:
Stumbled upon this while working on something else: why were these unused arguments defined and documented in the docstring? Also, is the docstring accurate? Should we review this? 
https://github.com/ansys/pyfluent/blob/1145fd8c563f9d2d11e52daad4d2f7d250623770/src/ansys/fluent/core/streaming_services/monitor_streaming.py#L124-L143

Also, this method is copy-pasted below in the general `StreamingService` class and doesn't seem to be used at all here, is this for another package or perhaps deprecated? I removed the block below locally and it doesn't seem to cause any issues that I've noticed.
https://github.com/ansys/pyfluent/blob/1145fd8c563f9d2d11e52daad4d2f7d250623770/src/ansys/fluent/core/streaming_services/streaming.py#L88-L104

seanpearsonuk:
@raph-luc From mem, I'm pretty sure that this is due to polymorphism.

hpohekar:
`refresh` method does not exist in streaming.py now.

`refresh` method from monitors_streaming.py is being used in session_solver.py.

hpohekar:
@seanpearsonuk 

We have verified that though `refresh` arguments are unused here but they are essential for callback signature that we are registering.

If we remove them from signature then tests fail or hang forever.

https://github.com/ansys/pyfluent/pull/3681

https://github.com/ansys/pyfluent/actions/runs/13024030601/job/36331468048

seanpearsonuk:
Callback registration requires the specific callback interface:

```Python
   def register_callback(
       self,
       event_name: TEvent | str,
       callback: Callable,
       *args,
       **kwargs,
   ):
       """Register the callback.

       Parameters
       ----------
       event_name : TEvent or str
           Event to register the callback to.
       callback : Callable
           Callback to register. If the custom arguments,
           args and kwargs, are empty then the callback
           signature must be precisely <function>(session, event_info).
           Otherwise, the arguments for args and/or kwargs
           must precede the other arguments in the signature.
```
@hpohekar Note though that there is an annotation error in monitor_streaming.py:
```Python
    def refresh(self, session, event_info) -> None:
        """Refresh plots on-initialized and data-read events.

        This method is registered with the EventsManager and is called
        to refresh plots whenever on-initialized and data-read events occur.

        Parameters
        ----------
        session : str
            Session object.
        event_info : object
            Event info object.
```

[...]

PR #2222, opened 2023-11-09
Fixing Slurm/HPC issues

raph-luc:
Closes https://github.com/ansys/pyfluent/issues/2223, see that tracker for more details

Tested in CDC, and the user tested in OTTC, confirmed to solve the issues

Two issues fixed:
- PyFluent was freezing due to streaming thread locked up by `.join` with no timeout
- Watchdog was being launched even though Fluent is remote, and it shouldn't be

### DIFF
--- a/src/ansys/fluent/core/streaming_services/streaming.py
+++ b/src/ansys/fluent/core/streaming_services/streaming.py
@@ -1,6 +1,9 @@
 import itertools
+import logging
 import threading
 from typing import Callable, Optional
+
+logger = logging.getLogger("pyfluent.networking")
 
 
 class StreamingService:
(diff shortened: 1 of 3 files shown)


src/ansys/fluent/core/streaming_services/streaming.py
```diff
@@ -78,30 +81,14 @@ def start(self, *args, **kwargs) -> None:
...
-            Event info object.
-
-        Returns
-        -------
-        None
-        """
-        with self._lock_refresh:
-            self.stop()
-            self.start()
-
```
raph-luc, line 105: See https://github.com/ansys/pyfluent/issues/2221

mkundu1 (approved):

dnwillia-work (approved):

Issue #2233, opened 2023-11-10
Download file from examples gives SSL error on RH7


millerj97:
### 🔍 Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### 🐞 Description of the bug

When using the below command on RH7;
```
examples.download_file("mixing_elbow.pmdb", "pyfluent/mixing_elbow", save_path='.')  
```

The following error is reported;
```
Checking if specified file already exists...
File does not exist. Downloading specified file...
Traceback (most recent call last):
  File "/home/jomiller/ansys_inc/v241/commonfiles/CPython/3_10/linx64/Release/python/lib/python3.10/urllib/request.py", line 1348, in do_open
    h.request(req.get_method(), req.selector, req.data, headers,
  File "/home/jomiller/ansys_inc/v241/commonfiles/CPython/3_10/linx64/Release/python/lib/python3.10/http/client.py", line 1282, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/home/jomiller/ansys_inc/v241/commonfiles/CPython/3_10/linx64/Release/python/lib/python3.10/http/client.py", line 1328, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/home/jomiller/ansys_inc/v241/commonfiles/CPython/3_10/linx64/Release/python/lib/python3.10/http/client.py", line 1277, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/home/jomiller/ansys_inc/v241/commonfiles/CPython/3_10/linx64/Release/python/lib/python3.10/http/client.py", line 1037, in _send_output
    self.send(msg)
  File "/home/jomiller/ansys_inc/v241/commonfiles/CPython/3_10/linx64/Release/python/lib/python3.10/http/client.py", line 975, in send
    self.connect()
  File "/home/jomiller/ansys_inc/v241/commonfiles/CPython/3_10/linx64/Release/python/lib/python3.10/http/client.py", line 1454, in connect
    self.sock = self._context.wrap_socket(self.sock,
  File "/home/jomiller/ansys_inc/v241/commonfiles/CPython/3_10/linx64/Release/python/lib/python3.10/ssl.py", line 513, in wrap_socket
    return self.sslsocket_class._create(
  File "/home/jomiller/ansys_inc/v241/commonfiles/CPython/3_10/linx64/Release/python/lib/python3.10/ssl.py", line 1071, in _create
[...]

PR #2234, opened 2023-11-10
Removing deprecated `urlretrieve` to fix `download_file` issue

raph-luc:
Closes https://github.com/ansys/pyfluent/issues/2233

Fixes issues that were occurring on Red Hat and CentOS machines (and HPC environments), I was able to reproduce the issues and test the changes locally as well.

Also removed redundant `print`s and replaced them with logging.

Confirmed with @millerj97 that this branch solves the issues on their machines as well

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,7 @@ install_requires = [
     "pyyaml>=6.0",
     "docker>=6.1.3",
     "psutil>=5.9.5",
+    "requests>=2.31.0",
 ]
 
 extras_require = {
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -33,6 +33,7 @@
     "pyyaml>=6.0",
     "docker>=6.1.3",
     "psutil>=5.9.5",
+    "requests>=2.31.0",
```
raph-luc, line 36: `docker` already required this package, so it will already be present in any recent PyFluent environments

seanpearsonuk (approved):

mkundu1 (approved):

PR #2245, opened 2023-11-14
Bump version 0.18.2

raph-luc:


setup.py
```diff
@@ -33,6 +33,7 @@
     "pyyaml>=6.0",
     "docker>=6.1.3",
     "psutil>=5.9.5",
+    "requests>=2.31.0",
```
raph-luc, line 36: `docker` already had the same `requests` dependency, so this is not a new dependency, it is essentially just being made explicit here

PR #2255, opened 2023-11-15
Pythonic Syntax for __getattribute__ Error.

ansaboutin:
Added a python_path function to format a __getattribute__ runtime error to python syntax.

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -128,6 +128,7 @@ class Base:
 
     _name = None
     fluent_name = None
+    _python_name = None
 
     @property
     def parent(self):
@@ -152,7 +153,7 @@ class Base:
         By default, this returns the object's static name. If the object is a child of a
         named object, the object's name is returned.
         """
-        return getattr(self, "_python_name") or self.__class__.__name__
+        return getattr(self, "_python_name", None) or "." + self.__class__.__name__
 
     @property
     def path(self) -> str:
@@ -180,7 +181,7 @@ class Base:
         ppath = self._parent.python_path
         if not ppath:
             return self.python_name
-        return ppath + "." + self.python_name
+        return ppath + self.python_name
 
     def get_attrs(self, attrs, recursive=False) -> Any:
         """Get the requested attributes for the object."""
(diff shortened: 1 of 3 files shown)


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -159,6 +170,20 @@ def path(self) -> str:
...
+
+        Constructed in python syntax from 'python_path' and the parents python
+        path.
+        """
+        if self._parent is None:
+            return "<session>"
+        ppath = self._parent.python_path
+        if not ppath:
+            return self.python_name
+        return ppath + self.python_name
```
mkundu1, line 185:
```suggestion
        return ppath + "." + self.python_name
```

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -145,6 +145,17 @@ def obj_name(self) -> str:
...
+    @property
+    def python_name(self) -> str:
+        """Python name of this object.
+
+        By default, this returns the object's static name. If the object is a child of a
+        named object, the object's name is returned.
+        """
+        if self._name:
+            return self._python_name
+        return "." + self.__class__.__name__
```
mkundu1, line 157:
```suggestion
        return getattr(self, "_python_name") or self.__class__.__name__
```

src/ansys/fluent/core/solver/flobject.py
ansaboutin: I did undo this change as I was getting a "." before the bracket:
image

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -145,6 +146,15 @@ def obj_name(self) -> str:
             return self.fluent_name
         return self._name
 
+    @property
+    def python_name(self) -> str:
+        """Python name of this object.
+
+        By default, this returns the object's static name. If the object is a child of a
+        named object, the object's name is returned.
```
seanpearsonuk, line 154: But we can also return the class name prefixed with a dot. That doesn't look right.

dnwillia-work (approved):

mkundu1 (approved):

prmukherj (approved):

PR #2256, opened 2023-11-16
Fix generic datamodel

mkundu1:
Fix generic datamodel when generated datamodel classes are not available


seanpearsonuk (approved):

prmukherj (approved):

mkundu1:
> Is this for when trying to access datamodel attributes without running codegen first?

Yeah. Also, we are planning to test some datamodel rpc based on the flserver datamodel, we don't generate the flserver datamodel files.

raph-luc (approved):

PR #2257, opened 2023-11-18
Transform cached state for display names or internal names

mkundu1:
The cached state returned from `DataModelCache.get_state(rules, obj, name_key)` will contain display names or internal names based on the parameter  `name_key`


pyproject.toml
```diff
@@ -10,9 +10,6 @@ requires = [
 testpaths = ["tests"]
 addopts = """
 -v
---cov=ansys.fluent
---cov-report=xml:cov_xml.xml
---cov-report=html
```
mkundu1, line 15: Moved these to Makefile so that debugging from IDE works (https://pytest-cov.readthedocs.io/en/latest/debuggers.html)

prmukherj (approved):

mkundu1:
@prmukherj Please wait till I investigate the test failures.

seanpearsonuk (approved):

prmukherj:
Thank you @mkundu1 for the input. I'll check this.

seanpearsonuk (approved):

mkundu1:
@prmukherj @seanpearsonuk 
Among the remaining failing tests, one type of failure seems to be due to the code queries for `TaskList` and `InactiveTaskList` parameters which return internal names. They don't match with the state as the state is received with display names from the cache.
There is another type of failure which I've locally tested to fix by adding some sleep before the assert, e.g. the following assert in `tests/test_meshing_workflow.py::test_extended_wrapper`:
```
assert import_geometry.arguments.get_state(explicit_only=True) == dict(
    FileName=None
)
```
The assert is executed before the cache is updated with the streamed state.

prmukherj:
@mkundu1, @seanpearsonuk , what should be our approach then? Shall we fix the test along this new behavior? The earlier behavior seems inaccurate then, right?

PR #2262, opened 2023-11-20
Strongly Typed Version Checks

ansaboutin:
Moved the FluentVersion Enum class.

Is there a cleaner way to accomplish the same access style for `FluentVersion.current`?

It looks like the version values that come back from a session are ints. Should the FluentVersion values be changed to int for consistency?
image

There is another fluent_version.py in utils. Should these be consolidated?

### DIFF
--- a/src/ansys/fluent/core/__init__.py
+++ b/src/ansys/fluent/core/__init__.py
@@ -14,7 +14,6 @@ logging.configure_env_var()
 from ansys.fluent.core._version import __version__  # noqa: F401
 from ansys.fluent.core.launcher.launcher import (  # noqa: F401
     FluentMode,
-    FluentVersion,
     connect_to_fluent,
     launch_fluent,
 )
--- a/src/ansys/fluent/core/launcher/fluent_version.py
+++ b/src/ansys/fluent/core/launcher/fluent_version.py
@@ -1,0 +1,87 @@
+"""Provides the Fluent version."""
+
+from enum import Enum
+import os
+
+supported_versions = ["24.2.0", "24.1.0", "23.2.0", "23.1.0", "22.2.0"]
+
+
+def get_current_version() -> str:
+    """Return the version string corresponding to the most recent, available ANSYS
+    installation.
+
+    The returned value is the string component of one of the members of the
+    FluentVersion class.
+
+    Returns
+    -------
+    str
+        Ansys version string
+
+    Raises
+    ------
+    RuntimeError
+        If an Ansys version cannot be found.
+    """
+    for v in supported_versions:
+        if "AWP_ROOT" + "".join(str(v).split("."))[:-1] in os.environ:
+            return v
+
+    raise RuntimeError("An Ansys version cannot be found.")
+
+
+class FluentVersion(Enum):
+    """An enumeration over supported Fluent versions."""
+
+    _24R2 = supported_versions[0]
+    _24R1 = supported_versions[1]
+    _23R2 = supported_versions[2]
+    _23R1 = supported_versions[3]
+    _22R2 = supported_versions[4]
+
+    current = get_current_version()
+
+    @classmethod
+    def _missing_(cls, version):
+        if isinstance(version, (float, str)):
+            version = str(version) + ".0"
+            for v in FluentVersion:
+                if version == v.value:
+                    return FluentVersion(version)
+            else:
+                raise RuntimeError(
+                    f"The specified version '{version[:-2]}' is not supported."
+                    + " Supported versions are: "
+                    + ", ".join([ver.value for ver in FluentVersion][::-1])
+                )
+
+    def __str__(self):
+        return str(self.value)
+
+    def __lt__(self, other):
+        if isinstance(other, FluentVersion):
+            return self.value < other.value
+        return RuntimeError(
+            "'<' is only supported between instances of 'FluentVersion' and 'FluentVersion'."
+        )
+
+    def __le__(self, other):
+        if isinstance(other, FluentVersion):
+            return self.value <= other.value
+        return RuntimeError(
+            "'<=' is only supported between instances of 'FluentVersion' and 'FluentVersion'."
+        )
+
+    def __gt__(self, other):
+        if isinstance(other, FluentVersion):
+            return self.value > other.value
+        return RuntimeError(
+            "'>' is only supported between instances of 'FluentVersion' and 'FluentVersion'."
+        )
+
+    def __ge__(self, other):
+        if isinstance(other, FluentVersion):
+            return self.value >= other.value
+        return RuntimeError(
+            "'>=' is only supported between instances of 'FluentVersion' and 'FluentVersion'."
+        )
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -19,6 +19,7 @@ from ansys.fluent.core.launcher.fluent_container import (
     configure_container_dict,
     start_fluent_container,
 )
+from ansys.fluent.core.launcher.fluent_version import FluentVersion
 import ansys.fluent.core.launcher.watchdog as watchdog
 from ansys.fluent.core.scheduler import build_parallel_options, load_machines
 from ansys.fluent.core.session import _parse_server_info_file
@@ -56,56 +57,6 @@ def check_docker_support():
     except docker.errors.DockerException:
         return False
     return True
-
-
-class FluentVersion(Enum):
-    """An enumeration over supported Fluent versions."""
-
-    version_24R1 = "24.1.0"
-    version_23R2 = "23.2.0"
-    version_23R1 = "23.1.0"
-    version_22R2 = "22.2.0"
-
-    @classmethod
-    def _missing_(cls, version):
-        if isinstance(version, (float, str)):
-            version = str(version) + ".0"
-            for v in FluentVersion:
-                if version == v.value:
-                    return FluentVersion(version)
-            else:
-                raise RuntimeError(
-                    f"The specified version '{version[:-2]}' is not supported."
-                    + " Supported versions are: "
-                    + ", ".join([ver.value for ver in FluentVersion][::-1])
-                )
-
-    def __str__(self):
-        return str(self.value)
-
-
-def get_ansys_version() -> str:
-    """Return the version string corresponding to the most recent, available ANSYS
-    installation.
-
-    The returned value is the string component of one of the members of the
-    FluentVersion class.
-
-    Returns
-    -------
-    str
-        Ansys version string
-
-    Raises
-    ------
-    RuntimeError
-        If an Ansys version cannot be found.
-    """
-    for v in FluentVersion:
-        if "AWP_ROOT" + "".join(str(v).split("."))[:-1] in os.environ:
-            return str(v)
-
-    raise RuntimeError("An Ansys version cannot be found.")
 
 
 def get_fluent_exe_path(**launch_argvals) -> Path:
--- a/src/ansys/fluent/core/systemcoupling.py
+++ b/src/ansys/fluent/core/systemcoupling.py
@@ -5,6 +5,8 @@ import os
 from typing import List
 import xml.etree.ElementTree as XmlET
 
+from ansys.fluent.core.launcher.fluent_version import FluentVersion
+
 
 @dataclass
 class Variable:
@@ -42,10 +44,9 @@ class SystemCoupling:
     def __init__(self, solver):
         self._solver = solver
         # version check - this requires Fluent 2024 R1 or newer.
-        fluent_version = self._solver.get_fluent_version()
-        if float(fluent_version[:-2]) < 24.1:
+        if FluentVersion.current < FluentVersion._24R1:
             raise RuntimeError(
-                f"Fluent version is {fluent_version}. PySystemCoupling integration requires Fluent 24.1.0 or later."
+                f"Fluent version is {FluentVersion.current}. PySystemCoupling integration requires Fluent 24.1.0 or later."
             )
 
     @property
--- a/src/ansys/fluent/core/utils/search.py
+++ b/src/ansys/fluent/core/utils/search.py
@@ -3,7 +3,7 @@ from pathlib import Path
 import pickle
 from typing import Any, Optional
 
-from ansys.fluent.core.launcher.launcher import FluentVersion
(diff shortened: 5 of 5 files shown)


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1208,7 +1210,7 @@ def get_cls(name, info, parent=None, version=None):
             "user_creatable", False
         )
 
-        if version == "222":
```
seanpearsonuk, line 1211: @ansaboutin Because flobject.py is used in Fluent as well, we would need to bring the FluentVersion class into Fluent's codebase. In doing so, we need to keep the imports as simple as possible. So for PyFluent, we can put a proxy module in the same folder as flobject.py to handle the modue import.

src/ansys/fluent/core/utils/fluent_version.py
```diff
@@ -20,3 +37,97 @@ def get_version_for_file_name(version: Optional[str] = None, session=None):
         version = get_version(session)
 
     return "".join(version.split(".")[0:2])
+
+
+class FluentVersion(Enum):
```
seanpearsonuk, line 42: @ansaboutin we want to have this class in a separate module so that we can import it cleanly from flobject.py without dragging in other unwanted dependencies that are unneeded in the Fluent context.

src/ansys/fluent/core/utils/fluent_version.py
```diff
@@ -20,3 +37,97 @@ def get_version_for_file_name(version: Optional[str] = None, session=None):
...
+    def __int__(self):
+        """Return the version as a number (e.g. 232)"""
+        if self.value:
+            return int(self.value.replace(".", "")[:-1])
+        return 0
+
+    def __str__(self):
+        """Return the version path (e.g. "AWP_ROOT232")"""
+        return str(f"AWP_ROOT{int(self)}")
+
```
seanpearsonuk, line 114: @ansaboutin Shall we use integer comparisons for all the following? Then we can allow arbitrary string representation while maintaining strictness that successive versions always have higher integer values.

src/ansys/fluent/core/solver/flversion.py
seanpearsonuk, line 1: @ansaboutin could you try just adding a .py file here to import from utils?

src/ansys/fluent/core/solver/flversion.py
```diff
@@ -0,0 +1,3 @@
+from ansys.fluent.core.utils.fluent_version import FluentVersion
+
+FluentVersion
```
seanpearsonuk, line 3: @ansaboutin is this line required?

src/ansys/fluent/core/utils/fluent_version.py
```diff
@@ -20,3 +37,97 @@ def get_version_for_file_name(version: Optional[str] = None, session=None):
...
+        raise AnsysVersionNotFound()
+
+    def __int__(self):
+        """Return the version as a number (e.g. 232)"""
+        if self.value:
+            return int(self.value.replace(".", "")[:-1])
+        return 0
+
+    def __str__(self):
```
mkundu1, line 111: Perhaps, we can be explicit here, like `awp_var()`. AWP_ROOT is not the only string representation of Fluent versions, the docker image version (like `v24.1.0`) is another example of string representation.

src/ansys/fluent/core/utils/fluent_version.py
```diff
@@ -20,3 +37,97 @@ def get_version_for_file_name(version: Optional[str] = None, session=None):
...
+        """Return the version path (e.g. "AWP_ROOT232")"""
+        return str(f"AWP_ROOT{int(self)}")
+
+    def __lt__(self, other):
```
mkundu1, line 115: [@functools.total_ordering](https://docs.python.org/3/library/functools.html#functools.total_ordering) can be used to reduce number of operator implementations. The error messages need to be made generic which is probably fine.

src/ansys/fluent/core/solver/flversion.py
ansaboutin: @seanpearsonuk Without the `FluentVersion` line the pre-commit check will not let me have an unused import. Is there a better way to get around it?

seanpearsonuk: @ansaboutin 
Try `# noqa: F401` at the end of the import line

seanpearsonuk, line 3:
```suggestion
from ansys.fluent.core.utils.fluent_version import FluentVersion # noqa: F401
```

src/ansys/fluent/core/utils/fluent_version.py
seanpearsonuk: 🥇

src/ansys/fluent/core/utils/fluent_version.py
ansaboutin: Yes I can do that. Should a method for the image string be added also? Are there any other strings of version used?

mkundu1: Yeah, I guess we'll need another method for the image tag (can be part of a different PR). No, I'm not aware of anything else ATM.

src/ansys/fluent/core/utils/fluent_version.py
```diff
@@ -20,3 +38,99 @@ def get_version_for_file_name(version: Optional[str] = None, session=None):
...
+
+    FluentVersion.v23R2.value == '23.2.0'
+    """
+
+    v24R2 = "24.2.0"
+    v24R1 = "24.1.0"
+    v23R2 = "23.2.0"
+    v23R1 = "23.1.0"
+    v22R2 = "22.2.0"
+    default = ""
```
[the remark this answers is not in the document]
raph-luc: This `default` value is also getting parsed as a "supported" version, see the first lone comma here:
`RuntimeError: The specified version '25.2.0' is not supported. Supported versions are: , 22.2.0, 23.1.0, 23.2.0, 24.1.0, 24.2.0`

src/ansys/fluent/core/utils/fluent_version.py
```diff
@@ -20,3 +38,99 @@ def get_version_for_file_name(version: Optional[str] = None, session=None):
...
+                    return FluentVersion(version)
+            else:
+                raise RuntimeError(
+                    f"The specified version '{version[:-2]}' is not supported."
+                    + " Supported versions are: "
+                    + ", ".join([v.value for v in FluentVersion][::-1])
+                )
+
+    @classmethod
+    def current(cls):
```
raph-luc, line 87:
```suggestion
    @staticmethod
    def current():
```

src/ansys/fluent/core/utils/fluent_version.py
```diff
@@ -20,3 +38,99 @@ def get_version_for_file_name(version: Optional[str] = None, session=None):
...
+
+    def img_tag(self):
+        """Get the Fluent version in the image tag format.
+
+        Returns
+        -------
+        str
+            Image tag (e.g. "v23.2.0")
+        """
+        return str(f"v{self.value}")
```
raph-luc, line 125:
```suggestion
        return f"v{self.value}"
```

src/ansys/fluent/core/utils/fluent_version.py
```diff
@@ -20,3 +38,99 @@ def get_version_for_file_name(version: Optional[str] = None, session=None):
...
+            FluentVersion member corresponding to the newest Fluent version.
+
+        Raises
+        ------
+        AnsysVersionNotFound
+            If an Ansys version cannot be found.
+        """
+        for version in FluentVersion:
+            if version.awp_var() in os.environ:
+                return version
```
raph-luc, line 103: So `FluentVersion.current()` walks through the environment variables in the system and returns the latest installed.

Would it make more sense and be more intuitive to rename so we have a method `FluentVersion.get_latest_installed()` for this rather than `FluentVersion.current()`?

src/ansys/fluent/core/utils/fluent_version.py
```diff
@@ -20,3 +38,99 @@ def get_version_for_file_name(version: Optional[str] = None, session=None):
...
+        if isinstance(version, (int, float, str)):
+            version = str(version)
+            if len(version) == 3:
+                version = version[:2] + "." + version[2:]
+            version += ".0"
+            for v in FluentVersion:
+                if version == v.value:
+                    return FluentVersion(version)
+            else:
+                raise RuntimeError(
```
raph-luc, line 80: Should we use a higher level exception rather than a `RuntimeError`? Can use the existing `AnsysVersionNotFound`, with a custom message.

src/ansys/fluent/core/utils/fluent_version.py
```diff
@@ -1,9 +1,27 @@
...
 from typing import Optional
 
 import ansys.fluent.core as pyfluent
 
 
+class AnsysVersionNotFound(RuntimeError):
+    """Provides the error when Ansys version is not found."""
+
+    def __init__(self):
+        super().__init__("Verify the value of the 'AWP_ROOT' environment variable.")
```
raph-luc, line 13: Nitpick: we are only raising `AnsysVersionNotFound` in a single location. Would it make more sense to go with:
```suggestion
class AnsysVersionNotFound(RuntimeError):
    """Provides the error when Ansys version is not found."""
    pass
```
and when raising the exception:
`raise AnsysVersionNotFound("Verify the value of the 'AWP_ROOT' environment variable.")`
Less lines and cleaner in my opinion.

I think it only really makes sense to consolidate the error message as you suggest here when the same error message is used in multiple locations. But I'm OK with leaving it as it is too.

Same for `ComparisonError` below this one in the code, also only used in a single location.

src/ansys/fluent/core/solver/flobject.py
raph-luc: I think this change should be reverted: it is minor, and this is going to make things unnecessarily more complicated. Leaving `flobject.py` unchanged seems better.

src/ansys/fluent/core/utils/fluent_version.py
```diff
@@ -20,3 +38,99 @@ def get_version_for_file_name(version: Optional[str] = None, session=None):
...
+
+    v24R2 = "24.2.0"
+    v24R1 = "24.1.0"
+    v23R2 = "23.2.0"
+    v23R1 = "23.1.0"
+    v22R2 = "22.2.0"
+    default = ""
+
+    @classmethod
+    def _missing_(cls, version):
```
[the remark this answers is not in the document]
ansaboutin: Here is says class method am I misunderstanding something? 
https://docs.python.org/3/library/enum.html#enum.Enum._missing_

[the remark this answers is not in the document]
mkundu1: We should replace `FluentVersion` with `cls` in the implementation.

tests/conftest.py
```diff
@@ -8,10 +8,10 @@
 import pytest
 
 from ansys.fluent.core.data_model_cache import DataModelCache
-from ansys.fluent.core.launcher.launcher_utils import FluentVersion
+from ansys.fluent.core.utils.fluent_version import FluentVersion
 
 _fluent_versions = list(FluentVersion)
-_fluent_release_version = _fluent_versions[1]
+_fluent_release_version = _fluent_versions[2].value
```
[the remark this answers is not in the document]
mkundu1: This can be `_fluent_versions[1]` or `_fluent_versions[2]` depending on the time of the year. Maybe we can be explicit here: `FluentVersion,v23R2`.

src/ansys/fluent/core/utils/fluent_version.py
seanpearsonuk: @raph-luc it's a specific error message bound to a specifc error type. The idea is that the code that raises the error is no more verbose than necessary.

src/ansys/fluent/core/utils/fluent_version.py
```diff
@@ -20,3 +37,88 @@ def get_version_for_file_name(version: Optional[str] = None, session=None):
...
+    @classmethod
+    def current_release(cls):
+        """Return the version member of the current release.
+
+        Returns
+        -------
+        FluentVersion
+            FluentVersion member corresponding to the latest release.
+        """
+        return cls.v232
```
raph-luc, line 109: To make maintaining this easier, so we don't have to dig into this file when a new Fluent release comes out, we could have a constant indicating the latest released Fluent version maybe in `src/ansys/fluent/core/_version.py` together with the PyFluent version?

Alternatively, a module constant defined at the top of this module with this information would also be better I think.

Then `FluentVersion.current_release()` parses this new variable and does not need to be separately maintained.

The doc build would ideally also reference this new constant, but this would be something to be discussed further and investigated (as doc build version specification is currently hardcoded in the YML files).

src/ansys/fluent/core/utils/fluent_version.py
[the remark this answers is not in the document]
dnwillia-work: I think it looks good now right.

src/ansys/fluent/core/utils/fluent_version.py
dnwillia-work: I think this is resolved now too.

src/ansys/fluent/core/utils/fluent_version.py
ansaboutin: Current release string added to _version.py

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1307,7 +1307,7 @@ def _gethash(obj_info):
 
 
 # pylint: disable=missing-raises-doc
-def get_root(flproxy, version: str = "") -> Group:
+def get_root(flproxy, version: str = "23.2.0") -> Group:
```
raph-luc, line 1310: What happens if we keep `version` as an empty string `""` as it was before?

Is it only `tests\test_flobject.py` that is calling this without specifying the version? If that is the case then this should be fixed there in the tests, and not here

raph-luc: Addressed by https://github.com/ansys/pyfluent/pull/2262/commits/d4e2d4c1aff4aa99a9d68e98442414360607305f

raph-luc (approved):
LGTM, just one last idea/suggestion to refactor the tests so they are easier to maintain, making use of the new FluentVersion: https://github.com/ansys/pyfluent/pull/2298
(can be merged before or after this PR)

seanpearsonuk (approved):

PR #2263, opened 2023-11-21
Expose settings root implicitly.

prmukherj:
Settings root to be exposed implicitly. 

For fluent versions till 23.1 there are a fixed set of settings attributes like: "file", "mesh", "setup", "solution", "results", "parametric_studies", "current_parametric_study", "parameters", "parallel", "report", and "server". For versions 23.2 and above, these attributes are directly taken from root and populated. 

This removes hard coding of the attributes which were affected by changes to settings-api.


seanpearsonuk (approved):

raph-luc (commented):
Question: The way we are providing `dict(root)` here is going to be missing attributes initially, compared to before. For example, with these changes I get:
``` 
>>> session = pf.launch_fluent()
>>> dir(session)
['__bool__', '__call__', '__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__enter__', '__eq__', '__exit__', '__format__', '__ge__', '__getattr__', '__getattribute__', '__getstate__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', '_batch_ops_service', '_build_from_fluent_connection', '_field_data_service', '_lck', '_monitors_service', '_pim_methods', '_preferences', '_reduction_service', '_remote_file_handler', '_root', '_se_service', '_settings_api_root', '_settings_root', '_settings_service', '_sync_from_future', '_system_coupling', '_tui', '_tui_service', '_version', '_workflow', '_workflow_se', 'build_from_fluent_connection', 'connection_properties', 'create_from_server_info_file', 'datamodel_events', 
'datamodel_service_se', 'datamodel_service_tui', 'download', 'error_state', 'events_manager', 'events_service', 'execute_tui', 'exit', 'field_data', 'field_data_streaming', 'field_info', 'fluent_connection', 'force_exit', 'force_exit_container', 'get_fluent_version', 'health_check_service', 'id', 'journal', 'monitors_manager', 'preferences', 'read_case', 'read_case_lightweight', 'reduction', 'rp_vars', 'scheme_eval', 'settings_service', 'start_journal', 'stop_journal', 'svar_data', 'svar_info', 'svar_service', 'system_coupling', 'transcript', 'tui', 'upload', 'version', 'workflow', 'write_case']
```
It is missing `file`, `mesh`, `setup`, `solution`, etc., regardless of existing codegen files (same behavior in v222 and v242). If I do any settings API call (which requests the static info), then after that point the `dir()` method includes the previously missing attributes/objects, as expected.

For comparison, in the `main` branch without these changes doing the same thing I get:
```
 >>> session = pf.launch_fluent()
>>> dir(session)
['__call__', '__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__enter__', '__eq__', '__exit__', '__format__', '__ge__', '__getattribute__', '__getstate__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', '_batch_ops_service', '_build_from_fluent_connection', '_field_data_service', '_lck', '_monitors_service', '_pim_methods', '_preferences', '_reduction_service', '_remote_file_handler', '_root', '_se_service', '_settings_root', '_settings_service', '_sync_from_future', '_system_coupling', '_tui', '_tui_service', '_version', '_workflow', '_workflow_se', 'build_from_fluent_connection', 'connection_properties', 'create_from_server_info_file', 'current_parametric_study', 'datamodel_events', 'datamodel_service_se', 'datamodel_service_tui', 'error_state', 'events_manager', 'events_service', 'execute_tui', 'exit', 'field_data', 'field_data_streaming', 'field_info', 'file', 'fluent_connection', 'force_exit', 'force_exit_container', 'get_fluent_version', 'health_check_service', 'id', 'journal', 'mesh', 'monitors_manager', 'parallel', 'parameters', 'parametric_studies', 'preferences', 'read_case', 'read_case_lightweight', 'reduction', 'report', 'results', 'rp_vars', 'scheme_eval', 'server', 'settings_service', 'setup', 'solution', 'start_journal', 'stop_journal', 'svar_data', 'svar_info', 'svar_service', 'system_coupling', 'transcript', 'tui', 'version', 'workflow', 'write_case']
```
Which as you can see has all of the missing attributes from the start. 

Is this a known limitation of this approach? 

Should we improve the `__dir__()` method so that it gets the attributes and properly populates the dir? As @mkundu1 suggested, I believe this could be done using the codegen files without needing to get the static info from the server.

tests/test_session.py
```diff
@@ -202,7 +226,7 @@ def test_create_session_from_launch_fluent_by_passing_ip_and_port_and_password()
     )
     # check a few dir elements
     session_dir = dir(session)
-    for attr in ("field_data", "field_info", "setup", "solution"):
+    for attr in ("field_data", "field_info"):
```
raph-luc, line 229: As this is testing the root/session object, shouldn't we be going in the other direction and making sure that more of the attributes are correctly exposed by `dir()`, rather than less? 

See also my other question. For example, I think the test should ideally be making sure that `file`, `mesh`, `results`, etc., are also exposed, in addition to `setup` and `solution` and the others; and not ignoring the fact that `setup` and `solution` aren't exposed anymore.

prmukherj:
Good point @raph-luc. I have updated the dir() with static info call. We are avoiding the use of codegen files for solver. Instead of that we are using static info call once to populate the list.

tests/test_session.py
seanpearsonuk: Thanks, @raph-luc. It's important to have good coverage for this behaviour.

prmukherj: @seanpearsonuk and @raph-luc, this test needs to be updated as it is using a static session data. New implementation makes a static info. call. That is the difference. Now, dir(solver) has the settings root methods like file, etc. exposed as well but in a live running session where valid settings api data is present and not a dummy session like the one used in this test.

raph-luc: Got it, thanks for the explanation @prmukherj. 

With that in mind, I would say we should add a new test that instantiates a real Fluent `session` object, and checks if `dir(session)` indeed has all the attributes that we expect it to have.

In addition, as the test we are discussing here `test_create_session_from_launch_fluent_by_passing_ip_and_port_and_password` is not launching Fluent and is also no longer using `launch_fluent` (though I understand it did in the past), we should rename it to something like `test_session_from_mocked_fluent_using_ip_and_port_and_password`, as its current name can be misleading.

raph-luc (approved):

PR #2252, opened 2023-11-15
Factory pattern in launch_fluent

hpohekar:
- [x] Apply Factory Pattern in launch_fluent()
- [x] Test launch_fluent() in local session
- [x] Test launch_fluent() in Ansys Lab
- [x] Add docstrings
- [x] Construct launcher object of given type
- [x] Make launcher interface callable
- [x] Launch Standalone independently
- [x] Add tests for StandaloneLauncher
- [x] Launch PIMLauncher independently
- [x] Add tests for PIMLauncher 
- [x] Launch DockerLauncher independently
- [x] Add tests for DockerLauncher 
- [x] Move each Launcher to separate files.
- [x] Resolve import issues
- [x] Create a separate module for shared launcher utilities
- [x] Test local sessions again after restructuring
- [x] Test pim sessions again after restructuring
- [x] Test `connect_to_fluent` - found 1 bug - resolved in current PR https://github.com/ansys/pyfluent/issues/2272

PR #2269, opened 2023-11-21
Support Slurm options in launch_fluent

mkundu1:
```
>>> slurm = pyfluent.launch_fluent(scheduler_options={"scheduler": "slurm"})
>>> slurm
<ansys.fluent.core.launcher.launcher.SlurmFuture object at 0x2b9238af63b0>
>>> slurm.pending(), slurm.running(), slurm.done()
(True, False, False)  # before Fluent is launched
>>>  slurm.pending(), slurm.running(), slurm.done()
(False, True, False) # after Fluent is launched
>>> session = slurm.result()
>>> session.exit()
>>> slurm.pending(), slurm.running(), slurm.done()
(False, False, True)
```

### DIFF
--- a/src/ansys/fluent/core/fluent_connection.py
+++ b/src/ansys/fluent/core/fluent_connection.py
@@ -269,6 +269,7 @@ class FluentConnection:
         self.error_state = ErrorState()
         self._data_valid = False
         self._channel_str = None
+        self._slurm_job_id = None
         self.finalizer_cbs = []
         if channel is not None:
             self._channel = channel
@@ -299,6 +300,8 @@ class FluentConnection:
         # At this point, the server must be running. If the following check_health()
         # throws, we should not proceed.
         self.health_check_service.check_health()
+
+        self._slurm_job_id = launcher_args and launcher_args.get("slurm_job_id")
 
         self._id = f"session-{next(FluentConnection._id_iter)}"
 
@@ -387,6 +390,9 @@ class FluentConnection:
         )
         FluentConnection._monitor_thread.cbs.append(self._finalizer)
 
+    def _close_slurm(self):
+        subprocess.run(["scancel", f"{self._slurm_job_id}"])
+
     def force_exit(self):
         """Immediately terminates the Fluent client, losing unsaved progress and data.
 
@@ -443,6 +449,9 @@ class FluentConnection:
                 stdout=subprocess.DEVNULL,
                 stderr=subprocess.DEVNULL,
             )
+        elif self._slurm_job_id:
+            logger.debug("Fluent running inside Slurm, closing Slurm session...")
+            self._close_slurm()
         else:
             logger.error("Could not find cleanup file.")
 
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -80,7 +80,7 @@ def launch_fluent(
     precision: Optional[str] = None,
     processor_count: Optional[int] = None,
     journal_file_names: Union[None, str, list[str]] = None,
-    start_timeout: int = 60,
+    start_timeout: Optional[int] = None,
     additional_arguments: Optional[str] = "",
     env: Optional[Dict[str, Any]] = None,
     start_container: Optional[bool] = None,
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -126,7 +126,8 @@ def launch_fluent(
         journal(s). The default is ``None``.
     start_timeout : int, optional
         Maximum allowable time in seconds for connecting to the Fluent
-        server. The default is ``60``.
+        server. The default is ``60`` if Fluent is launched outside a Slurm environment,
+        no timeout if Fluent is launched within a Slurm environment.
     additional_arguments : str, optional
         Additional arguments to send to Fluent as a string in the same
         format they are normally passed to Fluent on the command line.
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -21,6 +21,7 @@ from ansys.fluent.core.launcher.launcher_utils import (
     _process_kwargs,
 )
 from ansys.fluent.core.launcher.pim_launcher import PIMLauncher
+from ansys.fluent.core.launcher.slurm_launcher import SlurmFuture, SlurmLauncher
 from ansys.fluent.core.launcher.standalone_launcher import StandaloneLauncher
 import ansys.fluent.core.launcher.watchdog as watchdog
 from ansys.fluent.core.session_meshing import Meshing
@@ -52,7 +53,7 @@ def create_launcher(fluent_launch_mode: str = None, **kwargs):
     DisallowedValuesError
         If an unknown Fluent launch mode is passed.
     """
-    allowed_options = ["container", "pim", "standalone"]
+    allowed_options = ["container", "pim", "standalone", "slurm"]
     if (
         not isinstance(fluent_launch_mode, str)
         or str(fluent_launch_mode) not in allowed_options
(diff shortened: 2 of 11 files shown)


src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -672,6 +814,10 @@ def launch_fluent(
         which means an independent watchdog process is run to ensure
         that any local GUI-less Fluent sessions started by PyFluent are properly closed (or killed if frozen)
         when the current Python process ends.
+    slurm : bool, optional
```
mkundu1, line 817: can make it as `slurm_dict` with other slurm options similar to `container_dict`

seanpearsonuk:
@mkundu1 We just need to be aware if there are potential conflicts with @hpohekar's branch.

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -558,6 +580,122 @@ def __init__(self, launch_string):
         super().__init__(details)
 
 
+class SlurmFuture:
```
mkundu1, line 583: Closely follow the [Future interface](https://docs.python.org/3/library/concurrent.futures.html#future-objects)

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -296,9 +297,16 @@ def __init__(
...
         )
-        # At this point, the server must be running. If the following check_health()
-        # throws, we should not proceed.
-        self.health_check_service.check_health()
+        if launcher_args:
+            slurm_job_id = launcher_args.get("slurm_job_id")
+            if slurm_job_id:
+                self.slurm_job_id = slurm_job_id
+                while not self.health_check_service.is_serving:
+                    time.sleep(1)
```
raph-luc, line 305: I believe we should add a timeout here, otherwise this can hang indefinitely

mkundu1: This is not executed on the same thread where launch_fluent is called. User will be able to cancel this from that thread.
```
slurm = pyfluent.launch_fluent(scheduler_options={"scheduler": "slurm", "port": 29000})
slurm.cancel()
```

raph-luc: I see, but it would still be good practice to add a timeout here, no? 

If the job is already in the "RUNNING" state and it hasn't been able to connect after say 60 seconds, it should probably timeout and raise an exception as there is something certainly wrong?

mkundu1: Right, and we should automatically cancel the job in that case.

hpohekar:
@mkundu1 @seanpearsonuk 

API codegen is failing for 24R2 in this and my PR.

This PR - https://github.com/ansys/pyfluent/actions/runs/6957536941/job/18945517597?pr=2269

image

`-------------------------------------------`

My PR - https://github.com/ansys/pyfluent/actions/runs/6967211289/job/18958781377?pr=2252

image

hpohekar:
> > @mkundu1 @seanpearsonuk
> > API codegen is failing for 24R2 in this and Factory pattern PR.
> > This PR - https://github.com/ansys/pyfluent/actions/runs/6957536941/job/18945517597?pr=2269
> > image
> > `-------------------------------------------`
> > Factory pattern PR - https://github.com/ansys/pyfluent/actions/runs/6967211289/job/18958781377?pr=2252
> > #2252
> > datamodelgen.py is executing successfully.
> > image
> 
> This should not block merging, I already have a Fluent side PR fix the codegen.

@mkundu1 Thank you.

src/ansys/fluent/core/session.py
```diff
@@ -240,7 +246,17 @@ def create_from_server_info_file(
...
-        ip, port, password = _parse_server_info_file(server_info_file_name)
+        scheduler_options = (
+            connection_kwargs
+            and connection_kwargs.get("launcher_args")
+            and connection_kwargs["launcher_args"].get("scheduler_options")
+        )
+        if scheduler_options and scheduler_options["scheduler"] == "slurm":
+            ip = _parse_slurm_host_file(server_info_file_name)
+            port = scheduler_options["port"]
+            password = None
```
raph-luc, line 257: Is this only because the default server-info file location is put in a tmp dir by default in PyFluent? In a scheduler environment, I believe we can just write the sifile to the current directory, as typically the submission node home path will be mounted in the same way to the compute node, as will any working directories being used.

Setting `-sifile=server-info.txt` by default in scheduler environments should work fine in most cases, we can print something like `logger.warning("Writing server information file to the current working directory")`, and if for some reason an user needs a custom specification, then they can specify it (or ask for help to know how to do it, at least we have the warning so they know what to ask)

Currently in a Slurm environment if you manually launch Fluent with `-sifile=server-info.txt` and then `connect_to_fluent(server_info_filepath="serverinfo.txt")`, it already works by default without having to specify anything else

mkundu1: Right, I overlooked the fact that the `pwd` is linked (there is a `-scheduler_workdir` launch option, but we can implement that later). So, we can simply write the server-info file in `pwd` and parse it as usual in this first implementation.

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -583,8 +720,9 @@ def launch_fluent(
     cwd: Optional[str] = None,
     topy: Optional[Union[str, list]] = None,
     start_watchdog: Optional[bool] = None,
+    scheduler_options: dict = None,
```
raph-luc, line 723:
```suggestion
    scheduler_options: Optional[dict] = None,
```

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -558,6 +579,122 @@ def __init__(self, launch_string):
...
+
+        logger.debug(f"Launching Fluent with command: {launch_cmd}")
+        proc = subprocess.Popen(launch_cmd, **kwargs)
+        slurm_job_id = _get_slurm_job_id(proc)
+        logger.info(f"Slurm job id = {slurm_job_id}")
+        self._argvals["slurm_job_id"] = slurm_job_id
+        return slurm_job_id
+
+    def _launch(self, slurm_job_id) -> Union[Meshing, PureMeshing, Solver, SolverIcing]:
+        _await_fluent_launch(self._server_info_file_name, -1, self._sifile_last_mtime)
```
raph-luc, line 680: The way that `_await_fluent_launch` is being called here, there is no timeout and the logging calls aren't ideal:
image

Should we support the existing default timeout (currently 1 minute), with user-configurable Slurm queue timeout through the existing `launch_fluent(start_timeout=...)` argument? Could change the behavior so that setting this timeout to 0 means it will wait indefinitely (with adjusted logging reports without the "X seconds remaining" part), or have a different default behavior when launching Fluent in a Slurm queue (but still respecting what the user specifies in `start_timeout`).

raph-luc (approved):

src/ansys/fluent/core/launcher/launcher.py
mkundu1: For Slurm, we need to set the default for `start_timeout` such that it will wait indefinitely. I don't know what is a good value for a default, I've heard things can wait for a week.

mkundu1: I'm not sure what is a good value for the default, so I'll set the default of `start_timeout` such that it will wait indefinitely.

src/ansys/fluent/core/launcher/launcher_utils.py
```diff
@@ -347,7 +359,10 @@ def _await_fluent_launch(
             raise TimeoutError("The launch process has timed out.")
         time.sleep(1)
         start_timeout -= 1
-        logger.info(f"Waiting for Fluent to launch...{start_timeout} seconds remaining")
+        if start_timeout >= 0:
+            logger.info(
+                f"Waiting for Fluent to launch...{start_timeout} seconds remaining"
+            )
```
raph-luc, line 365: Minor suggestion, I think it will be helpful to have a logging call indicating that PyFluent is now going to be waiting for Fluent to launch, regardless of timeout configuration (particularly now that we will support waiting indefinitely).
```suggestion
        logger.info(f"Waiting for Fluent to launch...")
        if start_timeout >= 0:
            logger.info(
                f"...{start_timeout} seconds remaining"
            )
```

seanpearsonuk (approved):

PR #2270, opened 2023-11-22
Refactor SVARData to make initialization light-weight.

prmukherj:
Initialization of Svar Data is made light-weight using a flag.


seanpearsonuk:
@prmukherj please can you add a test. The issue gives a hint about the kind of stuff to do. Also, would it help to add data validity checks before trying access?

cc @ajain-work

prmukherj:
@seanpearsonuk, test for SVAR Data is there in test_svars.py. Do we need any other specific test?

seanpearsonuk:
> @seanpearsonuk, test for SVAR Data is there in test_svars.py. Do we need any other specific test?

re the scenario reported in the issue. That needs to be callable without data being available. Everything must work safely (such as invalid calls) without data available. Must not crash.

src/ansys/fluent/core/services/svar.py
```diff
@@ -435,11 +435,12 @@ def __init__(
         self._service = service
         self._svar_info = svar_info
```
seanpearsonuk, line 437: In the old version, we do the work once in the constructor. In the new version, we do the work on demand and we want to ensure that we still only do it once. But, there seems to be no check about that.

prmukherj: Right @seanpearsonuk. However, we don't want to do it once as it fetches the Svars Info every time there are some updates to the data. That's why the check has been removed now. Are we missing something?

cc. @ajain-work

seanpearsonuk: OK, I was just comparing this with the original code, which I suppose was wrong in this respect.

prmukherj: Yes, @seanpearsonuk. Thanks to @ajain-work for pointing out this issue :).

seanpearsonuk: OK, thanks for clarifying. It completely makes sense - it was only that it was a bit mysterious. Anyway, no problem, let's proceed.

Issue #2275, opened 2023-11-23
Mixture/species transport model API should replicate behavior of materials change/create API


ansjsia:
### 📝 Description of the feature

The settings API calls to create new mixtures for the species transport model should be made consistent with that of the materials change/create API. Currently, it is not possible to create new mixtures without resorting to TUI commands.

### 💡 Steps for implementing the feature

These are the changes that would need to be implemented, as far as I can tell at this time:
- Fix the `volumetric_species.delete()` method (https://github.com/ansys/pyfluent/issues/2259), so that existing species in a mixture can be deleted. Currently, it is only possible to add species.
- Create a settings API method to copy a mixture from Fluent's database, similar to `copy_by_name()` for species. Currently, the only non-GUI way is to execute `/define/materials/change-create` in the TUI.
- Allow new mixtures to be created from an empty dictionary by executing `root.setup.materials.mixture['my-mixture'] = {}` in the same way that you can for species (this currently throws an error: https://github.com/ansys/pyfluent/issues/2274). `mixture-template` could be used as the default for any empty fields.

dnwillia-work:
Thanks @ansjsia.  I did spend a tiny bit of time trying this myself.  I loaded up a single phase case without the species module activated and found the following:
```
#
# Copy the materials in from the database to make a custom mixture (this works)
#
solver.setup.materials.database.copy_by_name(name="methane", type="fluid")
solver.setup.materials.database.copy_by_name(name="propane", type="fluid")

#
# Turn on species, it works but it also deletes the materials I just copied in from the database
#
solver.setup.models.species.model.option = "species-transport"

#
# Copy the materials again, this seems to work ok
#
solver.setup.materials.database.copy_by_name(name="methane", type="fluid")
solver.setup.materials.database.copy_by_name(name="propane", type="fluid")

#
# Make handles to the newly copied materials
#
methane = solver.setup.materials.fluid["methane"]
propane = solver.setup.materials.fluid["propane"]

#
# Now try and make a mixutre out of those two, this does not work.
#
# This gives that error message and leaves a mixture called 'comgas' in the group
# with three materials: water-vapor, oxygen and nitrogen
#
solver.setup.materials.mixture["comgas"] = { "meth" : methane, "prop" : propane }
RuntimeError: Key 'meth' is invalid

#
# I'm able to delete the incorrectly configured mixture
#
>>> solver.setup.materials.mixture.delete(name_list=["comgas"])

#
# Tried creating like this but it also does not work
#
[...]

Issue #2274, opened 2023-11-23
Creating new mixture with empty dictionary breaks mixture API


ansjsia:
### 🔍 Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### 🐞 Description of the bug

Using `solver.setup.materials.mixture['my-mixture'] = {}` makes all future attempts to get the state of the `mixture` object throw:
```
Error: CAR: invalid argument [1]: wrong type [not a pair]
Error Object: #f
...
RuntimeError: CAR: invalid argument [1]: wrong type [not a pair]
Error Object: #f
```

This is also inconsistent with the behavior of the API for creating new fluid or solid materials, where setting the state using an empty dictionary would populate the fields with the default material (air for fluids, aluminum for solids).

### 📝 Steps to reproduce

1. Activate the species transport model: `root.setup.models.species.model.option = 'species-transport'`
2. Run `root.setup.materials.mixture['my-mixture'] = {}`
3. Calling `root.setup.materials.mixture['my-mixture'].get_state()` or `root.setup.materials.mixture.get_state()` will throw a `RuntimeError`.

### 💻 Which operating system are you using?

Windows

### 📀 Which ANSYS version are you using?

24R1

### 🐍 Which Python version are you using?

3.11

### 📦 Installed packages

```shell
aiohttp==3.8.6
aiosignal==1.3.1
ansys-api-fluent==0.3.18
ansys-api-platform-instancemanagement==1.0.0
ansys-api-systemcoupling==0.1.0
-e git+https://github.com/ansys/pyfluent.git@c812ed4ce4fd03c4c02af9d88a2d47912c2c38a5#egg=ansys_fluent_core
ansys-fluent-visualization==0.7.1
ansys-platform-instancemanagement==1.1.2
-e git+https://github.com/ansys/pysystem-coupling.git@a445b8a153839b0abbe518be9b87099dde9470e2#egg=ansys_systemcoupling_core
anyio==4.0.0
appdirs==1.4.4
argon2-cffi==23.1.0
argon2-cffi-bindings==21.2.0
arrow==1.3.0
asttokens==2.4.1
async-lru==2.0.4
async-timeout==4.0.3
attrs==23.1.0
Babel==2.13.1
beautifulsoup4==4.12.2
bleach==6.1.0
certifi==2023.7.22
cffi==1.16.0
charset-normalizer==3.3.2
colorama==0.4.6
comm==0.1.4
contourpy==1.2.0
[...]

Issue #2259, opened 2023-11-20
Error in deleting species from mixture template (species transport)


sravanansys:
### 🔍 Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### 🐞 Description of the bug

I am activating the species transport model and adding additional species to it. But when I am trying to delete the unnecessary species I am getting an error.

`>>> fluent.setup.materials.mixture['mixture-template'].species.volumetric_species.delete(name_list=['o2'])

Error: eval: unbound variable
Error Object: name
Traceback (most recent call last):
  File "<pyshell#83>", line 1, in <module>
    fluent.setup.materials.mixture['mixture-template'].species.volumetric_species.delete(name_list=['o2'])
  File "D:\sravan\pyansys\pyfluent\src\ansys\fluent\core\solver\flobject.py", line 991, in __call__
    return self.flproxy.execute_cmd(self._parent.path, self.obj_name, **newkwds)
  File "D:\sravan\pyansys\pyfluent\src\ansys\fluent\core\services\settings.py", line 122, in _fn
    return fn(self, *args, **kwds)
  File "D:\sravan\pyansys\pyfluent\src\ansys\fluent\core\services\settings.py", line 320, in execute_cmd
    response = self._service_impl.execute_cmd(request)
  File "D:\sravan\pyansys\pyfluent\src\ansys\fluent\core\services\error_handler.py", line 15, in func
    raise RuntimeError(ex.details()) from None
RuntimeError: eval: unbound variable
Error Object: name`

### 📝 Steps to reproduce

Activate species transport model and try to delete one species from the mixture template.

`fluent.setup.models.species.model.option='species-transport'
fluent.setup.models.species.options.diffusion_energy_source=False
fluent.setup.materials.mixture['mixture-template'].species.volumetric_species.delete(name_list=['o2'])`

### 💻 Which operating system are you using?

Windows

### 📀 Which ANSYS version are you using?

2024R1

### 🐍 Which Python version are you using?

3.9

### 📦 Installed packages

```shell
ansys-api-fluent==0.3.18
ansys-api-platform-instancemanagement==1.0.0
-e git+https://github.com/ansys/pyfluent.git@5c81a342aff8cf995babed8011661bd16122265a#egg=ansys_fluent_core
-e git+https://github.com/ansys/pyfluent-visualization.git@b907538e399406d96da587a38d8518697746593d#egg=ansys_fluent_visualization
[...]

PR #2280, opened 2023-11-27
Add test for species transport model

ansjsia:
Test the following features of the species transport model:

- Turning on the species transport model
- Checking the mixtures object command list
- Creating a custom mixture with custom species (using the TUI)
- Copying an existing mixture
- Reassigning a cellzone material to a different mixture
- Deleting an existing mixture

Currently, the species transport model cannot be used without resorting to TUI commands due to the following open issues with the settings API:

- Creating/copying mixtures: https://github.com/ansys/pyfluent/issues/2275
- Listing mixtures, species, and properties: https://github.com/ansys/pyfluent/issues/2277
- Deleting a species from a mixture: https://github.com/ansys/pyfluent/issues/2259
- Creating a mixture from an empty dictionary: https://github.com/ansys/pyfluent/issues/2274


tests/test_solvermode/test_speciestransport.py
```diff
@@ -0,0 +1,61 @@
...
+@pytest.mark.quick
+@pytest.mark.setup
+@pytest.mark.fluent_version("latest")
+def test_species_material(load_mixing_elbow_mesh):
+    solver_session = load_mixing_elbow_mesh
+
+    # Test turning on species transport model
+    species_mdl = solver_session.setup.models.species.model
+
+    assert species_mdl.option.get_state() == "off"
```
dnwillia-work, line 13: Do you have to call `.get_state()` here?  I would think this should work:
`assert species_mdl.option == "off"`

tests/test_solvermode/test_speciestransport.py
dnwillia-work, line 1: You should name the module `test_species_model.py`.  A bit shorter and add an underscore between the two words.

tests/test_solvermode/test_speciestransport.py
```diff
@@ -0,0 +1,61 @@
+import pytest
+
+
+@pytest.mark.quick
+@pytest.mark.setup
+@pytest.mark.fluent_version("latest")
+def test_species_material(load_mixing_elbow_mesh):
```
dnwillia-work, line 7: Maybe name it `test_create_mixture`?

ansjsia: What if we called it `test_change_create_mixture` since I'm also editing, copying, and deleting objects?

raph-luc: I would say `test_species_materials` (added `s` at the end) is slightly better, as you are testing `setup.species` and `setup.materials` (`mixture` is under `materials` as well). 

Alternatively, if you would like to compete for the longest test names with Mainak, a very detailed `test_species_materials_mixture_create_change_copy_delete` would be great, as these are all the operations you are testing here

But any one of these suggested sound OK to me

tests/test_solvermode/test_speciestransport.py
raph-luc: What does `species_mdl.option` return? Compare it to that instead of the string `"off"`

ansjsia: `species_mdl.option` is of type `ansys.fluent.core.solver.settings_241.option.option`. Calling `species_mdl.option()` returns a `str`.

raph-luc: Right, thanks for the details. And does `species_mdl.option()` return the same as `species_mdl.option.get_state()`?

You are correct that using `get_state()` is more explicit, but in my understanding PyFluent users should generally not need these more API-backend calls such as `get_state`, `get_attr` and so on. In other words, we do not really want the more explicit commands here, we want to showcase the more simplified/easier-to-use interfaces if possible

ansjsia: Yes, just calling the object itself returns the same thing. I'll replace the `get_state()` call with just `()`.

tests/test_solvermode/test_speciestransport.py
dnwillia-work: It's true that the test is asserting a number of things but for me the main point of it is to illustrate how to create a new mixture from scratch.  For me a longer name is a bit cluttered and `test_species_materials` does not seem accurate enough.

raph-luc: I think that moving the creation part into a function and breaking this down into two or three separate tests might be a little excessive, but could work as well.

And I agree, I previously meant that the suggested `test_create_mixture` or `test_change_create_mixture` sound fine to me too

ansjsia:
@dnwillia-work @raph-luc For your consideration in the latest commit. I've renamed the file to `test_species_model.py` and split it up into multiple tests. I created a fixture such that only one solver instance would be used across all tests, ensuring that we only start and shut down Fluent once for all six tests.

The tests are:
1. Activate the species transport model
2. Check the command list
3. Create a custom mixture from custom species
4. Duplicate a mixture
5. Change the mixture assigned to a cellzone
6. Delete a mixture

The problem is that if one of these tests fails, all the following tests will fail as well, like dominoes. I'm not sure if this is acceptable. If it isn't, a potential solution is to implement recognizing an `incremental` test marker in our `conftests.py` file that would skip all remaining tests in the class if one fails: https://docs.pytest.org/en/latest/example/simple.html#incremental-testing-test-steps

tests/test_solvermode/test_speciestransport.py
raph-luc, line 1: Make sure to remove this file from the branch, as it is no longer applicable

tests/test_solvermode/test_species_model.py
```diff
@@ -0,0 +1,103 @@
...
+        "pyfluent/mixing_elbow",
+    )
+    solver_session = pyfluent.launch_fluent(mode="solver")
+    solver_session.file.read_case(file_name=cas_path)
+    yield solver_session
+    solver_session.exit()
+
+
+@pytest.mark.quick
+@pytest.mark.setup
```
raph-luc, line 20: We are not using these markers currently, so IMO they are just unnecessary copy-pasted repetition. I would remove them:
```suggestion
```

tests/test_solvermode/test_species_model.py
```diff
@@ -0,0 +1,103 @@
...
+@pytest.fixture(scope="module")
+def solver_session():
+    cas_path = download_file(
+        "mixing_elbow.cas.h5",
+        "pyfluent/mixing_elbow",
+    )
+    solver_session = pyfluent.launch_fluent(mode="solver")
+    solver_session.file.read_case(file_name=cas_path)
+    yield solver_session
+    solver_session.exit()
```
raph-luc, line 16: We have usually been specifying fixtures in separate files as they are re-used in multiple places, but as this one is specific to this test module, having it here makes sense to me too. 

Also, good that you are re-using one session for multiple tests, rather than launching a separate session for each test

tests/test_solvermode/test_speciestransport.py
dnwillia-work, line 1: I guess you don't need this file now?

tests/test_solvermode/test_species_model.py
dnwillia-work: What do you think, should we also remove them from `pyproject.toml`?  That's where they are defined:

https://github.com/ansys/pyfluent/blob/53aa0270d1ab0d9bdc91c0b6d749297357656c8a/pyproject.toml#L16

tests/test_solvermode/test_species_model.py
```diff
@@ -0,0 +1,103 @@
...
+    assert elbow_zone.material() == "custom-mixture-copy"
+
+
+@pytest.mark.quick
+@pytest.mark.setup
+@pytest.mark.fluent_version(">=24.1")
+def test_delete_mixture(solver_session):
+    materials = solver_session.setup.materials
+    materials.mixture.delete(name_list=["custom-mixture"])
+    assert "custom-mixture" not in materials.mixture.keys()
```
raph-luc, line 103: The way these are structured now, these aren't "unit" tests, i.e. they aren't isolated. 

If I run `test_delete_mixture` on its own for example, it is always going to fail, and this is not how we've been structuring the tests. Here they only work if all the tests run, and if they run in the specific order defined in this file.

I think keeping everything as one test like we had before was ultimately better

tests/test_solvermode/test_speciestransport.py
dnwillia-work: Yeah, the commits don't lie, looks like you did not.  Ideally what you would have done in this case is `git mv old_name.py new_name.py` then implemented additional changes into `new_name.py`.

tests/test_solvermode/test_species_model.py
dnwillia-work: Yeah, agree, unit tests must work in isolation and should not have dependencies on other unit tests.

tests/test_solvermode/test_species_model.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+@pytest.fixture(scope="module")
+def solver_session():
+    cas_path = download_file(
+        "mixing_elbow.cas.h5",
+        "pyfluent/mixing_elbow",
+    )
+    solver_session = pyfluent.launch_fluent(mode="solver")
+    solver_session.file.read_case(file_name=cas_path)
```
dnwillia-work, line 14: May as well add the lightweight case read option here.

tests/test_solvermode/test_species_model.py
raph-luc: That would make sense to me, the only markers we are actually using in the GitHub CIs are: `nightly`, `fluent-version`, `solvermode`, `codegen_required` and `standalone`.

In my understanding, we could remove all of these as they aren't being used, and aren't providing any helpful info that the test name or module name don't already provide:
https://github.com/ansys/pyfluent/blob/53aa0270d1ab0d9bdc91c0b6d749297357656c8a/pyproject.toml#L17-L22

tests/test_solvermode/test_speciestransport.py
raph-luc: If you deleted it manually using File Explorer for example, you'd need to add the file delete command to the commit with `git add tests/test_solvermode/test_speciestransport.py`

Either moving it with `git mv` as Dan suggested, or removing it directly with `git rm` (which deletes the file and adds it to the commit) would also have worked

tests/test_solvermode/test_species_model.py
ansjsia: Yeah, I created a specific fixture for this because the original mixing elbow fixture is function-scoped, but I needed a module-scoped fixture to run multiple tests on the same solver object.

ansjsia:
@raph-luc @dnwillia-work I've recombined everything into one test in the latest commit. Hopefully the name and comments are sufficiently descriptive.

tests/test_solvermode/test_species_model.py
raph-luc: Discussion in https://github.com/ansys/pyfluent/discussions/2287

raph-luc (approved):

tests/test_solvermode/test_species_model.py
dnwillia-work: OK, so lets leave `setup` but remove everything else that are not used currently.

tests/test_solvermode/test_species_model.py
dnwillia-work: Yeah, sorry, I mean case settings only mode:
https://ansyshelp.ansys.com/account/secured?returnurl=/Views/Secured/corp/v232/en/flu_ug/flu_ug_CaseSettingsOnly.html?q=case%20settings

raph-luc: Ohhh, that makes perfect sense, thanks for the reference!

dnwillia-work: yeah, I did not design this part of the API.  You should add an issue about your feedback :-).

dnwillia-work: @ansjsia I guess this thread is not relevant now since you removed the fixture.  Maybe you can add a fixture that provides lightweight case read mode later.

dnwillia-work (approved):

tests/test_solvermode/test_species_model.py
```diff
@@ -0,0 +1,72 @@
+import pytest
+
+
+@pytest.mark.setup
+@pytest.mark.fluent_version("latest")
```
dnwillia-work, line 5: You missed setting 24.1 or newer here.

ansjsia: Oops, good catch. Fixed.

tests/test_solvermode/test_species_model.py
ansjsia: Absolutely, I'll make a separate PR for this.

ansjsia:
@dnwillia-work @raph-luc The unit testing check for 23R2 is stalling (timing out after 6 hours) consistently on the test `test_meshing_workflow.py::test_watertight_workflow_children`, which seems completely unrelated to this PR. It seems to fail when I run it locally on 23R2. Any insight?

raph-luc:
@ansjsia I believe it wasn't `test_watertight_workflow_children` that stalled (as the logs say that it passed, and it should pass if you try it locally as well), it was the test after that one: `test_watertight_workflow_dynamic_interface`.

The team is aware of these tests that have been randomly stalling, we aren't sure of what is going on yet.

I set the failed test to re-run, let's see if it hangs again or if it passes. Changes in `main` branch are set to skip this test, we can merge them here if it fails again

PR #2282, opened 2023-11-28
Specify the data path through env var when the default is not writable (by docker process)

mkundu1:


mkundu1:
> What is the use-case of this additional env var? The correct way to launch Fluent containers from PyFluent is documented here: https://fluent.docs.pyansys.com/version/stable/api/general/launcher/fluent_container.html
> 
> Going to discuss in the issue tracker

Yes, this is already covered in config_dict, the github issue is something else (in addition to the mount dir issue)

PR #2285, opened 2023-11-29
Hide solver lite mode

hpohekar:
`SolverLite` mode is missing from Fluent launch modes.

We have mentioned here about SolverLite mode but we don't support it yet.

image


hpohekar:
> Should we also remove the [src\ansys\fluent\core\session_solver_lite.py](https://github.com/ansys/pyfluent/blob/main/src/ansys/fluent/core/session_solver_lite.py) file? There doesn't seem to be anything specific implemented there yet, it seems to be just a placeholder, and it is not being used or referenced anywhere else in the code.

We just want to hide it for now. We don't want to remove it.

prmukherj (approved):

Issue #2206, opened 2023-11-03
Settings classes for InputFile(s)/OutputFile(s)


mkundu1:
From the chat:

- [x] Add classes for InputFile(s)/OutputFile(s) in settings API - RL
- [x] Generate Python classes for InputFile(s)/OutputFile(s) during settings codegen. It can be a command/query argument or parameter in the settings API. The upload/download operations will be incorporated within the command/query's call method or the parameter's set_state method. - HP
- [ ] Generate Python classes for InputFile(s)/OutputFile(s) during datamodel codegen and incorporate the upload/download operations. - HP
- [x] Take the simple workflow: geometry -> MESHER -> mesh -> SOLVER -> data. Up to now, we are assuming that everything written is downloaded, but if you consider that the mesh is an intermediate file and can be very large, then downloading might not be desirable for multiple reasons. Apart from use of network bandwidth and local storage, it can delay the workflow significantly if executed as a synchronous step. Perhaps in PyFluent, we extend each one of these APIs with an argument that allows up/download to be disabled for the particular file operation. - HP

Issue #2195, opened 2023-11-02
Generalising the upload download capability across APIs


seanpearsonuk:
We would like to generalise this capability. Right now it is available for some methods at the session level. For it to work across APIs, we need to do something to identify which arguments are relevant. One option is for APIs to identify input files and output files via static attributes which we could then use in codegen to generate classes derived from InputFile and OutputFile which in turn would be derived from String base classes. Where we are getting the arguments in the relevant `__call__` method in flobject.py, we would need to recognize instances of those types and take the necessary additional actions. We would need to the same thing for data model APIs too. We can prototype this before actually requesting any API changes (e.g., we can hack the codegen so that it "finds" those attributes present for some known paths; for data model, we can actually change the FDL before running codegen). Please let me have your feedback on these thoughts.  

EDIT 09 Jan 2024, Agreed approach:
- don't mention filePurpose in *.proto. Instead introduce a map of attribute name to value at each info _node_
- file types are input, output an inout
- persist with the approach of generating classes based on the filePurpose attribute value

seanpearsonuk:
@hpohekar for simplicity, let's extend this issue to summarise the outstanding pieces:
- support for in-out files. Consensus is that this is supported by changing the attribute type transmitted from the server from string to string list. The only values in that list are the current two values for input and output
- simplify the approach for plugging in a file transfer service (eliminate the wrapper)
- move the _test transfer service_ from the src to the tests folder. Such test-specific code should always live there
- allow injection through `launch_fluent` and `launch_remote_fluent`:
```
class MyFileService:
  def upload(self, file_name):
    # ....

  def download(self, file_name):
    # ....

file_transfer_service = MyFileService(...)
session = pf.launch_fluent(file_transfer_service=file_transfer_service)
```

PR #2289, opened 2023-11-30
feat: Upload/Download across all API

hpohekar:
Closes - https://github.com/ansys/pyfluent/issues/2206 and https://github.com/ansys/pyfluent/issues/2195

For cloud platform like the Ansys Lab we need to upload file (case/data/mesh) to server to be able to read it in commands like read_case/ReadCase or read_mesh/ReadMesh and we need to download file from the server after executing commands like write_case/WriteCase or write_mesh/WriteMesh etc.  

Right now we have `<session>.upload()` and `<session>.download()` methods to upload/download files explicitly but we want to handle this internally.

Therefore to achieve this we have added filePurpose/filepurpose attribute to distinguish between input/output file types. Based on the file purpose we are generating settings classes by adding mixin class to the bases of existing file classes.

So if file purpose is input then we will upload the file and if it is output then we will download it.

We have integrated `remote_file_handler` in both settings and datamodel `__call__` methods to achieve this capability seamlessly.

We have added a `TransferRequestRecorder` class as a mock file service class to replicate Ansys Lab scenario locally.


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1007,6 +1033,20 @@ class Command(Action):
 
     def __call__(self, **kwds):
         """Call a query with the specified keyword arguments."""
+        if hasattr(self, "file_name"):
+            file_purpose = self.file_name.get_attr(_InlineConstants.file_purpose)
+        elif hasattr(self, "file_name_list"):
+            file_purpose = self.file_name_list.get_attr(_InlineConstants.file_purpose)
+        elif hasattr(self, "filename"):
+            file_purpose = self.filename.get_attr(_InlineConstants.file_purpose)
+        print(f"\nfile_purpose = {file_purpose}\n")
```
hpohekar, line 1042: We are going to remove this line.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1007,6 +1033,20 @@ class Command(Action):
...
+
+        if file_purpose == "input" and pypim.is_configured():
+            self.remote_file_handler.upload(file_name=kwds["file_name"])
+            print(
+                f"\nfile - {kwds['file_name']} uploaded for {str(self.__class__.__name__)}.\n"
+            )
```
hpohekar, line 1048: We are going to remove this line.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1021,7 +1061,15 @@ def __call__(self, **kwds):
...
                 if response in ["n", "N", "no"]:
                     return
-        return self.flproxy.execute_cmd(self._parent.path, self.obj_name, **newkwds)
+        cmd = self.flproxy.execute_cmd(self._parent.path, self.obj_name, **newkwds)
+
+        if file_purpose == "output" and pypim.is_configured():
+            self.remote_file_handler.download(file_name=kwds["file_name"])
+            print(
+                f"file - {kwds['file_name']} downloaded for {str(self.__class__.__name__)}."
+            )
```
hpohekar, line 1070: We are going to remove this line.

raph-luc:
> We are unable to test it in the Ansys Lab as it does not support 24R2 yet.

@hpohekar Right, can we mock/test the implementation in another way that does not rely on Ansys Lab? Otherwise we will indeed need to wait to properly support PyPIM/Ansys Lab.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1007,6 +1033,20 @@ class Command(Action):
...
         """Call a query with the specified keyword arguments."""
+        if hasattr(self, "file_name"):
+            file_purpose = self.file_name.get_attr(_InlineConstants.file_purpose)
+        elif hasattr(self, "file_name_list"):
+            file_purpose = self.file_name_list.get_attr(_InlineConstants.file_purpose)
+        elif hasattr(self, "filename"):
+            file_purpose = self.filename.get_attr(_InlineConstants.file_purpose)
+        print(f"\nfile_purpose = {file_purpose}\n")
+
+        if file_purpose == "input" and pypim.is_configured():
```
raph-luc, line 1044: I believe we should not be adding such specific code in `flobject.py`, can we hook or wrap around the `file_purpose` attribute at a higher level?

mkundu1, line 1044: `pypim.is_configured()` will not work in pyconsole. `remote_file_handler` can provide a method to check whether remote is configured or not.

hpohekar:
> > We are unable to test it in the Ansys Lab as it does not support 24R2 yet.
> 
> @hpohekar Right, can we mock/test the implementation in another way that does not rely on Ansys Lab? Otherwise we will indeed need to wait to properly support PyPIM/Ansys Lab.

@raph-luc can we mock/test the implementation in another way that does not rely on Ansys Lab? - Yes, I am working on it.

src/ansys/fluent/core/solver/flobject.py
hpohekar: Yes, sure.

hpohekar: @mkundu1 @seanpearsonuk 

I installed `ansys-platform-instancemanagement` in pyconsole as follow.

`D:\ANSYSDev\vNNN\commonfiles\CPython\3_10\winx64\Release\python\python.exe -m pip install ansys-platform-instancemanagement`

Now we can check `pypim.is_configured()` in pyconsole.

image

We can request build team to install `ansys-platform-instancemanagement` in both `thirdparty` and `commonfiles` python as it is one of the basic requirement to support upload/download across all API.

hpohekar: > `pypim.is_configured()` will not work in pyconsole. `remote_file_handler` can provide a method to check whether remote is configured or not.

@mkundu1 Is there any other reason due to which `pypim.is_configured()` will not work in pyconsole ?

hpohekar: We are using remote file handler's method now.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1006,7 +1030,18 @@ class Command(Action):
...
+        """Call a command with the specified keyword arguments."""
+        file_purpose = None
+        if hasattr(self, "file_name"):
+            file_purpose = self.file_name.get_attr(_InlineConstants.file_purpose)
+        elif hasattr(self, "file_name_list"):
+            file_purpose = self.file_name_list.get_attr(_InlineConstants.file_purpose)
+        elif hasattr(self, "filename"):
+            file_purpose = self.filename.get_attr(_InlineConstants.file_purpose)
+
+        if file_purpose == "input" and self.remote_file_handler.is_pim_configured():
```
seanpearsonuk, line 1042: `is_pim_configured()` is too concrete. The normal contract is this: if a non-None service is passed in, I shall call its methods like upload, download, whatever. The one who is passing it in has a responsibility to pass in something viable or nothing at all. Then, there is a clear expectation that a command invocation to read a file can incorporate the upload required step. If upload fails due to some implementation detail of the particular transfer service (e.g., pim has failed in some way) then the service throws an exception back up to the caller of the command.

src/ansys/fluent/core/session_solver.py
```diff
@@ -150,7 +150,9 @@ def _root(self):
         """Root settings object."""
         if self._settings_root is None:
             self._settings_root = settings_get_root(
-                flproxy=self._settings_service, version=self.version
+                flproxy=self._settings_service,
```
seanpearsonuk, line 153: Your architecture is ready to allow alternative file services to be injected. It's  a simple matter of requiring an object with upload and download methods. We could allow it to be inject at the pyfluent level statically such that it's simply picked up by each session object. That would be follow-up work.

hpohekar: What is the objective of tracking upload and download requests other than unit testing ? because `assert len(file_service.uploads) == 0` will be the case always for local session and in the GitHub.

src/ansys/fluent/core/solver/flobject.py
seanpearsonuk: The dependency of the core settings framework on the file transfer service needs to be abstract. It should not mention pim (or any other such specific names). The framework will not depend on pim, and alternative file transfer services can be used just as well.

src/ansys/fluent/core/session_solver.py
seanpearsonuk: tracking upload and download requests is what the mock service would do in the test

src/ansys/fluent/core/solver/flobject.py
hpohekar: Ok. Got it, we don't want to depend explicitly on pim file service, file service can be anything. I will update this change now. Thank you.

seanpearsonuk: When the file service is abstract like that, it also means ypu can use test fakes/mocks, as I commented below. The point I am making is that methods like upload in a mock can track what calls are being made, and do nothing else. That way you can check that the flobject framework is making the appropriate requests.

src/ansys/fluent/core/session_solver.py
hpohekar: I just asked this doubt because if we pass mock_instance to remote_file_handler in test then it returns remote_file_handler as a `NoneType` object. We can do upload/download using mock instance hence we can not track uploads/downloads, therefore we are using temporary `purpose` argument in the Ansys Lab.

seanpearsonuk: Your file service should have a simple interface such that alternative file services can easily be substituted. Right? If that alternative is a mock, you can implement it how you like. The mock will only be used in unit tests.

hpohekar: Yes, I agree. Thank you.

src/ansys/fluent/core/utils/file_transfer_service.py
```diff
@@ -137,6 +158,15 @@ class RemoteFileHandler:
...
         self._transfer_service = transfer_service
+        self._service_type = (
+            ServiceType.PIM
+            if isinstance(transfer_service, PimFileTransferService)
+            else None
+        )
+
+    def is_configured(self):
+        if self._service_type == ServiceType.PIM:
+            return pypim.is_configured()
```
hpohekar, line 169: `is_configured()` check is generic now. It is based on supported file transfer service types.

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1150,8 +1156,27 @@ def __call__(self, *args, **kwds) -> Any:
...
+                dict(
+                    FileName=os.path.basename(kwds["FileName"])
+                    if self.service.remote_file_handler.is_configured()
+                    else kwds["FileName"]
+                )
+            )
         _convert_value_to_variant(kwds, request.args)
+        file_purpose = None
+        if hasattr(self.create_instance(), "get_attr"):
+            file_purpose = self.create_instance().get_attr("FileName/filePurpose")
```
raph-luc, line 1170: I believe we should ideally do something like:
```
command_instance = self.create_instance()
file_purpose = command_instance.get_attr("FileName/filePurpose")
...
command_instance.delete()
```
Otherwise we would create an instance every time a command like this is called, and I don't know when/if Fluent would automatically delete or garbage collect it. 

However, I'm not sure if we have an easy way to delete created command instances from PyFluent like the `command_instance.delete()` I used as an example? Are you already aware of how to do this @hpohekar?

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1005,8 +1030,26 @@ def get_completer_info(self, prefix="", excluded=None) -> List[List[str]]:
...
+            kwds.update(
+                dict(
+                    file_name=os.path.basename(kwds["file_name"])
+                    if self.remote_file_handler.is_configured()
+                    else kwds["file_name"]
+                )
+            )
+        file_purpose = purpose if purpose else file_purpose
+        if file_purpose == "input" and self.remote_file_handler.is_configured():
+            self.remote_file_handler.upload(file_name=kwds["file_name"])
```
seanpearsonuk, line 1052: @hpohekar 
Key point # 1: We shall use the purpose attribute to generate classes, not as a runtime switch. Refer back to #2195 where it is stated
Key point # 2: Irrespective of which argument names are used or not used in the API, we shall not depend on those names in the logic. That was not part of the design, and none of the PyFluent code shall be written on that basis
Key point # 3: `is_configured` shall not be part of the service interface. This has already been pointed out. Such a method is not required because either a valid service is provided or none at all (I made a similar statement in a previous comment).

src/ansys/fluent/core/utils/file_transfer_service.py
hpohekar: Removed.

src/ansys/fluent/core/solver/flobject.py
hpohekar: @seanpearsonuk 

Key point # 3: - removed.

Key point # 2: - How to handle following PyPim behavior ?

image

hpohekar: related to Key point # 2:

We can remove following code but then Ansys Lab reads file from absolute path i.e. current working directory only. If we are OK with this behavior then we will remove it.

image

hpohekar: related to Key point # 1:

Ok. We will continue our earlier approach - https://github.com/ansys/pyfluent/pull/2238/files

We will use `file_purpose` attribute instead of `type` to generate classes.

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1150,9 +1156,27 @@ def __call__(self, *args, **kwds) -> Any:
...
+            kwds.update(
+                dict(
+                    FileName=os.path.basename(kwds["FileName"])
+                    if bool(self.service.remote_file_handler)
+                    else kwds["FileName"]
+                )
+            )
         _convert_value_to_variant(kwds, request.args)
+        file_purpose = None
+        if hasattr(self.create_instance(), "get_attr"):
```
seanpearsonuk, line 1169: You keep calling `create_instance` here. This is triggering remote calls that are actually creating command instances in the server.

src/ansys/fluent/core/solver/flobject.py
seanpearsonuk: Re: "Key point # 2: - How to handle following PyPim behavior ?", the comment is about depending on API argument names like `file_name`, `FileName` etc. It's nothing to do with PyPIM.

src/ansys/fluent/core/utils/file_transfer_service.py
```diff
@@ -181,14 +246,32 @@ def download(self, file_name: str, before_downloaded: Optional[Callable] = None)
...
+        )
+
+    def __bool__(self):
+        if self._service_type == ServiceType.PIM:
+            return self._transfer_service.is_pim_configured()
+        if self._service_type == ServiceType.Test:
+            return False
+
+
+class MyFileService:
```
seanpearsonuk, line 260: @hpohekar You probably ought to call it MockFileService (for instance). I used this name to make the general point.

src/ansys/fluent/core/services/datamodel_se.py
hpohekar: Yes, right. Therefore most of the tests are failing.

src/ansys/fluent/core/utils/file_transfer_service.py
hpohekar: Yes, Thank you.

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1150,9 +1156,27 @@ def __call__(self, *args, **kwds) -> Any:
         request.path = convert_path_to_se_path(self.path)
         request.command = self.command
         request.wait = True
+        if "FileName" in kwds:
```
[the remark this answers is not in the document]
hpohekar: codegen approach will resolve most of the issues.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1005,8 +1029,18 @@ def get_completer_info(self, prefix="", excluded=None) -> List[List[str]]:
...
+    def __call__(self, purpose: Optional[str] = None, **kwds):
+        """Call a command with the specified keyword arguments."""
+        file_purpose = None
+        if hasattr(self, "file_name"):
+            file_purpose = self.file_name.get_attr(_InlineConstants.file_purpose)
+        elif hasattr(self, "file_name_list"):
+            file_purpose = self.file_name_list.get_attr(_InlineConstants.file_purpose)
+        elif hasattr(self, "filename"):
+            file_purpose = self.filename.get_attr(_InlineConstants.file_purpose)
+        file_purpose = purpose if purpose else file_purpose
```
seanpearsonuk, line 1041: We are computing `file_purpose` and then only using it if `purpose` is not set. Please consider this point. Please also extract all this functionality into a separate method. This is complicting the implementation of `__call__` .

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1299,17 +1309,25 @@ def __init__(
...
+    def __call__(self, purpose: Optional[str] = None, *args, **kwds) -> Any:
         """Execute the command.
 
         Returns
         -------
         Any
             Return value.
         """
-        return self.service.execute_command(
+        if "Read" in self.__class__.__name__ and bool(
```
seanpearsonuk, line 1320: @hpohekar We need to eliminate these checks for "Read" and "Write" in the runtime code. We should be generating different types in the codegen instead.

src/ansys/fluent/core/solver/flobject.py
hpohekar: @seanpearsonuk Yes this is old code. We have extracted file_purpose in settings static info. Now we are working for meshing datamodel static info. We will update this code soon.

src/ansys/fluent/core/services/datamodel_se.py
hpohekar: Yes, as mentioned here, it's temporary. We are working on meshing datamodel static info. We will update this soon.

src/ansys/fluent/core/services/settings.py
```diff
@@ -274,6 +274,11 @@ def _extract_static_info(self, info: SettingsModule.StaticInfo) -> dict[str, Any
             ret["object-type"] = self._extract_static_info(info.object_type)
         if info.help:
             ret["help"] = info.help
+        try:
+            if info.file_purpose:
+                ret["file_purpose"] = info.file_purpose
```
seanpearsonuk, line 279: @hpohekar Looking above, I note that existing keys contain "-" and not "_".

hpohekar: Right, it should be "file-purpose". Thank you.

src/ansys/fluent/core/solver/flobject.py
hpohekar: `purpose` is temporary for testing in Ansys Lab. We will remove it soon. We have moved this functionality into a separate method.

src/ansys/fluent/core/services/datamodel_se.py
hpohekar: We have removed these checks. Now we are extracting file_purpose from datamodel static info.

src/ansys/fluent/core/services/settings.py
hpohekar: It's replaced with `attrs` now.

src/ansys/fluent/core/utils/file_transfer_service.py
```diff
@@ -137,6 +211,12 @@ class RemoteFileHandler:
 
     def __init__(self, transfer_service: Optional[Any] = None):
         self._transfer_service = transfer_service
+        if isinstance(transfer_service, PimFileTransferService):
+            self._service_type = ServiceType.PIM
+        elif isinstance(transfer_service, TransferRequestRecorder):
+            self._service_type = ServiceType.Test
+        else:
+            self._service_type = ServiceType.Null
```
seanpearsonuk, line 219: I think it would be good for the team to meet and discuss this design, so that we can all be on the same page to understand it. It's quite surprising/unconventional to see a subset of the possible types coupled together in this way.

hpohekar: Sure. I will explain our purpose to add this code with our entire team and it would be great if we get better solution/ideas for this than current one. Thank you.

seanpearsonuk: The test class belongs in the `tests` area. Any class with a matching interface can be used and doesn't need to be all defined in one place.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1169,6 +1240,24 @@ def allowed_values(self):
             return []
 
 
+class _InputFilePurposeMixin:
+    def file_purpose(self):
```
mkundu1, line 1244: I don't think we need this function, we can simply check `isinstance(argument, _InputFilePurposeMixin)`.

hpohekar: Yes, removed. Thank you.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -126,6 +132,16 @@ def flproxy(self):
...
 
+    @property
+    def remote_file_handler(self):
+        """Remote file handler.
+
+        Supports file upload and download.
+        """
+        if self._remote_file_handler is None:
+            return self._parent.remote_file_handler
+        return self._remote_file_handler
```
hpohekar, line 143: If we use `return self._remote_file_handler or self._parent.remote_file_handler` then we get recursion error.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -372,17 +388,49 @@ class String(SettingsBase[str], Textual):
     _state_type = str
 
 
-class Filename(SettingsBase[str], Textual):
+class InputOutputFileBase:
+    def is_input(self):
+        if self.file_purpose() == "input":
+            return True
+        return False
```
seanpearsonuk, line 395: Convoluted logic: a single-line does this.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -372,17 +388,49 @@ class String(SettingsBase[str], Textual):
     _state_type = str
 
 
-class Filename(SettingsBase[str], Textual):
+class InputOutputFileBase:
```
seanpearsonuk, line 391: I haven't looked at the other changes in this file, but the design already looks different to what we discussed. It's unclear how this class would fit into a rationale design.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -372,17 +388,49 @@ class String(SettingsBase[str], Textual):
...
+        return False
+
+    def before_execute(self, value):
+        try:
+            self.remote_file_handler.upload(file_name=value)
+        except AttributeError:
+            pass
+
+    def is_output(self):
+        if self.file_purpose() == "output":
```
seanpearsonuk, line 404: how are we covering in-out?

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -372,17 +388,49 @@ class String(SettingsBase[str], Textual):
...
-class Filename(SettingsBase[str], Textual):
+class InputOutputFileBase:
+    def is_input(self):
+        if self.file_purpose() == "input":
+            return True
+        return False
+
+    def before_execute(self, value):
+        try:
+            self.remote_file_handler.upload(file_name=value)
```
seanpearsonuk, line 399: We're leaving it to the caller decide whether upload/download is appropriate. That isn't what was discussed.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -372,17 +388,49 @@ class String(SettingsBase[str], Textual):
...
 
 
-class Filename(SettingsBase[str], Textual):
+class InputOutputFileBase:
+    def is_input(self):
+        if self.file_purpose() == "input":
+            return True
+        return False
+
+    def before_execute(self, value):
```
seanpearsonuk, line 397: No before_execute in the base?

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1010,8 +1058,14 @@ def get_completer_info(self, prefix="", excluded=None) -> List[List[str]]:
...
-    def __call__(self, **kwds):
-        """Call a query with the specified keyword arguments."""
+    def __call__(self, purpose: Optional[str] = None, **kwds):
+        """Call a command with the specified keyword arguments."""
+        for arg, value in kwds.items():
+            argument = getattr(self, arg)
+            if (
+                hasattr(argument, "is_input") and argument.is_input()
+            ) or purpose == "input":
+                argument.before_execute(value)
```
seanpearsonuk, line 1068: This is different to the design I shared with you.

src/ansys/fluent/core/solver/flobject.py
hpohekar: Yes, updated. Thank you.

src/ansys/fluent/core/solver/flobject.py
hpohekar: We haven't added in-out attribute in Fluent yet. We have checked input/output for now.

src/ansys/fluent/core/solver/flobject.py
hpohekar: As per our latest discussion, we have decided to get things working first and not to use mixins.

src/ansys/fluent/core/solver/flobject.py
hpohekar: As per our latest discussion, we have decided to use single base class (InputOutputFileBase) for Filename and FilenameList class instead of mixins.

src/ansys/fluent/core/solver/flobject.py
hpohekar: We are doing upload before command execution and download after command execution by checking whether file_purpose is input/output therefore I have implemented  self.remote_file_handler.upload in before_execute() and self.remote_file_handler.download in after_execute.

src/ansys/fluent/core/solver/flobject.py
hpohekar: If we don't add `if (hasattr(argument, "is_input") and argument.is_input())` check then we get attribute error for arguments other than `file_name` or `file_name_list`. 

Current Fluent changes are not in Ansys Lab therefore we have added `or purpose == "input"` condition to test it.

src/ansys/fluent/core/solver/flobject.py
hpohekar: I updated the code now.

codegen/settingsgen.py
```diff
@@ -134,6 +134,7 @@ def _populate_hash_dict(name, info, cls, api_tree):
 
     cls_touple = (
         name,
+        cls.__bases__,
```
seanpearsonuk, line 137: just noting that the existing name above seems to be misspelled. `cls_tuple` must have been intended.

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1429,9 +1449,15 @@ def __call__(self, *args, **kwds) -> Any:
         Any
             Return value.
         """
-        return self.service.execute_command(
+        file_purpose = self._get_file_purpose()
+        if file_purpose == "input" and bool(self.service.remote_file_handler):
```
seanpearsonuk, line 1453: @hpohekar What's the rationale for doing getting the command's file purpose in the case of datamodel commands?

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1421,6 +1431,16 @@ def __init__(
         else:
             self.path = path
 
+    def _get_file_purpose(self):
+        static_info = self._get_static_info()
+        try:
+            file_purpose = static_info["singletons"]["File"]["commands"][
```
seanpearsonuk, line 1437: This doesn't look right. What's special about singletons? This is separating singletons from named objects for instance, right? Why even do we need to go to that level? Cn we not just access the local static info automatically?

src/ansys/fluent/core/services/datamodel_se.py
hpohekar: We had extracted it on command level for both settings and datamodel. We have updated it for settings and we will update this for datamodel.

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1429,9 +1449,15 @@ def __call__(self, *args, **kwds) -> Any:
         Any
             Return value.
         """
-        return self.service.execute_command(
+        file_purpose = self._get_file_purpose()
+        if file_purpose == "input" and bool(self.service.remote_file_handler):
+            self.service.remote_file_handler.upload(file_name=kwds["FileName"])
+        command = self.service.execute_command(
```
seanpearsonuk, line 1455: We should still follow the `beforeExecute` `afterExecute` pattern even if only to keep the code tidy.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -49,6 +49,7 @@ class _InlineConstants:
     user_creatable = "user-creatable?"
     allowed_values = "allowed-values"
     file_purpose = "file-purpose"
+    file_purpose = "file-purpose"
```
seanpearsonuk, line 52: these are duplicate statements

codegen/settingsgen.py
hpohekar: `cls_tuple` is inside `_populate_hash_dict`.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1042,14 +1093,21 @@ def __call__(self, **kwds):
...
 
 class CommandWithPositionalArgs(Action):
     """Command Object."""
 
     def __call__(self, *args, **kwds):
-        """Call a query with the specified keyword arguments."""
+        """Call a command with the specified keyword arguments."""
+        for arg, value in kwds.items():
+            argument = getattr(self, arg)
+            argument.before_execute(value)
```
seanpearsonuk, line 1110:
```mermaid
  classDiagram
      Action <|-- Query
      Action <|-- BaseCommand  :
        execute_command
      BaseCommand <|-- Command :
        _execute_command
      BaseCommand <|-- CommandWithPositionalArgs :
        _execute_command
```

src/ansys/fluent/core/services/settings.py
```diff
@@ -259,11 +259,20 @@ def _extract_static_info(self, info: SettingsModule.StaticInfo) -> dict[str, Any
                 child.name: self._extract_static_info(child.value)
                 for child in info.arguments
             }
+            try:
+                for child in info.arguments:
+                    ret["arguments"][child.name].update(
```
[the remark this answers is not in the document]
hpohekar: I thought already we had done check for arguments and we are adding file_purpose on argument level so that we can just update the arguments with respective file purpose.

src/ansys/fluent/core/solver/flobject.py
hpohekar: Yes, I pulled latest changes and this was added during resolving conflicts. I'm cleaning stuff like this now. it's in progress.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1042,14 +1093,21 @@ def __call__(self, **kwds):
...
+        return cmd
 
 
 class CommandWithPositionalArgs(Action):
     """Command Object."""
 
     def __call__(self, *args, **kwds):
-        """Call a query with the specified keyword arguments."""
+        """Call a command with the specified keyword arguments."""
+        for arg, value in kwds.items():
```
seanpearsonuk, line 1108: I just want to check that only `kwds`, and not `args` is relevant here.

codegen/settingsgen.py
seanpearsonuk: Be it wherever, can we not update it?

src/ansys/fluent/core/services/datamodel_se.py
seanpearsonuk: Yes I would prefer that.

codegen/settingsgen.py
hpohekar: @seanpearsonuk  I meant indentation is correct we will correct it's name.

hpohekar: @seanpearsonuk Yes we will update it, I just read intended as indented so I have written that it's indented correctly.

src/ansys/fluent/core/services/datamodel_se.py
hpohekar: Yes, sure.

src/ansys/fluent/core/services/settings.py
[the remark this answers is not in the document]
hpohekar: We get following error if we check above condition.

image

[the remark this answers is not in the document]
hpohekar: What is special about attrs? - `attrs` is not a separate field in static info. 

Can't you do in the outer block - It's inside arguments therefore we are doing in arguments block.

src/ansys/fluent/core/services/datamodel_se.py
hpohekar: @seanpearsonuk 

`_stored_static_info` is empty until we read a case/mesh.

image

`_stored_static_info` is returns value only after reading a case or mesh file. We need file_purpose before execution of command.

We are already using static info using `static_info = self._get_static_info()` 

image

According to above static info structure we are accessing `file_purpose`.

Do we have any other way to access file_purpose ?

src/ansys/fluent/core/services/settings.py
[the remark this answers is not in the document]
mkundu1: The `"type"` and `"help"` are also added for arguments, so `"attrs"` should also be added at the same level. We'll be already inside the arguments while adding those in the inner call.

tests/test_meshing_queries_fdl.py
```diff
@@ -2025,7 +2025,7 @@ def test_meshing_utilities(new_mesh_session):
 
     assert meshing_session.meshing_utilities.replace_cell_zone_suffix(
         cell_zone_id_list=[87], old_suffix="fluid", new_suffix="fluid-new", merge=True
-    ) == [False]
+    ) == ["*the-non-printing-object*"]
```
mkundu1, line 2028: Should this be fixed on Fluent side?

hpohekar: @mkundu1 It was returning [False] but unexpectedly now returns ["*the-non-printing-object*"] 

image

As per last discussion in PyFluent Coordination meeting it should get fixed earlier but no idea why it is still there.

src/ansys/fluent/core/services/settings.py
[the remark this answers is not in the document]
hpohekar: Done. Thank you.

src/ansys/fluent/core/services/settings.py
```diff
@@ -263,7 +263,12 @@ def _extract_static_info(self, info: SettingsModule.StaticInfo) -> dict[str, Any
             ret["object-type"] = self._extract_static_info(info.object_type)
         if info.help:
             ret["help"] = info.help
-
+        try:
```
mkundu1, line 266: wondering why/when this will throw

hpohekar: @mkundu1 @seanpearsonuk 

If we use `if info.HasField("attrs"):` we get following error.

image

mkundu1: Good point, `HasField` is not supported for repeated/map fields. Perhaps in that case we can remove the if condition altogether as `info.attrs` should always exists as map, IIUC.

hpohekar: Yes, verified. Following change works fine.

image

Thank you.

src/ansys/fluent/core/solver/flobject.py

src/ansys/fluent/core/solver/flobject.py
seanpearsonuk, line 1: AFAICS flobject.py, is in good shape, and we need to address some specific points in datamodel_se.py. This can be done collaboratively. Hopefully @mkundu1 can assist, and I'm also happy to help.

src/ansys/fluent/core/services/datamodel_se.py
hpohekar: @seanpearsonuk 

As per our latest discussion we are unable to access FileName as follows.

read_mesh = meshing_session.meshing.File.ReadMesh.create_command_instance()`
`read_mesh.FileName` -> gives key error for ReadMesh.

src/ansys/fluent/core/solver/flobject.py
hpohekar: Yes, sure.

src/ansys/fluent/core/services/datamodel_se.py

src/ansys/fluent/core/solver/flobject.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/services/datamodel_se.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/services/datamodel_se.py
seanpearsonuk: @hpohekar I frequently get confused when I see resolved conversations that refer to outstanding issues.

hpohekar: @seanpearsonuk I'm not sure which changes are you looking at. I have updated the code and latest changes are as follows.

image

hpohekar: @seanpearsonuk https://github.com/ansys/pyfluent/pull/2289/files#diff-c29fea6870a8113071deda7cddca1079fab43d6f0e3229ea44cd9e697f35c6d9

seanpearsonuk: Thanks @hpohekar - that looks better. GitHub was still showing me the older version. Apologies.

hpohekar: No issues at all.

mkundu1 (approved):

prmukherj (approved):

src/ansys/fluent/core/utils/file_transfer_service.py
```diff
@@ -103,14 +147,43 @@ def download(self, file_name: str, local_file_name: Optional[str] = None):
...
+                )
+        if self.is_configured():
+            for file in files:
+                if os.path.isfile(file):
+                    print(f"\nFile already exists. File path:\n{file}\n")
+                else:
+                    self.download(os.path.basename(file), local_file_name=".")
+
 
 class RemoteFileHandler:
```
seanpearsonuk, line 188: as discussed, `RemoteFileHandler` will be removed (redundant). @hpohekar

hpohekar: We will integrate this changes in a separate PR.

PR #2296, opened 2023-12-04
Upgrade unit tests to use lightweight case load where possible

ansjsia:
To speed up solver mode unit tests, tests that don't require loading the mesh have been set to call `root.file.read` with the parameter `lightweight_setup=True`.

### DIFF
--- a/tests/test_solvermode/test_models.py
+++ b/tests/test_solvermode/test_models.py
@@ -5,12 +5,12 @@ import pytest
 @pytest.mark.quick
 @pytest.mark.setup
 @pytest.mark.fluent_version("latest")
-def test_solver_models(load_mixing_elbow_mesh):
-    solver_session = load_mixing_elbow_mesh
+def test_solver_models(load_mixing_elbow_settings_only):
+    solver_session = load_mixing_elbow_settings_only
     models = solver_session.setup.models
-    assert not models.energy.enabled()
-    models.energy.enabled = True
     assert models.energy.enabled()
+    models.energy.enabled = False
+    assert not models.energy.enabled()
     models.multiphase.models = "vof"
     assert models.multiphase.models() == "vof"
     models.viscous.model = "laminar"
--- a/tests/test_solvermode/test_models.py
+++ b/tests/test_solvermode/test_models.py
@@ -26,8 +26,8 @@ def test_solver_models(load_mixing_elbow_settings_only):
 @pytest.mark.quick
 @pytest.mark.setup
 @pytest.mark.fluent_version("latest")
-def test_disk_2d_models(load_disk_mesh):
-    solver_session = load_disk_mesh
+def test_disk_2d_models(load_disk_settings_only):
+    solver_session = load_disk_settings_only
     models = solver_session.setup.models
     solver_session.setup.general.solver.two_dim_space = "axisymmetric"
     solver_session.setup.general.solver.two_dim_space = "swirl"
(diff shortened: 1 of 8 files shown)


tests/test_solvermode/test_models.py
```diff
@@ -5,12 +5,12 @@
...
 @pytest.mark.fluent_version("latest")
-def test_solver_models(load_mixing_elbow_mesh):
-    solver_session = load_mixing_elbow_mesh
+def test_solver_models(load_mixing_elbow_settings_only):
+    solver_session = load_mixing_elbow_settings_only
     models = solver_session.setup.models
-    assert not models.energy.enabled()
-    models.energy.enabled = True
     assert models.energy.enabled()
+    models.energy.enabled = False
```
dnwillia-work, line 12: Any reason this changed, the logic was reversed before

ansjsia: When loading a mesh file, energy is disabled by default, but the case file has energy transport turned on. To test changing the energy transport model I figured the way to solve this would be to test turning it off. We would still be checking the `True` and `False` states, just in the opposite order.

tests/test_settings_api.py
```diff
@@ -9,8 +9,11 @@
 def test_setup_models_viscous_model_settings(new_solver_session) -> None:
     solver_session = new_solver_session
     case_path = download_file("elbow_source_terms.cas.h5", "pyfluent/mixing_elbow")
-    solver_session.file.read_case(file_name=case_path)
-    solver_session.solution.initialization.hybrid_initialize()
```
ansjsia, line 13: The test seems to be able to set up the viscous model normally without initializing the case. Is there a reason the original test initialized the case?

dnwillia-work: Not that I know of.

dnwillia-work (approved):

raph-luc:
Running solvermode tests here: https://github.com/ansys/pyfluent/actions/runs/7146060041

If any tests fail we can compare with others to see if they were already failing, or if they are new failures: https://github.com/ansys/pyfluent/actions/workflows/weekly-solvermode-test-run.yml

seanpearsonuk:
> > Running solvermode tests here: https://github.com/ansys/pyfluent/actions/runs/7146060041
> > If any tests fail we can compare with others to see if they were already failing, or if they are new failures: https://github.com/ansys/pyfluent/actions/workflows/weekly-solvermode-test-run.yml
> 
> @raph-luc I'm noticing these solver mode tests are being run with 24.2 rather than 24.1. 2 of the 4 test failures are caused by changes in API object names (`test_boundaries.py::test_boundaries_elbow` and `test_species_model.py::test_change_create_mixture`). Should we delay updating them until 24.1 is released?
> 
> `test_boundaries.py::test_boundaries_periodic` is throwing an `AssertionError` when comparing two dictionaries. I'll have to do some digging to see if this is caused by an API change.
> 
> Would you have any idea on how to investigate `test_general.py::test_solver_import_mixingelbow` throwing an `_InactiveRpcError`?

@ansjsia @raph-luc 
I believe that we should test against stable APIs here, so yes, we should avoid touching 24.2 at the present time.

seanpearsonuk (approved):

raph-luc:
@ansjsia when comparing the test results from [this branch](https://github.com/ansys/pyfluent/actions/runs/7146060041/job/19462983934) with those [from `main`](https://github.com/ansys/pyfluent/actions/runs/7146293836/job/19463690613), we can see that the test failures from this branch are not due to the changes here: these tests were already failing before with the exact same errors. We shouldn't address those in this PR.

raph-luc (approved):

PR #2300, opened 2023-12-07
Revise test markers

ansjsia:
Implement the following changes in test markers to remove markers with little utility and to improve clarity for intended use:

1. Change the name of the `setup` marker to `settings_only`.
2. If the test initializes or solves the case, remove the `settings_only` marker.
3. Update the description of the `settings_only` marker to “Read and modify the case settings only without loading the mesh, initializing, or solving the case.”
4. Remove the following markers from all tests that use them and from `pyproject.toml`:
   1. `integration` (6 tests)
   2. `quick` (11 tests)
   3. `solve` (no tests)
   4. `mesh` (1 test)
   5. `optislang` (2 tests)

~This should not be merged until after PR https://github.com/ansys/pyfluent/pull/2296 is merged because it makes changes to whether some tests initialize or solve.~
EDIT: https://github.com/ansys/pyfluent/pull/2296 has been merged.


seanpearsonuk (approved):

raph-luc (approved):

tests/test_solvermode/test_post_vector.py
```diff
@@ -1,7 +1,7 @@
 import pytest
 
 
-@pytest.mark.setup
+@pytest.mark.settings_only
```
raph-luc, line 4: This one isn't really `settings_only`, right? It is also reading a .dat file which is probably more "heavyweight"

Same for `test_change_create_mixture`?

ansjsia: `test_post_vector` doesn't seem to require the `.dat.h5` file so I've changed it to use lightweight case read. I've also changed `test_change_create_mixture` to use lightweight case read, must have missed that one in the last PR.

Issue #2310, opened 2023-12-11
Using `case_file_name` argument to `pyfluent.launch_fluent()` throws exception


ansjsia:
### 🔍 Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### 🐞 Description of the bug

With the development PyFluent build, passing an argument to `case_file_name` in `pyfluent.launch_fluent()` throws:

```
TypeError                                 Traceback (most recent call last)
File [~\GitHub\pyfluent\src\ansys\fluent\core\launcher\standalone_launcher.py:264](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/jsia/OneDrive%20-%20ANSYS%2C%20Inc/Sandbox/~/GitHub/pyfluent/src/ansys/fluent/core/launcher/standalone_launcher.py:264), in StandaloneLauncher.__call__(self)
    263     else:
--> 264         session.read_case(self.case_file_name, self.lightweight_mode)
    265 if self.case_data_file_name:

TypeError: Solver.read_case() takes 2 positional arguments but 3 were given
```

This bug does not occur with PyFluent 0.18.1.

### 📝 Steps to reproduce

1. Call `pyfluent.launch_fluent(case_file_name=...)` where `...` is a case file of your choice.
2. PyFluent throws a `TypeError`.

### 💻 Which operating system are you using?

Windows

### 📀 Which ANSYS version are you using?

24R1

### 🐍 Which Python version are you using?

3.11

### 📦 Installed packages

```shell
accessible-pygments==0.0.4
alabaster==0.7.13
ansys-api-fluent==0.3.19
ansys-api-platform-instancemanagement==1.0.0
-e git+https://github.com/ansys/pyfluent.git@b0835a5e85185f158e875c928055a30dcf684c21#egg=ansys_fluent_core      
ansys-platform-instancemanagement==1.1.2
ansys-sphinx-theme==0.12.5
asttokens==2.4.1
attrs==23.1.0
autodocsumm==0.2.11
Babel==2.13.1
beartype==0.16.4
beautifulsoup4==4.12.2
bleach==6.1.0
build==1.0.3
certifi==2023.11.17
cfgv==3.4.0
charset-normalizer==3.3.2
colorama==0.4.6
comm==0.2.0
contourpy==1.2.0
coverage==7.3.2
cycler==0.12.1
debugpy==1.8.0
decorator==5.1.1
defusedxml==0.7.1
distlib==0.3.7
[...]

PR #2312, opened 2023-12-11
Fix issue with passing `case_file_name` argument to `launch_fluent()` throwing `TypeError`

ansjsia:
Fixes https://github.com/ansys/pyfluent/issues/2310 by replacing the call to `root.read_case()` with `root.file.read()` using keyword arguments so that they are passed in correctly.

### DIFF
--- a/src/ansys/fluent/core/launcher/standalone_launcher.py
+++ b/src/ansys/fluent/core/launcher/standalone_launcher.py
@@ -260,12 +260,10 @@ class StandaloneLauncher:
             if self.case_file_name:
                 if self.meshing_mode:
                     session.tui.file.read_case(self.case_file_name)
+                elif self.lightweight_mode:
+                    session.read_case_lightweight(self.case_file_name)
                 else:
-                    session.file.read(
-                        file_type="case",
-                        file_name=self.case_file_name,
-                        lightweight_setup=self.lightweight_mode,
-                    )
+                    session.read_case(self.case_file_name)
             if self.case_data_file_name:
                 if not self.meshing_mode:
                     session.file.read(
--- a/src/ansys/fluent/core/launcher/standalone_launcher.py
+++ b/src/ansys/fluent/core/launcher/standalone_launcher.py
@@ -59,6 +59,7 @@ class StandaloneLauncher:
         cwd: Optional[str] = None,
         topy: Optional[Union[str, list]] = None,
         start_watchdog: Optional[bool] = None,
+        scheduler_options: Optional[dict] = None,
         **kwargs,
     ):
         """Launch Fluent session in standalone mode.
@@ -176,6 +177,8 @@ class StandaloneLauncher:
         _process_invalid_args(dry_run, "standalone", argvals)
         args = _get_argvals(argvals, mode)
         argvals.update(args)
+        if argvals["start_timeout"] is None:
+            argvals["start_timeout"] = 60
         for arg_name, arg_values in argvals.items():
             setattr(self, arg_name, arg_values)
         self.argvals = argvals
@@ -204,7 +207,7 @@ class StandaloneLauncher:
         sifile_last_mtime = Path(server_info_file_name).stat().st_mtime
         if self.env is None:
             setattr(self, "env", {})
-        kwargs = _get_subprocess_kwargs_for_fluent(self.env)
+        kwargs = _get_subprocess_kwargs_for_fluent(self.env, self.argvals)
         if self.cwd:
             kwargs.update(cwd=self.cwd)
         launch_string += _build_journal_argument(self.topy, self.journal_file_names)
--- a/src/ansys/fluent/core/launcher/standalone_launcher.py
+++ b/src/ansys/fluent/core/launcher/standalone_launcher.py
@@ -263,11 +263,15 @@ class StandaloneLauncher:
                 elif self.lightweight_mode:
                     session.read_case_lightweight(self.case_file_name)
                 else:
-                    session.read_case(self.case_file_name)
+                    session.file.read(
+                        file_type="case",
+                        file_name=self.case_file_name,
+                    )
             if self.case_data_file_name:
                 if not self.meshing_mode:
                     session.file.read(
-                        file_type="case-data", file_name=self.case_data_file_name
+                        file_type="case-data",
+                        file_name=self.case_data_file_name,
                     )
                 else:
                     raise RuntimeError(
--- a/src/ansys/fluent/core/launcher/standalone_launcher.py
+++ b/src/ansys/fluent/core/launcher/standalone_launcher.py
@@ -261,7 +261,11 @@ class StandaloneLauncher:
                 if self.meshing_mode:
                     session.tui.file.read_case(self.case_file_name)
                 else:
-                    session.read_case(self.case_file_name, self.lightweight_mode)
+                    session.file.read(
+                        file_type="case",
+                        file_name=self.case_file_name,
+                        lightweight_setup=self.lightweight_mode,
+                    )
             if self.case_data_file_name:
                 if not self.meshing_mode:
                     session.file.read(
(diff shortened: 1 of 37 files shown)


ansjsia:
I think this reveals another issue with the source code. The docstring for `launch_fluent()` states the following for the `lightweight_mode` parameter:

```
lightweight_mode : bool, optional
        Whether to run in lightweight mode. In lightweight mode, the lightweight settings are read into the
        current Fluent solver session. The mesh is read into a background Fluent solver session which will
        replace the current Fluent solver session once the mesh read is complete and the lightweight settings
        made by the user in the current Fluent solver session have been applied in the background Fluent
        solver session. This is all orchestrated by PyFluent and requires no special usage.
        This parameter is used only when ``case_file_name`` is provided. The default is ``False``.
```

However, the existing code in `standalone_launcher.py` carries this out by making the following call:

```python
session.file.read_case(self.case_file_name, self.lightweight_mode)
```

Which passes the value of `lightweight_mode` to the `lightweight_setup` parameter of `read_case()`, which does not exist. If the intention is to pass the argument to `lightweight_setup`, this will not actually load the mesh in the background, conflicting with the documentation.

Does the code or the documentation for `pyfluent.launch_fluent()` need to be updated?

seanpearsonuk:
> I think this reveals another issue with the source code. The docstring for `launch_fluent()` states the following for the `lightweight_mode` parameter:
> 
> ```
> lightweight_mode : bool, optional
>         Whether to run in lightweight mode. In lightweight mode, the lightweight settings are read into the
>         current Fluent solver session. The mesh is read into a background Fluent solver session which will
>         replace the current Fluent solver session once the mesh read is complete and the lightweight settings
>         made by the user in the current Fluent solver session have been applied in the background Fluent
>         solver session. This is all orchestrated by PyFluent and requires no special usage.
>         This parameter is used only when ``case_file_name`` is provided. The default is ``False``.
> ```
> 
> However, the existing code in `standalone_launcher.py` carries this out by making the following call:
> 
> ```python
> session.file.read_case(self.case_file_name, self.lightweight_mode)
> ```
> 
> Which passes the value of `self.lightweight_mode` to the `lightweight_setup` parameter of `read_case()` (which does not exist). If the intention is to pass the argument to `lightweight_setup`, this will not actually load the mesh in the background, conflicting with the documentation.
> 
> Does the code or the documentation for `pyfluent.launch_fluent()` need to be updated?



> I think this reveals another issue with the source code. The docstring for `launch_fluent()` states the following for the `lightweight_mode` parameter:
> 
> ```
> lightweight_mode : bool, optional
>         Whether to run in lightweight mode. In lightweight mode, the lightweight settings are read into the
>         current Fluent solver session. The mesh is read into a background Fluent solver session which will
>         replace the current Fluent solver session once the mesh read is complete and the lightweight settings
>         made by the user in the current Fluent solver session have been applied in the background Fluent
>         solver session. This is all orchestrated by PyFluent and requires no special usage.
>         This parameter is used only when ``case_file_name`` is provided. The default is ``False``.
> ```
> 
> However, the existing code in `standalone_launcher.py` carries this out by making the following call:
> 
> ```python
> session.file.read_case(self.case_file_name, self.lightweight_mode)
> ```
> 
> Which passes the value of `self.lightweight_mode` to the `lightweight_setup` parameter of `read_case()` (which does not exist). If the intention is to pass the argument to `lightweight_setup`, this will not actually load the mesh in the background, conflicting with the documentation.
> 
> Does the code or the documentation for `pyfluent.launch_fluent()` need to be updated?

@ansjsia It seems that this must have got broken with the recent PR that added `standalone_launcher.py` etc. The documentation states the intent correctly but the recent code changes are diverging from that.

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -261,7 +261,11 @@ def __call__(self):
                 if self.meshing_mode:
                     session.tui.file.read_case(self.case_file_name)
                 else:
-                    session.read_case(self.case_file_name, self.lightweight_mode)
+                    session.file.read(
+                        file_type="case",
+                        file_name=self.case_file_name,
+                        lightweight_setup=self.lightweight_mode,
+                    )
```
raph-luc, line 268: In my understanding, the previous version wasn't set up correctly. 

We can still support the PyFluent-handled "lightweight mode" (which seems to not be quite the same as the Fluent settings-only "lightweight setup" mode), I believe these changes have an example of how it could be set up so it works correctly: https://github.com/ansys/pyfluent/pull/2268/files

seanpearsonuk: https://github.com/ansys/pyfluent/pull/2252/files#r1424056929

ansjsia: @seanpearsonuk Thanks for clarifying. It seems like PyFluent "lightweight mode" (loading the mesh in the background) should be called `lightweight_setup` and Fluent's settings-only mode (not loading the mesh at all, previously referred to as "lightweight setup") should actually be called `settings_only` and they're getting confused with each other.

Perhaps the Fluent launcher and read methods, as well as their documentation, should explicitly distinguish between a `lightweight_setup` mode (load the mesh in the background) and a `settings_only` mode (settings only, do **not** load the mesh in the background). It should emulate the Fluent GUI:

image

What would be the best path forward? I wonder if it would be best to condense all the `solver.read_*()` and `solver.file.read_*()` methods into just `solver.file.read()` to ensure that there's only one way to read files (in the spirit of [PEP 20](https://peps.python.org/pep-0020/)).

raph-luc: @ansjsia I would say that in the short term, to fix the `TypeError` issue when specifying `case_file_name` you correctly identified here, we should make the changes as outlined in the links Sean and I shared above (both links have the same solution). 

Also, can you please add a test covering the `case_file_name` argument so we make sure we don't break this again in the future? 

Then, after discussion, and in a separate PR, we will hopefully be able to properly address the "lightweight/lite mode/setup settings-only" naming confusion (which I expect will require more extensive changes).

ansjsia:
@raph_luc @seanpearsonuk The latest push restructures the code to match the samples you gave and adds unit tests for reading case/data files into Fluent at launch. However, I'm running into an issue with the `root.read_case_lightweight()` method where it throws an error about not being able to read the mesh (see https://github.com/ansys/pyfluent/issues/2313 ).

It doesn't seem to affect reading the case settings though, as the new unit tests I put in are still passing.

seanpearsonuk:
> where it throws an error about not being able to read the mesh

@ansjsia (cc: @raph-luc) 

Please could you try to test doing a lightweight read of the same case directly in Fluent. I believe that it is possible in the TUI - not in the GUI last I saw,

ansjsia:
@seanpearsonuk I've been able to do a settings-only read using the following methods:

- In the launcher GUI
- After launch in the GUI
- In the TUI by calling `file/read-case-settings`
- In PyFluent by calling `root.file.read(..., lightweight_setup=True)`

Since `root.read_case_lightweight()` calls `root.file.read()` under the hood, the issue seems to be with creating a session to load the mesh in the background.

tests/test_launcher.py
```diff
@@ -75,6 +76,66 @@ def test_container_launcher():
...
 
 
+@pytest.mark.standalone
+def test_case_load():
+    # Test that launch_fluent() works with a case file as an argument
+    _, cas_path = download_input_file(
+        "pyfluent/mixing_elbow",
+        "mixing_elbow.cas.h5",
+    )
+    session = pyfluent.launch_fluent(case_file_name=cas_path)
```
raph-luc, line 86: As @ansjsia identified, the `standalone` marker is required as these tests won't work in the GitHub workflows, because we don't currently support `case_file_name` for container runs, and we don't have tests with a standalone Fluent install. See https://github.com/ansys/pyfluent/issues/2314.

raph-luc (approved):

seanpearsonuk (approved):
Great work @ansjsia, thanks.

PR #2325, opened 2023-12-21
Examples download files to own folder

ansjsia:
In response to https://github.com/ansys/pyfluent/issues/915

Now, when `examples.download_file()` is called without an argument to `save_path`, example files will be downloaded to a directory of their own to avoid interfering with other examples.

### DIFF
--- a/src/ansys/fluent/core/examples/downloads.py
+++ b/src/ansys/fluent/core/examples/downloads.py
@@ -162,11 +162,11 @@ def download_file(
     if save_path is None:
         if directory:
             # Get name of last folder in directory
-            directory_tip = Path(directory).name
+            directory_tip = os.path.basename(directory)
         else:
             # Strip all suffixes from file name
             directory_tip = file_name.split(".")[0]
-        save_path = Path(pyfluent.EXAMPLES_PATH) / directory_tip
+        save_path = os.path.join(pyfluent.EXAMPLES_PATH, directory_tip)
 
     url = _get_file_url(file_name, directory)
     return _retrieve_file(url, file_name, save_path, return_without_path)
--- a/src/ansys/fluent/core/examples/downloads.py
+++ b/src/ansys/fluent/core/examples/downloads.py
@@ -134,7 +134,6 @@ def download_file(
 
     Examples
     --------
-    TODO: Update these
     >>> from ansys.fluent.core import examples
     >>> file_path = examples.download_file("bracket.iges", "geometry")
     >>> file_path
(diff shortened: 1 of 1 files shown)


src/ansys/fluent/core/examples/downloads.py
```diff
@@ -150,6 +159,15 @@ def download_file(
...
             return_without_path = False
 
+    if save_path is None:
+        if directory:
+            # Get name of last folder in directory
+            directory_tip = Path(directory).name
+        else:
+            # Strip all suffixes from file name
+            directory_tip = file_name.split(".")[0]
+        save_path = Path(pyfluent.EXAMPLES_PATH) / directory_tip
```
raph-luc, line 169: I think we should use a temporary directory with a random name here, otherwise when different tests use the same downloaded files, this might also cause issues

As an example, you can see how I've generated random strings in the past in `src\ansys\fluent\core\launcher\watchdog.py`

ansjsia: I see it. To make it more user-readable, what if we appended `_n` to the folder name, where `n` is an incrementally increasing number, if a folder with that name already exists? The caveat is that if a script downloads multiple files by calling `download_file` for each one, they will end up distributed across multiple folders.

But perhaps this may not be an issue at all since in the `example-data` repository, the only files that are not in a directory are `.gitattributes`, `.gitignore`, `LICENSE`, and `README.md`. I only added consideration of a case where both `directory` and `save_path` are `None` in the event that `example-data` is restructured in the future.

raph-luc: I think that is a good idea too, and worth trying, certainly better than a fully random suffix appended to the file name, though with bit of IO cost checking if folder exists or not until finding an unused name, which I think is fine

ansjsia: I think we already have logic preventing this case from occurring. If the file already exists, it will not be downloaded again.

raph-luc: That is not what I meant, the issue isn't necessarily whether the file already exists, the issue is whether the different examples/tests will be running in the same folder or not (as in the issue tracker linked). 

Therefore, as you said in the PR description: "when `examples.download_file()` is called without an argument to `save_path`, example files will be downloaded to a directory of their own to avoid interfering with other examples."

mkundu1: Another way of doing this is
1. Each test creates a temporary working directory.
2. The input files are downloaded within a common folder and copied to the test's working directory.
3. The working directory is deleted after the test is run.

This will avoid downloading the same file multiple times and also multiple tests can use the same input file without any interference.

PR #2334, opened 2024-01-02
Update mixing_elbow_settings_api example to run in solver mode.

MaxJPRey:


raph-luc (approved):

raph-luc:
FYI the style check failure is because the branch name should start with `docs` in this case rather than `doc` https://actions.docs.ansys.com/version/stable/style-actions/index.html#branch-name-style-action

But can just ignore it, this style check failure should not block merging

mkundu1 (approved):

hpohekar (approved):

PR #2338, opened 2024-01-09
refactor: Remove wait_on_refresh

prmukherj:
Just commenting out the "_wait_on_refresh" method from the workflow.py code and running all the related tests suggest that it is running fine. So, we can get rid of this method with some level of confidence. What are your view?

Got rid of all instances of sleep.

cc @seanpearsonuk , @mkundu1

### DIFF
--- a/src/ansys/fluent/core/workflow.py
+++ b/src/ansys/fluent/core/workflow.py
@@ -294,7 +294,6 @@ class BaseTask:
             return ArgumentWrapper(self, attr)
         except Exception as ex:
             logger.debug(str(ex))
-        # self._command_source._wait_on_refresh()
         return self._task_objects.get(attr, None)
 
     def __setattr__(self, attr, value):
@@ -911,7 +910,6 @@ class WorkflowWrapper:
         )  # or self._task_with_cmd_matching_help_string(attr)
         if obj:
             return obj
-        # self._wait_on_refresh()
         return self._task_objects.get(attr, None)
 
     def __dir__(self):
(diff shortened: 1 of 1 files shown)


seanpearsonuk:
@prmukherj yes, we can trust the tests - the objective was to remove this code and if it can be done in the simplest way then even better.

src/ansys/fluent/core/workflow.py
```diff
@@ -925,20 +925,20 @@ def __call__(self):
...
-            orig_refresh_count = refresh_count
-            sleep(20 * time_unit)
-            while self._refreshing or self._refresh_count > refresh_count:
-                refresh_count = self._refresh_count
-                sleep(time_unit)
-            if self._refresh_count > orig_refresh_count:
-                self._wait_on_refresh(time_unit=time_unit, skip_check=True)
-        if not skip_check:
-            logger.debug("_wait_on_refresh time taken {time() - t0}")
+    # def _wait_on_refresh(self, time_unit=0.1, skip_check=False):
```
seanpearsonuk, line 928: @prmukherj   Just noticing other sleep calls. Please see ln 986, and see if we need those.

prmukherj: I have noticed that already. This sleep can be removed as well but it will do this step repeatedly: logger.debug("Call _refresh_task_accessors"). So keeping the small sleep here might be useful, but it can be removed otherwise. What do you think @seanpearsonuk ?

seanpearsonuk: I would hope that we can access all the task information immediately after creating the workflow. So, please let's try it. If it doesn't work then we will need a better architectural solution.

prmukherj: ok

seanpearsonuk (approved):

hpohekar (approved):

PR #2342, opened 2024-01-11
refactor: Make new methods in datamodel_se consistent with PyStateEng…

prmukherj:
…ine (server).

### DIFF
--- a/src/ansys/fluent/core/services/datamodel_se.py
+++ b/src/ansys/fluent/core/services/datamodel_se.py
@@ -386,7 +386,7 @@ class DatamodelService(StreamingService):
         response = self._impl.get_object_names(request)
         return response.names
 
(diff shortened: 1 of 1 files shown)


seanpearsonuk:
> Make new methods in datamodel_se consistent with PyStateEngine.
> 
> A couple of methods: 'delete_child_objects' and 'delete_all_child_objects' were applied at the container level in the PyStateEngine. However, in pyfluent they were exposed at object level. This was made consistent by exposing these methods at container level in pyfluent as well.
> 
> The signature of these methods changed in pyfluent to take object type as argument as well. The application can be seen in the updated test cases.

🥇 for the detailed description

hpohekar (approved):

mkundu1 (approved):

PR #2357, opened 2024-01-16
feat: workflow updates.

prmukherj:
Here we have made an attempt to remove the string aspects in workflow objects.

Now, meshing.workflow.import_geometry is equivalent to meshing.workflow.task("Import Geometry").

If the user mistypes an attribute, he gets a suggestion of the allowed atrributes in form of python names.


seanpearsonuk:
@prmukherj, the present request is to analyse the usage and recommend further changes. The implemented feature already exists.

prmukherj:
@seanpearsonuk, will do that. Yes, I noticed that this feature exists.

PR #2362, opened 2024-01-17
refactor: Make flobject consistent with server.

prmukherj:
This PR can be merged once server-side changes are in place.

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -33,7 +33,7 @@ settings_logger = logging.getLogger("pyfluent.settings_api")
 
 
 class InactiveObjectError(RuntimeError):
-    """Raised on attempt to access an inactive object."""
+    """Inactive object access.."""
 
     def __init__(self, python_path):
         super().__init__(f"'{python_path}' is currently inactive.")
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -33,10 +33,10 @@ settings_logger = logging.getLogger("pyfluent.settings_api")
 
 
 class InactiveObjectError(RuntimeError):
-    """Provides the error when the object is inactive."""
+    """Raised on attempt to access an inactive object."""
 
-    def __init__(self):
-        super().__init__("Object is not active.")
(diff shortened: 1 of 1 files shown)


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -33,10 +33,10 @@
 
 
 class InactiveObjectError(RuntimeError):
-    """Provides the error when the object is inactive."""
+    """Raised on attempt to access an inactive object."""
```
hpohekar, line 36: Class descriptions begin with verbs ending with "s" or "es".

prmukherj: Then something like gets raised? Doesn't sound right.
cc. @seanpearsonuk

prmukherj: What about: "Provides an error message on attempt to access an inactive object" ?

hpohekar: Provides the error when an inactive object is accessed.

seanpearsonuk: @prmukherj either what you have written or
"Inactive object access"
```
>>> help(ValueError)
Help on class ValueError in module builtins:

class ValueError(Exception)
 |  Inappropriate argument value (of correct type).
...

>>> help(RuntimeError)
Help on class RuntimeError in module builtins:

class RuntimeError(Exception)
 |  Unspecified run-time error.
...

class TypeError(Exception)
 |  Inappropriate argument type.
```
etc

I don't think we actually want to say "Provides the error when..." as we have in the current code.

hpohekar (approved):

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -33,10 +33,10 @@
 
 
 class InactiveObjectError(RuntimeError):
-    """Provides the error when the object is inactive."""
+    """Inactive object access.."""
```
seanpearsonuk, line 36: Looks like a bonus .
```suggestion
    """Inactive object access."""
```

seanpearsonuk (approved):

mkundu1 (approved):

Issue #2351, opened 2024-01-15
Error when running PyFluent in HPC/Slurm environment


christospliakos:
### 🔍 Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### 🐞 Description of the bug

So I'm following the example of using PyFluent with a scheduler on an HPC machine (https://fluent.docs.pyansys.com/version/stable/user_guide/launching_ansys_fluent.html#scheduler-support). I specify the path where FLUENT is installed (tried different paths as well) but no luck.

The way that I'm launching fluent from the python script is: 
`solver = launch_fluent(mode="solver", precision='double', show_gui=False, gpu=True, start_timeout=60)`
Bellow are the bash file and the slurm-out file. 

### 📝 Steps to reproduce

**bash file:**
> 
> #!/bin/bash
> #SBATCH --job-name=FLUENT-2023R2-gpu-case
> #SBATCH --partition=ampere
> #SBATCH --gres=gpu:1
> #SBATCH --nodes=1
> #SBATCH --ntasks-per-node=1
> #SBATCH --time=2:00:00
> 
> module load intel-oneapi-compilers/2022.0.2-sygdcrc python/3.10.10-tuelrz
> module load ansys/2023R2
> 
> #CREATE HOSTFILE
> echo "Creating hostfile..."
> srun hostname > ${SLURM_JOBID}.hostfile
> 
> #ACTIVATE VENV
> echo "Activating virtual environment..."
> . ./myenv/bin/activate
> 
> #EXPORT FLUENT SO IT CAN BE FOUND
> echo "Exporting AWP_ROOT232..."
> export AWP_ROOT232=/mnt/apps/prebuilt/ansys/2023R2/ansys_inc/v232/
> 
> #RUN PYFLUENT
> echo "Running PyFluent..."
> python benchmark_wing.py
> 
> #AS SOON AS RUN IS COMPLETE, DELETE HOSTFILE
> echo "Cleaning up hostfile..."
> rm -f ${SLURM_JOBID}.hostfile

**Slurm Out file:**

> remove ansys 2023R2 (PATH LD_LIBRARY_PATH ANSYS_ROOT ANSYSLI_SERVERS ANSYSLMD_LICENSE_FILE CFX5RSH ANSWAIT)
[...]

PR #2373, opened 2024-01-19
docs: Document additional slurm options

mkundu1:
`scheduler_options` argument of `launch_fluent`:
image

`Launch or connect to Fluent` section of User guide:
image

### DIFF
--- a/src/ansys/fluent/core/launcher/slurm_launcher.py
+++ b/src/ansys/fluent/core/launcher/slurm_launcher.py
@@ -2,8 +2,9 @@
 
 Examples
 --------
->>> # The keys scheduler_headnode, scheduler_queue and scheduler_account are optional
->>> # and should be specified in a similar manner to Fluent's scheduler options.
+Note that the keys ``scheduler_headnode``, ``scheduler_queue`` and ``scheduler_account``
+are optional and should be specified in a similar manner to Fluent's scheduler options.
+
 >>> slurm = pyfluent.launch_fluent(
 ...   scheduler_options={
 ...     "scheduler": "slurm",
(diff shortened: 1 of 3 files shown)


raph-luc:
Should we also update this doc page with the new way to manage and launch Slurm jobs: https://fluent.docs.pyansys.com/version/stable/user_guide/launching_ansys_fluent.html#scheduler-support 
?

See https://github.com/ansys/pyfluent/issues/2351

hpohekar (approved):

seanpearsonuk (approved):

src/ansys/fluent/core/launcher/slurm_launcher.py
```diff
@@ -2,12 +2,22 @@
 
 Examples
 --------
->>> slurm = pyfluent.launch_fluent(scheduler_options={"scheduler": "slurm"})
+>>> # The keys scheduler_headnode, scheduler_queue and scheduler_account are optional
+>>> # and should be specified in a similar manner to Fluent's scheduler options.
```
raph-luc, line 6: I believe that removing this from the code block, and changing it from a code comment to text, should work better:
```suggestion
Note that the keys ``scheduler_headnode``, ``scheduler_queue`` and ``scheduler_account`` are 
optional and should be specified in a similar manner to Fluent's scheduler options.

```

raph-luc: Should we instead link to the more complete documentation at `doc/source/user_guide/launching_ansys_fluent.rst`, rather than repeating this info here?

raph-luc (approved):

src/ansys/fluent/core/launcher/slurm_launcher.py
mkundu1: I feel as the note explains the arguments of the following code, it is better to place it close to the code. I've moved it from comment to text.

PR #2380, opened 2024-01-23
feat: GRPC error interceptor.

prmukherj:


seanpearsonuk (commented):
@prmukherj I didn't check every file, but it looks good.

seanpearsonuk (approved):

src/ansys/fluent/core/services/interceptors.py
```diff
@@ -94,6 +94,31 @@ def intercept_unary_unary(
...
+
+    def _intercept_call(
+        self,
+        continuation: Any,
+        client_call_details: grpc.ClientCallDetails,
+        request: Any,
+    ) -> Any:
+        response = continuation(client_call_details, request)
+        if response.exception() is not None and response.code() != grpc.StatusCode.OK:
+            raise RuntimeError(response.details(), response.code())
```
mkundu1, line 109: I think it is better to display only error details string instead of a tuple. With the following code you can get the original behaviour (won't require to change the unitests):
```
grpc_ex = response.exception()
ex = RuntimeError(grpc_ex.details())
ex.__context__ = grpc_ex
raise ex from None
```

prmukherj: Thank you for this suggestion @mkundu1. Implemented the same.

mkundu1 (approved):

PR #2384, opened 2024-01-23
fix: error-code 0 from Fluent indicates server running without error

mkundu1:


src/ansys/fluent/core/streaming_services/events_streaming.py
```diff
@@ -50,7 +50,11 @@ def _process_streaming(self, id, stream_begin_method, started_evt, *args, **kwar
                 event_name = response.WhichOneof("as")
                 with self._lock:
                     self._streaming = True
-                    if event_name == "errorevent":
+                    # error-code 0 from Fluent indicates server running without error
```
mkundu1, line 53: I've confirmed this from Fluent code.

raph-luc:
Why is the server transmitting an error event with a message saying that no error occurred? Certainly odd. We should probably revise this in the server side as well, though this fix in PyFluent makes sense to me too.

prmukherj (approved):

mkundu1:
> Why is the server transmitting an error event with a message saying that no error occurred? Certainly odd. We should probably revise this in the server side as well, though this fix in PyFluent makes sense to me too.

`error-code = 0` is used in workspace client code for some post-connect actions. We can look into it when/if we plan to revise the workspace code.

PR #2385, opened 2024-01-23
refactor: Clean up the launcher code. (#2375)

prmukherj:
* refactor: Clean up the launcher code.

* Settings root should be re-created in lightweight mode

* refactor: more clean up. (#2376)

* refactor: more clean up.

* Minor fix

* slurm launcher

---------

### DIFF
--- a/src/ansys/fluent/core/launcher/launcher_utils.py
+++ b/src/ansys/fluent/core/launcher/launcher_utils.py
@@ -128,10 +128,10 @@ class FluentMode(Enum):
     """An enumeration over supported Fluent modes."""
 
     # Tuple: Name, Solver object type, Meshing flag, Launcher options
-    MESHING_MODE = ("meshing", Meshing, True, [])
-    PURE_MESHING_MODE = ("pure-meshing", PureMeshing, True, [])
-    SOLVER = ("solver", Solver, False, [])
-    SOLVER_ICING = ("solver-icing", SolverIcing, False, [("fluent_icing", True)])
+    MESHING_MODE = Meshing
+    PURE_MESHING_MODE = PureMeshing
+    SOLVER = Solver
+    SOLVER_ICING = SolverIcing
 
     @staticmethod
     def get_mode(mode: str) -> "FluentMode":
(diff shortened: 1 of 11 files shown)


hpohekar (approved):

mkundu1 (approved):

src/ansys/fluent/core/launcher/launcher_utils.py
```diff
@@ -128,10 +128,10 @@ class FluentMode(Enum):
     """An enumeration over supported Fluent modes."""
 
     # Tuple: Name, Solver object type, Meshing flag, Launcher options
```
seanpearsonuk, line 130: This comment looks out of place here.

I'm not sure about these changes.

src/ansys/fluent/core/launcher/launcher_utils.py
```diff
@@ -153,12 +153,27 @@ def get_mode(mode: str) -> "FluentMode":
...
         for m in FluentMode:
-            if mode == m.value[0]:
+            if mode == m.value._internal_name:
                 return m
-        else:
-            raise DisallowedValuesError(
-                "mode", mode, ["meshing", "pure-meshing", "solver", "solver-icing"]
-            )
+        raise DisallowedValuesError(
+            "mode", mode, ["meshing", "pure-meshing", "solver", "solver-icing"]
```
seanpearsonuk, line 159: I didn't notice this list previously. These items are already known, and this will break in maintenance.

prmukherj: @seanpearsonuk, shall we not pass the allowed values then?

seanpearsonuk: All I mean here is that this is a list that is effectively already known: it can be evaluated based on the enum, correct? If we hard-code it then it can get out of step with the correct value.

prmukherj: What about somethig like:
allowed_modes = []
        for m in FluentMode:
            allowed_modes.append(m.value._internal_name)
            if mode == m.value._internal_name:
                return m
        raise DisallowedValuesError(
            "mode", mode, allowed_modes
        )

seanpearsonuk: Anything like that as long as it works.

prmukherj: @hpohekar, actually we are already looping over these once. So, I didn't create another loop.

seanpearsonuk (approved):

PR #2399, opened 2024-01-30
test: nonexistent_attrs in meshing.workflow

mkundu1:
The issue was occurring when codegen was run and then these steps were followed.

Earlier:
```
>>> hasattr(meshing.workflow, "xyz")
True
>>> meshing.workflow.xyz is None
True
```
Now:
```
>>> hasattr(meshing.workflow, "xyz")
False
>>> meshing.workflow.xyz is None
raises AttributeError
```


src/ansys/fluent/core/workflow.py
```diff
@@ -911,7 +911,9 @@ def __getattr__(self, attr):
         )  # or self._task_with_cmd_matching_help_string(attr)
         if obj:
             return obj
-        return self._task_objects.get(attr, None)
+        if attr in self._task_objects:
+            return self._task_objects[attr]
+        super().__getattribute__(attr)
```
seanpearsonuk, line 916: @mkundu1 Is this line meant to return the value returned by the call?

prmukherj: @seanpearsonuk, yes. This provides an attribute error for non-existing attr instead of returning None as well.

mkundu1: Yes, this should return the value @prmukherj

prmukherj: Done.

src/ansys/fluent/core/workflow.py
```diff
@@ -911,7 +911,9 @@ def __getattr__(self, attr):
         )  # or self._task_with_cmd_matching_help_string(attr)
         if obj:
             return obj
-        return self._task_objects.get(attr, None)
+        if attr in self._task_objects:
+            return self._task_objects[attr]
+        return super().__getattribute__(attr)
```
seanpearsonuk, line 916:
```suggestion
        return self._task_objects.get(attr) or super().__getattribute__(attr)
```

tests/test_meshing_workflow.py
```diff
@@ -741,3 +741,11 @@ def test_modified_workflow(new_mesh_session):
         task_display_names.append(name)
 
     assert set(task_display_names) == task_object_display_names
+
+
+def test_nonexistent_attrs(new_mesh_session):
+    meshing = new_mesh_session
+    assert not hasattr(meshing.workflow, "xyz")
+    with pytest.raises(AttributeError) as msg:
+        meshing.workflow.xyz()
```
seanpearsonuk, line 750:
```suggestion
        meshing.workflow.xyz
```
The test is that access of `xyz` is sufficeint to raise an exception.

tests/test_meshing_workflow.py
```diff
@@ -741,3 +741,11 @@ def test_modified_workflow(new_mesh_session):
...
+    assert msg.value.args[0] == "'MeshingWorkflow' object has no attribute 'xyz'"
```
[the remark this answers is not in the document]
prmukherj: @seanpearsonuk, We need to do something else here?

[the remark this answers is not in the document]
seanpearsonuk: Is it good enough that it caught an attribute error?

Or, if you do want to go deeper in the testing then I would use a custom exception type with methods to access the specific data (e.g., object type, attribute name), which are definitely not arbitrary.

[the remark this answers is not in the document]
prmukherj: Thank you @seanpearsonuk for the suggestion. Actually, earlier non-existing attrs were returning None according to the old implementation, and so we decided to just raise an attribute error to handle that case. This is what we are testing in this current test. Should we do anything else here?

prmukherj (approved):

hpohekar (approved):

PR #2415, opened 2024-02-02
fix: Fix an issue in command creation inside singleton

mkundu1:
The class variable `_full_static_info` at `PyCommand` stores static info for all datamodels, starting from root.
The instance variable `_static_info` at `PyCommand` stores the static-info at command level.
The instance variable `static_info` at `PyCommandArguments` stores the static-info at command arguments level.

### DIFF
--- a/src/ansys/fluent/core/services/datamodel_se.py
+++ b/src/ansys/fluent/core/services/datamodel_se.py
@@ -1,7 +1,7 @@
 """Wrappers over StateEngine based datamodel gRPC service of Fluent."""
 from enum import Enum
 import functools
-import itertools
+from itertools import chain, repeat
 import logging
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1420,6 +1417,7 @@ def __init__(
             self.path = []
         else:
             self.path = path
+        self._static_info = None  # command's static info
```
mkundu1, line 1420: `__full_static_info` stores static info for all datamodels, starting from root at class-level. `_static_info` stores the static-info at command level which is passed with `create_instance`.

seanpearsonuk (approved):

prmukherj (approved):

hpohekar (approved):

PR #2421, opened 2024-02-06
refactor: Make version method private

hpohekar:
`version` is specifically used for codegen. We have renamed it to `_version` so that it is private now.

### DIFF
--- a/doc/source/api/solver/settings.rst
+++ b/doc/source/api/solver/settings.rst
@@ -336,5 +336,5 @@ The following list summarizes common wildcards:
 
 Root object
 -----------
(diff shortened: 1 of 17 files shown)


seanpearsonuk (approved):

prmukherj (approved):

mkundu1:
Some additional internal cleanups which can be considered (can be done in a separate PR)
1. Store a `FluentVersion` object in `connection_properties` of `FluentConnection` and pass that around the services.
2. I think `get_version_for_file_name` is just `FluentVersion.number`.

PR #2422, opened 2024-02-06
fix: issues in workflow and case reader

prmukherj:
This PR fixes a warning that can be seen in case reader while reading project files. Alongside, this also makes the meshing taskobject arguments non read-only. So, we can set it's values. It also has tests for the same.

### DIFF
--- a/tests/test_meshing_workflow.py
+++ b/tests/test_meshing_workflow.py
@@ -255,10 +255,9 @@ def test_accessors_for_argument_sub_items(new_mesh_session):
     assert not import_geom.arguments.FileName()
     import_geom.arguments.FileName = "xyz.txt"
     assert import_geom.arguments.FileName() == "xyz.txt"
-    with pytest.warns(
-        UserWarning, match="No attribute named 'File' in 'Import Geometry'."
-    ):
+    with pytest.raises(AttributeError) as msg:
         import_geom.arguments.File = "sample.txt"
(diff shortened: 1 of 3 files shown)


seanpearsonuk (approved):

hpohekar (approved):

tests/test_meshing_workflow.py
```diff
@@ -234,10 +234,41 @@ def test_accessors_for_argument_sub_items(new_mesh_session):
...
-    assert import_geom.arguments.FileName.is_read_only()
-    assert import_geom.arguments.CadImportOptions.OneZonePer.is_read_only()
+    assert not import_geom.arguments.FileName.is_read_only()
+    assert not import_geom.arguments.FileName()
+    import_geom.arguments.FileName = "xyz.txt"
+    assert import_geom.arguments.FileName() == "xyz.txt"
+    with pytest.warns(
+        UserWarning, match="No attribute named 'File' in 'Import Geometry'."
+    ):
+        import_geom.arguments.File = "sample.txt"
```
mkundu1, line 261: Usually Python throws `AttributeError` in such scenarios.

prmukherj: Actually we are excepting AttributeError and passing warning so as not to stop the code execution.

seanpearsonuk (approved):

mkundu1 (approved):

PR #2429, opened 2024-02-06
fix: Accept GPU IDs in -gpu parameter of launch_fluent

mkundu1:


src/ansys/fluent/core/launcher/fluent_launcher_options.json
```diff
@@ -49,16 +49,6 @@
         },
         "fluent_format": "{}"
     },
-    "gpu": {
```
mkundu1, line 52: Complex arguments of type like `bool|list` cannot be handled here.

seanpearsonuk (approved):

hpohekar (approved):

prmukherj (approved):

PR #2436, opened 2024-02-09
refactor: Fluent GPU Solver is only supported for 3D.

hpohekar:
Fluent GPU Solver is only supported for 3D. Now we are raising `GPUSolverSupportError` exception if `gpu=True` and version is not "3d".

image

`------------------------------------------`

image

### DIFF
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -227,9 +227,8 @@ def launch_fluent(
     The allocated machines and core counts are queried from the scheduler environment and
     passed to Fluent.
     """
-    if gpu:
-        if version != "3d":
-            raise GPUSolverSupportError()
+    if version != "3d" and gpu:
+        raise GPUSolverSupportError()
     _process_kwargs(kwargs)
     del kwargs
     fluent_launch_mode = _get_fluent_launch_mode(
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -227,8 +227,8 @@ def launch_fluent(
     The allocated machines and core counts are queried from the scheduler environment and
     passed to Fluent.
     """
-    if version and gpu:
-        if version != "3d" and gpu:
+    if gpu:
+        if version != "3d":
             raise GPUSolverSupportError()
     _process_kwargs(kwargs)
     del kwargs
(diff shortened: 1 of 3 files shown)


src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -226,6 +227,9 @@ def launch_fluent(
     The allocated machines and core counts are queried from the scheduler environment and
     passed to Fluent.
     """
+    if version and gpu:
+        if version != "3d" and gpu:
+            raise GPUSolverSupportError()
```
prmukherj, line 232: @hpohekar, please add a test for this.

seanpearsonuk: Why two ifs like this? Why not just the second one?

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -226,6 +227,9 @@ def launch_fluent(
     The allocated machines and core counts are queried from the scheduler environment and
     passed to Fluent.
     """
+    if version and gpu:
```
seanpearsonuk, line 230: Why two ifs like this? Why not just the second one?

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -226,6 +227,9 @@ def launch_fluent(
...
+        if version != "3d" and gpu:
```
seanpearsonuk, line 231: Why two ifs like this? Why not just the second one?

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -226,6 +227,8 @@ def launch_fluent(
     The allocated machines and core counts are queried from the scheduler environment and
     passed to Fluent.
     """
+    if version != "3d" and gpu:
```
seanpearsonuk, line 230: 👍

src/ansys/fluent/core/launcher/launcher.py
hpohekar: @seanpearsonuk I thought explicit string based comparison should be checked only if both version and gpu is given. Anyway I updated it now. Thank you.

src/ansys/fluent/core/launcher/launcher.py
hpohekar: > @hpohekar, please add a test for this.

Done. Thank you.

mkundu1 (approved):

src/ansys/fluent/core/launcher/launcher.py
hpohekar: @seanpearsonuk  No. If we move current check to _process_invalid_args then checking current condition (`if version != "3d" and gpu:`) after execution of following 6 statements is redundant and therefore to avoid this we have added this check at first line of function.

image

seanpearsonuk (approved):

PR #2440, opened 2024-02-10
feat: get git build details.

prmukherj:
Get build details from git.

image

While importing pyfluent we get build details of pyfluent. On the other hand while launching or connecting to a session we get the Fluent build details.


src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -226,6 +227,7 @@ def launch_fluent(
     The allocated machines and core counts are queried from the scheduler environment and
     passed to Fluent.
     """
+    print(get_build_details_string())
```
prmukherj, line 230: What do you think about this?

seanpearsonuk (approved):

hpohekar (approved):

src/ansys/fluent/core/get_build_details.py
```diff
@@ -0,0 +1,43 @@
...
+        pass
+    return build_details
+
+
+def build_version_string():
+    build_details = get_build_version()
+    build_string = ""
+    for key, value in build_details.items():
+        build_string += key + ":" + f" {value}  "
+    print(build_string)
```
seanpearsonuk, line 43:
```suggestion
def build_version_string():
    print("  ".join([f"{k}: {v}" for k,v in get_build_version().items()]))
```
- Note that this function is misnamed. Its name should start with the word print. There's nothing wrong with retaining `build_version_string()` and printing the result.
- I have replaced the mixture of concatenation and fstring with pure fstring, which is more readable and more flexible.

mkundu1 (approved):
Looking at the pandas and dask, they expose a module variable `__git_version__` alongwith `__version__`.
https://github.com/pandas-dev/pandas/blob/main/pandas/__init__.py#L190
https://github.com/dask/dask/blob/main/dask/__init__.py#L20

`__version__` and `__git_version__` together will provide the complete information about the build. Looks like they are both using the [versioneer](https://pypi.org/project/versioneer/) package to get the git information.

Sorry if this is bit late, perhaps something to consider as future improvement.

PR #2428, opened 2024-02-06
fix: Avoid child classes as attributes of parent class

mkundu1:
To be merged after #2289 (some adjustment will be needed with that PR's changes)

PR #2448, opened 2024-02-13
feat: Backward compatibility for settings API

mkundu1:
This PR adds the backward compatbility support in PyFluent settings API using the `child-aliases`/`command-aliases`/`query-aliases` attributes in settings API.

A tree of alias-objects mimicking the original object tree is constructed on-access basis during runtime. For an object hierarchy `A -> B -> C`, If `A1` is an alias of `A`, `A1.B` and `A1.C` are also aliases and the backward compatibility message for set_state is shown at all levels. The `_Alias` base class prints the backward compatibility message on various journaling points using the scheme_eval function which is passed through the `get_root()` function.

Please see the Fluent side PR for more details on flobject.py change.

I'm also incorporating changes from #2428 which simplifies some code requiring access to original settings object classes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,7 +147,7 @@ build-doc-source:
 	@sudo rm -rf doc/source/api/meshing/tui
 	@sudo rm -rf doc/source/api/solver/datamodel
 	@sudo rm -rf doc/source/api/solver/tui
-	#@sudo rm -rf doc/source/api/solver/_autosummary/settings
+	@sudo rm -rf doc/source/api/solver/_autosummary/settings
 	@sudo rm -rf /home/ansys/.local/share/ansys_fluent_core/examples
 	@pip install -r requirements/requirements_doc.txt
 	@xvfb-run make -C doc html
--- a/Makefile
+++ b/Makefile
@@ -147,7 +147,7 @@ build-doc-source:
 	@sudo rm -rf doc/source/api/meshing/tui
 	@sudo rm -rf doc/source/api/solver/datamodel
 	@sudo rm -rf doc/source/api/solver/tui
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -147,7 +147,7 @@ build-doc-source:
 	@sudo rm -rf doc/source/api/meshing/tui
 	@sudo rm -rf doc/source/api/solver/datamodel
 	@sudo rm -rf doc/source/api/solver/tui
-	@sudo rm -rf doc/source/api/solver/_autosummary/settings
+	#@sudo rm -rf doc/source/api/solver/_autosummary/settings
```
mkundu1, line 150: temporarily to test doc build

prmukherj (approved):

hpohekar (approved):

seanpearsonuk (approved):

tests/test_settings_api.py
```diff
@@ -139,3 +140,71 @@ def test_api_upgrade(new_solver_session, capsys):
     case_path = download_file("Static_Mixer_main.cas.h5", "pyfluent/static_mixer")
     solver.tui.file.read_case(case_path)
     "<solver_session>.file.read_case" in capsys.readouterr().out
+
+
+@pytest.mark.skip(reason="Skipping till docker image is updated")
```
mkundu1, line 145: As there is a significant change in flobject.py, I'll merge bith Fluent and PyFluent PR at the same time. I'll enable the new test after Fluent docker image is updated.

mkundu1: I a;so need to push a dev version as the new PyConsole implmentation will not work with newer Fluent + older PyFluent.

PR #2462, opened 2024-02-16
feat: Snake case workflow arguments access.

prmukherj:
Arguments access in meshing workflow is made snake case now.
For example:
```
>>> watertight.create_surface_mesh.CFDSurfaceMeshControls.MaxSize.set_state()
```
is now
```
>>> watertight.create_surface_mesh.cfd_surface_mesh_controls.max_size.set_state()
```
For more details please look into the updated docs and tests.

### DIFF
--- a/src/ansys/fluent/core/workflow.py
+++ b/src/ansys/fluent/core/workflow.py
@@ -16,10 +16,13 @@ AttrMap = {}
 
 
 def camel_to_snake_case(input_arg: str) -> str:
+    """Convert camel case input string to snake case output string."""
     return re.sub("((?<=[a-z0-9])[A-Z]|(?!^)[A-Z](?=[a-z]))", r"_\1", input_arg).lower()
 
 
 def populate_and_use_snake_case(attr: str, attr_dict):
+    """Populate the snake-case attribute map and
+    return snake case of the passed attribute."""
     if attr.islower():
         for key in attr_dict.keys():
             if camel_to_snake_case(str(key)) not in AttrMap:
@@ -148,6 +151,7 @@ class BaseTask:
                 _ordered_children=[],
                 _task_list=[],
                 _task_objects={},
+                _dynamic_interface=command_source._dynamic_interface,
             )
         )
 
@@ -304,8 +308,9 @@ class BaseTask:
         self._command_source.DeleteTasks(ListOfTasks=[self.name()])
 
     def __getattr__(self, attr):
-        snake_attr = populate_and_use_snake_case(str(attr), self.arguments())
-        attr = snake_attr if snake_attr else attr
+        if self._dynamic_interface:
+            snake_attr = populate_and_use_snake_case(str(attr), self.arguments())
+            attr = snake_attr if snake_attr else attr
         try:
             result = getattr(self._task, attr)
             if result:
@@ -326,8 +331,20 @@ class BaseTask:
             setattr(self._task, attr, value)
 
     def __dir__(self):
+        arg_list = []
+        for arg in self.arguments():
+            if self._dynamic_interface:
+                arg_list.append(camel_to_snake_case(arg))
+            else:
+                arg_list.append(arg)
+
         return sorted(
-            set(list(self.__dict__.keys()) + dir(type(self)) + dir(self._task))
+            set(
+                list(self.__dict__.keys())
+                + dir(type(self))
+                + dir(self._task)
+                + arg_list
+            )
         )
 
     def add_child_to_task(self):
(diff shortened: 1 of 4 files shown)


prmukherj:
@seanpearsonuk and @mkundu1, the dir needs updating to expose arguments as well?

src/ansys/fluent/core/workflow.py
```diff
@@ -11,6 +12,27 @@
...
+    return re.sub("((?<=[a-z0-9])[A-Z]|(?!^)[A-Z](?=[a-z]))", r"_\1", input_arg).lower()
+
+
+def populate_and_use_snake_case(attr: str, attr_dict):
+    """Populate the snake-case attribute map and return snake case of the passed
+    attribute."""
+    if attr.islower():
+        for key in attr_dict.keys():
+            if camel_to_snake_case(str(key)) not in AttrMap:
+                AttrMap[camel_to_snake_case(str(key))] = str(key)
```
seanpearsonuk, line 29: - This doesn't keep the result of `cache_to_snake_case()` which would avoid an unnecessary second call.
- The functionality is not specific to workflow.

src/ansys/fluent/core/workflow.py
```diff
@@ -11,6 +12,27 @@
 from ansys.fluent.core.data_model_cache import DataModelCache
 from ansys.fluent.core.services.datamodel_se import PyCallableStateObject, PyMenuGeneric
 
+AttrMap = {}
```
seanpearsonuk, line 15: This could be an attribute of `populate_and_use_snake_case()`.

src/ansys/fluent/core/workflow.py
```diff
@@ -11,6 +12,27 @@
...
+
+
+def camel_to_snake_case(input_arg: str) -> str:
+    """Convert camel case input string to snake case output string."""
+    return re.sub("((?<=[a-z0-9])[A-Z]|(?!^)[A-Z](?=[a-z]))", r"_\1", input_arg).lower()
+
+
+def populate_and_use_snake_case(attr: str, attr_dict):
```
seanpearsonuk, line 23: This function does not require a `dict` to be passed in. The caller should call `keys()` is the source object happens to be a `dict`.

src/ansys/fluent/core/workflow.py
```diff
@@ -754,7 +805,7 @@ def add_child_and_update(self, state=None):
         state : Optional[dict]
             Optional state.
         """
-        self.add_child(state)
+        self._add_child(state)
```
seanpearsonuk, line 808: How does this terminate?

src/ansys/fluent/core/workflow.py
```diff
@@ -3,14 +3,51 @@
...
+camel_to_snake_case.cache = {}
+
+
+def snake_to_camel_case(snake_case_str: str, camel_case_strs: Iterable):
+    """Populate the snake-case attribute map and return camel case of the passed
+    attribute."""
+    try:
+        return snake_to_camel_case.cache[snake_case_str]
+    except KeyError:
+        if snake_case_str.islower():
```
mkundu1, line 36: In which scenaio this is not a lowercase?

src/ansys/fluent/core/workflow.py
```diff
@@ -3,14 +3,51 @@
...
+            for camel_case_str in camel_case_strs:
+                _snake_case_str = camel_to_snake_case(str(camel_case_str))
```
mkundu1, line 38: When is this not a string? It is better if these conversion function simply operates on strings (i.e. should not try to do more than that like getting string representation of arbitrary python objects).

src/ansys/fluent/core/workflow.py
```diff
@@ -3,14 +3,51 @@
...
+                if _snake_case_str not in snake_to_camel_case.cache:
+                    snake_to_camel_case.cache[_snake_case_str] = str(camel_case_str)
+
+        try:
+            return snake_to_camel_case.cache[snake_case_str]
+        except KeyError:
+            return
```
mkundu1, line 45: If the result must be one of `camel_case_strs`, prefer to throw exception. That will catch programming error elsewhere.

src/ansys/fluent/core/workflow.py
```diff
@@ -3,14 +3,51 @@
...
+    attribute."""
+    try:
+        return snake_to_camel_case.cache[snake_case_str]
+    except KeyError:
+        if snake_case_str.islower():
+            for camel_case_str in camel_case_strs:
+                _snake_case_str = camel_to_snake_case(str(camel_case_str))
+                if _snake_case_str not in snake_to_camel_case.cache:
+                    snake_to_camel_case.cache[_snake_case_str] = str(camel_case_str)
+
```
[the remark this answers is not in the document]
prmukherj: Done, Thank you.

mkundu1 (approved):

seanpearsonuk (approved):

PR #2466, opened 2024-02-16
feat:                   units follow-up with testing

seanpearsonuk:
- **Follow-up on initial units integration into PyFluent.**
- upgrade ansys-units dependency
- enable a temporarily disabled test step due to this upgrade
- extend the testing
- fix an flobject assignment issue where exceptions were inadvertently hidden
- treat None value for units-quantity to indicate dimensionless
- implement exception handling
- rename `get_state_as_quantity()` (too long) as `as_quantity()`
- make the generation of the table data stricter

### DIFF
--- a/src/ansys/fluent/core/solver/flunits.py
+++ b/src/ansys/fluent/core/solver/flunits.py
@@ -82,6 +82,8 @@ Unhandled:
  'wave-length': 'Angstrom'}
 """
 
+from typing import Optional
+
 _fl_unit_table = {
     "acceleration": "m s^-2",
     "angle": "radian",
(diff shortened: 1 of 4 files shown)


src/ansys/fluent/core/solver/flunits.py
```diff
@@ -43,9 +93,7 @@ def make_python_fl_unit_table(scheme_unit_table):
     "collision-rate": "m^-3 s^-1",
     "concentration": "kmol m^-3",
```
seanpearsonuk, line 94: mol, not kmol is an SI unit so this is wrong

src/ansys/fluent/core/solver/flunits.py
```diff
@@ -109,14 +146,9 @@ def make_python_fl_unit_table(scheme_unit_table):
     "pressure-time-deriv-sqr": "Pa^2 s^-2",
     "pressure-time-derivative": "Pa s^-1",
     "resistance": "m^-1",
-    "site-density": "kgmol m^-2",
     "soot-formation-constant-unit": "kg N^-1 m^-1 s^-1",
-    "soot-limiting-nuclei-rate": "1e+15-particles/m3-s",
     "soot-linear-termination": "m^3 s^-1",
-    "soot-oxidation-constant": "kg m kgmol^-1 K^-0.5 s^-1",
-    "soot-pre-exponential-constant": "1.e15 kg^-1 s^-1",
     "soot-sitespecies-concentration": "kmol m^-3",
```
seanpearsonuk, line 151: mol not kmol is an SI unit so this is wrong

src/ansys/fluent/core/solver/flunits.py
```diff
@@ -59,13 +107,9 @@ def make_python_fl_unit_table(scheme_unit_table):
     "elec-charge": "A h",
```
seanpearsonuk, line 107: this is wrong. 
I'm going to add an extra constraint: Instantiate a unit quantity for each unit and compare its value to the SI value.

seanpearsonuk: This is all done and updated

prmukherj (approved):

hpohekar (approved):

PR #2478, opened 2024-02-20
feat: state with units

seanpearsonuk:
- **Add `RealList` to units support**.
- Generalize the implementations for `Real` and `RealList` in a shared mixin base class.
- Change the behaviour of method `real_with_units`. Previously it returned None if the units were undefined or if the value were not `float` (or `int`). That would be particularly unsatisfactory for nested state, so we construct the tuple unconditionally now.
- Added test steps for the `RealList`.
- ansys.units update is required, as commented in test code.
- added code and test for **nested** state with units. **This means a lot more changes since the approvals!**

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -581,10 +581,13 @@ class SettingsBase(Base, Generic[StateT]):
     def _add_units_to_state(self, state):
         if isinstance(state, collections.abc.Mapping):
             for k, v in state.items():
+                child = None
                 if isinstance(self, collections.abc.Mapping):
-                    child = self[k]
-                else:
-                    child = getattr(self, k, None)
+                    try:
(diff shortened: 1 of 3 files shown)


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -640,9 +648,12 @@ class Boolean(SettingsBase[bool], Property):
     _state_type = bool
 
 
-class RealList(SettingsBase[RealListType], Numerical):
+class RealList(SettingsBase[RealListType], RealNumerical):
     """A ``RealList`` object representing a real list setting."""
 
+    base_set_state = SettingsBase[RealType].set_state
```
seanpearsonuk, line 654: mistake

mkundu1 (approved):

prmukherj (approved):

hpohekar (approved):

PR #2483, opened 2024-02-22
docs: update units doc

seanpearsonuk:
- **update units docs**
- reflect that we now support `RealList` as well as `Real`.
- ...and container level `state_with_units`.
- remove the part that implied that `ansys-units` is an optional dependency, which was misleading.


hpohekar (approved):

doc/source/api/solver/settings.rst
```diff
@@ -107,18 +107,21 @@ and ``NamedObject`` types, the state value is a dictionary. For the
...
+The ``Real`` and ``RealList`` parameter objects in PyFluent can incorporate units
+alongside values. If an object supports units, you can retrieve its value and units
+as an ``ansys.units.Quantity`` object using the ``as_quantity`` method. Alternatively,
+you can obtain the same information as a tuple by calling the ``state_with_units``
+method. If the container object supports units, calling ``state_with_units`` on it
+returns a ``dict`` where relevant values are represented as tuples containing both
+the value and units.
+
+Both ``ansys.units.Quantity`` objects and value-unit tuples can be used with the
+``set_state`` method of ``Real`` or ``RealList`` objects.
```
PipKat, line 119:
```suggestion
The ``Real`` and ``RealList`` parameter objects in PyFluent can incorporate units
alongside values. If an object supports units, you can retrieve its value and units
as an ``ansys.units.Quantity`` object using the ``as_quantity()`` method. Alternatively,
you can obtain the same information as a tuple by calling the ``state_with_units()``
method. If the container object supports units, calling the ``state_with_units()`` method on it
returns a dictionary where relevant values are represented as tuples containing both
the value and units.

Both ``ansys.units.Quantity`` objects and value-unit tuples can be used with the
``set_state()`` method of ``Real`` or ``RealList`` objects.
```
Minor edits as method names are supposed to include the ``()`` after them. Nice working on meeting all other style objectives!

seanpearsonuk: Thank you, @PipKat: I had the same thought, and instead followed the approach previously adopted in this file. I will accept this suggested change and update other instances herein.

PR #2486, opened 2024-02-22
refactor: Fix error handling msg in flobject.

prmukherj:
Improve error handling message.

### DIFF
--- a/src/ansys/fluent/core/filereader/case_file.py
+++ b/src/ansys/fluent/core/filereader/case_file.py
@@ -194,7 +194,7 @@ class CaseVariable:
                 allowed_name_error_message(
                     context="config-vars",
                     trial_name=name,
-                    allowed_values=list(self._variables.keys()),
+                    allowed_values=list(self._variables),
                 )
             )
 
--- a/src/ansys/fluent/core/filereader/case_file.py
+++ b/src/ansys/fluent/core/filereader/case_file.py
@@ -192,7 +192,9 @@ class CaseVariable:
         except KeyError:
             raise ValueError(
                 allowed_name_error_message(
-                    "config-vars", name, list(self._variables.keys())
+                    context="config-vars",
+                    trial_name=name,
+                    allowed_values=list(self._variables.keys()),
                 )
             )
 
--- a/src/ansys/fluent/core/solver/error_message.py
+++ b/src/ansys/fluent/core/solver/error_message.py
@@ -18,7 +18,7 @@ def allowed_name_error_message(
     """Provide an error message with the closest names matching the 'trial_name' from
     the 'allowed_values' list."""
     if not message:
-        message = f"'{context}' has no attribute '{trial_name}'."
+        message = f"'{context}' has no attribute '{trial_name}'"
     message += ".\n"
     matches = None
     if allowed_values:
--- a/src/ansys/fluent/core/solver/error_message.py
+++ b/src/ansys/fluent/core/solver/error_message.py
@@ -18,7 +18,7 @@ def allowed_name_error_message(
     """Provide an error message with the closest names matching the 'trial_name' from
     the 'allowed_values' list."""
     if not message:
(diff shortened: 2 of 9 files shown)


src/ansys/fluent/core/filereader/case_file.py
```diff
@@ -192,7 +192,9 @@ def __call__(self, name: str = ""):
         except KeyError:
             raise ValueError(
                 allowed_name_error_message(
-                    "config-vars", name, list(self._variables.keys())
+                    context="config-vars",
+                    trial_name=name,
+                    allowed_values=list(self._variables.keys()),
```
seanpearsonuk, line 197:
```suggestion
                    allowed_values=list(self._variables),
```

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1015,7 +1019,9 @@ def __getattr__(self, name: str):
         except KeyError:
             raise AttributeError(
                 allowed_name_error_message(
-                    "Settings objects", name, self.get_active_child_names()
+                    context="Settings objects",
```
seanpearsonuk, line 1022: Probably, this should be raising _from_.

src/ansys/fluent/core/solver/error_message.py
```diff
@@ -10,11 +10,16 @@ def closest_allowed_names(trial_name: str, allowed_names: str) -> List[str]:
...
+    allowed_values: Any,
+    context: Optional[str] = None,
+    trial_name: Optional[str] = None,
+    message: Optional[str] = None,
 ) -> str:
     """Provide an error message with the closest names matching the 'trial_name' from
     the 'allowed_values' list."""
-    message = f"{trial_name} is not an allowed {context} name.\n"
+    if not message:
+        message = f"{trial_name} is not an allowed {context} name"
```
seanpearsonuk, line 21: Why only generate the message conditionally? This is the whole point if this function.

src/ansys/fluent/core/solver/flobject.py
seanpearsonuk, line 1022: Does this generate f"{trial_name} is not an allowed Settings objects name"?

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1202,7 +1208,9 @@ def __getitem__(self, name: str) -> ChildTypeT:
                 )
             raise KeyError(
                 allowed_name_error_message(
-                    "Settings objects", name, self.get_object_names()
+                    context="Settings objects",
```
seanpearsonuk, line 1211: same question here

src/ansys/fluent/core/solver/flobject.py
prmukherj: yes

src/ansys/fluent/core/solver/error_message.py
prmukherj, line 21:
```suggestion
        message = f"'{context}' has no attribute '{trial_name}'."
```

hpohekar (approved):

src/ansys/fluent/core/solver/error_message.py
```diff
@@ -10,11 +10,16 @@ def closest_allowed_names(trial_name: str, allowed_names: str) -> List[str]:
...
+    allowed_values: Any,
+    context: Optional[str] = None,
+    trial_name: Optional[str] = None,
+    message: Optional[str] = None,
 ) -> str:
     """Provide an error message with the closest names matching the 'trial_name' from
     the 'allowed_values' list."""
-    message = f"{trial_name} is not an allowed {context} name.\n"
+    if not message:
+        message = f"'{context}' has no attribute '{trial_name}'"
```
seanpearsonuk, line 21: We are still not spot on here. E.g., _Settings objects has_. So in those cases we should send the object name, equivalent to the other flobject scenarios.

seanpearsonuk (approved):

PR #2490, opened 2024-02-23
feat: units percent and ohms and handling nested quantities

seanpearsonuk:
- Extend units coverage by massaging the Fluent table a bit more.
- Fix an issue for setting quantities at `x` where `x.value` is a `Real`*. WARNING: this is a hack.

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -595,7 +595,14 @@ class SettingsBase(Base, Generic[StateT]):
     def set_state(self, state: Optional[StateT] = None, **kwargs):
         """Set the state of the object."""
         with self.while_setting_state():
-            return self.flproxy.set_var(self.path, self.to_scheme_keys(kwargs or state))
+            if isinstance(state, (tuple, ansys_units.Quantity)) and hasattr(
+                self, "value"
+            ):
+                self.value.set_state(state, **kwargs)
+            else:
+                return self.flproxy.set_var(
+                    self.path, self.to_scheme_keys(kwargs or state)
+                )
 
     @staticmethod
     def _print_state_helper(state, out, indent=0, indent_factor=2):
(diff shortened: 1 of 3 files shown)


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -595,7 +595,14 @@ def get_state(self) -> StateT:
     def set_state(self, state: Optional[StateT] = None, **kwargs):
         """Set the state of the object."""
         with self.while_setting_state():
-            return self.flproxy.set_var(self.path, self.to_scheme_keys(kwargs or state))
+            if isinstance(state, (tuple, ansys_units.Quantity)) and hasattr(
```
seanpearsonuk, line 598: We get an exception for tuple, so that case is catchable, but for `Quantity` the code falls through to defaulting to string and the object is implicitly convertible, so no exception is catchable - we just end up sending a rubbish expression. Hence this hack.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -595,7 +595,14 @@ def get_state(self) -> StateT:
     def set_state(self, state: Optional[StateT] = None, **kwargs):
         """Set the state of the object."""
         with self.while_setting_state():
```
seanpearsonuk, line 597: I noticed that we don't have the with part in the new overriding `set_state()` method.

tests/test_flobject.py
```diff
@@ -1035,63 +1032,6 @@ def test_ansys_units_integration(load_mixing_elbow_mesh):
     )
 
 
-@pytest.mark.fluent_version(">=24.1")
```
seanpearsonuk, line 1038: In fact, it's a dependency, so all this is a bit pointless

mkundu1 (approved):

prmukherj (approved):

hpohekar (approved):

PR #2492, opened 2024-02-26
feat: Expose workflow and dm methods in snake_case.

prmukherj:


src/ansys/fluent/core/workflow.py
```diff
@@ -1178,6 +1183,181 @@ def refresh_after_sleep(_):
...
+        None
+
+        Raises
+        ------
+        RuntimeError
+            If command_name does not match a task name.
+        """
+        self._populate_help_string_command_id_map()
+        if command_name not in self._help_string_command_id_map:
+            raise RuntimeError(
```
prmukherj, line 1289: ValueError here?

src/ansys/fluent/core/workflow.py
```diff
@@ -380,14 +381,55 @@ def __dir__(self):
...
         return self._task.InsertCompoundChildTask()
 
+    def get_next_possible_tasks(self) -> str:
+        """Get the list of possible names of commands that can be inserted as tasks
+        after this current task is executed."""
+        next_tasks = []
+        for task in self._task.GetNextPossibleTasks():
+            next_tasks.append(camel_to_snake_case(task))
+
+        return next_tasks
```
seanpearsonuk, line 402: list comprehension preferred

src/ansys/fluent/core/workflow.py
```diff
@@ -1109,7 +1157,7 @@ def __dir__(self):
         """Override the behaviour of dir to include attributes in WorkflowWrapper and
         the underlying workflow."""
         arg_list = [camel_to_snake_case(arg) for arg in dir(self._workflow)]
-        return list(
+        dir_list = list(
             dict.fromkeys(
```
seanpearsonuk, line 1161: redundant?

src/ansys/fluent/core/workflow.py
```diff
@@ -1121,6 +1169,14 @@ def __dir__(self):
...
             )
         )
+        unwanted = [
+            "reset_workflow",
+            "initialize_workflow",
+            "load_workflow",
+            "create_new_workflow",
+        ]
+        dir_list = [e for e in dir_list if e not in unwanted]
+        return dir_list
```
seanpearsonuk, line 1179: could sort here

src/ansys/fluent/core/workflow.py
```diff
@@ -1121,6 +1169,14 @@ def __dir__(self):
                 )
             )
         )
+        unwanted = [
+            "reset_workflow",
+            "initialize_workflow",
+            "load_workflow",
+            "create_new_workflow",
+        ]
+        dir_list = [e for e in dir_list if e not in unwanted]
```
seanpearsonuk, line 1178: actually both args could be `set`s, leading to a bit of a nice rewrite over all here

src/ansys/fluent/core/workflow.py
```diff
@@ -1178,6 +1234,197 @@ def refresh_after_sleep(_):
...
+        -------
+        None
+
+        Raises
+        ------
+        AttributeError
+            Not implemented in the new structure.
+        """
+        raise AttributeError(
+            "'reset_workflow' is not implemented in the new workflow structure."
```
seanpearsonuk, line 1254: as discussed, we'll remove all these directly.

src/ansys/fluent/core/workflow.py
```diff
@@ -1178,6 +1234,197 @@ def refresh_after_sleep(_):
...
+                    command_obj_instance = command_obj.create_instance()
+                    help_str = command_obj_instance.get_attr("helpString")
+                    if help_str and help_str.islower():
+                        self._help_string_command_id_map[help_str] = command
+                    del command_obj_instance
+
+    def get_possible_tasks(self):
+        """Get the list of possible names of commands that can be inserted as tasks."""
+        self._populate_help_string_command_id_map()
+        return list(self._help_string_command_id_map.keys())
```
seanpearsonuk, line 1335: `keys()` call not needed

src/ansys/fluent/core/workflow.py
```diff
@@ -1178,6 +1234,197 @@ def refresh_after_sleep(_):
...
+            List of task items.
+
+        Returns
+        -------
+        None
+
+        Raises
+        ------
+        RuntimeError
+            If command_name does not match a task name.
```
seanpearsonuk, line 1379:
```suggestion
            If command_name does not match a task name. None of the tasks is deleted.
```

src/ansys/fluent/core/workflow.py
```diff
@@ -1178,6 +1234,197 @@ def refresh_after_sleep(_):
...
+        for task_name in list_of_tasks:
+            try:
+                camel_case_list_of_tasks.append(
+                    self._help_string_command_id_map[task_name]
+                )
+            except KeyError:
+                raise RuntimeError(
+                    f"'{task_name}' is not an allowed command task.\n"
+                    "Please use 'get_allowed_task_list()' to view list of allowed command tasks."
+                )
```
seanpearsonuk, line 1392: from?

src/ansys/fluent/core/workflow.py
```diff
@@ -1178,6 +1234,197 @@ def refresh_after_sleep(_):
...
+        """
+        self._populate_help_string_command_id_map()
+        camel_case_list_of_tasks = []
+        for task_name in list_of_tasks:
+            try:
+                camel_case_list_of_tasks.append(
+                    self._help_string_command_id_map[task_name]
+                )
+            except KeyError:
+                raise RuntimeError(
```
seanpearsonuk, line 1389: ValueError

src/ansys/fluent/core/workflow.py
seanpearsonuk, line 1: @prmukherj `_add_child` / `add_child_and_update` seems to require the state keys in camel case. I can only see tests using it in test_meshing_workflow.py rather than in test_new_meshing_workflow.py. test_new_meshing_workflow.py should contain all the enhanced-level tests.

src/ansys/fluent/core/workflow.py
```diff
@@ -1109,18 +1155,14 @@ def __dir__(self):
...
-                )
-            )
+        dir_list = set(
+            list(self.__dict__.keys())
+            + dir(type(self))
+            + arg_list
+            + self.child_task_python_names()
         )
+        dir_list = dir_list - self._unwanted_attrs
+        return sorted(dir_list)
```
seanpearsonuk, line 1165:
```suggestion
        dir_set = set(
            list(self.__dict__.)
            + dir(type(self))
            + arg_list
            + self.child_task_python_names()
        )
        dir_set = dir_set - self._unwanted_attrs
        return sorted(filter(None, dir_set))
```
A few changes here again - the last line is the key one!

seanpearsonuk: N.b. python names of dynamic tasks are None so they had to be filtered out here. Otherwise `-` raises!

src/ansys/fluent/core/workflow.py
```diff
@@ -1002,6 +1041,13 @@ def __init__(self, workflow: PyMenuGeneric, command_source: PyMenuGeneric) -> No
         self._main_thread_ident = None
         self._task_objects = {}
         self._dynamic_interface = False
+        self._help_string_command_id_map = {}
+        self._unwanted_attrs = {
+            "reset_workflow",
+            "initialize_workflow",
+            "load_workflow",
+            "create_new_workflow",
```
seanpearsonuk, line 1049:
```suggestion
            "reset_workflow",
            "initialize_workflow",
            "load_workflow",
            "create_new_workflow",
            "fault_tolerant",
            "part_management",
            "pm_file_management",
            "service",
            "task_object",
            "watertight",
            "workflow"]
```
I also got confused here because `rename()` is a datamodel method (I did not recognize it thus immediately) and `Rename()`/`rename()` exists at task level. So, although it is clear that `rename()` is a datamodel method here, when we are dealing with tasks we surely have a collision!

@prmukherj

seanpearsonuk: @prmukherj I just noticed that we are still exposing `get_attrib_value()` which can remove altogether from datamodel interface!

src/ansys/fluent/core/workflow.py
prmukherj: Done that, thank you. I have updated the tests in test_meshing_workflow.py as well. Actually, there are two ways. None of the tests with new structure should work in the old manner. So, I think we can keep the tests in the original place and keep the pure object-oriented tests here in test_new_meshing_workflow.py. If you think otherwise, then we can just move these tests to the new file.

seanpearsonuk:
@prmukherj if you merge this into the other branch now then we can view all the changes in a single PR

src/ansys/fluent/core/workflow.py
prmukherj: Move these tests to new file.

PR #2493, opened 2024-02-26
feat: support inout file purpose

hpohekar:
Support inout file purpose.

### DIFF
--- a/src/ansys/fluent/core/services/datamodel_se.py
+++ b/src/ansys/fluent/core/services/datamodel_se.py
@@ -22,6 +22,7 @@ from ansys.fluent.core.services.interceptors import (
     WrapApiCallInterceptor,
 )
 from ansys.fluent.core.services.streaming import StreamingService
+from ansys.fluent.core.solver.error_message import allowed_name_error_message
 
 Path = list[tuple[str, str]]
 _TValue = Union[None, bool, int, float, str, Sequence["_TValue"], dict[str, "_TValue"]]
@@ -47,6 +48,22 @@ def _get_value_from_message_dict(
         else:
             d = next(filter(None, (d.get(x) for x in k)))
     return d
+
+
+class DisallowedFilePurpose(ValueError):
+    """Provides the error when the specified file purpose is not in allowed values."""
+
+    def __init__(
+        self,
+        context: Optional[Any] = None,
+        name: Optional[Any] = None,
+        allowed_values: Optional[Any] = None,
+    ):
+        super().__init__(
+            allowed_name_error_message(
+                context=context, trial_name=name, allowed_values=allowed_values
+            )
+        )
 
 
 class InvalidNamedObject(RuntimeError):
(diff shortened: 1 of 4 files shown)


src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -49,6 +50,22 @@ def _get_value_from_message_dict(
     return d
 
 
+class DisallowedFilePurpose(ValueError):
+    """Provides the error when the specified file purpose is not in allowed values."""
```
seanpearsonuk, line 54:
```suggestion
    """Raised when the specified file purpose is not in allowed values."""
```

seanpearsonuk (approved):
Just one comment.

N.b. we applied the same re-phrasing of exception class docstring in ansys-units. Maybe in a separate PR< you'd l;ike to do the same for pyfluent.

@hpohekar

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -49,6 +50,22 @@ def _get_value_from_message_dict(
     return d
 
 
+class DisallowedFilePurpose(ValueError):
+    """Is raised when the specified file purpose is not in allowed values."""
```
PipKat, line 54:
```suggestion
    """Provides the error raised when the specified file purpose is not in the allowed values."""
```
Descriptions for classes should begin with a verb ending in "s" or "es" for consistency.

hpohekar: @seanpearsonuk @PipKat 

Yes, Thanks for the suggestion. 

I agree that descriptions for classes should begin with a verb ending in "s" or "es". We have followed this guideline in all custom exception classes. ("Provides the error...")

I think we are following this guideline in current case as well because "Is" also a supporting _verb_ ending with "s". 

We are planning to change all docstrings from "Provides the error when the..." to "Is raised when the..." because we think that "Is raised..." is more appropriate.

@PipKat  Could you please confirm that whether "Is raised..." follows documentation guideline or not, because we want to update all docstrings in new format ("Is raised...") and we also want to follow documentation guidelines.

Thank you.

seanpearsonuk: The description begins _Is_ which is a verb ending with s, so there is no issue.

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1443,6 +1460,13 @@ def _get_file_purpose(self, arg):
             elif file_purpose == "output":
                 if _OutputFile not in self.__class__.__bases__:
                     self.__class__.__bases__ += (_OutputFile,)
+            elif file_purpose == "inout":
+                if _InOutFile not in self.__class__.__bases__:
+                    self.__class__.__bases__ += (_InOutFile,)
+            else:
```
seanpearsonuk, line 1466: check if file_purpose is not None outside.

src/ansys/fluent/core/services/datamodel_se.py
seanpearsonuk: @PipKat We will avoid gratuitous use of _is_, but it works very well in this narrow case. The doc-string reflects exactly how you would explain the class to someone conversationally whereas _provides the error_ does not.

prmukherj (approved):

PR #2498, opened 2024-02-27
refactor: rename remote_file_handler to file_transfer_service

hpohekar:
Renamed `remote_file_handler` to `file_transfer_service`.

### DIFF
--- a/src/ansys/fluent/core/session_solver.py
+++ b/src/ansys/fluent/core/session_solver.py
@@ -77,16 +77,17 @@ class Solver(BaseSession):
     def __init__(
         self,
         fluent_connection,
-        remote_file_handler: Optional[Any] = None,
+        file_transfer_service: Optional[Any] = None,
     ):
         """Solver session.
 
         Args:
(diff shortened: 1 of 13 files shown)


src/ansys/fluent/core/session_solver.py
```diff
@@ -77,16 +77,17 @@ class Solver(BaseSession):
     def __init__(
         self,
         fluent_connection,
-        remote_file_handler: Optional[Any] = None,
+        file_transfer_service: Optional[Any] = None,
```
mkundu1, line 80: This will need a Fluent-side change. I was passing a PimFileTransferService instance as remote_file_handler due to a previous issue, now we can remove that keyword argument from Fluent code.

seanpearsonuk (approved):

src/ansys/fluent/core/session_solver.py
hpohekar: @mkundu1 Done. Thank you. !469694, t994337

mkundu1 (approved):

PR #2499, opened 2024-02-27
docs: Start all custom exception class docstring with Is raised...

hpohekar:
Replaced "Provides the error ...." with "Raised ...." in all custom exception classes docstring.

### DIFF
--- a/src/ansys/fluent/core/exceptions.py
+++ b/src/ansys/fluent/core/exceptions.py
@@ -6,7 +6,7 @@ from ansys.fluent.core.solver.error_message import allowed_name_error_message
 
 
 class DisallowedValuesError(ValueError):
-    """Is raised when an argument value is not in the allowed values."""
+    """Raised when an argument value is not in the allowed values."""
 
     def __init__(
         self,
@@ -22,13 +22,13 @@ class DisallowedValuesError(ValueError):
 
 
 class InvalidArgument(ValueError):
-    """Is raised when an argument value is inappropriate."""
+    """Raised when an argument value is inappropriate."""
 
     pass
 
 
 class SurfaceSpecificationError(ValueError):
-    """Is raised when both ``surface_ids`` and ``surface_names`` are provided."""
+    """Raised when both ``surface_ids`` and ``surface_names`` are provided."""
 
     def __init__(self):
         super().__init__("Provide either 'surface_ids' or 'surface_names'.")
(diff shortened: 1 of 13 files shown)


prmukherj (approved):

mkundu1:
"Raised when ..." sounds better.

The build-in exceptions follows the same format - https://docs.python.org/3/library/exceptions.html#concrete-exceptions.

seanpearsonuk (approved):

PipKat:
> "Raised when ..." sounds better.
> 
> The built-in exceptions follow the same format - https://docs.python.org/3/library/exceptions.html#concrete-exceptions.

I'd be onboard with using "Raised" and following the Python org format.

seanpearsonuk:
> > "Raised when ..." sounds better.
> > The built-in exceptions follow the same format - https://docs.python.org/3/library/exceptions.html#concrete-exceptions.
> 
> I'd be OK with using "Raised" to follow the Python org format, although I'd prefer using "Raises" to be consistent with our own docstring style.

Yes "Raised when

> > "Raised when ..." sounds better.
> > The built-in exceptions follow the same format - https://docs.python.org/3/library/exceptions.html#concrete-exceptions.
> 
> I'd be OK with using "Raised" to follow the Python org format, although I'd prefer using "Raises" to be consistent with our own docstring style.

Yes, _Raised when_ is our intention here and _Is raised when_ was meant as a compromise. I don't know how to start a docstring for an exception class with _Raises_; @PipKat?

PipKat:
> > > "Raised when ..." sounds better.
> > > The built-in exceptions follow the same format - https://docs.python.org/3/library/exceptions.html#concrete-exceptions.
> > 
> > 
> > I'd be OK with using "Raised" to follow the Python org format, although I'd prefer using "Raises" to be consistent with our own docstring style.
> 
> Yes "Raised when
> 
> > > "Raised when ..." sounds better.
> > > The built-in exceptions follow the same format - https://docs.python.org/3/library/exceptions.html#concrete-exceptions.
> > 
> > 
> > I'd be OK with using "Raised" to follow the Python org format, although I'd prefer using "Raises" to be consistent with our own docstring style.
> 
> Yes, _Raised when_ is our intention here and _Is raised when_ was meant as a compromise. I don't know how to start a docstring for an exception class with _Raises_; @PipKat?

@seanpearsonuk Point taken. To use "Raises" we'd need to rewrite as this: "Raises the error that occurs when ...." I know this seems wordy and might possibly cause line length issues for some error classes.  However, without the brief description including what is raised (an error), it isn't clear on its own. I know that the type of error appears after the class name, so I won't waste anymore of your time. Go ahead and use whatever you decide is best.

hpohekar:
@mkundu1 @seanpearsonuk @prmukherj 

We have used "Raised when..." everywhere.

PR #2502, opened 2024-02-27
fix: version check and scp reading when running inside a container

ochernuk:
When running Fluent within a container, the version check in System Coupling module was throwing an exception. Fixed by using the correct version query.

SCP file written by Fluent needs to be downloaded so it can be read in by PyFluent. Adjusted the code to use that.

### DIFF
--- a/src/ansys/fluent/core/systemcoupling.py
+++ b/src/ansys/fluent/core/systemcoupling.py
@@ -75,24 +75,22 @@ class SystemCoupling:
         def get_scp_string() -> str:
             """Get SCP file contents in the form of the XML string."""
 
-            fluent_cwd = self._solver.connection_properties.cortex_pwd
-            scp_file_name = os.path.join(fluent_cwd, "fluent.scp")
+            scp_file_name = "fluent.scp"
             self._solver.setup.models.system_coupling.write_scp_file(
                 file_name=scp_file_name
             )
-            local_scp_file_name = scp_file_name
-            if self._solver.connection_properties.inside_container:
-                local_scp_file_name = os.path.join(os.getcwd(), "fluent.scp")
-                self._solver.download(scp_file_name, local_scp_file_name)
+
+            # download the file locally in case Fluent is in a container
+            self._solver.download(scp_file_name)
 
             assert os.path.exists(
-                local_scp_file_name
+                scp_file_name
             ), "ERROR: could not create System Coupling .scp file"
 
-            with open(local_scp_file_name, "r") as f:
+            with open(scp_file_name, "r") as f:
                 xml_string = f.read()
 
-            os.remove(local_scp_file_name)
+            os.remove(scp_file_name)
             return xml_string
 
         def get_name(xml_element) -> str:
--- a/src/ansys/fluent/core/systemcoupling.py
+++ b/src/ansys/fluent/core/systemcoupling.py
@@ -72,6 +72,7 @@ class SystemCoupling:
         self._solver.setup.models.system_coupling.init_and_solve()
 
     def __get_syc_setup(self) -> dict:
+
         def get_scp_string() -> str:
             """Get the SCP file contents in the form of an XML string."""
 
--- a/src/ansys/fluent/core/systemcoupling.py
+++ b/src/ansys/fluent/core/systemcoupling.py
@@ -72,7 +72,6 @@ class SystemCoupling:
         self._solver.setup.models.system_coupling.init_and_solve()
 
     def __get_syc_setup(self) -> dict:
-
         def get_scp_string() -> str:
             """Get the SCP file contents in the form of an XML string."""
 
--- a/src/ansys/fluent/core/systemcoupling.py
+++ b/src/ansys/fluent/core/systemcoupling.py
@@ -83,7 +83,7 @@ class SystemCoupling:
             local_scp_file_name = scp_file_name
             if self._solver.connection_properties.inside_container:
                 local_scp_file_name = os.path.join(os.getcwd(), "fluent.scp")
-                f.download(scp_file_name, local_scp_file_name)
+                self._solver.download(scp_file_name, local_scp_file_name)
 
             assert os.path.exists(
                 local_scp_file_name
--- a/src/ansys/fluent/core/systemcoupling.py
+++ b/src/ansys/fluent/core/systemcoupling.py
@@ -86,7 +86,7 @@ class SystemCoupling:
 
             assert os.path.exists(
                 scp_file_name
-            ), "ERROR: could not create System Coupling .scp file"
+            ), "ERROR: could not create System Coupling SCP file"
 
             with open(scp_file_name, "r") as f:
                 xml_string = f.read()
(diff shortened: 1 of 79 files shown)


src/ansys/fluent/core/systemcoupling.py
```diff
@@ -73,19 +74,25 @@ def solve(self) -> None:
...
+
             fluent_cwd = self._solver.connection_properties.cortex_pwd
             scp_file_name = os.path.join(fluent_cwd, "fluent.scp")
             self._solver.setup.models.system_coupling.write_scp_file(
                 file_name=scp_file_name
             )
+            local_scp_file_name = scp_file_name
+            if self._solver.connection_properties.inside_container:
+                local_scp_file_name = os.path.join(os.getcwd(), "fluent.scp")
+                f.download(scp_file_name, local_scp_file_name)
```
raph-luc, line 86: Shouldn't this be something like
```suggestion
                self._solver.download(scp_file_name, local_scp_file_name)
```
?

src/ansys/fluent/core/systemcoupling.py
```diff
@@ -73,19 +74,25 @@ def solve(self) -> None:
...
         def get_scp_string() -> str:
             """Get SCP file contents in the form of the XML string."""
+
             fluent_cwd = self._solver.connection_properties.cortex_pwd
             scp_file_name = os.path.join(fluent_cwd, "fluent.scp")
             self._solver.setup.models.system_coupling.write_scp_file(
                 file_name=scp_file_name
             )
+            local_scp_file_name = scp_file_name
+            if self._solver.connection_properties.inside_container:
```
raph-luc, line 84: Would `<session>.fluent_connection._remote_instance` be a better way to check if remote?

Alternatively, I think we don't even need to check if remote here, in my understanding `<session>.download()` won't do anything nor raise errors if PyPIM isn't configured and if it isn't running remote. @hpohekar is this correct?

Otherwise we'd probably need a more elegant way to check if running remote

iboyd-ansys: I think this is correct. I have been looking into the Fluent implementation just recently and my understanding is that `download` reduces to a no-op if PyPIM isn't configured.

hpohekar: @raph-luc No.

We raise `PyPIMConfigurationError` if PyPIM is not configured.

image

image

raph-luc: Ok thanks for the clarification @hpohekar, then in my understanding the description of `PimFileTransferService` is incorrect as it implies these should be no-ops:
https://github.com/ansys/pyfluent/blob/caddd75a20c8f72975295b4f3d297fd2a2588962/src/ansys/fluent/core/utils/file_transfer_service.py#L15-L19

cc @seanpearsonuk 

@ochernuk @iboyd-ansys in this case checking `<session>.fluent_connection._remote_instance` is probably the way to go for now

iboyd-ansys: My apologies - it turns out I was looking at an older version of the PyFluent code. It looks to me that it used to behave more like the way @raph-luc described.

seanpearsonuk: @raph-luc The intention has always been that `upload()` and `download()` are no-ops if no service is available. That's so that the same script works both in local and remote contexts. This was certainly the status in the past, and there ought to have been a test for that (I believe that there was). @hpohekar Has there been a conscious change of behaviour in this respect?

src/ansys/fluent/core/systemcoupling.py
```diff
@@ -73,19 +74,25 @@ def solve(self) -> None:
     def __get_syc_setup(self) -> dict:
         def get_scp_string() -> str:
             """Get SCP file contents in the form of the XML string."""
```
PipKat, line 76:
```suggestion
            """Get SCP file contents in the form of an XML string."""
```

src/ansys/fluent/core/systemcoupling.py
hpohekar: @seanpearsonuk @raph-luc 

Earlier we had only single file transfer service (PimFileTransferService). Now we can plug in any file transfer service which has upload/download service therefore we are raising appropriate exception if particular service is not configured.

seanpearsonuk: @hpohekar OK, I see. Please could you make a simple change so that upload an dowload calls can just do nothing and not raise. That is still valid. It still comes back to the script reuse requirement.

hpohekar: @seanpearsonuk Yes, sure.

seanpearsonuk: Thanks

src/ansys/fluent/core/systemcoupling.py
```diff
@@ -44,9 +44,10 @@ class SystemCoupling:
     def __init__(self, solver):
         self._solver = solver
         # version check - this requires Fluent 2024 R1 or newer.
-        if FluentVersion.get_latest_installed() < FluentVersion.v241:
+        current_fluent_version = FluentVersion(self._solver.get_fluent_version())
+        if current_fluent_version < FluentVersion.v241:
```
seanpearsonuk, line 48: Hmm, `get_fluent_version()` should return a `FluentVersion`. The caller can get the `value` attribute which is the string currently returned, I will add a PyFluent issue for this. We don't want to change this at the moment.

dnwillia-work (approved):

src/ansys/fluent/core/systemcoupling.py
```diff
@@ -72,12 +73,16 @@ def solve(self) -> None:
...
-            scp_file_name = os.path.join(fluent_cwd, "fluent.scp")
+            """Get the SCP file contents in the form of an XML string."""
+
+            scp_file_name = "fluent.scp"
             self._solver.setup.models.system_coupling.write_scp_file(
                 file_name=scp_file_name
             )
+
+            # download the file locally in case Fluent is in a container
+            self._solver.download(scp_file_name)
```
raph-luc, line 84: Note this is going to raise errors when running locally and/or without PyPIM, until we improve this behavior as discussed in the other comment chain

raph-luc (approved):

src/ansys/fluent/core/systemcoupling.py
```diff
@@ -72,12 +73,17 @@ def solve(self) -> None:
...
                 file_name=scp_file_name
             )
+
+            # download the file locally in case Fluent is in a container
+            if self._solver.fluent_connection._remote_instance != None:
+                self._solver.download(scp_file_name)
+
             assert os.path.exists(
                 scp_file_name
             ), "ERROR: could not create System Coupling .scp file"
```
PipKat, line 89:
```suggestion
            ), "ERROR: could not create System Coupling SCP file"
```
When only specifying the extension part of a file type, use all capital letters.

PR #2503, opened 2024-02-28
Maint/refactor workflow wrappers

prmukherj:


src/ansys/fluent/core/workflow.py
```diff
@@ -938,7 +938,7 @@ def __init__(
         command_source: Union[ClassicWorkflow, EnhancedWorkflow],
         task: str,
     ) -> None:
-        """Initialize CompoundTask.
+        """Initialize a compound task.
```
prmukherj, line 941:
```suggestion
        """Initialize CompoundTask.
```

src/ansys/fluent/core/workflow.py
```diff
@@ -149,7 +149,7 @@ def __init__(
         command_source: Union[ClassicWorkflow, EnhancedWorkflow],
         task: str,
     ) -> None:
-        """Initialize BaseTask.
+        """Initialize the basetask.
```
prmukherj, line 152:
```suggestion
        """Initialize BaseTask.
```

src/ansys/fluent/core/meshing/meshing_workflow.py
```diff
@@ -38,53 +38,86 @@ def __init__(
         self,
         workflow: PyMenuGeneric,
         meshing: PyMenuGeneric,
-        part_management: PyMenuGeneric,
-        pm_file_management: PyMenuGeneric,
     ) -> None:
-        """Initialize MeshingWorkflow.
+        """Initialize EnhancedMeshingWorkflow.
```
seanpearsonuk, line 42: 👍

src/ansys/fluent/core/meshing/meshing_workflow.py
```diff
@@ -38,53 +38,86 @@ def __init__(
...
-            Part-management file-management object.
+            The meshing object.
         """
         super().__init__(workflow=workflow, command_source=meshing)
-        self._is_ftm = False
-        self._part_management = part_management
-        self._pm_file_management = pm_file_management
+
+
+class WaterTightMeshingWorkflow(EnhancedMeshingWorkflow):
```
seanpearsonuk, line 54:
```suggestion
class WatertightMeshingWorkflow(EnhancedMeshingWorkflow):
```

src/ansys/fluent/core/meshing/meshing_workflow.py
```diff
@@ -38,53 +38,86 @@ def __init__(
...
+    """Provides water-tight meshing specialization of the workflow wrapper."""
+
+    def __init__(self, workflow: PyMenuGeneric, meshing: PyMenuGeneric) -> None:
```
seanpearsonuk, line 57: 👍

src/ansys/fluent/core/meshing/meshing_workflow.py
```diff
@@ -38,53 +38,86 @@ def __init__(
...
+        """Initialize WaterTightMeshingWorkflow.
```
seanpearsonuk, line 58:
```suggestion
        """Initialize WatertightMeshingWorkflow.
```

src/ansys/fluent/core/meshing/meshing_workflow.py
```diff
@@ -38,53 +38,86 @@ def __init__(
...
+        super().__init__(workflow=workflow, meshing=meshing)
 
     def watertight(self, dynamic_interface: bool) -> None:
         """Initialize a watertight workflow.
 
         Parameters
         ----------
         dynamic_interface : bool
-            xxx
+            Flag to expose object-oriented behaviour.
```
seanpearsonuk, line 75: Is this flag ever false now? In other words, can we remove it?

src/ansys/fluent/core/meshing/meshing_workflow.py
```diff
@@ -38,53 +38,86 @@ def __init__(
...
         """
 
         self._new_workflow(
             name="Watertight Geometry", dynamic_interface=dynamic_interface
         )
-        self._is_ftm = False
+
+
+class FaultTolerantMeshingWorkflow(EnhancedMeshingWorkflow):
```
seanpearsonuk, line 83: 👍

src/ansys/fluent/core/meshing/meshing_workflow.py
```diff
@@ -38,53 +38,86 @@ def __init__(
...
+        self._pm_file_management = pm_file_management
 
     def fault_tolerant(self, dynamic_interface: bool):
         """Initialize a fault-tolerant workflow.
 
         Parameters
         ----------
         dynamic_interface : bool
-            xxx
+            Flag to expose object-oriented behaviour.
```
seanpearsonuk, line 116: of course, same question here

src/ansys/fluent/core/session_base_meshing.py
```diff
@@ -4,7 +4,8 @@
 from ansys.fluent.core.fluent_connection import FluentConnection
 from ansys.fluent.core.meshing.meshing_workflow import (
     ClassicMeshingWorkflow,
-    EnhancedMeshingWorkflow,
+    FaultTolerantMeshingWorkflow,
+    WaterTightMeshingWorkflow,
```
seanpearsonuk, line 8:
```suggestion
    WatertightMeshingWorkflow,
```

src/ansys/fluent/core/session_base_meshing.py
```diff
@@ -146,10 +147,20 @@ def workflow(self):
         return self._old_workflow
 
     @property
-    def new_workflow(self):
+    def new_wt_workflow(self):
```
seanpearsonuk, line 150: Since we are always returning a stored instance, new_ is not appropriate. We can drop that part and expand the contraction: `watertight_workflow`.

src/ansys/fluent/core/session_base_meshing.py
```diff
@@ -146,10 +147,20 @@ def workflow(self):
...
         if not self._new_workflow:
-            self._new_workflow = EnhancedMeshingWorkflow(
+            self._new_workflow = WaterTightMeshingWorkflow(
+                self._workflow_se,
+                self.meshing,
+            )
+        return self._new_workflow
+
+    @property
+    def new_ft_workflow(self):
```
seanpearsonuk, line 160: `fault_tolerant_workflow`

seanpearsonuk: If it's always the same cached object, how can you truly switch workflows? Even if you maintain two separate objects for wt and ft respectively, you still need to be careful to avoid a caching problem when you try to revert back to a given workflow type because the cached object would be invalid.

src/ansys/fluent/core/workflow.py
```diff
@@ -1303,7 +1299,7 @@ def insert_new_task(self, command_name: str):
         )
 
     def delete_tasks(self, list_of_tasks: list[str]):
-        """Delete the list of tasks passed as an argument.
+        """Delete the list of tasks passed as argument.
```
seanpearsonuk, line 1302:
```suggestion
        """Delete the provided list of tasks.
```

src/ansys/fluent/core/workflow.py
```diff
@@ -1370,7 +1366,7 @@ def create_composite_task(self, list_of_tasks: list[str]):
 
 
 class ClassicWorkflow:
-    """Wraps a meshing workflow object.
+    """Wraps meshing workflow object.
```
seanpearsonuk, line 1369: Original version reads better,

tests/test_new_meshing_workflow.py
seanpearsonuk, line 1: If there aren't any tests that switch between workflows then we should add that. It should switch multiple times, including reinitialising on the same type.

src/ansys/fluent/core/session_pure_meshing.py
```diff
@@ -109,13 +109,13 @@ def workflow(self):
...
-        return self._base_meshing.new_workflow
+        self._base_meshing.new_wt_workflow.watertight(dynamic_interface)
+        return self._base_meshing.new_wt_workflow
 
     def fault_tolerant(self, dynamic_interface=True):
         """Get a new fault-tolerant workflow."""
-        self._base_meshing.new_workflow.fault_tolerant(dynamic_interface)
-        return self._base_meshing.new_workflow
+        self._base_meshing.new_ft_workflow.fault_tolerant(dynamic_interface)
+        return self._base_meshing.new_ft_workflow
```
seanpearsonuk, line 118: is all this code here because the other meshing mode inherits from it?

prmukherj: yes

seanpearsonuk: OK good

tests/test_new_meshing_workflow.py
```diff
@@ -920,3 +921,65 @@ def downstream_names(task):
...
+            == f"'WatertightMeshingWorkflow' object has no attribute '{attr}'"
+        )
+
+    watertight.import_geometry.file_name.set_state(import_file_name)
+    watertight.import_geometry.length_unit = "in"
+    watertight.import_geometry()
+
+    assert watertight.import_geometry.file_name()
+    # Resets the workflow:
+    watertight.watertight()
```
seanpearsonuk, line 952: This should be via an explict reset command.

prmukherj: yes, sounds good. I also need to understand some concerns with reseting in the meeting today.

tests/test_new_meshing_workflow.py
```diff
@@ -634,7 +635,7 @@ def test_watertight_workflow_dynamic_interface(mixing_elbow_geometry, new_mesh_s
         watertight.create_volume_mesh
     assert (
         msg.value.args[0]
-        == "'EnhancedMeshingWorkflow' object has no attribute 'create_volume_mesh'"
+        == "'WatertightMeshingWorkflow' object has no attribute 'create_volume_mesh'"
```
seanpearsonuk, line 638: - We don't want such assertions.
- `pytest.raises()` has an optional `match` argument which supports regex.
- From Python 3.10, AttributeError has `obj` and `name` attributes which can be checked directly.

tests/test_new_meshing_workflow.py
```diff
@@ -920,3 +921,65 @@ def downstream_names(task):
         "Enclose Fluid Regions (Capping)",
         "Create Regions",
     ]
+
+
+@pytest.mark.skip("Randomly failing in CI")
```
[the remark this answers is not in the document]
seanpearsonuk: Maybe this one is failing because it does not sleep whereas the above test does sleep

tests/test_new_meshing_workflow.py
seanpearsonuk: - regex approah might be best right now as it fits neatly into the pytest framework, thereby eliminating boilerplate.

prmukherj: Done something with this already, but will look into it further.

prmukherj: Will remove the error msg string check.

prmukherj: Done.

tests/test_new_meshing_workflow.py
prmukherj: Done.

tests/test_new_meshing_workflow.py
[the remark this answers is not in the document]
prmukherj: I have tried both de-activating the data-model cache and adding custom sleep upto 30 s, but both could not resolve this random issue on CI. For now I have deactivated all these 3 tests in CI and I am testing them locally.
cc. @seanpearsonuk, @mkundu1, @hpohekar.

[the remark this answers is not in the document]
hpohekar: Yes. @prmukherj These tests are failing in poetry PR as well. https://github.com/ansys/pyfluent/pull/2512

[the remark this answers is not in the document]
prmukherj: @hpohekar, I think these are not the same. Poetry PR is failing for some other reason.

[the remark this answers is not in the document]
hpohekar: @prmukherj Ok. I will check it. I thought poetry PR contains changes related to pyproject.toml file only, no source code has changed.

tests/test_new_meshing_workflow.py
prmukherj: Done.

mkundu1:
Regarding the naming conventions like `ClassicMeshingWorkflow, EnhancedMeshingWorkflow`, does a naming `ClassicMeshingWorkflow, MeshingWorkflow` sound better? The unprefixed name will be prominent in a standard search and will be understood as the most preferred to use.

@prmukherj @seanpearsonuk

src/ansys/fluent/core/meshing/faulttolerant.py
```diff
@@ -16,19 +13,10 @@ def fault_tolerant_workflow(**launch_args) -> EnhancedMeshingWorkflow:
```
mkundu1, line 13: Can we pass just the meshing session to this function?

`launch_args` can be complex and it is better to handle that complexity in a single place, i.e. in `launch_fluent()` method. If we allow all laucher args in this function, we need to document them here also as this is a user-facing function (and also about the arg vals which are not supported in meshing mode).

IMHO, the wrapper should not be responsible for managing the session instance, i.e. if user needs to use the session instance while or after working with the wrapper, is there a way to do that?

prmukherj: I think this makes sense. I have removed the launch part from here. I don't remember exactly why it was added here, @seanpearsonuk, was there a valid reason that we are missing now?

seanpearsonuk:
> Regarding the naming conventions like `ClassicMeshingWorkflow, EnhancedMeshingWorkflow`, does a naming `ClassicMeshingWorkflow, MeshingWorkflow` sound better? The unprefixed name will be prominent in a standard search and will be understood as the most preferred to use.
> 
> @prmukherj @seanpearsonuk

Yes, this makes sense. Thanks, @mkundu1.

src/ansys/fluent/core/meshing/faulttolerant.py
seanpearsonuk: @prmukherj we should make the changes suggested by @mkundu1.

prmukherj: Already pushed @seanpearsonuk

prmukherj: Regarding the renaming I was awaiting your feedback. Will do it soon. Thank you.

hpohekar (approved):

mkundu1 (approved):

Issue #629, opened 2022-07-12
Move to poetry or flit


seanpearsonuk:

hpohekar:
@seanpearsonuk What are the required changes ?

seanpearsonuk:
@hpohekar, This is an old issue, based on a recommendation from @mkundu1. We can wait to hear from him whether any change should be planned regarding this.

PR #2511, opened 2024-02-29
build: Use poetry

hpohekar:
Added poetry in pyproject.toml. Closes https://github.com/ansys/pyfluent/issues/629


hpohekar:
https://github.com/ansys/pyfluent/pull/2512

PR #2512, opened 2024-02-29
build: Use poetry

hpohekar:
Added poetry in pyproject.toml. Closes https://github.com/ansys/pyfluent/issues/629


pyproject.toml
```diff
@@ -1,10 +1,97 @@
...
+    { include = "tui_222", from = "src/ansys/fluent/core/meshing", to = "ansys/fluent/core/meshing" },
+    { include = "tui_231", from = "src/ansys/fluent/core/meshing", to = "ansys/fluent/core/meshing" },
+    { include = "tui_232", from = "src/ansys/fluent/core/meshing", to = "ansys/fluent/core/meshing" },
+    { include = "tui_241", from = "src/ansys/fluent/core/meshing", to = "ansys/fluent/core/meshing" },
+    { include = "tui_242", from = "src/ansys/fluent/core/meshing", to = "ansys/fluent/core/meshing" },
+    { include = "settings_222", from = "src/ansys/fluent/core/solver", to = "ansys/fluent/core/solver" },
+    { include = "settings_231", from = "src/ansys/fluent/core/solver", to = "ansys/fluent/core/solver" },
+    { include = "settings_232", from = "src/ansys/fluent/core/solver", to = "ansys/fluent/core/solver" },
+    { include = "settings_241", from = "src/ansys/fluent/core/solver", to = "ansys/fluent/core/solver" },
+    { include = "settings_242", from = "src/ansys/fluent/core/solver", to = "ansys/fluent/core/solver" },
```
mkundu1, line 39: We need to include codegen folders because we are installing before and after codegen?

There is also tui_*.py in solver folder and fluent_version_*.py in core folder.

Makefile
```diff
@@ -25,112 +26,144 @@ unittest: unittest-dev-241
 unittest-dev-222:
 	@echo "Running unittests"
 	@sudo rm -rf /home/ansys/.local/share/ansys_fluent_core/examples
-	@pip install -r requirements/requirements_tests.txt
+	@pip uninstall poetry -y
+	@pip install poetry>=1.8.2
```
mkundu1, line 30: Is it better to remove this repetation with a requirements_build.txt?

pyproject.toml
mkundu1, line 1: How are we creating the reader and tests packages?

Makefile
hpohekar: The main issue that even after installing poetry twice in `Install pyfluent` and `Install again after codegen` still we are getting following error in each unit test check.

image

To resolve above error we added above duplication otherwise we don't need those 2 lines each time.

hpohekar: If we are going to retain requirements_*.txt then use of poetry will not serve it's purpose fully because I think it replaces setup.py, requirements_*.txt.

pyproject.toml
hpohekar: Yes, Thank you.

Makefile
mkundu1: Yes, poetry needs to be installed separately in the VM that is executing the unittests, because poetry is not a dependency of PyFluent. I meant to replace the above lines in all places with `pip install -r requirements/requirements_build.txt`. `requirements_build.txt` would contain `poetry>=1.8.2`.

pyproject.toml
```diff
@@ -1,10 +1,97 @@
...
+pytest-sphinx = "^0.5.0"
+sphinx-autobuild = "^2021.3.14"
+sphinx-autodoc-typehints = "^1.25.3"
+sphinx-copybutton = "^0.5.2"
+sphinx-gallery = "^0.15.0"
+sphinx-notfound-page = "^1.0.0"
+sphinxcontrib-websupport = "^1.2.7"
+sphinxemoji = "^0.3.1"
+sphinx-toggleprompt = "^0.5.2"
+autodocsumm = "^0.2.12"
```
mkundu1, line 79: Our doc dependencies have version matching clause (`==`) in the requirements file and those are updated by dependabot through a PR. In past, we have seen issues when a later version is automatically picked up during PR runs. I think it is safer to retain the version matchig clause for doc dependecies (may same for tests dependencies).

Makefile
hpohekar: Done. Thank you.

pyproject.toml

pyproject.toml
```diff
@@ -1,10 +1,83 @@
...
+
+[tool.poetry.group.test]
+optional = true
+
+[tool.poetry.group.test.dependencies]
+pytest-mock = "==3.12.0"
+pytest-cov = "==4.1.0"
+pytest = "==8.0.2"
+h5py = "==3.10.0"
+reader = "==3.7"
```
mkundu1, line 75: what is this package?

pyproject.toml
```diff
@@ -1,10 +1,83 @@
...
+autodocsumm = "==0.2.12"
+
+[tool.poetry.group.test]
+optional = true
+
+[tool.poetry.group.test.dependencies]
+pytest-mock = "==3.12.0"
+pytest-cov = "==4.1.0"
+pytest = "==8.0.2"
+h5py = "==3.10.0"
```
mkundu1, line 74: h5py is a dependency of the optional package "reader". We can to install pyfluent with the "reader" optional package during unitesting and remove h5py from test dependencies.

pyproject.toml
hpohekar: @mkundu1 Yes, I verified this. 

We have commented disabled API generated files section in `.gitignore`

image

We even don't need `include = [{ path = "src/ansys/fluent/core/solver/settings_242", format = ["sdist", "wheel"] },]` the generated files are included by poetry.

Thank you very much.

pyproject.toml
hpohekar: h5py was specified in test requirements.

image

pyproject.toml
mkundu1: @hpohekar We shouldn't modify the `.gitignore` file, instead we should reinstate the explicit includes with globbing patterns.

pyproject.toml
mkundu1: I know, the suggestion is about general improvement.

pyproject.toml
hpohekar: It's reader package that needs h5py. 

@mkundu1 I think we can specify it follows.

[tool.poetry.dependencies]
h5py = { version = "^3.10.0", optional = true }

[tool.poetry.extras]
reader= ["h5py "]

pyproject.toml
hpohekar: Ok. sure. We will include all of them using globbing patterns. Thank you.

pyproject.toml
hpohekar: Ok. I see. We will remove it. Thank you.

hpohekar:
@mkundu1 

We have already installed all required dependencies but still it shows that pytest is not available.

https://github.com/ansys/pyfluent/actions/runs/8143729062/job/22257679049

image

pyproject.toml

pyproject.toml

pyproject.toml

pyproject.toml
hpohekar: We are building "reader" and "tests" optional packages as follows. (any 1 way)

image

```
poetry install --extras "reader tests"
poetry install -E reader -E tests
poetry install --all-extras
```

mkundu1:
I believe changing

pip install -q --force-reinstall dist/*.whl > /dev/null

to

pip install -q --force-reinstall dist/*.whl[reader] > /dev/null

in ci.yml should fix the current h5py related issue.

hpohekar:
> I believe changing
> 
> ```
> pip install -q --force-reinstall dist/*.whl > /dev/null
> ```
> 
> to
> 
> ```
> pip install -q --force-reinstall dist/*.whl[reader] > /dev/null
> ```
> 
> in ci.yml should fix the current h5py related issue.

Right. Thanks a lot.

.github/workflows/ci.yml
```diff
@@ -508,7 +511,9 @@ jobs:
 
       - name: Unit Testing
         if: ${{ !contains(github.event.pull_request.title, '[skip tests]') }}
-        run: make unittest-dev-${{ matrix.version }}
+        run: |
+          make install-test
```
mkundu1, line 515: Should this also be called in nightly/weekly testsing CIs?

mkundu1 (approved):
Great work! thanks

seanpearsonuk (approved):

hpohekar:
> Great work! thanks

Thank you very much.

hpohekar:
@mkundu1 We have added `make install-test` in all `Unit Testing` checks. `make install-test` was added in ci.yml only. Thank you.

PR #2524, opened 2024-03-02
feat: Minimize number of fluent launches during codegen

mkundu1:
Reducing number of times Fluent is launched from 8 to 3 while running allapigen.py. The session instances  for each mode (solver, meshing and solver-icing) are stored in a dictionary which is reused accross all *gen.py files.

### DIFF
--- a/codegen/datamodelgen.py
+++ b/codegen/datamodelgen.py
@@ -4,6 +4,7 @@ from pathlib import Path
 import shutil
 from typing import Any, Dict
 
+from ansys.fluent.core import FluentMode, launch_fluent
 from ansys.fluent.core.session import BaseSession as Session
 from ansys.fluent.core.utils.fluent_version import (
     FluentVersion,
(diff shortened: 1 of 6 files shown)


prmukherj (approved):

hpohekar (approved):

codegen/datamodelgen.py
```diff
@@ -172,24 +174,32 @@ def _populate_static_info(self):
...
+            if FluentMode.MESHING_MODE not in self.sessions:
+                self.sessions[FluentMode.MESHING_MODE] = launch_fluent(
+                    mode=FluentMode.MESHING_MODE
+                )
+            session = self.sessions[FluentMode.MESHING_MODE]
             for _, info in self._static_info.items():
                 if "meshing" in info.modes:
                     info.static_info = self._get_static_info(info.rules, session)
-            session.exit()
+            session.exit()  # exiting the meshing session here as it won't be required during allapigen anymore
```
seanpearsonuk, line 187: @mkundu1 should we also remove this session from the sessions at this point?

PR #2531, opened 2024-03-04
test: Replace some fixtures with scoped fixture to reduce unittest runtime

mkundu1:
Fixtures with session scope are created only once during a unittest session. I have used them in tests which don't modify the Fluent session or read a case file immediately after using the fixture.

Some fixtures, which are used in 1 or 2 modules, are kept under module scope. We need to also limit the number of simulateous sessions in github runners (to reduce memory usage).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 from contextlib import nullcontext
 import functools
 import operator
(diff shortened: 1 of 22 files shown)


tests/conftest.py
```diff
@@ -119,3 +119,17 @@ def helpers(monkeypatch):
...
     if exitstatus == 5:
         session.exitstatus = 0
+
+
+# tests_by_fixture = defaultdict(list)
+
+
+# def pytest_collection_finish(session):
+#     for k, v in sorted(tests_by_fixture.items(), key=lambda t: len(t[1]), reverse=True):
+#         print(k, len(v))
```
mkundu1, line 129: some debug print, keeping for future use

PR #2546, opened 2024-03-07
refactor: Refactor the Launcher Utilities.

prmukherj:
Restructure the launcher_utils.py

### DIFF
--- a/tests/test_launcher.py
+++ b/tests/test_launcher.py
@@ -20,10 +20,10 @@ from ansys.fluent.core.launcher.launcher_utils import (
     _build_journal_argument,
     _is_windows,
     check_docker_support,
-    get_fluent_exe_path,
 )
 from ansys.fluent.core.launcher.process_launch_string import (
     _build_fluent_launch_args_string,
+    get_fluent_exe_path,
 )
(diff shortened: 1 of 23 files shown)


tests/test_launcher.py
```diff
@@ -8,24 +8,30 @@
...
+    is_windows,
+)
+from ansys.fluent.core.launcher.process_launch_string import (
+    _build_fluent_launch_args_string,
     get_fluent_exe_path,
 )
+from ansys.fluent.core.launcher.pyfluent_enums import (
+    FluentLinuxGraphicsDriver,
+    FluentMode,
+    FluentUI,
```
seanpearsonuk, line 31: **Aside** Just looking at this now, which we settled on by committee. It's non-obvious IMHO, and as type name, you'd think it actually represents a UI itself. Compare with FluentMode and LaunchMode which are explicitly modes and therefore much clearer. @mkundu1 @raph-luc

seanpearsonuk (approved):

tests/test_launcher.py
prmukherj: Sounds right @seanpearsonuk. Thank you.

mkundu1 (approved):

PR #2576, opened 2024-03-15
fix: Bug in arg setting for Enhanced Meshing Workflow.

prmukherj:
Arg settings were not reflecting to gui.

### DIFF
--- a/src/ansys/fluent/core/workflow.py
+++ b/src/ansys/fluent/core/workflow.py
@@ -713,7 +713,12 @@ class ArgumentWrapper(PyCallableStateObject):
         if attr in self.__dict__:
             self.__dict__[attr] = value
         else:
-            getattr(self, attr).set_state(value)
+            if self._dynamic_interface:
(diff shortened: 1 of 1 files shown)


src/ansys/fluent/core/workflow.py
```diff
@@ -713,7 +713,12 @@ def __setattr__(self, attr, value):
         if attr in self.__dict__:
             self.__dict__[attr] = value
         else:
-            getattr(self, attr).set_state(value)
+            if self._dynamic_interface:
+                camel_attr = snake_to_camel_case(
+                    str(attr), self._get_camel_case_arg_keys()
+                )
+                attr = camel_attr if camel_attr else attr
```
seanpearsonuk, line 720: is this any different to `attr = camel_attr or attr` (as per my preference)

seanpearsonuk (approved):

hpohekar (approved):

PR #2590, opened 2024-03-19
feat: Unstable settings warning

mkundu1:
PyFluent will show warning when an unstable API is used.

There is a Fluent side PR to add the attribute `webui-release-active?` in static info.


src/ansys/fluent/core/session_solver.py
```diff
@@ -60,7 +60,7 @@ def _import_settings_root(root):
         api_keys = root.child_names
 
     for root_item in api_keys:
-        _class_dict[root_item] = getattr(root, root_item)
```
mkundu1, line 63: to suppress the warning from internal call

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -991,24 +992,26 @@ def get_completer_info(self, prefix="", excluded=None) -> List[List[str]]:
...
-                        parents += parent
-            except AttributeError:
-                pass
-        if len(path_list):
-            print(f"\n {str(name)} can be accessed from the following paths: \n")
-            for path in path_list:
-                print(path)
-        if len(parents):
-            return f"\n {name} is a child of {parents} \n"
+        with warnings.catch_warnings(category=UnstableSettingWarning):
```
mkundu1, line 995: to suppress the warning from internal call

hpohekar (approved):

prmukherj (approved):

seanpearsonuk (approved):

PR #2594, opened 2024-03-20
fix: Fix EventSubscription when an event cannot be subscribed.

mkundu1:
This will fix the error seen from `EventScubscription.__del__` in recent code. I'm not able to write a unittest as pytest doesn't seem to detect the error (or the warning which is converted from the error).

Note: this error will not be present in PyConsole


hpohekar (approved):

prmukherj (approved):

prmukherj:
Thank you for this fix @mkundu1.

prmukherj:
I can confirm the issue no longer exists after this fix. Thank you.

seanpearsonuk (approved):

Issue #2607, opened 2024-03-22
examples.download_file not raising exceptions on failure


seanpearsonuk:
Troubleshooting tests that specify a non-existent file (e.g. directory not existing) where the failure is a low-level Fluent error. Problem is that download_file fails to raise.

PR #2608, opened 2024-03-22
fix: Raise an exception if remote file does not exist

hpohekar:
closes https://github.com/ansys/pyfluent/issues/2607

### DIFF
--- a/src/ansys/fluent/core/examples/downloads.py
+++ b/src/ansys/fluent/core/examples/downloads.py
@@ -15,6 +15,13 @@ import ansys.fluent.core as pyfluent
 logger = logging.getLogger("pyfluent.networking")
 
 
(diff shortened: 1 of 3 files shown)


src/ansys/fluent/core/examples/downloads.py
```diff
@@ -15,6 +15,13 @@
 logger = logging.getLogger("pyfluent.networking")
 
 
+class RemoteFileNotFoundError(FileNotFoundError):
+    """Raised on an attempt to download a non-existing remote file."""
```
seanpearsonuk, line 19:
```suggestion
    """Raised on an attempt to download a non-existent remote file."""
```

prmukherj (approved):

mkundu1 (approved):

PR #2623, opened 2024-03-26
fix: processor_count

mkundu1:


prmukherj (approved):

mkundu1:
Looks like `additional_arguments` is not supported for container launch. I can reproduce it locally. For now, I'll open a separate defect to address it after the release.

Issue #2643, opened 2024-04-02
utils/generic.py: confusing mix of functionality


seanpearsonuk:
- contains unrelated functionalities
- each group of functionality should be in a separate module with a relevant name
- `loop` is a shared object instantiated at module scope, only relevant to some functions
- `utils` namespace name is not helpful. Are all these things meant to be public?

hpohekar:
@seanpearsonuk 

Regarding `Are all these things meant to be public?` - We are not including these modules in rst. Therefore they are not available in public documentation. If we want, we can make them private explicitly.

hpohekar:
@seanpearsonuk

Regarding `utils namespace name is not helpful.` - Do we want to rename it to something like `utilities` ?

PR #2649, opened 2024-04-03
refactor: generic.py restructuring

hpohekar:
closes https://github.com/ansys/pyfluent/issues/2643

1. Removed `generic.py` as it's name was not helping to clarify it's usage.
2. `in_notebook` moved to `check_in_notebook.py`
3.  `loop` related functions moved to `event_loop.py`.
4. Singleton metaclasses moved to `singlton_meta.py`
5. `timing` moved to `timing_decorator.py`


hpohekar:
@seanpearsonuk @mkundu1 @prmukherj 

We can rename recently added modules. Please share if you have any suggestions. Thank you.

src/ansys/fluent/core/utils/singleton_meta.py
```diff
@@ -0,0 +1,20 @@
+"""Provides a module for Metaclasses."""
```
hpohekar, line 1: Shall we move this to `src/ansys/fluent/core/meta.py` ?

src/ansys/fluent/core/utils/timing_decorator.py
```diff
@@ -0,0 +1,16 @@
+"""Provides a module for timing function decorator."""
```
hpohekar, line 1: Shall we combine `event_loop.py` and `timing_decorator.py` into a new single `decorators.py` module ? because all of them are decorators.

src/ansys/fluent/core/utils/singleton_meta.py
prmukherj: Don't think so. meta.py has some very definite purpose and it should not contain generic methods.

src/ansys/fluent/core/utils/timing_decorator.py
seanpearsonuk: The code in `event_loop.py` is accessing the shared variable, loop. Therefore, only code that needs to access this variable should be in the same module.

hpohekar: Ok. We will keep them as separate files. Thank you.

src/ansys/fluent/core/utils/singleton_meta.py
seanpearsonuk: @hpohekar asks a very useful question. I believe that we should only either agree or propose an alternative. We can see by searching the code that all these classes are only used in the context of **external** tools for postprocessing, visualization and user interfaces. In the codebase currently, these classes are located amongst general-purpose PyFluent code in modules headed by docstrings like "this is generic code", which doesn't help to clarfiy the intent nor the current usage. It's also unlikely that any of this code will be reused directly in PyFluent itself. We already have the separate `post_objects` namespaces which is serving a similar purpose. We *could* rename that folder and put all the aforementioned code in there.

hpohekar: @seanpearsonuk Just now I discussed this point with @ajain-work. We can remove `singleton_meta.py` because we are not using it anywhere in PyFluent.

We can move  `src/ansys/fluent/core/meta.py` to `post_objects` because it is visualization specific only.

seanpearsonuk: @hpohekar class in `singleton_meta.py` are used in pyfluent-visualization.

hpohekar: @seanpearsonuk Right. I discussed this point with @ajain-work again. So he suggested that we can keep both `singleton_meta.py` and `meta.py` under `post_objects`. So we will also move separated modules under `post_objects` namespace. Do we want to rename `post-objects` ?

mkundu1 (approved):
Should we move those singular functions within a module to an __init__.py in the parent module? We won't loose anything in terms of namespaces if we do that.

prmukherj (approved):

src/ansys/fluent/core/post_objects/check_in_notebook.py
```diff
@@ -0,0 +1,13 @@
+"""Provides a module to check if application is running in notebook."""
+
+
+def in_notebook():
+    """Function to check if application is running in notebook."""
```
PipKat, line 5:
```suggestion
    """Checks if the application is running in notebook."""
```

src/ansys/fluent/core/post_objects/singleton_meta.py
```diff
@@ -0,0 +1,20 @@
+"""Provides a module for Metaclasses."""
+
+from abc import ABCMeta
+
+
+class SingletonMeta(type):
+    """Meta class for singleton type."""
```
PipKat, line 7:
```suggestion
    """Provides the metaclass for the singleton type."""
```
Google search shows it as one word, not capitalized.

src/ansys/fluent/core/post_objects/singleton_meta.py
```diff
@@ -0,0 +1,20 @@
+"""Provides a module for Metaclasses."""
```
PipKat, line 1:
```suggestion
"""Provides a module for metaclasses."""
```
Google search shows it as one word, not capitalized.

src/ansys/fluent/core/post_objects/singleton_meta.py
```diff
@@ -0,0 +1,20 @@
...
+    _single_instance = None
+
+    def __call__(cls, *args, **kwargs):
+        if not cls._single_instance:
+            cls._single_instance = super(SingletonMeta, cls).__call__(*args, **kwargs)
+        return cls._single_instance
+
+
+class AbstractSingletonMeta(ABCMeta, SingletonMeta):
+    """Meta class for abstract singleton type."""
```
PipKat, line 18:
```suggestion
    """Provides the metaclass for the abstract singleton type."""
```

src/ansys/fluent/core/post_objects/timing_decorator.py
```diff
@@ -0,0 +1,16 @@
+"""Provides a module for timing function decorator."""
```
PipKat, line 1:
```suggestion
"""Provides a module for the timing function decorator."""
```

src/ansys/fluent/core/post_objects/timing_decorator.py
```diff
@@ -0,0 +1,16 @@
...
+
+import time
+
+
+def timing(func):
+    """Timing function decorator."""
```
PipKat, line 7: Description should start with a verb ending in "s" or "es". Describe what the function does.

src/ansys/fluent/core/utils/event_loop.py
```diff
@@ -0,0 +1,26 @@
+"""Provides a module for event loop."""
```
PipKat, line 1:
```suggestion
"""Provides a module for the event loop."""
```

src/ansys/fluent/core/utils/event_loop.py
```diff
@@ -0,0 +1,26 @@
...
+
+import asyncio
+from functools import partial
+
+loop = asyncio.get_event_loop()
+
+
+def execute_in_event_loop_threadsafe(f):
+    """Decorator to execute function in an event loop from another thread."""
```
PipKat, line 10: Start all descriptions with verbs ending in "s" or "es"

src/ansys/fluent/core/post_objects/timing_decorator.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/utils/event_loop.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/utils/singleton_meta.py
seanpearsonuk: Renaming `post-objects` seems like a good option, as mentioned in an earlier post.

seanpearsonuk (approved):

PR #2660, opened 2024-04-04
fix: Add default for argument_names in commands.

prmukherj:
For commands with no arguments, argument_names attribute returns an empty list instead of throwing an attribute error.

### DIFF
--- a/tests/test_flobject.py
+++ b/tests/test_flobject.py
@@ -1166,5 +1166,5 @@ def test_default_argument_names_for_commands(load_static_mixer_settings_only):
 
     assert solver.results.graphics.contour.rename.argument_names == ["new", "old"]
     assert solver.results.graphics.contour.delete.argument_names == ["name_list"]
-    # The below is the default behaviour when no arguments are associated with the command.
+    # The following is the default behavior when no arguments are associated with the command.
     assert solver.results.graphics.contour.list.argument_names == []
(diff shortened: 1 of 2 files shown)


mkundu1 (approved):

tests/test_flobject.py
```diff
@@ -1146,3 +1146,25 @@ def test_static_info_hash_identity(new_solver_session):
...
+        "make_a_copy",
+        "display",
+        "copy",
+        "add_to_graphics",
+        "clear_history",
+    ]
+
+    assert solver.results.graphics.contour.rename.argument_names == ["new", "old"]
+    assert solver.results.graphics.contour.delete.argument_names == ["name_list"]
+    # The below is the default behaviour when no arguments are associated with the command.
```
PipKat, line 1169:
```suggestion
    # The following is the default behavior when no arguments are associated with the command.
```
Avoid location indicators like above and below. Use words like previous, earlier, later, following, and subsequent instead. Also use U.S. spellings.

seanpearsonuk:
Just waiting on some description

prmukherj:
> Just waiting on some description

Updated the description:  "For commands with no arguments, argument_names attribute returns an empty list instead of throwing an attribute error."

seanpearsonuk (approved):

PR #2662, opened 2024-04-05
fix: Camel case convertor.

prmukherj:


hpohekar (approved):

tests/test_new_meshing_workflow.py
```diff
@@ -1220,3 +1221,10 @@ def test_new_meshing_workflow_without_dm_caching(
     with pytest.raises(RuntimeError):
         fault_tolerant.import_cad_and_part_management.arguments()
     assert watertight.import_geometry.arguments()
+
```
seanpearsonuk, line 1224: We must add more test cases here.

prmukherj: Sure @seanpearsonuk. Will do. Could you please suggest some edge cases as well?
Thank you.

seanpearsonuk: - sequences of upper-case letters at different locations
- sequences of digits at different locations
- underscores at different locations
- starting with lower case
- special characters
- combinations of the above
- single character
- empty strings

prmukherj: @seanpearsonuk, could you please have a look now?

seanpearsonuk (approved):

PR #2670, opened 2024-04-09
fix: data model attribute updates

seanpearsonuk:
Data model event streaming code sends the object as callback argument, which was incorrectly interpreted as the value. The response is supposed to contain the value so a more complete solution would change that callback interface to pass the value. For now I am just erasing the cache item. 

An additional - previously ignored - issue concerns the accumulation of callbacks for a single attribute, which are never cleared. The object returned from `add_on_attribute_changed()` can be used to unregister. An object should do so upon destruction.  I am assuming that that is necessary. In terms of accumulation of callbacks for a live objects, we could track what is already registered.


src/ansys/fluent/core/services/datamodel_se.py
seanpearsonuk, line 1: Changes is this file merged from a separate branch of @mkundu1 referenced in the issue.

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -831,15 +852,25 @@ def _get_remote_attr(self, attrib: str) -> Any:
...
+            f"Entering _get_cached_attr for {self.path} and {attrib} ({self.cached_attrs})"
+        )
         cached_val = self.cached_attrs.get(attrib)
+        logger.debug(f"In _get_cached_attr, got value {cached_val} from cache")
         if cached_val is None:
+            logger.debug("In _get_cached_attr, get value from remote")
             cached_val = self._get_remote_attr(attrib)
+            logger.debug(f"In _get_cached_attr, got value {cached_val} from remote")
             try:  # will fail for Fluent 23.1 or before
+                print("in _get_cached_attr, register the callback")
```
prmukherj, line 865: This should be removed or moved to logger right?

mkundu1:
@seanpearsonuk Thanks for investgating the issue. I'm fixing this in #2673 by consuming the attribute value from the stream (we already have the complete server-side infrastructure).

I'll look into the lifetime issue of the subscription objects spearately.

seanpearsonuk:
@mkundu1 is doing the more architectural fix from the calling side - passing only the value and not the object

PR #2681, opened 2024-04-11
fix: workflow validation

seanpearsonuk:
Fix for some enhanced meshing workflow issues, one of which is reported on the Fluent side.

1. When an invalid task argument is set, an exception is not immediately triggered because the relevant server code does not do rules validation at the key-value level of dictionary parameters. The invalidity only becomes apparent at the command level. So, with this change, I ensure that when the task arguments are changed, the command arguments are also updated. That way the exception is triggered right away upon the relevant call. Also, the state is repaired within the same assignment implementation.

2. Code like `my_task.some_arg = 1` did not follow the same path through the code as `my_task.arguments.update_dict(dict(some_arg = 1))`, which is now rectified. I suspect that there could be similar issues for other methods at the same level.

### DIFF
--- a/src/ansys/fluent/core/workflow.py
+++ b/src/ansys/fluent/core/workflow.py
@@ -868,9 +868,6 @@ class CommandTask(BaseTask):
         return cmd
 
     def _command(self):
-        cmd = self._cmd
-        del cmd
-        self._cmd = None
         if not self._cmd:
             self._cmd = _new_command_for_task(self._task, self._source)
         return self._cmd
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/workflow.py
```diff
@@ -555,15 +559,13 @@ def set_state(self, args: dict) -> None:
...
         """
-        if self._dynamic_interface:
-            self.get_state()
-            camel_args = {}
-            for key, val in args.items():
-                camel_args[self._snake_to_camel_map[key]] = val
-            self._task.Arguments.set_state(camel_args)
-        else:
-            self._task.Arguments.set_state(args)
+        self._assign(args, "set_state")
```
seanpearsonuk, line 568: This is a bit ugly but the prime alternative was to use a lambda. I don't want to pass in the whole method directly to avoid triggering any code prematurely.

prmukherj (approved):

hpohekar (approved):

PR #2690, opened 2024-04-13
fix: Unsubscribe datamodel subscriptions on object deletion.

mkundu1:
There is a corresponding Fluent side PR fixing some issues in the datamodel event tags.

All corresponding datamodel subscription objects are unsubscribed while deleting the datamodel object. Subscription objects which are not of on-deleted type are unsubscribed just before the datamodel object is deleted. On-deleted type subscripton objects are unsubscribed through a final on-deleted callback after the datamodel object is deleted.

### DIFF
--- a/src/ansys/fluent/core/services/datamodel_se.py
+++ b/src/ansys/fluent/core/services/datamodel_se.py
@@ -4,8 +4,8 @@ from enum import Enum
 import functools
 import itertools
 import logging
-from threading import RLock
 import os
+from threading import RLock
 from typing import Any, Callable, Iterator, NoReturn, Optional, Sequence, Union
 
 from google.protobuf.json_format import MessageToDict, ParseDict
@@ -393,9 +393,9 @@ class SubscriptionList:
         Parameters
         ----------
         tag : str
-            Subscription tag
+            Subscription tag.
         subscription : EventSubscription
-            Subscription object
+            Subscription object.
         """
         with self._lock:
             self._subscriptions[tag] = subscription
@@ -406,7 +406,7 @@ class SubscriptionList:
         Parameters
         ----------
         tag : str
-            Subscription tag
+            Subscription tag.
         """
         with self._lock:
             self._subscriptions.pop(tag, None)
@@ -420,9 +420,9 @@ class SubscriptionList:
         Parameters
         ----------
         rules : str
-            Datamodel object rules
+            Datamodel object rules.
         path : str
-            Datamodel object path
+            Datamodel object path.
         deletion_stage : {"before", "after"}
             All subscription objects except those of on-deleted type are unsubscribed
             before the datamodel object is deleted. On-deleted subscription objects are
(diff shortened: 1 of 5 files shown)


src/ansys/fluent/core/streaming_services/datamodel_event_streaming.py
```diff
@@ -53,30 +53,30 @@ def _process_streaming(self, id, stream_begin_method, started_evt, *args, **kwar
...
-                                )
-                            ):
-                                cb[1](cb[0])
-                            elif response.HasField("commandExecutedEventResponse"):
-                                command = response.commandExecutedEventResponse.command
-                                args = _convert_variant_to_value(
-                                    response.commandExecutedEventResponse.args
-                                )
-                                cb[1](cb[0], command, args)
+                    cb = self._cbs.get(response.tag, None)
```
mkundu1, line 56: As tags are unique, we don't need to loop through the callback dictionary. Also, some callbacks can unsubscribe and delete other callbacks during which we cannot loop through the dictionary.

tests/test_datamodel_service.py
```diff
@@ -487,3 +493,267 @@ def test_read_ony_set_state(new_mesh_session):
...
+                    pass
+
+            class X(PyTextual):
+                pass
+
+            class C(PyCommand):
+                pass
+
+
+def _create_datamodel_root(session, rules_str) -> PyMenu:
```
mkundu1, line 546: This will create a test datamodel in server.

hpohekar (approved):

prmukherj (approved):

seanpearsonuk (approved):

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -369,11 +372,79 @@ def unsubscribe(self) -> None:
...
+
+    def add(self, tag: str, subscription: EventSubscription) -> None:
+        """Add a subscription object.
+
+        Parameters
+        ----------
+        tag : str
+            Subscription tag
+        subscription : EventSubscription
+            Subscription object
```
PipKat, line 397:
```suggestion
            Subscription tag.
        subscription : EventSubscription
            Subscription object.
```

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -369,11 +372,79 @@ def unsubscribe(self) -> None:
...
+        with self._lock:
+            self._subscriptions[tag] = subscription
+
+    def remove(self, tag: str) -> None:
+        """Remove a subscription object.
+
+        Parameters
+        ----------
+        tag : str
+            Subscription tag
```
PipKat, line 408:
```suggestion
            Subscription tag.
```

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -369,11 +372,79 @@ def unsubscribe(self) -> None:
...
+    ) -> None:
+        """Unsubscribe corresponding subscription objects while the datamodel object is
+        being deleted.
+
+        Parameters
+        ----------
+        rules : str
+            Datamodel object rules
+        path : str
+            Datamodel object path
```
PipKat, line 424:
```suggestion
            Datamodel object rules.
        path : str
            Datamodel object path.
```

PR #2695, opened 2024-04-15
ci: Reduce the number of Fluent version to run for each test

mkundu1:
Following are the 3 categories of unittests:

1. **Tests which do not require to run Fluent at all or can be run with any stable Fluent version:**  Such tests aren't marked with the `fluent_version` marker and will be run only with the latest released Fluent version during PRs and with the latest released and the current development Fluent versions during nightly. If these tests get broken due to a recent Fluent API change, that will be caught during nightly.
2.  **Tests which depend on the current development Fluent version or are tracking Fluent API change:** Such tests are marked as `fluent_version("develop")` and will be run only with the current development Fluent version.
3. **Tests which depend on a specific released Fluent version:** Such tests are marked as `fluent_version("==<version>")` and will be run only with that specific Fluent version.

After this PR is merged, we only need to add the `fluent_version` marker for tests which depends on the current development Fluent version or a specific released Fluent version. We should make sure a test is not marked to run with more than one Fluent version unless there is very specific scenario.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -48,9 +48,9 @@ def pytest_runtest_setup(item):
     if fluent_version_markers:
         spec = fluent_version_markers[0].args[0]
         # Tests which depend on the current development Fluent version or tracking
-        # Fluent API change are marked as fluent_version("latest"). They are run only
+        # Fluent API change are marked as fluent_version("develop"). They are run only
         # with the development Fluent version.
-        if spec == "latest":
+        if spec == "develop":
             fluent_version_spec = f"=={_fluent_develop_version}"
         # Tests which depend on a specific released Fluent version are marked as
         # fluent_version("==<version>"). They are run only with that specific Fluent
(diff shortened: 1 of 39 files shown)


prmukherj (approved):

seanpearsonuk (approved):

mkundu1:
@prmukherj There is a Fluent version check in the `CompundTask.add_child_and_update` method in workflow.py. Let me know if we need to add specific Fluent version marker in some tests to cover that code.

raph-luc (approved):

hpohekar (approved):

prmukherj:
> @prmukherj There is a Fluent version check in the `CompundTask.add_child_and_update` method in workflow.py. Let me know if we need to add specific Fluent version marker in some tests to cover that code.

Not required @mkundu1, that is to just provide a warning to users using that argument for earlier versions. We don't need to test it for any specific versions.

dnwillia-work (approved):

mkundu1:
Superseded by #2708

Issue #2693, opened 2024-04-15
Launcher code refactoring


seanpearsonuk:
This is a follow-up to #2611. 

1. The current structure of the code is:
```
def launch_fluent(...):
    <argument processing>
    launcher = create_launcher()
    return launcher()
```

So that each launcher class is directly usable, we should clear all _<argument processing>_ out of _launch_fluent()_ and instead have it done within the scope of each launcher instantiation wherever necessary. This would be done via shared code.

2. Identify and eliminate any conditions like `if mode == launch_mode_xyz` (except for the point in the code where the type of launcher is decided!). The aim is to handle any such conditions not through boolean logic but through polymorphism, which has been facilitated by the introduction of the polymorphic launcher classes. Some of those conditionals were still recently present in the code, but I don't know if they were all eliminated in the last round of changes.  Please check by searching for uses of the relevant enum.

PR #2702, opened 2024-04-17
refactor: launcher code refactoring

hpohekar:
closes https://github.com/ansys/pyfluent/issues/2693

1. Added `_get_graphics_driver`, `_get_ui_mode` and `_validate_gpu`.
2. Added `show_gui` to Standalone constructor signature.


src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -302,37 +294,12 @@ def launch_fluent(
...
-    )
     fluent_launch_mode = _get_fluent_launch_mode(
         start_container=start_container,
         container_dict=container_dict,
         scheduler_options=scheduler_options,
     )
     del start_container
-    mode = _get_mode(mode)
     argvals = locals().copy()
     _process_invalid_args(dry_run, fluent_launch_mode, argvals)
```
hpohekar, line 304: We have already removed invalid args for `PIM` and `DockerContainer` launcher from constructor signature therefore `_process_invalid_args` is not required in these 2 launchers.

The invalid args are supported for `Standalone` launcher only therefore `if fluent_launch_mode != enums.LaunchMode.STANDALONE:` is not valid inside `Standalone` launcher.

The first condition check `if dry_run and fluent_launch_mode != enums.LaunchMode.CONTAINER:` is only possible in `launch_fluent` because `dry_run` is not supported for `Standalone` and `PIM` launcher.

Therefore I think `_process_invalid_args` should be as it is here.

mkundu1: Is it possbile to replace `_process_invalid_args` with a decorated `__init__` function in the individual launcher class? Some usage like this:
```
class PIMLauncher:
    @warn_unknown_keywards
    def __init__(...):  # this won't have `env` in the keyward list
        ...
        
launcher = PIMLauncher(..., env=...)
# returns a PIM launcher but will print a warning: Unknown keyword arguments {'env'} are not processed.
```
Make sure the `__init__` docstring is not changed by wrapping a decorator.

hpohekar: We will try this. Thank you.

src/ansys/fluent/core/launcher/pyfluent_enums.py
```diff
@@ -233,3 +249,35 @@ def _get_standalone_launch_fluent_version(
...
+
+def _get_ui_mode(
+    show_gui: Optional[bool] = None,
+):
+    """Get the graphics driver.
+
+    Parameters
+    ----------
+    show_gui: bool
+        Whether to show Fluent GUI.
```
PipKat, line 262:
```suggestion
        Whether to show the Fluent GUI.
```

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -123,6 +128,11 @@ def __init__(
             default is ``True``. You can stop and start the streaming of the
             Fluent transcript subsequently via the method calls, ``transcript.start()``
             and ``transcript.stop()`` on the session object.
+        show_gui : bool, optional
+            Whether to display the Fluent GUI. The default is ``None``, which does not
+            cause the GUI to be shown. If a value of ``False`` is
+            not explicitly provided, the GUI will also be shown if
```
PipKat, line 134:
```suggestion
            not explicitly provided, the GUI is shown if
```

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -123,6 +128,11 @@ def __init__(
...
+            the environment variable ``PYFLUENT_SHOW_SERVER_GUI`` is set to 1.
```
PipKat, line 135:
```suggestion
        show_gui : bool, optional
            Whether to display the Fluent GUI. The default is ``None``,
            in which case the GUI is not shown. If ``False`` is
            not explicitly provided, the GUI will be shown if
            the ``PYFLUENT_SHOW_SERVER_GUI`` environment
            variable is set to 1.
```

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -123,6 +128,11 @@ def __init__(
             default is ``True``. You can stop and start the streaming of the
             Fluent transcript subsequently via the method calls, ``transcript.start()``
             and ``transcript.stop()`` on the session object.
```
PipKat, line 130:
```suggestion
            Fluent transcript subsequently with calls to the ``transcript.start()``
            and ``transcript.stop()`` methods on the session object.
```
Google dev stye guide recommends not using "via".

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -123,6 +128,11 @@ def __init__(
...
+        show_gui : bool, optional
+            Whether to display the Fluent GUI. The default is ``None``, which does not
+            cause the GUI to be shown. If a value of ``False`` is
+            not explicitly provided, the GUI will also be shown if
+            the environment variable ``PYFLUENT_SHOW_SERVER_GUI`` is set to 1.
         case_file_name : str, optional
             If provided, the case file at ``case_file_name`` is read into the Fluent session.
```
PipKat, line 137:
```suggestion
            Name of the case file to read into the Fluent session.
```

src/ansys/fluent/core/launcher/pyfluent_enums.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/launcher/standalone_launcher.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/launcher/standalone_launcher.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/launcher/standalone_launcher.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/launcher/launcher.py
hpohekar: We have to introduce **kwargs in `def __init__(...):` then only it is possible to use `PIMLauncher(..., env=...)`, but we have already removed kwargs from all constructors and we don't want to use it.

hpohekar: We have found another solution for this. We have moved `_process_invalid_args` to `create_launcher` where all required arguments are available i.e. `dry_run`, `fluent_launch_mode`, `kwargs` for `_process_invalid_args`. 

Standalone - 

image

Container 

image

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -123,8 +128,15 @@ def __init__(
             default is ``True``. You can stop and start the streaming of the
             Fluent transcript subsequently via the method calls, ``transcript.start()``
             and ``transcript.stop()`` on the session object.
+        show_gui : bool, optional
+            Whether to display the Fluent GUI. The default is ``None``,
+            in which case the GUI is not shown. If ``False`` is
+            not explicitly provided, the GUI will be shown if
```
PipKat, line 134:
```suggestion
            not explicitly provided, the GUI is shown if
```
Use present tense verb.

hpohekar: Done. Thank you.

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -172,6 +184,12 @@ def __init__(
         The allocated machines and core counts are queried from the scheduler environment and
         passed to Fluent.
         """
+        if version == "2d" and gpu:
+            raise GPUSolverSupportError()
```
seanpearsonuk, line 188: @hpohekar I'll just take the above two lines as an example but this will apply generally to other code. We want to do this check not just for this launcher but for other launchers, too. And I see that you are already doing that, so that's good. But we shouldn't duplicate the code to do that in each launcher. We need to factor things out effectively.

hpohekar: @seanpearsonuk I thought our requirement is to keep all launchers independent and therefore minimum code should be inside a common function like `create_launcher` therefore we have kept `_process_invalid_args` only there. So now to remove all duplication we have moved it to `create_launcher`

seanpearsonuk: If some subset (>1) of launchers all agree on a generalised validation for a given combination of argument values then each of those launchers should call a common function, e.g., 
```
def validate_gpu(gpu, version):
  if version == "2d" and gpu:
          raise GPUSolverSupportError()
```
Once we have written each launcher code in terms of a sequence of such high-level validation calls, we can look at what we have and see how it should be refactored further - there are options.

hpohekar: So we want to add following function to pyfluent_enums.py and then we will call it in `create_launcher` right ?

```
def _validate_gpu(gpu, version):
  if version == "2d" and gpu:
          raise GPUSolverSupportError()
```

seanpearsonuk: My thinking is that it shall be possible to use each launcher independently which means that each one is in control of its argument validation.

hpohekar: Ok. We will call it inside of each launcher. Thank you.

src/ansys/fluent/core/launcher/pyfluent_enums.py
```diff
@@ -233,3 +249,35 @@ def _get_standalone_launch_fluent_version(
...
+
+    Returns
+    -------
+    ui_mode: UIMode
+        Fluent GUI mode.
+    """
+    ui_mode = None
+    if show_gui is not None:
+        warnings.warn(
+            "'show_gui' is deprecated, use 'ui_mode' instead",
```
PipKat, line 272:
```suggestion
            "'show_gui' is deprecated. Use 'ui_mode' instead.",
```
Punctuation should be used in message strings.

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -123,8 +125,15 @@ def __init__(
...
+            in which case the GUI is not shown. If ``False`` is
+            not explicitly provided, the GUI is shown if
+            the ``PYFLUENT_SHOW_SERVER_GUI`` environment
+            variable is set to 1.
         case_file_name : str, optional
-            If provided, the case file at ``case_file_name`` is read into the Fluent session.
+            Name of the case file to read into the
+            Fluent session. The default is ``None``.
         case_data_file_name : str, optional
             If provided, the case and data files at ``case_data_file_name`` are read into the Fluent session.
```
PipKat, line 138:
```suggestion
            Name of the case data file. If names of both a case file and case data file are provided, they are read into the Fluent session.
```

src/ansys/fluent/core/launcher/standalone_launcher.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/launcher/pyfluent_enums.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/launcher/standalone_launcher.py
hpohekar: Done. Thank you.

prmukherj (approved):

seanpearsonuk (approved):

PR #2707, opened 2024-04-17
fix: deleted kwargs - not for merge

mkundu1:
to test a PyConsole fix


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1462,6 +1462,13 @@ class Map(SettingsBase[DictStateType]):
     """A ``Map`` object representing key-value settings."""
```
PipKat, line 1462:
```suggestion
    """Provides the ``Map`` object representing key-value settings."""
```

PR #2708, opened 2024-04-18
ci: Add CI to run tests on old Fluent versions weekly

mkundu1:
The following CIs will be run regularly for unittests:
1. With develop version of Fluent - nightly
2. With the latest release version of Fluent - nightly
3. With older versions of Fluent - weekly
4. Without codegen - weekly
5. solvermode tests with develop version of Fluent - weekly


.github/workflows/compare-flobject.yml
mkundu1, line 1: Removing as we don't need to sync flobject.py after packaging PyFluent with Fluent

hpohekar (approved):

prmukherj (approved):

.github/workflows/test-run-old-versions-weekly.yml
raph-luc, line 1: Should we use a test `matrix` for the old versions run so that each version runs in a separate step and also parallel? Similar to what we do in `.github\workflows\ci.yml`. I think this would make it easier to keep track of failures in different versions, and also speed up the execution of the entire suite of tests.

In my understanding this was an issue we were having previously: the nightly test would fail because some specific version failed, and that would "obscure" failures in other versions that were sometimes more important because we were running all of these one after another in a single step, rather than parallel in separate steps.

raph-luc (approved):

.github/workflows/test-run-old-versions-weekly.yml
mkundu1: Yes, we should do that, thanks. The nightly tests are running on self-hosted runnes, but those will free during the weekends.

PR #2709, opened 2024-04-18
fix: Remote file transfer strategy refactoring

hpohekar:
closes https://github.com/ansys/pyfluent/issues/2710 https://github.com/ansys/pyfluent/issues/2711 https://github.com/ansys/pyfluent/issues/2721

1. Remote FTS was running in the background even after calling `session.exit()` or `session.force_exit()` in the wsl. Now it is deleted after calling `session.exit()` or `session.force_exit()`
2. Nightly Dev Test Fix - Runners are subject to change for each CI therefore this scenario is handled now.
3. More configuration options are provided for remote file transfer startegy.


prmukherj (approved):

src/ansys/fluent/core/utils/file_transfer_service.py
```diff
@@ -110,24 +110,29 @@ def download(
...
+            else [file_name]
+        )
+    elif isinstance(file_name, pathlib.PurePath):
+        files = (
+            [str(os.path.join(path, file_name.name))]
+            if os.path.isfile(os.path.join(path, file_name.name))
+            else [str(file_name)]
+        )
+    elif isinstance(file_name, list):
+        if os.path.exists(path):
```
raph-luc: What is going on here? Seems unrelated to the changes described in the PR. So `path` was hardcoded to work only in GitHub runners previously?

Also, `pathlib.Path` should be able to handle `str` paths with no issues in my understanding, do we need two different `if` clauses for `str` and `Path`?

Lastly, usually we don't need to convert a `pathlib.Path` to `str` as in line 126, in my understanding most Python tools should work with `pathlib.Path` out of the box (including all built-in functions like `open`).

raph-luc (commented):
Do we need these being printed to the user console? 
image

I believe these should only show up in debug logging calls

hpohekar:
> Do we need these being printed to the user console? image
> 
> I believe these should only show up in debug logging calls

Point 1 - related to `[2024-04-18 08:12:49.360737] [0x0000000000de8500] [info]    File transfer server started.`

We had a user requirement to get information about file upload/download progress therefore we are running this gRPC FTS container without `detach` mode.

Point 2 - related to 

```
>>> solver.exit()
>>> 67bfe7fb1fe2
```

This is docker's default behavior. Docker return ID of container when we use `docker kill <id>`

src/ansys/fluent/core/session.py
```diff
@@ -326,11 +326,19 @@ def exit(self, **kwargs) -> None:
...
+            pass
 
     def force_exit(self) -> None:
         """Immediately terminates the Fluent session, losing unsaved progress and
         data."""
         self._fluent_connection.force_exit()
+        try:
+            self._file_transfer_service.exit()
+        except AttributeError:
+            pass
```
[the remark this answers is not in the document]
hpohekar: Use is allowed to plug in own FTS. If user has not implemented method to exit respected FTS then we will get `Attribute` error. To handle this scenario we have done this.

raph-luc:
> This is docker's default behavior. Docker return ID of container when we use `docker kill <id>`

The [Docker API](https://docker-py.readthedocs.io/en/stable/) provides a container stop method that doesn't do this, can we use that or does the current implementation not allow this?

hpohekar:
> > This is docker's default behavior. Docker return ID of container when we use `docker kill <id>`
> 
> The [Docker API](https://docker-py.readthedocs.io/en/stable/) provides a container stop method that doesn't do this, can we use that or does the current implementation not allow this?

@raph-luc 

I have tried `docker stop` but gRPC FTS is not terminated, it is still running. It terminates only using `docker kill`

image

image

Thanks for the suggestion.

hpohekar:
> > > This is docker's default behavior. Docker return ID of container when we use `docker kill <id>`
> > 
> > 
> > The [Docker API](https://docker-py.readthedocs.io/en/stable/) provides a container stop method that doesn't do this, can we use that or does the current implementation not allow this?
> 
> @raph-luc
> 
> I have tried `docker stop` but gRPC FTS is not terminated, it is still running. It terminates only using `docker kill`
> 
> image
> 
> image
> 
> Thanks for the suggestion.

@raph-luc Ok. It is terminated now.

`docker stop` takes some time where as `docker kill` terminates it quickly. Ultimately both uses SIGKILL behind the scene. Thank you.

src/ansys/fluent/core/utils/file_transfer_service.py
```diff
@@ -257,7 +257,7 @@ def exit(self):
         ]
         for container_id in active_container_ids:
             try:
-                subprocess.Popen(f"docker kill {container_id}", shell=True)
+                subprocess.Popen(f"docker stop {container_id}", shell=True)
             except Exception:
                 pass
```
raph-luc, line 262: This entire `exit` method definition can be replaced with Python Docker API calls (starting from line 247), which is going to be much cleaner and less error prone than parsing shell output like this. I think this PR may be a good opportunity to review this `exit` method (or in a separate PR when time permits)

src/ansys/fluent/core/utils/file_transfer_service.py
hpohekar: There was a bug in locating files in GitHub (container mode) as the path was hard-coded. We have resolved this now on the basis of Fluent mode. If the Fluent is running is container mode then it will check `host_mount_path` as `path` to locate files if that path is already exists else it will use given file name directly.

src/ansys/fluent/core/session.py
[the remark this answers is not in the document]
raph-luc: Presumably if users set up their own FTS, then they would have to provide an `exit` method as well, right? Otherwise they will constantly have zombie services running in their machine. 

Doesn't that mean that we don't want to hide an `AttributeError` here, so that the user is aware of this issue?

src/ansys/fluent/core/utils/file_transfer_service.py
hpohekar: @raph-luc 

Yes, I verified that we don't need to convert `pathlib.Path` to `str`. We can use it directly. We will remove `str` for `pathlib.Path`.  Thank you.

```
>>> file_path = pathlib.Path("D:/Repos/pyfluent/bug_solver.cas.h5")                            
>>> file_path                                                                                  
WindowsPath('D:/Repos/pyfluent/bug_solver.cas.h5')
>>> solver = launch_fluent()
>>> solver.file.read_case(file_name=file_path)
pyfluent.settings_api WARNING: Mismatch between generated file and server object info. Dynamically created settings classes will be used.
Fast-loading "D:\ANSYSDev\vNNN\fluent\fluent24.2\\addons\afd\lib\hdfio.bin"
Done.

Reading from AAPQ13pUvWxQtxN:"D:\Repos\pyfluent\bug_solver.cas.h5" in NODE0 mode ...
  Reading mesh ...
       66417 nodes,     1 node zone  ...
  Done.


Building...
     mesh
     materials,
     interface,
     domains,
        mixture
     zones,
        wall-elbow
        wall-inlet
        outlet
        cold-inlet
        hot-inlet
        symmetry-xyplane
        interior--elbow-fluid
        elbow-fluid
     parallel,
Done.
>>> 
```

src/ansys/fluent/core/session.py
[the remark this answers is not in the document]
hpohekar: Earlier I had kept it as `self._file_transfer_service.exit()` only without try block. We can raise an exception with more information in `except` block instead of `pass` only.

src/ansys/fluent/core/utils/file_transfer_service.py

src/ansys/fluent/core/session.py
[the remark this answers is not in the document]
hpohekar: they would have to provide an exit method as well, right? - yes.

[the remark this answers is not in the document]
hpohekar: We have removed try block and user will get `AttributeError` now.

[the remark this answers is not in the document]
hpohekar: We will update `exit` method. Thank you.

src/ansys/fluent/core/session.py
```diff
@@ -326,11 +326,19 @@ def exit(self, **kwargs) -> None:
         """Exit session."""
         logger.debug("session.exit() called")
         self._fluent_connection.exit(**kwargs)
+        try:
+            self._file_transfer_service.exit()
+        except AttributeError:
+            pass
```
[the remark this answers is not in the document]
mkundu1: Yes, this should be in the finalizer, otherwise the issue will still remain when the session object is garbage collected. @hpohekar

hpohekar:
> Do we need these being printed to the user console? image
> 
> I believe these should only show up in debug logging calls

@raph-luc This is not appearing as we are using Docker SDK now. Thank you.

hpohekar:
> > This is docker's default behavior. Docker return ID of container when we use `docker kill <id>`
> 
> The [Docker API](https://docker-py.readthedocs.io/en/stable/) provides a container stop method that doesn't do this, can we use that or does the current implementation not allow this?

@raph-luc Done. Thank you.

src/ansys/fluent/core/utils/file_transfer_service.py
hpohekar: @raph-luc It's update now. We are using just `self.server.kill()` with help of Docker SDK. Thank you.

src/ansys/fluent/core/session.py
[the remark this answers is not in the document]
hpohekar: Yes, this is the remaining thing only. We will implement this soon. Thank you.

src/ansys/fluent/core/utils/file_transfer_service.py
```diff
@@ -139,26 +141,46 @@ class RemoteFileTransferStrategy(FileTransferStrategy):
...
+                    }
+                },
             )
         except Exception:
             self.host_port = random.randint(6000, 7000)
-            self.server = subprocess.Popen(
-                f"docker run -p {self.host_port}:50000 -v {self.host_mount_path}:{self.container_mount_path} ghcr.io/ansys/tools-filetransfer:latest",
-                shell=True,
+            self.ports = ports if ports else {"50000/tcp": self.host_port}
+            self.server = self.docker_client.containers.run(
```
raph-luc, line 174: Minor: I think `container` would be a better variable name
```suggestion
            self.container = self.docker_client.containers.run(
```

src/ansys/fluent/core/utils/file_transfer_service.py
```diff
@@ -240,20 +262,7 @@ def download(
...
-            id_port.split(" ")[0]
-            for id_port in id_ports
-            if str(self.host_port) in id_port
-        ]
-        for container_id in active_container_ids:
-            try:
-                subprocess.Popen(f"docker kill {container_id}", shell=True)
-            except Exception:
-                pass
+        self.server.kill()
```
raph-luc, line 265: I believe we should use `stop`  instead, as it already has a built-in configurable timeout of 10 seconds, after which it will send a kill signal (see [docs](https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.Container.stop)).

This together with the other change I suggested above would mean:
```suggestion
        self.container.stop()
```

src/ansys/fluent/core/utils/file_transfer_service.py
```diff
@@ -110,25 +110,27 @@ def download(
...
-        elif isinstance(file_name, list):
-            files = [str(file) for file in file_name]
+    if isinstance(file_name, str):
+        file_path_check = os.path.join(path, os.path.basename(file_name))
+        files = (
+            [str(file_path_check)] if os.path.isfile(file_path_check) else [file_name]
+        )
+    elif isinstance(file_name, pathlib.PurePath):
+        file_path_check = os.path.join(path, file_name.name)
+        files = [file_path_check] if os.path.isfile(file_path_check) else [file_name]
```
raph-luc, line 123: As mentioned earlier, I think we can still put these two `if` clauses together (`str` and  `Path`), by for example preparing the path with `file_name = Path(file_name)`

mkundu1:
Not sure if the permission issue in self-hosted runs
```
PermissionError: [Errno 13] Permission denied: '/home/ansys/.local/share/ansys_fluent_core/examples/serverinfo-ggcxfjqg.txt'
```
started coming from this PR. Probably the permission of EXAMPLES_PATH directory is somehow changed.

src/ansys/fluent/core/session.py
```diff
@@ -326,11 +326,13 @@ def exit(self, **kwargs) -> None:
         """Exit session."""
         logger.debug("session.exit() called")
         self._fluent_connection.exit(**kwargs)
+        self._file_transfer_service.exit()
 
     def force_exit(self) -> None:
         """Immediately terminates the Fluent session, losing unsaved progress and
```
PipKat, line 332:
```suggestion
        """Immediately terminate the Fluent session, losing unsaved progress and
```

src/ansys/fluent/core/utils/file_transfer_service.py

src/ansys/fluent/core/utils/file_transfer_service.py

src/ansys/fluent/core/utils/file_transfer_service.py

src/ansys/fluent/core/session.py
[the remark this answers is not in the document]
hpohekar: Done. Thank you.

mkundu1:
As I have tested, the file-transfer container doesn't support a graceful stop. I passed a large value to the `-t` parameter of `docker stop` and it is waiting for that many seconds. So, we can replace `docker stop` with `docker kill`, because `docker stop` will anyway do a forced stop (equivalent to `docker kill`) after a default grace period. This will reduce the time to exit() and will probably fix the failing test.

src/ansys/fluent/core/utils/file_transfer_service.py
```diff
@@ -150,37 +151,40 @@ def __init__(
         ----------
         image_name: str
             Name of the image.
+        image_tag : str, optional
+            Tag of the image.
         port: int, optional
-            Port for file transfer service to use.
+            Port for the file transfer service to use.
         container_mount_path: Union[str, Path], optional
             Path inside the container where host mount path will be mounted to.
```
PipKat, line 154:
```suggestion
            Path inside the container to mount the host mount path to.
```

src/ansys/fluent/core/utils/file_transfer_service.py
```diff
@@ -150,37 +151,40 @@ def __init__(
...
         host_mount_path: Union[str, Path], optional
             Existing path in the host operating system that will be available inside the container.
```
PipKat, line 156:
```suggestion
            Existing path in the host operating system to be available inside the container.
```

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -333,6 +335,8 @@ def __init__(
             The corresponding remote instance when Fluent is launched through
             PyPIM. This instance will be deleted when calling
             ``Session.exit()``.
+        file_transfer_service : optional
+            File transfer service. Uploads/downloads files to/from the server.
```
PipKat, line 339:
```suggestion
            File transfer service for uploading files to and
            downloading files from the server.
```
Avoid using slashes and meaning is often ambiguous.

src/ansys/fluent/core/session.py
```diff
@@ -328,7 +332,7 @@ def exit(self, **kwargs) -> None:
         self._fluent_connection.exit(**kwargs)
 
     def force_exit(self) -> None:
-        """Immediately terminates the Fluent session, losing unsaved progress and
+        """Immediately terminate the Fluent session, losing unsaved progress and
```
PipKat, line 335:
```suggestion
        """Immediately exit the Fluent session, losing unsaved progress and
```
I'd prefer the verb be exit to match the verb in the method name.

src/ansys/fluent/core/utils/file_transfer_service.py
```diff
@@ -134,26 +134,55 @@ class RemoteFileTransferStrategy(FileTransferStrategy):
...
     def __init__(
         self,
+        image_name: Optional[str] = None,
+        image_tag: Optional[str] = None,
+        port: Optional[int] = None,
         container_mount_path: Optional[str] = None,
         host_mount_path: Optional[str] = None,
     ):
-        self.host_port = random.randint(5000, 6000)
+        """Provides gRPC-based remote file transfer strategy.
```
PipKat, line 143:
```suggestion
        """Provides the gRPC-based remote file transfer strategy.
```

src/ansys/fluent/core/session.py
raph-luc: @PipKat there is a difference in meaning here, `exit` is a "graceful" exit, while `terminate` is an "externally forced" exit. Do you think something like the suggestion below would be better?
```suggestion
        """Forces the Fluent session to exit, losing unsaved progress and
```
Thanks for the help!

src/ansys/fluent/core/utils/file_transfer_service.py
raph-luc: @PipKat I am not sure how, it is literally "mounting" the "host mount path" inside the container. Maybe:
```suggestion
            Path inside the container for the host mount path.
```
Leaving the fact that it will be mounted implied (should be enough I think).

raph-luc:
@hpohekar @PipKat note that some of these doc suggestions are for docstrings copied from `src\ansys\fluent\core\launcher\fluent_container.py`. The final docstring updates should therefore be made in both locations (as applicable). Perhaps we can do that in a separate PR to avoid bundling unrelated changes in this PR.

PipKat:
> @hpohekar @PipKat note that some of these doc suggestions are for docstrings copied from `src\ansys\fluent\core\launcher\fluent_container.py`. The final docstring updates should therefore be made in both locations (as applicable). Perhaps we can do that in a separate PR to avoid bundling unrelated changes in this PR.

I'm fine with that. I assume that the changes only need to made in `src\ansys\fluent\core\launcher\fluent_container.py` and will then somehow eventually get copied to here? Is this something that you need or want me to do--or will you or the PyFluent team take care of it?

raph-luc:
@PipKat we will take care of it, what I meant is the changes to the docstrings being made here should later be applied to the location where these docstrings were copied from (`fluent_container.py`).

src/ansys/fluent/core/utils/file_transfer_service.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/session.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/fluent_connection.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/utils/file_transfer_service.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/utils/file_transfer_service.py
hpohekar: Done. Thank you.

hpohekar:
> As I have tested, the file-transfer container doesn't support a graceful stop. I passed a large value to the `-t` parameter of `docker stop` and it is waiting for that many seconds. So, we can replace `docker stop` with `docker kill`, because `docker stop` will anyway do a forced stop (equivalent to `docker kill`) after a default grace period. This will reduce the time to exit() and will probably fix the failing test.

Yes, right. All tests are passing after using 'docker kill`. Thanks a lot for having look into it.

hpohekar:
> Not sure if the permission issue in self-hosted runs
> 
> ```
> PermissionError: [Errno 13] Permission denied: '/home/ansys/.local/share/ansys_fluent_core/examples/serverinfo-ggcxfjqg.txt'
> ```
> 
> started coming from this PR. Looks like the permission of EXAMPLES_PATH directory is somehow changed:
> 
> ```
> ansys@pyfluent-ci-3:~$ ls -ltr /home/ansys/.local/share
> total 20
> drwxr-xr-x 3 ansys ansys 4096 Feb  9  2023 pyvista
> drwxr-xr-x 2 ansys ansys 4096 Jul 31  2023 ansys_tools_path
> drwxr-xr-x 3 ansys ansys 4096 Sep 22  2023 ansys_mapdl_core
> drwxr-xr-x 4 ansys ansys 4096 Mar 29 10:09 virtualenv
> drwxr-xr-x 4 ansys ansys 4096 Apr 18 19:19 ansys_fluent_core
> ansys@pyfluent-ci-3:~$ ls -ltr /home/ansys/.local/share/ansys_fluent_core/
> total 44
> drwxr-xr-x 4 ansys ansys 36864 Jun 13  2023 examples.bkp
> drwxr-xr-x 2 root  root   4096 Apr 18 19:19 examples
> ```
> 
> I have cleaned up the examples path within the runners, will keep checking what is going on.

@mkundu1 

This is happened again.

image

I have cleared `examples` folder and changed it's read-write permission then this issue is resolved.

No idea why this is happening to this PR particularly.

seanpearsonuk (approved):

PR #2728, opened 2024-04-22
fix: Mode switching in Enhanced meshing workflow.

prmukherj:
Switching workflows was causing issues. That's resolved.

### DIFF
--- a/tests/test_new_meshing_workflow.py
+++ b/tests/test_new_meshing_workflow.py
@@ -1483,11 +1483,8 @@ def test_independent_meshing_sessions(new_mesh_session, new_mesh_session_1):
     watertight = meshing_1.watertight()
     assert watertight.import_geometry.arguments()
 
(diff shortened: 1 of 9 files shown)


tests/test_new_meshing_workflow.py
```diff
@@ -1466,3 +1466,40 @@ def test_created_workflow(new_mesh_session):
...
+    watertight = meshing_1.watertight()
+    assert watertight.import_geometry.arguments()
+
+    ft = meshing_1.fault_tolerant()
+    assert ft.import_cad_and_part_management.arguments()
+
+    watertight = meshing_1.watertight()
+    assert watertight.import_geometry.arguments()
+
+    with pytest.raises(RuntimeError):
```
prmukherj, line 1486: @mkundu1, after the data-model state caching fix this should not raise the Runtime Error.

PR #2731, opened 2024-04-23
refactor: Update Enhanced Meshing workflow to cache data faster.

prmukherj:
Earlier initializing the pre-defined workflows like "watertight" was taking approximately 50 seconds. Now this timing has been reduced to take only approximately 4 seconds. This is by reducing number of server calls and using cached data.

### DIFF
--- a/src/ansys/fluent/core/utils/dictionary_operations.py
+++ b/src/ansys/fluent/core/utils/dictionary_operations.py
@@ -1,6 +1,30 @@
 """Performs some operations on Python dictionaries."""
 
+from typing import Any
 
-def get_first_dict_key_for_value(input_dict, value):
-    """Get the first dictionary key that matches a value."""
-    return next((key for key, val in input_dict.items() if val == value), None)
+
+def get_first_dict_key_for_value(input_dict: dict, value: Any):
+    """Get the first dictionary key that matches a value. Typical
+    usage is where the value is known to be unique in the input
+    dictionary.
+
+    Parameters
+    ----------
+    input_dict : dict
+    value : Any
+
+    Returns
+    -------
+    Any
+        The key associated with the first match.
+
+    Raises
+    ------
+    ValueError
+        If the value is absent from the dictionary.
+    """
+
+    try:
+        return next((key for key, val in input_dict.items() if val == value))
+    except StopIteration:
+        raise ValueError()
--- a/src/ansys/fluent/core/utils/dictionary_operations.py
+++ b/src/ansys/fluent/core/utils/dictionary_operations.py
@@ -1,0 +1,6 @@
+"""Performs some operations on python dictionaries."""
+
+
+def get_first_dict_key_for_value(input_dict, value):
+    """Get the first dictionary key that matches a value."""
+    return next(key for key, val in input_dict.items() if val == value)
--- a/src/ansys/fluent/core/utils/dictionary_operations.py
+++ b/src/ansys/fluent/core/utils/dictionary_operations.py
@@ -4,9 +4,8 @@ from typing import Any
 
 
 def get_first_dict_key_for_value(input_dict: dict, value: Any):
-    """Get the first dictionary key that matches a value. Typical
-    usage is where the value is known to be unique in the input
-    dictionary.
+    """Get the first dictionary key that matches a value. Typical usage is where the
+    value is known to be unique in the input dictionary.
 
     Parameters
     ----------
--- a/src/ansys/fluent/core/utils/dictionary_operations.py
+++ b/src/ansys/fluent/core/utils/dictionary_operations.py
@@ -3,4 +3,4 @@
 
 def get_first_dict_key_for_value(input_dict, value):
     """Get the first dictionary key that matches a value."""
-    return next(key for key, val in input_dict.items() if val == value)
+    return next((key for key, val in input_dict.items() if val == value), None)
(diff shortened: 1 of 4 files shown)


src/ansys/fluent/core/workflow.py
```diff
@@ -358,29 +358,28 @@ def python_name(self) -> str:
...
+                    this_command.get_attr("helpString")
                 )
+                self._cache_data(this_command)
             except Exception:
                 pass
         return self._python_name
 
+    def _cache_data(self, command):
+        _disp_text = command.get_attr("displayText")
+        if self._python_name in self._command_source._help_string_display_text_map:
```
seanpearsonuk, line 372: It looks like this map should be named according to `python_name` keys rather than `help_string` keys, because that's how it's being keyed here. E.g., `_python_name_to_display_text`.

prmukherj: Will do this once the behavior is fixed. Thank you.

src/ansys/fluent/core/workflow.py
```diff
@@ -356,37 +356,43 @@ def python_name(self) -> str:
...
-                        self._python_name
-                    ] = this_command.get_attr("displayText")
-                except Exception:
-                    pass
+            if self._command_source._dynamic_python_names:
+                display_name_map = self._command_source._python_name_display_text_map
+                if self.display_name() not in display_name_map.values():
+                    self._set_python_name()
+                else:
+                    self._python_name = list(display_name_map.keys())[
```
seanpearsonuk, line 364:
```suggestion
                    self._python_name = list(display_name_map)[
```

seanpearsonuk: We might add a utility function and use it here

```
>>> def get_first_dict_key_for_value(d, value):
...   return next(key for key, val in d.items() if val == value)
...
>>> get_first_dict_key_for_value({1:2, 3:4, 4:4}, 4)
3
```

prmukherj: @seanpearsonuk, for now shall we keep this locally and later move it to utils? Or, else we have to create a separate file in utils with this. Though I think the later to be a better idea as this can be generically used later as well. Please let me know your thoughts. Thank you.

src/ansys/fluent/core/utils/dictionary_operations.py
```diff
@@ -0,0 +1,6 @@
+"""Performs some operations on python dictionaries."""
+
+
+def get_first_dict_key_for_value(input_dict, value):
+    """Get the first dictionary key that matches a value."""
+    return next(key for key, val in input_dict.items() if val == value)
```
mkundu1, line 6: `next` takes a second argument for defualt. We can return None when the value is not found, then we can skip checking for conatinment in the caller code.

src/ansys/fluent/core/workflow.py
```diff
@@ -356,37 +357,43 @@ def python_name(self) -> str:
...
+
         return self._python_name
 
+    def _set_python_name(self):
+        try:
+            this_command = self._command()
+            # temp reuse helpString
+            self._python_name = camel_to_snake_case(this_command.get_attr("helpString"))
+            self._cache_data(this_command)
+        except Exception:
```
mkundu1, line 379: What kind of execption does the above code throw? Catching generic exceptions should be avoided if possible.

src/ansys/fluent/core/workflow.py
```diff
@@ -356,37 +357,43 @@ def python_name(self) -> str:
...
+            pass
+
+    def _cache_data(self, command):
+        _disp_text = command.get_attr("displayText")
```
mkundu1, line 383: Minor: generally we don't prefix `_` to local variable names

src/ansys/fluent/core/workflow.py
prmukherj: Ideally we should not remove omit the exception at this level. Now our entire enhanced meshing workflow is based on the "helpString" from the fdl file which forms the Python name for our workflow. If some issue is there in the extraction of the "helpString" we should throw that. What do you think?
@mkundu1, @seanpearsonuk

src/ansys/fluent/core/workflow.py
prmukherj: Done, Thank you.

src/ansys/fluent/core/utils/dictionary_operations.py
```diff
@@ -0,0 +1,6 @@
+"""Performs some operations on python dictionaries."""
```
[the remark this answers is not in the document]
prmukherj: Done, Thank you.

src/ansys/fluent/core/utils/dictionary_operations.py
prmukherj: Done, Thank you.

seanpearsonuk (approved):

src/ansys/fluent/core/utils/dictionary_operations.py
```diff
@@ -0,0 +1,29 @@
...
+
+    Parameters
+    ----------
+    input_dict : dict
+    value : Any
+
+    Returns
+    -------
+    Any
+        The key associated with the first match.
```
PipKat, line 18:
```suggestion
        Key associated with the first match.
```

mkundu1 (approved):

hpohekar (approved):

PR #2732, opened 2024-04-23
ci: pre-commit commit message

mkundu1:
This will fix the commit name issue in pre-commit PRs ([example](https://github.com/ansys/pyfluent/pull/2667)). I could not find a way to fix the branch name issue.

Also reducing the frequency of pre-commit update to monthly.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,11 @@
+ci:
+  autofix_commit_msg: |
+      ci: auto fixes from pre-commit.com hooks
+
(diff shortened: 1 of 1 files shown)


mkundu1:
> Regarding the branch name issue, should we ask [PyAnsys team](https://github.com/ansys/actions/issues) to add an exception for `pre-commit-ci-update-config` or allow us to configure exceptions for the branch names?

This should be already supported in the latest version of ansys/actions. Updated the latest version in this PR.

Issue #2735, opened 2024-04-24
Avoid file modification in case of gRPC remote FTS


hpohekar:
gRPC remote FTS modifies files after upload if that file is already existed.

PR #2736, opened 2024-04-24
fix: avoid gRPC FTS file modification

hpohekar:
closes https://github.com/ansys/pyfluent/issues/2735

Do not upload file if it's already exists on remote.

### DIFF
--- a/src/ansys/fluent/core/utils/file_transfer_service.py
+++ b/src/ansys/fluent/core/utils/file_transfer_service.py
@@ -234,12 +234,10 @@ class RemoteFileTransferStrategy(FileTransferStrategy):
         files = _get_files(file_name, self.host_mount_path)
         if self.client:
             for file in files:
-                is_file_exists_on_remote = self.file_exists_on_remote(
-                    os.path.basename(file)
-                )
-                if is_file_exists_on_remote:
+                is_file_on_remote = self.file_exists_on_remote(os.path.basename(file))
+                if is_file_on_remote:
                     print(f"\n{file} already uploaded.\n")
-                elif os.path.isfile(file) and not is_file_exists_on_remote:
+                elif os.path.isfile(file) and not is_file_on_remote:
                     self.client.upload_file(
                         local_filename=file,
                         remote_filename=(
--- a/src/ansys/fluent/core/utils/file_transfer_service.py
+++ b/src/ansys/fluent/core/utils/file_transfer_service.py
@@ -6,6 +6,7 @@ import pathlib
 import random
 import shutil
 from typing import Any, Callable, List, Optional, Protocol, Union  # noqa: F401
+import warnings
 
 from alive_progress import alive_bar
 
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/utils/file_transfer_service.py
```diff
@@ -229,7 +234,12 @@ def upload(
         files = _get_files(file_name, self.host_mount_path)
         if self.client:
             for file in files:
-                if os.path.isfile(file):
+                is_file_exists_on_remote = self.file_exists_on_remote(
+                    os.path.basename(file)
+                )
+                if is_file_exists_on_remote:
+                    print(f"\n{file} already uploaded.\n")
```
seanpearsonuk, line 241: Shall we use the `warnings` module? 

A more accurate message would say that a file with the *same name* *exists at the remote location*.   We don't know if it's the same file or if it were uploaded.

src/ansys/fluent/core/utils/file_transfer_service.py
```diff
@@ -229,7 +234,12 @@ def upload(
         files = _get_files(file_name, self.host_mount_path)
         if self.client:
             for file in files:
-                if os.path.isfile(file):
+                is_file_exists_on_remote = self.file_exists_on_remote(
```
seanpearsonuk, line 237: Either leave off the `is_` part or rename it `does_file_exist_on_remote` (which is better grammatically but too long and clumsy) or just `is_file_on_remote`.

hpohekar: Right, `is_file_on_remote` is correct and looks good. Done. Thank you.

src/ansys/fluent/core/utils/file_transfer_service.py
```diff
@@ -229,7 +234,12 @@ def upload(
...
+                    os.path.basename(file)
+                )
+                if is_file_exists_on_remote:
+                    print(f"\n{file} already uploaded.\n")
+                elif os.path.isfile(file) and not is_file_exists_on_remote:
```
seanpearsonuk: According to my understanding, this aligns the gRPC file transfer service behaviour with the PIM based behaviour that's (currently) implemented in a separate class. From that point of view, we can merge this now, and then consider what the overall behaviour should be and to what extent it should be configurable by the user.

src/ansys/fluent/core/utils/file_transfer_service.py
hpohekar: In whole source code we are not using warning module to check the existing files. `(examples.download(), PIMFileTransferService,)` Therefore we are not using it here. We have updated the message now. Thank you.

src/ansys/fluent/core/utils/file_transfer_service.py
hpohekar: Yes, sure.

src/ansys/fluent/core/utils/file_transfer_service.py
seanpearsonuk: I'm not sure about using print. One thing about using `warnings.warn()` is that allows specific message types to be disabled.

hpohekar: Right, I agree. We will update this now. Thank you.

seanpearsonuk: Overall, as this is a library, we must be careful about writing to stdout.

hpohekar: Yes, I agree.

mkundu1:
@hpohekar I think, in general, we still want to upload the file when it exists in the server as the client side file can be newer. A better check here may be not to upload the file if the local and remote absolute filepaths are identical. Nevertheless, feel free to merge the current PR to fix the immediate issues in the regressions, we can consider this in a later PR. CC: @seanpearsonuk

mkundu1 (approved):

prmukherj (approved):

PR #2737, opened 2024-04-25
test: test dict op

seanpearsonuk:
Add testing for and adjust new function.

### DIFF
--- a/src/ansys/fluent/core/utils/dictionary_operations.py
+++ b/src/ansys/fluent/core/utils/dictionary_operations.py
@@ -1,6 +1,30 @@
 """Performs some operations on Python dictionaries."""
 
+from typing import Any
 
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/utils/dictionary_operations.py
```diff
@@ -1,6 +1,30 @@
...
+        The key associated with the first match.
+
+    Raises
+    ------
+    ValueError
+        If the value is absent from the dictionary.
+    """
+
+    try:
+        return next((key for key, val in input_dict.items() if val == value))
```
seanpearsonuk, line 28: I removed the default-to-`None` strategy, which doesn't work because `None` is a valid key.

prmukherj (approved):

PR #2751, opened 2024-04-30
fix: rpvar fix for bool etc

seanpearsonuk:
Correct the argument checks in rp-var service.

### DIFF
--- a/tests/test_rp_vars.py
+++ b/tests/test_rp_vars.py
@@ -60,3 +60,19 @@ def test_rp_vars_allowed_values(new_solver_session_no_transcript) -> None:
         rp_vars("number-of-iterat")
 
     assert "number-of-iterations" in rp_vars.allowed_values()
+
+
+@pytest.mark.fluent_version(">=23.2")
+def test_rp_vars_boolean(new_solver_session_no_transcript) -> None:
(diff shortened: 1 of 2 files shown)


mkundu1 (approved):

tests/test_rp_vars.py
```diff
@@ -60,3 +60,19 @@ def test_rp_vars_allowed_values(new_solver_session_no_transcript) -> None:
...
+    var_name = "rp-lam?"
+    rp_vars = solver.rp_vars
+
+    var_val = rp_vars(var_name)
+    assert isinstance(var_val, bool)
+
+    for i in range(10):
+        var_val = not var_val
+        rp_vars(var_name, var_val)
+        assert rp_vars(var_name) == var_val
```
raph-luc, line 78: Curious: why do this 10 times, shouldn't 2 times be enough? Was it failing only when doing more than 2 `not` operations?

raph-luc (approved):

tests/test_rp_vars.py
seanpearsonuk: 2 is in principle enough for the known issue. It could have been written that way also, but I decided to add some iterations.

PR #2752, opened 2024-04-30
refactor: Exposing custom workflow exception

### LINKED ISSUE #2750: Minor: exception defined inside internal method
Quick question:
https://github.com/ansys/pyfluent/blob/1e1416aa5c9ee706159b58d152eab45a59cee44c/src/ansys/fluent/core/workflow.py#L66-L73

Isn't this definition for a custom `Exception` nested inside a non-public method kind of useless, in the sense that users won't easily be able to import and catch this `NewCommandError`? In which case this 5 lines definition could be replaced with just `Exception(f"Could not create command for task {task_name}")` as it would be caught using `except Exception:` regardless?

Otherwise I think this custom exception definition should be moved outside of the non-public method, so that it can more easily be imported if an user wants to catch it.

raph-luc:
Closes #2750

Renamed from `NewCommandError` to the more detailed but verbose `CommandInstanceCreationError`
Can now be imported as `from ansys.fluent.core.workflow import CommandInstanceCreationError` if a client wants to handle it
Slightly reworded the docstring as well

### DIFF
--- a/src/ansys/fluent/core/workflow.py
+++ b/src/ansys/fluent/core/workflow.py
@@ -18,11 +18,11 @@ from ansys.fluent.core.utils.dictionary_operations import get_first_dict_key_for
 from ansys.fluent.core.utils.fluent_version import FluentVersion
 
 
-class CommandInstanceCreationError(Exception):
+class CommandInstanceCreationError(RuntimeError):
     """Raised when an attempt to create an instance of a task command fails."""
 
     def __init__(self, task_name):
-        super().__init__(f"Could not create command instance for task {task_name}")
+        super().__init__(f"Could not create command instance for task {task_name}.")
 
 
 def camel_to_snake_case(camel_case_str: str) -> str:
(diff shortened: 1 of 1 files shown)


src/ansys/fluent/core/workflow.py
```diff
@@ -18,6 +18,13 @@
 from ansys.fluent.core.utils.fluent_version import FluentVersion
 
 
+class CommandInstanceCreationError(Exception):
```
mkundu1, line 21: Is it better to derive the custom exception from more specific exception class, `RuntimeError` in this case? If user wants to handle execptions in a general manner, they'll need to catch generic `Exception` in this case which is not a good practice.

mkundu1 (approved):

src/ansys/fluent/core/workflow.py
raph-luc: I agree that in this case inheriting from `RuntimeError` would work fine and be potentially more useful in some use cases, I am going to include that as part of the changes here. Thanks for the suggestion.

prmukherj (approved):

seanpearsonuk (approved):

PR #2762, opened 2024-05-01
fix: expose meshing attrs at session level

prmukherj:
Contains datamodel_se related set_state() updates.

### DIFF
--- a/src/ansys/fluent/core/session_pure_meshing.py
+++ b/src/ansys/fluent/core/session_pure_meshing.py
@@ -93,10 +93,6 @@ class PureMeshing(BaseSession):
                 )
                 self._fluent_connection.register_finalizer_cb(stream.stop)
 
-        for child_name in self._base_meshing.meshing().keys():
-            child = getattr(self._base_meshing.meshing, child_name)
-            self.__dict__[child_name] = child
-
     @property
     def tui(self):
         """Instance of ``main_menu`` on which Fluent's SolverTUI methods can be
--- a/src/ansys/fluent/core/session_pure_meshing.py
+++ b/src/ansys/fluent/core/session_pure_meshing.py
@@ -93,6 +93,10 @@ class PureMeshing(BaseSession):
                 )
                 self._fluent_connection.register_finalizer_cb(stream.stop)
 
+        for child_name in self._base_meshing.meshing().keys():
+            child = getattr(self._base_meshing.meshing, child_name)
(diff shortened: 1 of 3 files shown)


mkundu1 (approved):

seanpearsonuk (approved):

hpohekar (approved):

src/ansys/fluent/core/session_pure_meshing.py
```diff
@@ -93,6 +93,10 @@ def __init__(
                 )
                 self._fluent_connection.register_finalizer_cb(stream.stop)
 
+        for child_name in self._base_meshing.meshing().keys():
+            child = getattr(self._base_meshing.meshing, child_name)
+            self.__dict__[child_name] = child
```
seanpearsonuk, line 98: @prmukherj I have a few points to make here. We might need a team meeting to go through all this properly.
1. If we are only doing a **copy** on **initialisation** then we are not guaranteed to reference the full set of the child `meshing` attribute's relevant children at an arbitrary later time, as we would need to do. I believe that this is a live issue that you can demonstrate through additional testing.
2. We are _blindly_ copying all of the child `meshing` attribute's children into the parent `meshing` object. Are we sure that we are not overwriting anything inappropriately? Even if we are satisfied that that is not happening now, can we be confident that it won't happen in future, which could introduce subtle bugs, easily missed by testing?
3. A separate issue from the above one is that we might be copying additional, irrelevant attributes.
4. Do we need the parent `meshing` object to duplicate its child `meshing` attribute's interface? What we certainly need is compatibility for the sake of journaling but beyond that, extension of this interface goes against the deliberate design decision that we made in PyFluent.

tests/test_meshing_workflow.py
```diff
@@ -576,3 +576,19 @@ def test_new_2d_meshing_workflow(new_mesh_session):
     # Switch to solution mode
     solver = meshing.switch_to_solver()
     assert solver
+
+
+@pytest.mark.fluent_version(">=23.2")
+def test_high_level_objects(new_mesh_session):
```
seanpearsonuk, line 582: This test is fine as far as it goes but it only tests the existence of a singleton object at initialisation time. As pointed out in my other comment, there is more to test here.

prmukherj: @seanpearsonuk, sure. We can have a meeting once @mkundu1 is available today?

PR #2766, opened 2024-05-01
fix: datamodel set_state

prmukherj:


src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -886,9 +886,15 @@ def set_state(self, state: Optional[Any] = None, **kwargs) -> None:
...
         Raises
         ------
+        TypeError
+            If no arguments are provided to set_state.
         ReadOnlyObjectError
             If the object is read-only.
         """
+        if not state and not kwargs:
+            raise TypeError(
+                "set_state() requires either 'state' or 'kwargs' arguments."
```
seanpearsonuk, line 896: Do we know the use case for kwargs?

prmukherj: Yes @seanpearsonuk. Updated the test.
"meshing.GlobalSettings.set_state(EnableOversetMeshing=False)"

mkundu1 (approved):

src/ansys/fluent/core/services/datamodel_se.py
seanpearsonuk: Do flobject-based APIs support it, too?

mkundu1: Yes, https://github.com/ansys/pyfluent/blob/main/tests/test_flobject.py#L443

seanpearsonuk (approved):

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -886,9 +886,15 @@ def set_state(self, state: Optional[Any] = None, **kwargs) -> None:
 
         Raises
         ------
+        TypeError
+            If no arguments are provided to set_state.
```
PipKat, line 890:
```suggestion
            If no arguments are provided to the ``set_state()`` method.
```

src/ansys/fluent/core/services/datamodel_se.py
PipKat, line 896:
```suggestion
                "The `set_state()` method requires either 'state' or 'kwargs' arguments."
```

prmukherj: Thank you @mkundu1 .

PR #2767, opened 2024-05-01
feat: Reorganize codegen and generated API files

mkundu1:
1. The codegen module is moved to src/ansys/fluent/core so that it is packaged. A helper script is added at codegen/allapigen.py for users.
2. All API files will be generated under a single directory, src/ansys/fluent/core/generated for PyFluent. This is to make packaging of the API files in Fluent side easier. I think this is okay as users won't directly import these files. The generated API directory is controlled through a module level variable which will be overwritten in Fluent side.

This PR contains the high level changes of the source and packaging structure. The goal is to package the codegen module and use it in Fluent to generate the API files at some Fluent packaging path. There will be additional PRs to refactor the codegen code to use Fluent's local static info. We shall also create a zipped module for settings API files to include in Fluent package (not for PyFluent ATM as the expanded directory is useful as the .pyi files are read by intellisense).


.github/workflows/ci.yml
```diff
@@ -145,16 +145,16 @@ jobs:
         id: cache-api-code
         with:
           path: |
-            src/ansys/fluent/core/datamodel_231
-            src/ansys/fluent/core/fluent_version_231.py
-            src/ansys/fluent/core/meshing/tui_231.py
-            src/ansys/fluent/core/solver/settings_231
-            src/ansys/fluent/core/solver/tui_231.py
+            src/ansys/fluent/core/generated/datamodel_231
```
seanpearsonuk, line 148: Can we factor out these duplicated paths?

mkundu1: Currently we are maintaining separate caches for each Fluent versions which are updated at different frequency, so not changing that ATM.

seanpearsonuk (approved):

src/ansys/fluent/core/__init__.py
```diff
@@ -83,7 +84,11 @@ def version_info() -> str:
 # Whether stream and cache commands state
 DATAMODEL_USE_NOCOMMANDS_DIFF_STATE = True
 
+# Parent directory where codegen writes out API files
+GENERATED_API_DIR = (Path(__file__) / ".." / "generated").resolve()
```
raph-luc, line 88: Rather than inside a `"generated"` directory, thoughts about a name that is more specific such as `"API_files"` or something similar? 

`ansys/fluent/core/API_files` vs `ansys/fluent/core/generated`

raph-luc (approved):

src/ansys/fluent/core/__init__.py
```diff
@@ -83,7 +84,11 @@ def version_info() -> str:
...
 
+
+# TODO: cleanup the following and related code
```
PipKat, line 91:
```suggestion
# TODO: clean up the following and related code
```
Clean up is two words when a verb and one word as a noun.

src/ansys/fluent/core/codegen/datamodelgen.py
```diff
@@ -112,48 +109,44 @@ def __init__(
 
 
 class DataModelGenerator:
-    def __init__(self, version, pyfluent_path, sessions: dict):
+    """Datamodel API class generator."""
```
PipKat, line 112:
```suggestion
    """Provides the datamodel API class generator."""
```

src/ansys/fluent/core/codegen/tuigen.py
```diff
@@ -40,28 +39,23 @@
...
-        / f"tui_{version}.py"
-    ).resolve()
+def _get_tui_filepath(mode: str, version: str):
+    return (GENERATED_API_DIR / mode / f"tui_{version}.py").resolve()
 
 
 _INDENT_STEP = 4
 
 
+# TODO: Move doc specific variables to docgen
```
PipKat, line 52:
```suggestion
# TODO: Move doc-specific variables to docgen
```

hpohekar (approved):

prmukherj (approved):

src/ansys/fluent/core/__init__.py
seanpearsonuk: I think the facts that they are generated and that they are API files are both salient pieces of information. I'm fine with the current name, but I would also have nothing against `generated_api`.

mkundu1: For now, I won't change the name, will revisit the idea in the next refactoring.

PR #2743, opened 2024-04-29
refactor: gRPC remote FTS refactoring

hpohekar:
closes https://github.com/ansys/pyfluent/issues/2744

1. Added fix for Nightly Dev Test 24.2
2. Changed `host_mount_path`.
3. Resolved file modification issue.

```
>>> import ansys.fluent.core as pyfluent
>>> from ansys.fluent.core import examples
>>> case_file_name = examples.download_file("mixing_elbow.cas.h5", "pyfluent/mixing_elbow", return_without_path=False)
/mnt/c/Users/hpohekar/pyfluent/src/ansys/fluent/core/examples/downloads.py:73: UserWarning:
File already exists. File path:
/home/hpohekar/.local/share/ansys_fluent_core/examples/mixing_elbow.cas.h5

  warnings.warn(
>>> case_file_name
'/home/hpohekar/.local/share/ansys_fluent_core/examples/mixing_elbow.cas.h5'
>>> import os
>>> os.path.getsize(case_file_name)
3062081                                                                                <<<<<<<<---------------------
>>> solver = pyfluent.launch_fluent()
>>> solver.file.read_case(file_name=case_file_name)
Fast-loading "/ansys_inc/v242/fluent/fluent24.2.0/addons/afd/lib/hdfio.bin"
Done.

Reading from 83baa1ed3968:"/mnt/pyfluent/mixing_elbow.cas.h5" in NODE0 mode ...
  Reading mesh ...
       17822 cells,     1 cell zone  ...
          17822 mixed cells,  zone id: 87
       91581 faces,     7 face zones ...
           2168 polygonal wall faces,  zone id: 34
            268 polygonal wall faces,  zone id: 33
            155 polygonal pressure-outlet faces,  zone id: 32
            152 polygonal velocity-inlet faces,  zone id: 31
             55 polygonal velocity-inlet faces,  zone id: 30
           2001 polygonal symmetry faces,  zone id: 29
          86782 mixed interior faces,  zone id: 89
       66417 nodes,     1 node zone  ...
Warning: reading 2 partition grid onto 1 compute node machine.
         Combining every 2 partitions.
  Done.


Building...
     mesh
     materials,
     interface,
     domains,
        mixture
     zones,
        wall-elbow
        wall-inlet
        outlet
        cold-inlet
        hot-inlet
        symmetry-xyplane
        interior--elbow-fluid
        elbow-fluid
     parallel,
Done.
>>> os.path.getsize(case_file_name)
[...]

PR #2775, opened 2024-05-02
fix: configure remote grpc file transfer service

hpohekar:
closes https://github.com/ansys/pyfluent/issues/2782

Set `pyfluent.USE_FILE_TRANSFER_SERVICE=True` to use file transfer service.


src/ansys/fluent/core/__init__.py
```diff
@@ -84,6 +84,9 @@ def version_info() -> str:
 # Whether stream and cache commands state
 DATAMODEL_USE_NOCOMMANDS_DIFF_STATE = True
 
+# Whether to use remote gRPC file transfer service
+REMOTE_GRPC_FILE_TRANSFER_SERVICE = None
```
mkundu1, line 88: It is simpler if a boolean config is either True or False.

mkundu1:
I think we should keep the source code conditional based on the configuration without changing the tests.

hpohekar:
> I think we should keep the source code conditional based on the configuration without changing the tests. After fixing the underlying issue, we shall be able to run the same tests after switching the configuration.

@mkundu1 

The issue is that if we set `pyfluent.REMOTE_GRPC_FILE_TRANSFER_SERVICE=True` then 

`examples.download_file("Static_Mixer_Parameters.cas.h5", "pyfluent/static_mixer",)` returns `"Static_Mixer_Parameters.cas.h5"` only, which is incorrect because we are changing `host_mount_path` to `pyfluent.USER_DATA_PATH` to avoid file modification issue.

We need full path therefore we are updating the code as follows

```
if pyfluent.REMOTE_GRPC_FILE_TRANSFER_SERVICE:
    file_name = examples.download_file("Static_Mixer_Parameters.cas.h5", "pyfluent/static_mixer",, return_without_path=False)
else:
    file_name = examples.download_file("Static_Mixer_Parameters.cas.h5", "pyfluent/static_mixer",)
```

hpohekar:
> I think we should keep the source code conditional based on the configuration without changing the tests. After fixing the underlying issue, we shall be able to run the same tests after switching the configuration.

@mkundu1 

In following PR also we are changing `host_mount_path` and updating tests accordingly. (we are holding it for now due to hang issue)

https://github.com/ansys/pyfluent/pull/2743

hpohekar:
Ideally all tests should use full path but we had done this adjustment because we have set `pyfluent.EXAMPLES_PATH` as a `host_mount_path` before using remote gRPC FTS.

mkundu1:
> > I think we should keep the source code conditional based on the configuration without changing the tests. After fixing the underlying issue, we shall be able to run the same tests after switching the configuration.
> 
> @mkundu1
> 
> In following PR also we are changing `host_mount_path` and updating tests accordingly. (we are holding it for now due to hang issue)
> 
> #2743

Does it

src/ansys/fluent/core/__init__.py
hpohekar: Done. Thank you.

prmukherj (approved):

seanpearsonuk:
> I think we should keep the source code conditional based on the configuration without changing the tests. After fixing the underlying issue, we shall be able to run the same tests after switching the configuration.

Yes, I was going to make the same comment, thanks @mkundu1. It's a simple case of encapsulation.

And, @hpohekar, irrespective of this point, please look at the code you have written in all these tests, simplified as follows. Can you see the issue?
```
if x:
    y = foo(a, b, False)
else:
    y = foo(a, b, True)
```

seanpearsonuk (approved):

src/ansys/fluent/core/__init__.py
hpohekar: Done. Thank you.

hpohekar:
@mkundu1 @seanpearsonuk 

Closing due to consistent permission errors.

We have disabled remote gRPC FTS in [this PR.](https://github.com/ansys/pyfluent/pull/2780). 

Nightly Dev Test is not hanging now. https://github.com/ansys/pyfluent/actions/runs/8949142087/job/24583086511

hpohekar:
https://github.com/ansys/pyfluent/pull/2783

PR #2777, opened 2024-05-03
fix: settings allowed values check

hpohekar:
closes https://github.com/ansys/pyfluent/issues/2778

1. Updated attribute check in settings exception block.

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -1072,7 +1072,7 @@ class Group(SettingsBase[DictStateType]):
                 if allowed and value not in allowed:
                     raise allowed_values_error(name, value, allowed) from ex
             else:
-                raise ex
+                raise
 
 
 class WildcardPath(Group):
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1067,9 +1067,10 @@ def __setattr__(self, name: str, value):
...
         except Exception as ex:
-            allowed = attr.allowed_values()
-            if allowed and value not in allowed:
-                raise allowed_values_error(name, value, allowed) from ex
+            if hasattr(attr, "allowed_values"):
+                allowed = attr.allowed_values()
+                if allowed and value not in allowed:
+                    raise allowed_values_error(name, value, allowed) from ex
             else:
                 raise ex
```
h-krishnan, line 1075: Wouldn't a plain `raise` be sufficient here?

h-krishnan (approved):

src/ansys/fluent/core/solver/flobject.py
hpohekar: @h-krishnan Yes, but we think that it looks safe.

seanpearsonuk (approved):

seanpearsonuk:
@hpohekar please add a test. If you get stuck, please check with @h-krishnan. Please ensure your test fails with the original code. The test is important as we would be patching the product with this.

seanpearsonuk (approved):

prmukherj (approved):

mkundu1 (approved):

PR #2794, opened 2024-05-08
fix: datamodel doc

mkundu1:


prmukherj (approved):

hpohekar (approved):

mkundu1:
@hpohekar I think some of the path-related code in documentation generation can be simpler, if we pass the version string from the top level.

seanpearsonuk (approved):

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["poetry-core>=1.9.0"]
+requires = ["poetry-core>=1.0.0"]
```
mkundu1, line 2: Potential fix for `Unable to find installation candidates for rpds-py (0.18.1)` - recurring issue in dev docs CI

PR #2796, opened 2024-05-09
refactor: Remove 'show_gui' from launch_fluent arguments.

prmukherj:
Remove the ''show_gui'' argument and put a mechanism for migrating older scripts.

### DIFF
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -133,7 +133,7 @@ def deprecate(old_arg, new_arg, converter):
         """Holds the original method to perform operations on it."""
 
         def wrapper(*args, **kwargs):
-            """Performs the deprecation operation on the arguments of the original method."""
+            """Performs deprecation operation on the arguments of the original method."""
             if old_arg in kwargs:
                 converter(old_arg, new_arg, kwargs)
                 kwargs.pop(old_arg)
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -7,7 +7,6 @@ with gRPC.
 import logging
 import os
 from typing import Any, Dict, Optional, Union
-import warnings
 
 from ansys.fluent.core.fluent_connection import FluentConnection
 from ansys.fluent.core.launcher.container_launcher import DockerLauncher
@@ -127,18 +126,12 @@ def create_launcher(fluent_launch_mode: LaunchMode = None, **kwargs):
         return SlurmLauncher(**kwargs)
 
 
-def _show_gui_to_ui_mode_converter_func(old_arg, new_arg, kwargs):
-    warnings.warn(
-        "'show_gui' is deprecated. Use 'ui_mode' instead.",
-        PyFluentDeprecationWarning,
-    )
-    if kwargs[old_arg] is True and kwargs[new_arg] is None:
-        kwargs[new_arg] = UIMode.GUI
-
-
 #   pylint: disable=unused-argument
 @deprecate_arguments(
-    old_arg="show_gui", new_arg="ui_mode", converter=_show_gui_to_ui_mode_converter_func
+    old_arg="show_gui",
+    new_arg="ui_mode",
+    converter=lambda old_arg_val: UIMode.GUI if old_arg_val else None,
+    deprecation_class=PyFluentDeprecationWarning,
 )
 def launch_fluent(
     product_version: Optional[FluentVersion] = None,
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -30,7 +30,7 @@ from ansys.fluent.core.session_meshing import Meshing
 from ansys.fluent.core.session_pure_meshing import PureMeshing
 from ansys.fluent.core.session_solver import Solver
 from ansys.fluent.core.session_solver_icing import SolverIcing
-from ansys.fluent.core.utils.deprecate_args import deprecate_arguments
+from ansys.fluent.core.utils.deprecate_args import deprecate_argument
 from ansys.fluent.core.utils.fluent_version import FluentVersion
 from ansys.fluent.core.warnings import PyFluentDeprecationWarning
 
@@ -127,7 +127,7 @@ def create_launcher(fluent_launch_mode: LaunchMode = None, **kwargs):
 
 
 #   pylint: disable=unused-argument
-@deprecate_arguments(
+@deprecate_argument(
     old_arg="show_gui",
     new_arg="ui_mode",
     converter=lambda old_arg_val: UIMode.GUI if old_arg_val is True else None,
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -7,6 +7,7 @@ with gRPC.
 import logging
 import os
 from typing import Any, Dict, Optional, Union
+import warnings
 
 from ansys.fluent.core.fluent_connection import FluentConnection
 from ansys.fluent.core.launcher.container_launcher import DockerLauncher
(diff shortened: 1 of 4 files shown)


src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -125,7 +126,37 @@ def create_launcher(fluent_launch_mode: LaunchMode = None, **kwargs):
         return SlurmLauncher(**kwargs)
 
 
+def deprecate(old_arg, new_arg, converter):
```
prmukherj, line 129: This is a generic method that can be used to deprecate other arguments as well in future.

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -125,7 +126,37 @@ def create_launcher(fluent_launch_mode: LaunchMode = None, **kwargs):
...
+
+    return decorator
+
+
+def _show_gui_to_ui_mode_converter_func(old_arg, new_arg, kwargs):
+    warnings.warn(
+        "'show_gui' is deprecated. Use 'ui_mode' instead.",
+        PyFluentDeprecationWarning,
+    )
+    if kwargs[old_arg]:
```
seanpearsonuk, line 152: A few thoughts here. 
- We could create a general base class (or function) that issues the warning as long as it is provided with the names of the arguments.
- Should we be more strict about the type of `show_gui` (in the above line)? It's already the case that it can be any type. The same is true for other arguments like `start_transcript`.

prmukherj: yes @seanpearsonuk, we should do something regarding the type of "show_gui". Will update the code.

seanpearsonuk: Could you also check te same for similar arguments. Something as simple as `if x is True` for `if x` and `if x is False` for `if not x`.

src/ansys/fluent/core/launcher/launcher.py
seanpearsonuk: This function should be extracted from this module for reuse. We can embed the generalised warning within this function. (Almost) all warnings will have the same structure.

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -125,7 +127,19 @@ def create_launcher(fluent_launch_mode: LaunchMode = None, **kwargs):
...
+    warnings.warn(
+        "'show_gui' is deprecated. Use 'ui_mode' instead.",
+        PyFluentDeprecationWarning,
+    )
+    if kwargs[old_arg] is True and kwargs[new_arg] is None:
+        kwargs[new_arg] = UIMode.GUI
+
+
 #   pylint: disable=unused-argument
+@deprecate_arguments(
```
mkundu1, line 140: @prmukherj Can you please check if the docstring is changed?

prmukherj: Thank you Mainak, fixed it.

mkundu1 (approved):

hpohekar (approved):

src/ansys/fluent/core/utils/deprecate_args.py
```diff
@@ -0,0 +1,31 @@
...
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            """Performs deprecation operation on arguments of the original method."""
+            if old_arg in kwargs:
+                warnings.warn(
+                    f"'{old_arg}' is deprecated. Use '{new_arg}' instead.",
+                    deprecation_class,
+                )
+                val = converter(kwargs[old_arg])
+                if val is not None and kwargs[new_arg] is None:
```
seanpearsonuk, line 24: Bug fix.
```suggestion
                if val is not None and kwargs.get(new_arg) is None:
```

src/ansys/fluent/core/utils/deprecate_args.py
```diff
@@ -0,0 +1,31 @@
+"""Deprecates old argument and replaces them with corresponding new argument."""
```
raph-luc, line 1:
```suggestion
"""Module that provides a method to handle deprecated arguments."""
```

src/ansys/fluent/core/utils/deprecate_args.py
```diff
@@ -0,0 +1,31 @@
...
+
+import functools
+import warnings
+
+
+def deprecate_arguments(
+    old_arg, new_arg, converter, deprecation_class=DeprecationWarning
+):
+    """Deprecates old argument and replaces them with corresponding new argument."""
```
raph-luc, line 10:
```suggestion
    """Warns user that the argument provided is deprecated, and automatically replaces the 
    deprecated argument with the appropriate new argument."""
```

src/ansys/fluent/core/utils/deprecate_args.py
```diff
@@ -0,0 +1,31 @@
+"""Deprecates old argument and replaces them with corresponding new argument."""
+
+import functools
+import warnings
+
+
+def deprecate_arguments(
```
raph-luc, line 7: This only works for a single argument, right?
```suggestion
def deprecate_argument(
```

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -126,6 +127,12 @@ def create_launcher(fluent_launch_mode: LaunchMode = None, **kwargs):
 
 
 #   pylint: disable=unused-argument
+@deprecate_arguments(
+    old_arg="show_gui",
+    new_arg="ui_mode",
+    converter=lambda old_arg_val: UIMode.GUI if old_arg_val else None,
```
seanpearsonuk, line 133:
```suggestion
    converter=lambda old_arg_val: UIMode.GUI if old_arg_val is True else None,
```

src/ansys/fluent/core/utils/deprecate_args.py
```diff
@@ -0,0 +1,31 @@
...
+    old_arg, new_arg, converter, deprecation_class=DeprecationWarning
+):
+    """Deprecates old argument and replaces them with corresponding new argument."""
+
+    def decorator(func):
+        """Holds the original method to perform operations on it."""
+
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            """Performs deprecation operation on arguments of the original method."""
```
raph-luc, line 17:
```suggestion
            """Warns about the deprecated argument and replaces it with the new argument."""
```

src/ansys/fluent/core/utils/deprecate_args.py
prmukherj: Thank you @raph-luc.

src/ansys/fluent/core/utils/deprecate_args.py
prmukherj: Done. Thank you @raph-luc.

seanpearsonuk (approved):

src/ansys/fluent/core/utils/deprecate_args.py
```diff
@@ -0,0 +1,33 @@
...
+    def decorator(func):
+        """Holds the original method to perform operations on it."""
+
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            """Warns about the deprecated argument and replaces it with the new
+            argument."""
+            if old_arg in kwargs:
+                warnings.warn(
+                    f"'{old_arg}' is deprecated. Use '{new_arg}' instead.",
```
raph-luc, line 22: At this point we will be automatically replacing the deprecated argument with the new one, right? So I think a more appropriate wording would be:
```suggestion
                    f"'{old_arg}' is deprecated. Using '{new_arg}' instead.",
```
And can we move `val` definition before this warning, so that we can show it in the warning as well to better inform the user?

For example, putting these two suggestions together:
```suggestion
                    f"'{old_arg}' is deprecated. Using '{new_arg} = {val}' instead.",
```

src/ansys/fluent/core/utils/deprecate_args.py
```diff
@@ -0,0 +1,33 @@
...
+                    deprecation_class,
+                )
+                val = converter(kwargs[old_arg])
```
raph-luc, line 25: A slightly more verbose variable name would improve code clarity here I think, something like:
```suggestion
                new_value = converter(kwargs[old_arg])
```

raph-luc (approved):

PR #2802, opened 2024-05-10
build: Update support for python 3.12

prmukherj:
Update support for python 3.12.
Now we have an update for h5py parallel to python 3.12.


.github/workflows/ci.yml
```diff
@@ -110,7 +110,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.10'
+          python-version: '3.11'
```
RobPasMue, line 113: Why not moving this version number to the env section with a variable called MAIN_PYTHON_VERSION? That way, next time you only have to modify it once rather than multiple times throughout the entire file.

RobPasMue (approved):

hpohekar (approved):

mkundu1 (approved):

.github/workflows/ci.yml
prmukherj: Updated as per our discussion.

RobPasMue:
I'm not sure you can define the list as an env var though ... :(

.github/workflows/ci.yml
```diff
@@ -83,7 +85,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ['3.9', '3.10', '3.11']
+        python-version: ${{ env.SUPPORTED_PYTHON_VERSIONS }}
         should-release:
```
RobPasMue, line 89: This is what has to be reverted

RobPasMue: Have a look at the error: https://github.com/ansys/pyfluent/actions/runs/9210539031/workflow

prmukherj: Thank you.

RobPasMue (approved):

seanpearsonuk (approved):

PR #2806, opened 2024-05-10
feat: Prepare codegen to run with Fluent

mkundu1:
1. Remove the hardcoded session dependency from codegen. Instead, a collection of static-infos is passed to the codegen functions. In Fluent side, we can control how much static API should be generated.
2. The pyfluent module level variable `CODEGEN_OUTDIR` controls where API files are written out during codegen. In Fluent, this path is set to some path within Fluent package.
3. The pyfluent module level variable `CODEGEN_ZIP_SETTINGS` controls whether to zip the settings API files. This is to reduce number of files during Fluent packaging.

Draft Fluent PR - 491855

### DIFF
--- a/src/ansys/fluent/core/codegen/allapigen.py
+++ b/src/ansys/fluent/core/codegen/allapigen.py
@@ -1,21 +1,9 @@
 """Module to generate Fluent API classes."""
 
-import argparse
-import os
 from pathlib import Path
 import pickle
 
-from ansys.fluent.core import FluentMode, launch_fluent
-from ansys.fluent.core.codegen import (
-    datamodelgen,
-    print_fluent_version,
-    settingsgen,
-    tuigen,
-)
-from ansys.fluent.core.utils.fluent_version import (
-    FluentVersion,
-    get_version_for_file_name,
-)
+from ansys.fluent.core.codegen import datamodelgen, settingsgen, tuigen
 from ansys.fluent.core.utils.search import get_api_tree_file_name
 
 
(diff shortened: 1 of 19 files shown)


src/ansys/fluent/core/codegen/allapigen.py
```diff
@@ -24,37 +12,12 @@ def _update_first_level(d, u):
         d[k].update(u.get(k, {}))
 
 
-def generate():
+def generate(version: str, static_infos: dict):
     """Generate Fluent API classes."""
     api_tree = {"<meshing_session>": {}, "<solver_session>": {}}
-    parser = argparse.ArgumentParser(
```
mkundu1, line 30: I think this code is not used

src/ansys/fluent/core/codegen/datamodelgen.py
```diff
@@ -362,12 +370,42 @@ def _delete_generated_files(self):
...
 
 
-def generate(version, sessions: dict):
+def generate(version, static_infos: dict):
     """Generate datamodel API classes."""
-    return DataModelGenerator(version, sessions).write_static_info()
+    return DataModelGenerator(version, static_infos).write_static_info()
 
 
 if __name__ == "__main__":
```
mkundu1, line 378: We support running individual *gen files: `python src/ansys/fluent/core/codegen/datamodelgen.py`

src/ansys/fluent/core/session.py
```diff
@@ -47,8 +47,11 @@ def _parse_server_info_file(file_name: str):
 
 def _get_datamodel_attributes(session, attribute: str):
     try:
-        preferences_module = importlib.import_module(
-            f"ansys.fluent.core.generated.datamodel_{session._version}." + attribute
```
mkundu1, line 51: The hardcoded module path is removed. Now, module will be loaded from dynamic `CODEGEN_OUTDIR`.

prmukherj (approved):

seanpearsonuk (approved):

PR #2807, opened 2024-05-11
fix: launcher additional arguments

raph-luc:
`additional_arguments` was being evaluated in two different places: `_build_fluent_launch_args_string` and `_generate_launch_string`, and wasn't being passed properly to the container launcher

`additional_arguments` is also repeated inside `self.argvals` and `self.additional_arguments` in the launchers, not sure why, not changed in this PR

The suggestion in this PR is to evaluate `additional_arguments` only in `_build_fluent_launch_args_string`, so that it is also applicable to container launchers and not only launchers that call `_generate_launch_string` (slurm and standalone launchers currently)

### DIFF
--- a/src/ansys/fluent/core/launcher/process_launch_string.py
+++ b/src/ansys/fluent/core/launcher/process_launch_string.py
@@ -54,6 +54,7 @@ def _build_fluent_launch_args_string(**kwargs) -> str:
                 argval = fluent_map[json_key]
             launch_args_string += v["fluent_format"].replace("{}", str(argval))
     addArgs = kwargs["additional_arguments"]
+    launch_args_string += " " + addArgs
     if "-t" not in addArgs and "-cnf=" not in addArgs:
         parallel_options = build_parallel_options(
             load_machines(ncores=kwargs["processor_count"])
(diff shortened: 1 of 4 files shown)


src/ansys/fluent/core/launcher/process_launch_string.py
```diff
@@ -54,6 +54,8 @@ def _build_fluent_launch_args_string(**kwargs) -> str:
                 argval = fluent_map[json_key]
             launch_args_string += v["fluent_format"].replace("{}", str(argval))
     addArgs = kwargs["additional_arguments"]
+    if addArgs:
```
seanpearsonuk, line 57: Above (existing) variable name is camel case. It's also abbreviated in such a way that its name changes meaning.

seanpearsonuk (approved):

seanpearsonuk:
@raph-luc many thanks. Please could you go over this in the next meeting. In particular, please cover issues such as _`additional_arguments` is also repeated inside `self.argvals` and `self.additional_arguments` in the launchers, not sure why, not changed in this PR_. 

Thanks.

hpohekar (approved):

prmukherj (approved):

src/ansys/fluent/core/launcher/process_launch_string.py
raph-luc: Good catch, I agree, going to include the renaming of that symbol in this PR as well

raph-luc: Commit 682b5112

mkundu1 (approved):

PR #2808, opened 2024-05-11
refactor: more detailed deprecated argument warning message

raph-luc:


src/ansys/fluent/core/utils/deprecate_args.py
```diff
@@ -18,14 +27,16 @@ def wrapper(*args, **kwargs):
...
+                old_value = kwargs[old_arg]
+                warning_str = f"'{old_arg}' is deprecated. "
+                new_value = converter(kwargs[old_arg])
+                if kwargs.get(new_arg) is None:
+                    kwargs[new_arg] = new_value
+                    warning_str += f"Use '{new_arg} = {_str_repr(new_value)}' instead of '{old_arg} = {_str_repr(old_value)}'."
+                else:
+                    warning_str += f"Use only '{new_arg}' instead."
                 kwargs.pop(old_arg)
+                warnings.warn(warning_str, deprecation_class, stacklevel=2)
```
raph-luc, line 39: `stacklevel=2` means that the warning will indicate what call exactly resulted in the deprecation warning, rather than where in the code the deprecation warning is raised

Before:
`C:\ansysdev\pyfluent\src\ansys\fluent\core\utils\deprecate_args.py:39: PyFluentDeprecationWarning: ...`

Now with `stacklevel=2`:
`C:\tests\test_deprecation.py:3: PyFluentDeprecationWarning: ...`
Which indicates exactly the file name and line number of the call that includes the deprecated argument

mkundu1:
I think what we have here is combination of a warning and an information. We shall warn user that a deprecated arg-name is entered and what is to use instead. I think the message `'show_gui' is deprecated. Use 'ui_mode = "gui"' instead of 'show_gui = True'` is somewhat incomplete as we are not telling users what to do for `'show_gui = False'`. The previous message `'show_gui' is deprecated. Use 'ui_mode' instead.` is better as in that case user will probably look-up the usage of `ui_mode` in the documentation and will get the full picture.

We shall also inform the user about how the value is converted for the current input, otherwise user may think that the entered value has been discarded and try to re-execute the function again.

So, perhaps we can structure the code in this way:
1. Show warning for the deprecated arg-name usage.
2. Convert deprecated arg-name/value to new arg-name/value.
3. Show information about the above conversion.

raph-luc:
> So, perhaps we can structure the code in this way:
> 
> 1. Show warning for the deprecated arg-name usage.
> 2. Convert deprecated arg-name/value to new arg-name/value.
> 3. Show information about the above conversion.

@mkundu1 do you mean just using a warning print and a separate information print (I imagine you mean something like `logging.info()`) instead of putting it all together in a single warning message as I suggest here? That is closer to what I was originally suggesting as well (giving the user all the information in a more detailed way), but @seanpearsonuk raised a point that it might be too much information to provide: https://github.com/ansys/pyfluent/pull/2800#discussion_r1596603327 which is a fair point and that is partly why I suggest we be less verbose and more direct as in this PR.

Regardless, this PR already addresses all 3 points in my understanding, for point 1 we show a warning for the deprecated argument: `'show_gui' is deprecated.`;  point 2 is already handled by the `converter` function that @prmukherj had included and does the exact conversion from the previous argument to the new one, that is what we are using here; point 3 is the second part of the warning: `Use 'ui_mode = "gui"' instead of 'show_gui = True'.` which shows the exact conversion performed and what the user should use next time.

Let me know if you want to meet offline to discuss this further

src/ansys/fluent/core/utils/deprecate_args.py
```diff
@@ -3,13 +3,22 @@
...
+
 
 def deprecate_argument(
     old_arg, new_arg, converter, deprecation_class=DeprecationWarning
 ):
     """Warns user that the argument provided is deprecated, and automatically replaces
     the deprecated argument with the appropriate new argument."""
 
+    def _str_repr(var):
+        """Converts string or FluentEnum variable to quoted string representation."""
```
PipKat, line 16:
```suggestion
        """Converts a string or FluentEnum variable to quoted string representation."""
```

mkundu1 (approved):

prmukherj (approved):

seanpearsonuk (approved):

hpohekar (approved):

PR #2812, opened 2024-05-13
test: Add a test for System Coupling related settings

ochernuk:
Adding a simple test to check System Coupling related settings. It's probably worth adding a few more, but that can be done separately.

### DIFF
--- a/src/ansys/fluent/core/systemcoupling.py
+++ b/src/ansys/fluent/core/systemcoupling.py
@@ -90,21 +90,22 @@ class SystemCoupling:
                 file_name=scp_file_name
             )
 
-            # download the file locally in case Fluent is in a container
             if self._solver._fluent_connection._remote_instance != None:
+                # download the file locally in case Fluent is remote
+                # assume file transfer service is configured - download the file
+                self._solver.download(scp_file_name)
+            elif self._solver.connection_properties.inside_container:
+                # Right now, the way that PyFluent containers and tests are set up,
+                # the local Fluent container working directory will correspond to
+                # pyfluent.EXAMPLES_PATH in the host, so that is where the SCP file
+                # will be written.
                 examples_path_scp = os.path.join(pyfluent.EXAMPLES_PATH, scp_file_name)
                 if os.path.exists(examples_path_scp):
-                    # For container mode without file transfer service (e.g. on GitHub),
-                    # produced files are located in pyfluent.EXAMPLES_PATH. Just use
-                    # that path, instead of attempting to download the file from the container.
                     scp_file_name = examples_path_scp
-                else:
-                    # assume file transfer service is configured - download the file
-                    self._solver.download(scp_file_name)
 
             assert os.path.exists(
                 scp_file_name
-            ), "ERROR: could not create System Coupling SCP file"
+            ), f"ERROR: could not create System Coupling SCP file: {scp_file_name}"
 
             with open(scp_file_name, "r") as f:
                 xml_string = f.read()
--- a/src/ansys/fluent/core/systemcoupling.py
+++ b/src/ansys/fluent/core/systemcoupling.py
@@ -5,6 +5,7 @@ import os
 from typing import List
 import xml.etree.ElementTree as XmlET
 
+import ansys.fluent.core as pyfluent
 from ansys.fluent.core.utils.fluent_version import FluentVersion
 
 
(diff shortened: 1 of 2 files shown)


seanpearsonuk (approved):

mkundu1 (approved):

ochernuk:
@seanpearsonuk @mkundu1 @raph-luc wondering if you could offer any insight here. The test that I added is failing using v24.2.0 Fluent. Looks like it comes down to not being able to download the SCP file that Fluent generates from the container. So, either it's not using the download functionality correctly here:

https://github.com/ansys/pyfluent/blob/dcb2f69f181a0c9f68ac6515cf2b8aa8e97d167f/src/ansys/fluent/core/systemcoupling.py#L92

... or the `download()` function is not working on GitHub.

raph-luc:
@ochernuk You're right, and it's failing in v241 too not only v242. We definitely need some tests specifically for the File Transfer Service-related methods, I will make sure we bring this up in the team meeting tomorrow if others don't get to this first.

seanpearsonuk:
> @seanpearsonuk @mkundu1 @raph-luc wondering if you could offer any insight here. The test that I added is failing using v24.2.0 Fluent. Looks like it comes down to not being able to download the SCP file that Fluent generates from the container. So, either it's not using the download functionality correctly here:
> 
> https://github.com/ansys/pyfluent/blob/dcb2f69f181a0c9f68ac6515cf2b8aa8e97d167f/src/ansys/fluent/core/systemcoupling.py#L92
> 
> ... or the `download()` function is not working on GitHub.
> 
> (the assert on line 96 in the above link fails)

@hpohekar Is it expected to work? Do we have relevant tests?

hpohekar:
> > @seanpearsonuk @mkundu1 @raph-luc wondering if you could offer any insight here. The test that I added is failing using v24.2.0 Fluent. Looks like it comes down to not being able to download the SCP file that Fluent generates from the container. So, either it's not using the download functionality correctly here:
> > https://github.com/ansys/pyfluent/blob/dcb2f69f181a0c9f68ac6515cf2b8aa8e97d167f/src/ansys/fluent/core/systemcoupling.py#L92
> > 
> > ... or the `download()` function is not working on GitHub.
> > (the assert on line 96 in the above link fails)
> 
> @hpohekar Is it expected to work? Do we have relevant tests?

@seanpearsonuk We are not using any file transfer service by default. In case of container mode, specifically in GitHub the written file will be available in `pyfluent.EXAMPLES_PATH` directory by default. No need to download it explicitly in the GitHub test.

Right now it is failing at `self._solver.download(scp_file_name)` call because `file_transfer_service` is not provided.

If we still want to download it then we need add following code inside the test.

```
from ansys.fluent.core.utils.file_transfer_service import RemoteFileTransferStrategy
from pathlib import Path
pyfluent.USE_FILE_TRANSFER_SERVICE = True
session = pyfluent.launch_fluent(file_transfer_service=RemoteFileTransferStrategy())
file_name = Path(pyfluent.EXAMPLES_PATH) / scp_file_name
session.download(file_name=file_name ) # it will download that file in `pyfluent.USER_DATA_PATH` by default.
```

hpohekar:
> > @seanpearsonuk @mkundu1 @raph-luc wondering if you could offer any insight here. The test that I added is failing using v24.2.0 Fluent. Looks like it comes down to not being able to download the SCP file that Fluent generates from the container. So, either it's not using the download functionality correctly here:
> > https://github.com/ansys/pyfluent/blob/dcb2f69f181a0c9f68ac6515cf2b8aa8e97d167f/src/ansys/fluent/core/systemcoupling.py#L92
> > 
> > ... or the `download()` function is not working on GitHub.
> > (the assert on line 96 in the above link fails)
> 
> @hpohekar Is it expected to work? Do we have relevant tests?

@seanpearsonuk 

Is it expected to work? - No. It only works if we provide any file transfer service.

Do we have relevant tests? - Yes.

https://github.com/ansys/pyfluent/blob/main/tests/test_file_transfer_service.py#L42

seanpearsonuk:
> pyfluent.USE_FILE_TRANSFER_SERVICE = True
> session = pyfluent.launch_fluent(file_transfer_service=RemoteFileTransferStrategy())

That's great, many thanks for the detailed response, @hpohekar.

@mkundu1 (et al) should `file_transfer_service=RemoteFileTransferStrategy()` simply override the module-level setting?

raph-luc (approved):

Issue #2818, opened 2024-05-14
arg vals data management


seanpearsonuk:
Check this, as discussed, and update as necessary.

PR #2822, opened 2024-05-15
refactor: Launcher argvals refactoring

hpohekar:
closes https://github.com/ansys/pyfluent/issues/2818

### DIFF
--- a/src/ansys/fluent/core/launcher/container_launcher.py
+++ b/src/ansys/fluent/core/launcher/container_launcher.py
@@ -153,23 +153,23 @@ class DockerLauncher:
         del argvals["self"]
         if argvals["start_timeout"] is None:
             argvals["start_timeout"] = 60
-        for arg_name, arg_values in argvals.items():
-            setattr(self, arg_name, arg_values)
         self.argvals = argvals
-        self.new_session = self.mode.value[0]
+        self.new_session = self.argvals["mode"].value[0]
         self.file_transfer_service = file_transfer_service
 
     def __call__(self):
-        if self.mode == FluentMode.SOLVER_ICING:
+        if self.argvals["mode"] == FluentMode.SOLVER_ICING:
             self.argvals["fluent_icing"] = True
         args = _build_fluent_launch_args_string(**self.argvals).split()
-        if FluentMode.is_meshing(self.mode):
+        if FluentMode.is_meshing(self.argvals["mode"]):
             args.append(" -meshing")
-        if self.container_dict is None:
+        if self.argvals["container_dict"] is None:
             setattr(self, "container_dict", {})
-        if self.product_version:
-            self.container_dict["image_tag"] = f"v{self.product_version.value}"
-        if self.dry_run:
+        if self.argvals["product_version"]:
+            self.container_dict["image_tag"] = (
+                f"v{self.argvals['product_version'].value}"
+            )
+        if self.argvals["dry_run"]:
             config_dict, *_ = configure_container_dict(args, **self.container_dict)
             from pprint import pprint
 
@@ -190,7 +190,7 @@ class DockerLauncher:
             port=port,
             password=password,
             file_transfer_service=self.file_transfer_service,
-            cleanup_on_exit=self.cleanup_on_exit,
+            cleanup_on_exit=self.argvals["cleanup_on_exit"],
             slurm_job_id=self.argvals and self.argvals.get("slurm_job_id"),
             inside_container=True,
         )
@@ -199,12 +199,12 @@ class DockerLauncher:
             fluent_connection=fluent_connection,
             scheme_eval=fluent_connection._connection_interface.scheme_eval,
             file_transfer_service=self.file_transfer_service,
-            start_transcript=self.start_transcript,
+            start_transcript=self.argvals["start_transcript"],
         )
 
-        if self.start_watchdog is None and self.cleanup_on_exit:
+        if self.argvals["start_watchdog"] is None and self.argvals["cleanup_on_exit"]:
             setattr(self, "start_watchdog", True)
-        if self.start_watchdog:
+        if self.argvals["start_watchdog"]:
             logger.debug("Launching Watchdog for Fluent container...")
             watchdog.launch(os.getpid(), port, password)
 
--- a/src/ansys/fluent/core/launcher/container_launcher.py
+++ b/src/ansys/fluent/core/launcher/container_launcher.py
@@ -166,9 +166,9 @@ class DockerLauncher:
         if self.argvals["container_dict"] is None:
             self.argvals["container_dict"] = {}
         if self.argvals["product_version"]:
-            self.container_dict["image_tag"] = (
-                f"v{self.argvals['product_version'].value}"
-            )
+            self.argvals["container_dict"][
+                "image_tag"
+            ] = f"v{self.argvals['product_version'].value}"
         if self.argvals["dry_run"]:
             config_dict, *_ = configure_container_dict(
                 args, **self.argvals["container_dict"]
@@ -186,7 +186,7 @@ class DockerLauncher:
                 del config_dict_h
             return config_dict
 
-        port, password = start_fluent_container(args, self.container_dict)
+        port, password = start_fluent_container(args, self.argvals["container_dict"])
 
         fluent_connection = FluentConnection(
             port=port,
--- a/src/ansys/fluent/core/launcher/container_launcher.py
+++ b/src/ansys/fluent/core/launcher/container_launcher.py
@@ -164,13 +164,15 @@ class DockerLauncher:
         if FluentMode.is_meshing(self.argvals["mode"]):
             args.append(" -meshing")
         if self.argvals["container_dict"] is None:
-            setattr(self, "container_dict", {})
+            self.argvals["container_dict"] = {}
         if self.argvals["product_version"]:
             self.container_dict["image_tag"] = (
                 f"v{self.argvals['product_version'].value}"
             )
         if self.argvals["dry_run"]:
-            config_dict, *_ = configure_container_dict(args, **self.container_dict)
+            config_dict, *_ = configure_container_dict(
+                args, **self.argvals["container_dict"]
+            )
             from pprint import pprint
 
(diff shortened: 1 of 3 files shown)


mkundu1 (approved):

src/ansys/fluent/core/launcher/container_launcher.py
```diff
@@ -153,24 +153,26 @@ def __init__(
...
+        if self.argvals["container_dict"] is None:
+            self.argvals["container_dict"] = {}
+        if self.argvals["product_version"]:
+            self.argvals["container_dict"][
+                "image_tag"
+            ] = f"v{self.argvals['product_version'].value}"
+        if self.argvals["dry_run"]:
+            config_dict, *_ = configure_container_dict(
+                args, **self.argvals["container_dict"]
+            )
```
seanpearsonuk, line 175: Just wondering why some arguments are massaged in `__init__()`, others in `__call__`. The default would normally be `__init__()`.

I'll explain why this will become particularly pertinent here. The idea here is that users will use 
 and reuse _launcher_  _command objects_ directly (without going via `launch_fluent`). Before executing a command, the user should be able to query its argument state and see the argument values that will actually be used. 

@mkundu1 @prmukherj @hpohekar @raph-luc

mkundu1: Yes, the value of `self.argvals` should be final (as close as possible) at the end of `__init__()`.

hpohekar: Yes, I agree. Therefore I already explained in the previous comments ([on closed PR](https://github.com/ansys/pyfluent/pull/2820)) that we are using only `self.<argument_name>` inside `__call__` method not `self.argvals["<argument_name>"]` simultaneously.

image

raph-luc: I believe that is not the point being raised here @hpohekar. What I believe they are saying is that all this handling of `self.argvals` being done in `__call__()` should instead be done in `__init__()`, and I think that is a very good point.

Another benefit is that moving this logic to `__init__()` will save on having to prefix `self.` before every `argvals` reference, as `argvals` is a local variable there. 

So, for example, rather than having this in `__call__()`:
```
if self.argvals["mode"] == FluentMode.SOLVER_ICING:
    self.argvals["fluent_icing"] = True
```
We would have this in `__init__()`:
```
if argvals["mode"] == FluentMode.SOLVER_ICING:
    argvals["fluent_icing"] = True
```
(before `self.argvals = argvals`)

src/ansys/fluent/core/launcher/standalone_launcher.py
raph-luc, line 1: Related to this refactoring, should we also be reviewing these blocks of apparently copy-pasted code (similar in `DockerLauncher` and `PIMLauncher` in the `launcher.py` file):
https://github.com/ansys/pyfluent/blob/183b5c216b0c42030e55b7b25dad1e0768d6f914/src/ansys/fluent/core/launcher/launcher.py#L63-L86

And replacing these with for example
```
return StandaloneLauncher(**kwargs)
```
We can `kwargs.pop(key)` or `del kwargs[key]` as needed to remove arguments that we do not want to pass forward (but I think the launcher should be able to handle invalid arguments?)

Maybe this should be considered in a separate PR/issue

hpohekar: @raph-luc There is a reason to do this. Earlier we had done the same. i.e. `StandaloneLauncher(**kwargs)` but we realized that it gives scope to pass the unwanted arguments and to avoid this we are passing them explicitly.

All the signatures are different as per the required arguments for each launcher.

raph-luc: I believe I understand what you mean, that is why I mentioned `kwargs.pop(key)` above. This is what I meant: https://github.com/ansys/pyfluent/pull/2826

Please let me know what your thoughts are.

@mkundu1 @seanpearsonuk @prmukherj

src/ansys/fluent/core/launcher/pim_launcher.py
```diff
@@ -141,18 +141,16 @@ def __init__(
...
-            setattr(self, arg_name, arg_values)
         self.argvals = argvals
-        self.new_session = self.mode.value[0]
+        self.new_session = self.argvals["mode"].value[0]
 
-        if self.additional_arguments:
+        if self.argvals["additional_arguments"]:
             logger.warning(
                 "'additional_arguments' option for 'launch_fluent' is currently not supported "
                 "when starting a remote Fluent PyPIM client."
```
PipKat, line 150:
```suggestion
                "'additional_arguments' option for 'launch_fluent()' method is not supported "
                "when starting a remote Fluent PyPIM client."
```
Avoid words like currently to keep documentation "timeless".

src/ansys/fluent/core/launcher/pim_launcher.py
```diff
@@ -141,18 +141,16 @@ def __init__(
...
             )
 
-        if self.start_watchdog:
+        if self.argvals["start_watchdog"]:
             logger.warning(
                 "'start_watchdog' argument for 'launch_fluent' is currently not supported "
                 "when starting a remote Fluent PyPIM client."
```
PipKat, line 156:
```suggestion
                "'start_watchdog' argument for 'launch_fluent()' method is not supported "
                "when starting a remote Fluent PyPIM client."
```

src/ansys/fluent/core/launcher/pim_launcher.py
```diff
@@ -163,18 +161,18 @@ def __call__(self):
         logger.info(
             "Starting Fluent remotely. The startup configuration will be ignored."
```
PipKat, line 162:
```suggestion
            "Starting Fluent remotely. The startup configuration is ignored."
```

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -184,44 +184,46 @@ def __init__(
...
+            self.argvals["env"] = {}
+        kwargs = _get_subprocess_kwargs_for_fluent(self.argvals["env"], self.argvals)
+        if self.argvals["cwd"]:
+            kwargs.update(cwd=self.argvals["cwd"])
+        launch_string += _build_journal_argument(
+            self.argvals["topy"], self.argvals["journal_file_names"]
+        )
 
         if is_windows():
             # Using 'start.exe' is better, otherwise Fluent is more susceptible to bad termination attempts
```
PipKat, line 223:
```suggestion
            # Using 'start.exe' is better; otherwise Fluent is more susceptible to bad termination attempts.
```

src/ansys/fluent/core/launcher/pim_launcher.py

src/ansys/fluent/core/launcher/pim_launcher.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/launcher/pim_launcher.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/launcher/standalone_launcher.py
hpohekar: Done. Thank you.

prmukherj (approved):

src/ansys/fluent/core/launcher/container_launcher.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -184,103 +184,111 @@ def __init__(
...
-            setattr(self, arg_name, arg_values)
-        self.argvals = argvals
-        self.new_session = self.mode.value[0]
+        self.new_session = argvals["mode"].value[0]
         self.file_transfer_service = file_transfer_service
 
-    def __call__(self):
-        if self.lightweight_mode is None:
+        if argvals["lightweight_mode"] is None:
             # note argvals is no longer locals() here due to _get_session_info() pass
```
raph-luc, line 191:
```suggestion
```
This code comment is no longer relevant and can be removed

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -184,103 +184,111 @@ def __init__(
...
 
-        server_info_file_name = _get_server_info_file_name()
-        launch_string = _generate_launch_string(
-            self.argvals,
-            self.mode,
-            server_info_file_name,
+        self._server_info_file_name = _get_server_info_file_name()
+        self._launch_string = _generate_launch_string(
+            argvals,
+            argvals["mode"],
```
raph-luc, line 206: Should we also refactor `_generate_launch_string()` slightly so that we don't need to pass `argvals` and again `argvals["mode"]`?

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -184,103 +184,111 @@ def __init__(
...
 
-    def __call__(self):
-        if self.lightweight_mode is None:
+        if argvals["lightweight_mode"] is None:
             # note argvals is no longer locals() here due to _get_session_info() pass
-            self.argvals.pop("lightweight_mode")
-            setattr(self, "lightweight_mode", False)
-        fluent_version = _get_standalone_launch_fluent_version(self.product_version)
+            argvals.pop("lightweight_mode")
+            argvals["lightweight_mode"] = False
```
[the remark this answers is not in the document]
hpohekar: Done. Thank you.

src/ansys/fluent/core/launcher/standalone_launcher.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/launcher/standalone_launcher.py
hpohekar: https://github.com/ansys/pyfluent/issues/2833

Issue #2830, opened 2024-05-16
Add tests for known file transfer issues.


hpohekar:
Add tests for known file transfer issues.

PR #2831, opened 2024-05-16
fix: Add tests for known file transfer issues.

hpohekar:
closes https://github.com/ansys/pyfluent/issues/2830

Add tests for known file transfer issues.

We will try to find proper solution for these issues.

### DIFF
--- a/src/ansys/fluent/core/session.py
+++ b/src/ansys/fluent/core/session.py
@@ -364,10 +364,7 @@ class BaseSession:
         for relevant files."
 
     def upload(
-        self,
-        file_name: Union[list[str], str],
-        remote_file_name: Optional[str] = None,
-        command_name: Optional[str] = None,
+        self, file_name: Union[list[str], str], remote_file_name: Optional[str] = None
     ):
         """Upload a file to the server.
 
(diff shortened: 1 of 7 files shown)


src/ansys/fluent/core/session.py
```diff
@@ -374,10 +377,14 @@ def upload(
             Name of the local file to upload to the server.
         remote_file_name : str, optional
             remote file name, by default None
+        command_name : str, optional
```
mkundu1, line 380: This function should not be made specific to a command name.

hpohekar: We will find proper solution for this in future. Thank you.

tests/test_file_transfer_service.py
```diff
@@ -52,3 +56,39 @@ def test_remote_grpc_fts_container(monkeypatch, new_solver_session, new_mesh_ses
...
+        "mixing_elbow.cas.h5", "pyfluent/mixing_elbow"
+    )
+    data_file_name = examples.download_file(
+        "mixing_elbow.dat.h5", "pyfluent/mixing_elbow"
+    )
+    assert case_file_name
+    assert data_file_name
+    solver = pyfluent.launch_fluent(file_transfer_service=LocalFileTransferStrategy())
+
+    with pytest.raises(FileNotFoundError):
```
mkundu1, line 78: This line should not be there if the test is already skipped. When we fix the code, we shall just remove the skip marker.

mkundu1 (approved):

tests/test_file_transfer_service.py
raph-luc: Add a code comment instead above the line that fails to indicate it, so that we still have this documented. But without the `pytest.raises()`

raph-luc (approved):

tests/test_file_transfer_service.py
hpohekar: Done. Thank you.

prmukherj (approved):

PR #2836, opened 2024-05-17
fix: Support search_root properly with dynamic-path module import

mkundu1:
This fixes the issue with search_root introduced due to the dynamic path of the generated API. Now the modules imported from the dynamic path will be qualified with mode and version information in their names, so we won't loose any information for the search function.


tests/test_search.py
mkundu1, line 1: Reverting back #2806 changes as we now extract complete information about mode/version from `obj`.

hpohekar (approved):

seanpearsonuk (approved):

PR #2851, opened 2024-05-21
fix: Fix an issue when api files are not generated

mkundu1:


hpohekar (approved):

prmukherj (approved):

src/ansys/fluent/core/session_base_meshing.py
```diff
@@ -103,7 +103,7 @@ def _meshing_root(self):
                 CODEGEN_OUTDIR / f"datamodel_{self._version}" / "meshing.py",
             )
             meshing_root = meshing_module.Root(self._se_service, "meshing", [])
-        except ImportError:
+        except (ImportError, FileNotFoundError):
```
seanpearsonuk, line 106: Just noticed something in this class (nothing to do with these changes!). It looks a bit odd and unnecessary that we are using `@property` for these private implementations, which also leads to the curious looking implementation of the `meshing()` (property) method below. I think it just makes reading the code a little more awkward.

seanpearsonuk (approved):

PR #2861, opened 2024-05-23
fix: root affected level

mkundu1:
Distinguishing between `initialize` and `reinitialize`:
`initialize` - A different workflow is created, root affected callback is cleared.
`reinitialize` - The same workflow is reinitialized, root affected callback is not cleared.

### DIFF
--- a/src/ansys/fluent/core/session_pure_meshing.py
+++ b/src/ansys/fluent/core/session_pure_meshing.py
@@ -117,17 +117,17 @@ class PureMeshing(BaseSession):
 
     def watertight(self):
         """Get a new watertight workflow."""
-        self._base_meshing.watertight_workflow.reinitialize()
+        self._base_meshing.watertight_workflow.initialize()
         return self._base_meshing.watertight_workflow
 
     def fault_tolerant(self):
         """Get a new fault-tolerant workflow."""
-        self._base_meshing.fault_tolerant_workflow.reinitialize()
+        self._base_meshing.fault_tolerant_workflow.initialize()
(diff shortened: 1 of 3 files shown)


src/ansys/fluent/core/session_pure_meshing.py
```diff
@@ -117,17 +117,17 @@ def workflow(self):
 
     def watertight(self):
         """Get a new watertight workflow."""
-        self._base_meshing.watertight_workflow.reinitialize()
+        self._base_meshing.watertight_workflow.initialize()
```
mkundu1, line 120: Just curious, is there any benefit in returning the previous instance of the `WatertightMeshingWorkflow` class? If `watertight()` always creates a new instance of `WatertightMeshingWorkflow`, we could clear the root-affected callback at the `WatertightMeshingWorkflow.__init__` method.

seanpearsonuk: Yes, the current code seems unnecessarily complicated. Why is it telling objects to initialize and reinitialize etc?

prmukherj: We can discuss this once in today's meeting.

hpohekar (approved):

src/ansys/fluent/core/session_pure_meshing.py
prmukherj: @mkundu1, Please put in a github issue for refactoring this. Thank you

prmukherj (approved):

PR #2869, opened 2024-05-25
fix: extend unsubscribing cb for load and create wfs.

prmukherj:
1. Restructuring and clean up of enhanced meshing workflow.
2. Fix issues with '_refresh_task_accessors' in PyConsole.
3. Fix issues with initializing and re-initializing workflows.
4. Testing done both in PyFluent as well as PyConsole.


src/ansys/fluent/core/workflow.py
```diff
@@ -1558,7 +1557,10 @@ def refresh_after_sleep(_):
                     logger.debug("Already _refreshing, ...")
                 self._refreshing = True
                 logger.debug("Call _refresh_task_accessors")
-                _refresh_task_accessors(self)
+                try:
+                    _refresh_task_accessors(self)
+                except Exception:
```
seanpearsonuk, line 1562: Is there a more specific `Exception` derived class for which we know it is correct to `pass`?

seanpearsonuk:
I need a little reminding about the `initialize()` and `reinitialize()` methods. I would like to make sure that we need both of them (or even one of them). Also, there is a test comment mentioning a random CI issue connected to `reinitialize()` which I believe should be out of date now.

prmukherj:
> I need a little reminding about the `initialize()` and `reinitialize()` methods. I would like to make sure that we need both of them (or even one of them). Also, there is a test comment mentioning a random CI issue connected to `reinitialize()` which I believe should be out of date now.

As of now there is nothing called initialize. Reinitialize just resets the same workflow to it's original state.

src/ansys/fluent/core/workflow.py
prmukherj: Added a comment.

seanpearsonuk (approved):

PR #2870, opened 2024-05-27
fix: ssl certificate fix for nltk.download() on Linux

hpohekar:
closes https://github.com/ansys/pyfluent/issues/2871

Resolves ssl certificate error for `nltk.download()` on Linux.

image

### DIFF
--- a/src/ansys/fluent/core/utils/search.py
+++ b/src/ansys/fluent/core/utils/search.py
@@ -500,7 +500,16 @@ def _search_whole_word(
 
 def _download_nltk_data():
     """Download NLTK data on demand."""
+    import ssl
+
     import nltk
+
+    try:
+        _create_unverified_context = ssl._create_unverified_context
+    except AttributeError:
+        pass
+    else:
+        ssl._create_default_https_context = _create_unverified_context
 
     packages = ["wordnet", "omw-1.4"]
     for package in packages:


mkundu1 (approved):

src/ansys/fluent/core/utils/search.py
```diff
@@ -500,8 +500,17 @@ def _search_whole_word(
...
+    import ssl
+
     import nltk
 
+    try:
+        _create_unverified_context = ssl._create_unverified_context
+    except AttributeError:
+        pass
+    else:
+        ssl._create_default_https_context = _create_unverified_context
```
raph-luc, line 512:
```suggestion
    try:
        _create_unverified_context = ssl._create_unverified_context
        
    except AttributeError:
        pass
    else:
```

raph-luc, line 512: Minor thing: curious about the choice to use `else`, when does the exception happen? Is this code the same as:
```suggestion
    try:
        _create_unverified_context = ssl._create_unverified_context
        ssl._create_default_https_context = _create_unverified_context
    except AttributeError:
        pass
```
?

raph-luc (approved):

src/ansys/fluent/core/utils/search.py
hpohekar:
```
import ssl

try:
    _create_unverified_https_context = ssl._create_unverified_context
except AttributeError:
    # Legacy Python that doesn't verify HTTPS certificates by default
    pass
else:
    # Handle target environment that doesn't support HTTPS verification
    ssl._create_default_https_context = _create_unverified_https_context
```

Issue #2867, opened 2024-05-24
Mount the PyFluent working directory to Docker containers by default


raph-luc:
Ever since https://github.com/ansys/pyfluent/pull/359 we have been working with Fluent containers mounted to `pyfluent.EXAMPLES_PATH` in the host system. This was apparently fine as until recently, PyFluent with Fluent containers was mainly being used for the PyFluent CI and not many people external to the PyFluent team were using PyFluent with Fluent containers as well. 

Evidently, as seen in the recent uptick in user questions and bug reports related to Fluent containers, more and more people are using PyFluent with Fluent containers.

I do not believe that mounting the `pyfluent.EXAMPLES_PATH` to the Fluent container by default is useful or intuitive to users. I would suggest that this be reconsidered, and changed so that the current PyFluent working directory is mounted to the Fluent container working directory. In this way, both PyFluent and Fluent would be working in the exact same working directory (though each one sees it with a different absolute path). Note that currently PyFluent launches standalone Fluent by default also in the current PyFluent working directory, so this would mean the docker and standalone launchers would have behavior that is more similar to each other.

For the CI, I believe this means changes would also need to be made to the `examples.download_file()` method to make things seamless, as it is currently downloading to `pyfluent.EXAMPLES_PATH` by default, and it should then download to the current working directory (which is also more intuitive IMO). 

This would also break scripts from the users that may already be relying on this (odd) behavior of PyFluent mounting containers to the absolute path of `pyfluent.EXAMPLES_PATH` (and downloading example files to there as well).

Overall, I believe this would be a lot more intuitive to users. This change, with the changes for https://github.com/ansys/pyfluent/issues/2865, should make using PyFluent with Fluent containers a lot more user friendly.

In addition, making the example configuring `launch_fluent()` with `host_mount_path` more prominent would probably also be helpful (maybe adding more examples like that as well, now that Fluent containers seems to be a more common use-case).

@mkundu1 @hpohekar @prmukherj @seanpearsonuk

mkundu1:
Very good point, synchronizing the working directory will give a more intuitive behaviour. And if we make the default mount path configurable, we don't have to make much adjustments for the CIs.

seanpearsonuk:
@raph-luc @mkundu1 @prmukherj @hpohekar  I feel that it is important to expedite this work.  Any of you may self-assign.

raph-luc:
[...]

Issue #2865, opened 2024-05-24
Make the mapped path within container easily visible to users (either through logging or through a property at session level).


mkundu1:

raph-luc:
I think this could be through convenience functions as well, to more easily fetch what the PyFluent workdir is in the host system, and what the Fluent workdir is in the guest container system.

Some of the confusion here is probably unavoidable IMO due to the host-guest architecture, with PyFluent and Fluent in different systems, unless we wrap all TUI calls and file operations to transform the paths for the users so they don't have to think about this.

PR #2874, opened 2024-05-29
fix: update fluent container mount path and show them using logging

hpohekar:
closes https://github.com/ansys/pyfluent/issues/2867, https://github.com/ansys/pyfluent/issues/2865

1. Set `CONTAINER_MOUNT_PATH = None`
2. Set `CONTAINER_MOUNT_PATH = EXAMPLES_PATH ` using autouse fixture.
3. `host_mount_path` and `container_mount_path` are visible to user through logger.warning()

logging information of PyFluent and Fluent workdir is shown as follows.

```
>>> solver = launch_fluent()
pyfluent.launcher WARNING: Starting Fluent container mounted to /mnt/c/Users/hpohekar/pyfluent, with this path available as /mnt/pyfluent for the Fluent session running inside the container.
>>>
>>> solver.exit()
>>>
```


prmukherj (approved):

seanpearsonuk (approved):

src/ansys/fluent/core/__init__.py
```diff
@@ -71,6 +71,8 @@ def version_info() -> str:
 )
 EXAMPLES_PATH = os.path.join(USER_DATA_PATH, "examples")
 
+PYFLUENT_WORKING_DIR = Path(__file__.split("src")[0]).resolve()
```
raph-luc, line 74: This is the PyFluent source files directory, not the PyFluent working directory.

The current PyFluent working directory (which is the same as the current Python interpreter working directory) is provided by `os.getcwd()`.

I do not believe we need a `PYFLUENT_WORKING_DIR` constant, the value of the current working directory is always provided by `os.getcwd()`.

Is there a reason for this approach with the module constant that I am missing?

raph-luc (dismissed):

src/ansys/fluent/core/launcher/fluent_container.py
```diff
@@ -228,6 +228,9 @@ def configure_container_dict(
         host_mount_path = volumes_string.replace(":" + container_mount_path, "")
         logger.debug(f"host_mount_path: {host_mount_path}")
         logger.debug(f"container_mount_path: {container_mount_path}")
+    logger.info(
+        f"PyFluent workdir: Fluent workdir {host_mount_path}:{container_mount_path}"
+    )
```
raph-luc, line 233: I think this would be more appropriate as a warning so that it is more visible to users, and more clear and detailed so as to alleviate confusion that may happen when using Fluent containers:
```suggestion
    logger.warning(
        f"Starting Fluent container mounted to {host_mount_path}, with this path available as {container_mount_path} for the Fluent session running inside the container."
    )
```

src/ansys/fluent/core/__init__.py
hpohekar: @raph-luc 

1. `os.getcwd()` gives different values based on the file location from where it is being called.  Therefore we have used a constant to resolve this issue. 

2. `This is the PyFluent source files directory, not the PyFluent working directory.` - No. Fluent writes files in `Path(__file__.split("src")[0]).resolve()` and reads file with basename from this location.

Just verify 1 and 2 and let me know If I missed anything.

src/ansys/fluent/core/launcher/fluent_container.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/__init__.py
hpohekar: `fluent_container.py`

image

`downloads.py`

image

raph-luc: Regarding point 1, `os.getcwd()` specifically returns the current Python working directory, see the docs: "Return a unicode string representing the current working directory". The PyFluent current working directory should not be changing at all during execution, we'd need to specifically call `os.chdir(...)` for that, and we should never do that. The user has full control over the current Python interpreter working directory, and a Python library like PyFluent should typically not change it.

As to point 2, `Path(__file__.split("src")[0]).resolve()` inside `__init__.py` resolves to wherever PyFluent is installed in the environment. For development purposes, there is a `src` folder so this will be the Python source files. In a user install as I said above, it would return something like `PYFLUENT_WORKING_DIR = C:\test\.venv\Lib\site-packages\ansys\fluent\core\__init__.py` which is most certainly not what we want here.

hpohekar: @raph-luc 

I have experimented following thing.

pyfluent.PYFLUENT_WORKING_DIR = `Path(__file__.split("src")[0]).resolve()` (defined in __init__.py)
WORKING_DIR  = `os.getcwd()`         (defined in downloads.py)
WORK_DIR =  `os.getcwd()`               (defined in fluent_container .py)

cwd - `D:`

```
(env1) PS D:\> python
Python 3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)] on win32
>>> pyfluent.PYFLUENT_WORKING_DIR
WindowsPath('D:/pyfluent')
>>> from ansys.fluent.core.examples.downloads import WORKING_DIR
>>> WORKING_DIR
'D:\\'
>>> from ansys.fluent.core.launcher.fluent_container import WORK_DIR
>>> WORK_DIR
'D:\\'
>>>
```

If we run `downloads.py` and fluent_container.py` individually then it gives different results for `os.getcwd()` as follows.

`os.getcwd()`  = `D:\pyfluent\src\ansys\fluent\core\examples` for `downloads.py` 

`os.getcwd()`  = `D:\pyfluent\src\ansys\fluent\core\launcher` for `fluent_container.py` 

So we have verified that in an interpreter `os.getcwd()` gives correct and consistent results independent of inividual file locations.

We will update this now. 

Thank you.

hpohekar: Done. Thank you.

src/ansys/fluent/core/examples/downloads.py
raph-luc, line 1: @hpohekar if you look at the tests, with these changes there will still be issues with `pyfluent.EXAMPLES_PATH`. I think we can:

1. Either rework this so that during the CI tests, `pyfluent.EXAMPLES_PATH` is passed to `example.download_file()` so that it downloads to that directory rather than the default of downloading to the current working directory.

2. Or, and I would say preferably (but we should discuss this with everyone first), we can remove almost all mention of `pyfluent.EXAMPLES_PATH` from the code. I don't think there is a reason for it to exist anymore (it was actually only being used for the CI tests in my understanding). Examples files should be downloaded to where the end user specifies, or directly to the current working directory (as I believe most would expect). For the CI tests we can use a separate folder if needed, it could just be an `examples` folder inside the current working directory, I think that would be a lot cleaner and potentially avoid all the permission issues that I noticed we were having previously.

src/ansys/fluent/core/launcher/fluent_container.py
```diff
@@ -228,6 +228,9 @@ def configure_container_dict(
         host_mount_path = volumes_string.replace(":" + container_mount_path, "")
         logger.debug(f"host_mount_path: {host_mount_path}")
         logger.debug(f"container_mount_path: {container_mount_path}")
+    logger.warning(
+        f"Starting Fluent container mounted to {host_mount_path}, with this path available as {container_mount_path} for the Fluent session running inside the container."
+    )
```
raph-luc, line 233: Although I suggested this string, I think we should still briefly review this string with the rest of the team, to make sure it is clear and understandable enough to everyone. The wording can probably still be improved.

Also, @PipKat I believe your input would be appreciated here, if you can find the time.

src/ansys/fluent/core/examples/downloads.py
hpohekar: I think we should not keep `pyfluent.EXAMPLES_PATH` anymore. We have already done this. We have removed all mention of `pyfluent.EXAMPLES_PATH` because we should be consistent for all launchers, we should not patch code for GitHub only, by passing `pyfluent.EXAMPLES_PATH` to `download_file()` because it again introduces maintenance issues.

raph-luc: For the CI tests, we may need to create a folder, change the working directory to that folder, launch the container, run all the tests, then move out of that folder and delete it after running the tests. I believe this should be enough to deal with any remaining intermediate files

I think that we will also need to make similar changes to other `pyfluent-*` repos that were relying on `pyfluent.EXAMPLES_PATH` previously

hpohekar: It is also a good point to have `examples` folder inside cwd for CI runs, so that we can remove it whenever necessary.

raph-luc (dismissed):
Looks good to me, I think this is a tricky area to change, good job tackling this. Should still wait for additional feedback from others and double check for any potential testing issues that may happen, as Mainak warned

src/ansys/fluent/core/examples/downloads.py
raph-luc: Something like (haven't tested this code, just a quick idea that may be helpful later)
```
@pytest.fixture(autouse=True)
def run_before_each_test(request, monkeypatch):
    monkeypatch.chdir(Path(request.config.invocation_params.dir) / "test_workdir") 
```
might work, see https://docs.pytest.org/en/latest/reference/reference.html#pytest.Config.InvocationParams.dir

src/ansys/fluent/core/systemcoupling.py
```diff
@@ -97,9 +96,9 @@ def get_scp_string() -> str:
             elif self._solver.connection_properties.inside_container:
                 # Right now, the way that PyFluent containers and tests are set up,
                 # the local Fluent container working directory will correspond to
-                # pyfluent.EXAMPLES_PATH in the host, so that is where the SCP file
+                # os.getcwd() in the host, so that is where the SCP file
                 # will be written.
-                examples_path_scp = os.path.join(pyfluent.EXAMPLES_PATH, scp_file_name)
+                examples_path_scp = os.path.join(os.getcwd(), scp_file_name)
```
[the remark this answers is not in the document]
hpohekar: Yes, right. let's see.

src/ansys/fluent/core/examples/downloads.py
hpohekar: We will address this change in next PR. Thanks.

hpohekar:
We are getting following error and CI is failing. We will have to discuss proper solution for following error then we will resume this task.

image

https://github.com/ansys/pyfluent/actions/runs/9315232834/job/25640985375?pr=2874

src/ansys/fluent/core/launcher/fluent_container.py
```diff
@@ -186,7 +186,11 @@ def configure_container_dict(
         if file_transfer_service:
             host_mount_path = pyfluent.USER_DATA_PATH
         else:
-            host_mount_path = pyfluent.EXAMPLES_PATH
+            host_mount_path = (
+                pyfluent.EXAMPLES_PATH
+                if os.getenv("LOGNAME") == "ansys"
+                else os.getcwd()
+            )
```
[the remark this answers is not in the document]
raph-luc: To make this configurable through env vars, I would say we should instead add a new environment variable called something like `PYFLUENT_CONTAINER_MOUNT_PATH`, and set that to `pyfluent.EXAMPLES_PATH` (what it resolves to) during the CI tests in the runners

I think that using an existing unrelated env var is not a good approach

tests/test_launcher.py
```diff
@@ -450,3 +451,14 @@ def get_processor_count(solver):
...
+
+def test_container_warning_for_host_mount_path(caplog):
+    if check_docker_support():
+        solver = pyfluent.launch_fluent()
+        assert (
+            "this path available as /mnt/pyfluent for the Fluent session running inside the container."
+            in caplog.text
+        )
+        assert pyfluent.EXAMPLES_PATH in caplog.text
+        assert os.getcwd() not in caplog.text
```
raph-luc, line 464: I think the test should not assume that it will run in GitHub CI runners. I would say it would be best to make it more specific, configure the `container_dic` with the `host_mount_path` and `container_mount_path`, and assert both of those are in the warning message. 

For this test, can set `host_mount_path = os.getcwd()`. 

In this way, the test should work fine in both GitHub CI runners as well as when we run it locally in Windows or Linux machines outside of CI runners.

src/ansys/fluent/core/launcher/fluent_container.py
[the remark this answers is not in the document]
seanpearsonuk: Yes, the logic looks brittle.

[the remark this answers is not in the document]
mkundu1: I think a variable at pyfluent module level will be sufficient. The default velaue will be the cwd, an autouse fixture can be added to set the variable to EXAMPLES_PATH before any test.

tests/test_launcher.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/launcher/fluent_container.py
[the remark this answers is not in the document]
hpohekar: Done. Thanks a lot for all suggestions.

src/ansys/fluent/core/__init__.py
```diff
@@ -69,7 +69,9 @@ def version_info() -> str:
 USER_DATA_PATH = platformdirs.user_data_dir(
     appname="ansys_fluent_core", appauthor="Ansys"
 )
-EXAMPLES_PATH = os.path.join(USER_DATA_PATH, "examples")
+EXAMPLES_PATH = os.getcwd()
+
+PYFLUENT_CONTAINER_MOUNT_PATH = os.path.join(USER_DATA_PATH, "examples")
```
raph-luc, line 74:
```suggestion
EXAMPLES_PATH = os.path.join(USER_DATA_PATH, "examples")

CONTAINER_MOUNT_PATH = os.getcwd()
```

raph-luc: And in `fluent_container.py`, instead of 
```
host_mount_path = pyfluent.EXAMPLES_PATH
``` 
it should be 
```
host_mount_path = pyfluent.CONTAINER_MOUNT_PATH
```

tests/conftest.py
```diff
@@ -74,6 +75,7 @@ def run_before_each_test(
     monkeypatch: pytest.MonkeyPatch, request: pytest.FixtureRequest
 ) -> None:
     monkeypatch.setenv("PYFLUENT_TEST_NAME", request.node.name)
+    pyfluent.EXAMPLES_PATH = pyfluent.PYFLUENT_CONTAINER_MOUNT_PATH
```
raph-luc, line 78:
```suggestion
    pyfluent.CONTAINER_MOUNT_PATH = pyfluent.EXAMPLES_PATH
```

src/ansys/fluent/core/__init__.py
hpohekar: Done. Thank you.

tests/conftest.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/__init__.py
```diff
@@ -70,6 +70,7 @@ def version_info() -> str:
     appname="ansys_fluent_core", appauthor="Ansys"
 )
 EXAMPLES_PATH = os.path.join(USER_DATA_PATH, "examples")
+CONTAINER_MOUNT_PATH = os.getcwd()
```
raph-luc, line 73: One concern that I have with this approach, is that the user may import pyfluent, change working directories, and then expect to work with Fluent in the new working directory, which I would argue is what PyFluent should be supporting by default.  

I believe that would be the main argument against this approach setting `CONTAINER_MOUNT_PATH` on import, compared to just evaluating dynamically it at runtime.

src/ansys/fluent/core/launcher/fluent_container.py
```diff
@@ -186,7 +186,7 @@ def configure_container_dict(
         if file_transfer_service:
             host_mount_path = pyfluent.USER_DATA_PATH
         else:
-            host_mount_path = pyfluent.EXAMPLES_PATH
+            host_mount_path = pyfluent.CONTAINER_MOUNT_PATH
```
raph-luc, line 189: Should we, instead of using a module constant `pyfluent.CONTAINER_MOUNT_PATH`, use an environment variable called `PYFLUENT_CONTAINER_MOUNT_PATH`, as I initially suggested in https://github.com/ansys/pyfluent/pull/2874#discussion_r1622426411 ?

Something like this is what I had in mind:
```
host_mount_path = Path(os.getenv("PYFLUENT_CONTAINER_MOUNT_PATH", os.getcwd()))
```

tests/conftest.py
```diff
@@ -74,6 +75,7 @@ def run_before_each_test(
     monkeypatch: pytest.MonkeyPatch, request: pytest.FixtureRequest
 ) -> None:
     monkeypatch.setenv("PYFLUENT_TEST_NAME", request.node.name)
+    pyfluent.CONTAINER_MOUNT_PATH = pyfluent.EXAMPLES_PATH
```
raph-luc, line 78: Then here we'd set the environment variable, instead of a module constant

raph-luc (commented):
I would suggest that my suggestion in https://github.com/ansys/pyfluent/pull/2874#discussion_r1622426411 be considered again, using an environment variable instead of a module constant. See below for more details.

Sorry for all the back and forth.

src/ansys/fluent/core/__init__.py
mkundu1: Good point. We can get that working with the module variable also: set `CONTAINER_MOUNT_PATH = None` here and later `host_mount_path = pyfluent.CONTAINER_MOUNT_PATH or os.getcwd()`.

Just trying to avoid env var usage by users, unless it is the only way to do something.

hpohekar: Done. Thank you.

src/ansys/fluent/core/launcher/fluent_container.py
hpohekar: Done. Thank you.

tests/conftest.py
hpohekar: Done. Thank you.

hpohekar:
> I would suggest that my suggestion in [#2874 (comment)](https://github.com/ansys/pyfluent/pull/2874#discussion_r1622426411) be considered again, using an environment variable instead of a module constant. See below for more details.
> 
> Sorry for all the back and forth.

No issues at all. We need best solution and we can update the code each time. Thank you.

raph-luc (approved):

seanpearsonuk (approved):

PR #2886, opened 2024-06-02
fix: Add a method to set console loggers' level

mkundu1:
Please see the new FAQ entry to hide all PyFluent warnings from console.

### DIFF
--- a/src/ansys/fluent/core/logging.py
+++ b/src/ansys/fluent/core/logging.py
@@ -25,6 +25,24 @@ def root_config():
         logger.addHandler(ch)
 
 
+def set_console_loggers_level(level: Union[str, int]):
+    """Sets the level of all PyFluent loggers that write to console.
+
+    Parameters
+    ----------
+    level : str or int
+        Specified logging level to set PyFluent loggers to.
+
+    Notes
+    -----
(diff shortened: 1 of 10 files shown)


prmukherj (approved):

hpohekar (approved):

seanpearsonuk (approved):

src/ansys/fluent/core/logging.py
```diff
@@ -25,6 +25,24 @@ def root_config():
         logger.addHandler(ch)
 
 
+def set_console_loggers_level(level: Union[str, int]):
```
raph-luc, line 28: Minor suggestion, slightly clearer to me:
```suggestion
def set_console_logging_level(level: Union[str, int]):
```
We are setting the level of logging calls that output to console, the actual loggers (which propagate to console through the root pyfluent logger) will still be set to different levels

raph-luc (approved):

src/ansys/fluent/core/logging.py
```diff
@@ -25,6 +25,24 @@ def root_config():
...
+    """Sets the level of all PyFluent loggers that write to console.
```
raph-luc, line 29:
```suggestion
    """Sets the level of PyFluent logging being output to console.
```

PR #2892, opened 2024-06-03
refactor: Remove _MODE.

prmukherj:
MESHING_MODE -> MESHING
PURE_MESHING_MODE -> PURE_MESHING


seanpearsonuk (approved):

seanpearsonuk:
We ought to document this change prominently if the old version don't work any more.

prmukherj:
> We ought to document this change prominently if the old version don't work any more.

Is it a better idea to support backward compatibility?

mkundu1 (approved):

hpohekar (approved):

PR #2893, opened 2024-06-03
feat: Interrupt iterate.

prmukherj:
This will interrupt the iteration from PyFluent. 

To test this code, there is a module level variable: **SUPPORT_SOLVER_INTERRUPT**, which has to be set to **True**, since we are changing the default behavior of python in PyFleunt.


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1613,7 +1613,11 @@ def _execute_command(self, **kwds):
 
     def __call__(self, **kwds):
         """Call a command with the specified keyword arguments."""
-        return self.execute_command(**kwds)
+        try:
+            return self.execute_command(**kwds)
+        except KeyboardInterrupt:
+            if self.obj_name in ["iterate", "calculate", "dual_time_iterate"]:
```
seanpearsonuk, line 1619: This list of names is too concrete for a permanent solution.

prmukherj: We can actually change it to child names of "run_calculation".

seanpearsonuk: Which issue does that resolve?

seanpearsonuk: As discussed, we propose to check if parent has an interrupt method such that interrupt is explicitly a concept in the framework.

seanpearsonuk: @mkundu1 do you thoughts on this topic?

mkundu1: I think as `interrupt` is specific command in the settings API, it is better to call it explicitly `solver.settings.solution.run_calculation.interrupt()`. We can wrap it within a top-level method like `solver.interrupt()` (can be private).

Detecting interruptible commands through a settings API attribute is a good idea, we need to check with the settings API team.

seanpearsonuk: Thanks @mkundu1. We had touched on this and I think this is the best plan.

raph-luc: FYI @mkundu1 @seanpearsonuk there is an `interruptible?` attribute in the Settings API (and in the Rest API meshing datamodel wrapper), we make use of it in the WebUI

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1613,7 +1636,10 @@ def _execute_command(self, **kwds):
 
     def __call__(self, **kwds):
         """Call a command with the specified keyword arguments."""
-        return self.execute_command(**kwds)
+        try:
+            return self.execute_command(**kwds)
+        except KeyboardInterrupt:
+            self._interrupt()
```
mkundu1, line 1642: Here we shall check whether the command is interruptible. We won't need to try..except within the `_interrupt()` method.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1585,6 +1602,12 @@ def execute_command(self, *args, **kwds):
                 assert_type(ret, base_t._state_type)
             return ret
 
+    def _interrupt(self):
+        try:
+            self.root.solution.run_calculation.interrupt()
+        except AttributeError:
+            raise KeyboardInterrupt
```
raph-luc, line 1609: If I'm understanding this correctly, raising `KeyboardInterrupt` here doesn't seem appropriate. It should raise a more appropriate exception/error, and propagate whatever message was in the `AttributeError` that was caught

seanpearsonuk:
@prmukherj does this PR need to switch to draft while you make the discussed, additional changes?

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -169,6 +171,23 @@ def to_python_name(fluent_name: str) -> str:
...
+    if root is None:
+        root = obj
+    else:
+        root = _get_root_from_child_object(root)
+    return root
+
+
+def _interrupt(obj):
+    if pyfluent.SUPPORT_SOLVER_INTERRUPT:
+        root = _get_root_from_child_object(obj)
```
seanpearsonuk, line 185: Don't we also need to identify that the command is an "interruptable" command? - which we can determine from a hard-coded list (for now).

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -169,6 +169,15 @@ def to_python_name(fluent_name: str) -> str:
     return name
 
 
+def _get_root_from_child_object(obj):
```
prmukherj, line 172: This code snippet can be used to extract the root object from any given object.

seanpearsonuk: `_root()` could be a private `Base` method.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -2022,6 +2048,7 @@ def get_root(
         cls, _ = get_cls("", obj_info, version=version)
     root = cls()
     root.set_flproxy(flproxy)
+    root._set_interrupt(interrupt)
```
seanpearsonuk, line 2051: If we set an abstract interrupt handler then that handler can also encapsulate **any** logic therein. That means that the client is not forced to operate in terms of a static list of interruptible command paths. That's the Fluent logic, but it does automatically generalise.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -169,6 +169,15 @@ def to_python_name(fluent_name: str) -> str:
...
+    root = obj._parent
+    if root is None:
+        root = obj
+    else:
+        root = _get_root_from_child_object(root)
+    return root
+
+
 class Base:
     """Base class for settings and command objects.
```
PipKat, line 182:
```suggestion
    """Provides the base class for settings and command objects.
```
Class descriptions should start with a verb ending in "s" or "es"

src/ansys/fluent/core/solver/flobject.py

src/ansys/fluent/core/solver/flobject.py
prmukherj: Done, Thank you.

src/ansys/fluent/core/solver/flobject.py
prmukherj: Done, Thank you.

mkundu1 (approved):

src/ansys/fluent/core/solver/flobject.py
seanpearsonuk: Yes, this is what we discussed. The only tweak would be
```
self._root(self)._on_interrupt(self)
```
in the `except` block, which is more accurate IMHO.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -195,10 +195,20 @@ def __init__(self, name: Optional[str] = None, parent=None):
...
+        else:
+            return self._root(obj._parent)
+
     def set_flproxy(self, flproxy):
         """Set flproxy object."""
         self._setattr("_flproxy", flproxy)
 
+    def _set_interrupt(self, interrupt):
+        """Set interrupt method."""
+        self._setattr("_interrupt", interrupt)
```
seanpearsonuk, line 210:
```suggestion
    def _set_on_interrupt(self, on_interrupt):
        """Set interrupt method."""
        self._setattr("_on_interrupt", on_interrupt)
```

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1613,7 +1623,11 @@ def _execute_command(self, **kwds):
 
     def __call__(self, **kwds):
         """Call a command with the specified keyword arguments."""
-        return self.execute_command(**kwds)
+        try:
+            return self.execute_command(**kwds)
+        except KeyboardInterrupt:
+            self._root(self)._interrupt(self)
```
seanpearsonuk, line 1629:
```suggestion
            self._root(self)._on_interrupt(self)
```

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1640,7 +1654,11 @@ def _execute_command(self, *args, **kwds):
 
     def __call__(self, *args, **kwds):
         """Call a command with the specified keyword arguments."""
-        return self.execute_command(*args, **kwds)
+        try:
+            return self.execute_command(*args, **kwds)
+        except KeyboardInterrupt:
+            self._root(self)._interrupt(self)
```
seanpearsonuk, line 1660:
```suggestion
            self._root(self)._on_interrupt(self)
```

seanpearsonuk (approved):

PR #2913, opened 2024-06-05
refactor: Include 'version' and 'precision' into enums.

prmukherj:
In the launcher, make 'version' and 'precision' as enums.


tests/util/fixture_fluent.py
```diff
@@ -109,15 +109,17 @@ def launch_fluent_solver_2ddp():
         container_dict = {"host_mount_path": pyfluent.USER_DATA_PATH}
         file_transfer_service = RemoteFileTransferStrategy()
         solver_session = pyfluent.launch_fluent(
-            version="2d",
-            precision="double",
+            version=pyfluent.SolverVersion.TWO_DIMENSIONAL,
```
seanpearsonuk, line 112: The Fluent launcher calls it _Dimension_. _Version_ is a heavily overloaded term so this can get confusing and error-prone for users.

How about:
```
solver_2d = pyfluent.launch_fluent(dimension=pyfluent.DIMENSION.TWO)
solver_3d = pyfluent.launch_fluent(dimension=pyfluent.DIMENSION.THREE)
```
Meanwhile, we could support `version` for backward compatibility.

prmukherj: Absolutely agree. Will update it soon.

seanpearsonuk: N.b. `ONE` and `TWO` without any superfluous suffixes.

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -61,8 +63,8 @@ def __init__(
             FluentWindowsGraphicsDriver, FluentLinuxGraphicsDriver, str, None
         ] = None,
         product_version: Optional[FluentVersion] = None,
-        version: Optional[str] = None,
-        precision: Optional[str] = None,
+        dimension: Optional[Union[Dimension, str, None]] = None,
```
mkundu1, line 66: `Union[Dimension, str, None]` is sufficient.

tests/test_launcher.py
```diff
@@ -39,15 +39,15 @@ def test_gpu_version_error():
...
+            dimension="two",
             precision="single",
             processor_count=5,
             ui_mode="gui",
             gpu=True,
         )
         pyfluent.setup_for_fluent(
             mode="meshing",
-            version="2d",
+            dimension="2d",
```
prmukherj, line 50:
```suggestion
            dimension="two",
```

tests/util/fixture_fluent.py
```diff
@@ -129,7 +131,7 @@ def launch_fluent_solver_2ddp_t2():
         container_dict = {"host_mount_path": pyfluent.USER_DATA_PATH}
         file_transfer_service = RemoteFileTransferStrategy()
         solver_session = pyfluent.launch_fluent(
-            version="2d",
+            dimension="2d",
```
prmukherj, line 134:
```suggestion
            dimension="two",
```

tests/util/fixture_fluent.py
```diff
@@ -138,7 +140,7 @@ def launch_fluent_solver_2ddp_t2():
         )
     else:
         solver_session = pyfluent.launch_fluent(
-            version="2d", precision="double", processor_count=2, mode="solver"
+            dimension="2d", precision="double", processor_count=2, mode="solver"
```
prmukherj, line 143:
```suggestion
            dimension="two", precision="double", processor_count=2, mode="solver"
```

src/ansys/fluent/core/launcher/container_launcher.py
```diff
@@ -80,12 +82,14 @@ def __init__(
         product_version : FluentVersion, optional
             Version of Ansys Fluent to launch. Use ``FluentVersion.v241`` for 2024 R1.
             The default is ``None``, in which case the newest installed version is used.
-        version : str, optional
+        dimension : Dimension or str, optional
             Geometric dimensionality of the Fluent simulation. The default is ``None``,
-            in which case ``"3d"`` is used. Options are ``"3d"`` and ``"2d"``.
-        precision : str, optional
+            in which case ``"three"`` is used. Options are either the values of the ``Dimension``
+            enum or any of ``"three"`` and ``"two"``.
```
seanpearsonuk, line 88: I'm wondering why we're saying "three" now for the string counterpart. Is it relevant - isn't it still "3d"? I also thing thath the documentation should lead with the enum and not the string version. The enum is the _proper_ input.

prmukherj: It should be either "two" and "three" or the enums.

src/ansys/fluent/core/launcher/container_launcher.py
```diff
@@ -80,12 +82,14 @@ def __init__(
...
+        precision : Precision or str, optional
             Floating point precision. The default is ``None``, in which case ``"double"``
-            is used. Options are ``"double"`` and ``"single"``.
+            is used. Options are either the values of the ``Precision`` enum or any
+            of ``"double"`` and ``"single"``.
```
seanpearsonuk, line 92: I would make a similar point here. The most correct usage is `Precision.DOUBLE` etc. I would much prefer to see it documented that way.

prmukherj: ok, I agree. Will update the doc accordingly.
Thanks.

src/ansys/fluent/core/launcher/container_launcher.py
prmukherj: I think it makes more sense as to dimension="two" rather than dimension="2d", right?

src/ansys/fluent/core/launcher/process_launch_string.py
```diff
@@ -26,6 +31,20 @@ def _build_fluent_launch_args_string(**kwargs) -> str:
...
+    if dimension is None:
+        launch_args_string += f" {Dimension.THREE.value[0]}"
+    elif isinstance(dimension, str):
+        launch_args_string += f" {Dimension(dimension).value[0]}"
+    else:
+        launch_args_string += f" {dimension.value[0]}"
+    precision = kwargs.get("precision")
+    if precision is None:
+        launch_args_string += f"{Precision.DOUBLE.value[0]}"
+    elif isinstance(precision, str):
```
mkundu1, line 44: This branch is not required:
```
>>> assert pyfluent.Precision("single") == pyfluent.Precision.SINGLE
```

src/ansys/fluent/core/launcher/container_launcher.py
mkundu1: Ideally, the type should be enum or int.

prmukherj: Done, Thank you.
Now it supports enum and int.

src/ansys/fluent/core/launcher/container_launcher.py
prmukherj: Done, Thank you.

mkundu1 (approved):

seanpearsonuk (approved):

PR #2922, opened 2024-06-06
test: Reduce memory/time usage of some tests

mkundu1:
Test-steps that take more than 1 min currently:
```
196.26s setup    tests/parametric/test_parametric_workflow.py::test_parameters_list_function
177.29s setup    tests/test_data_model_cache.py::test_get_cached_values_in_command_arguments
112.47s call     tests/test_solution_variables.py::test_solution_variables_single_precision
100.88s call     tests/test_solution_variables.py::test_svars_single_precision
98.40s call     tests/test_fluent_fixes.py::test_monitors_list_set_data_637_974_1744_2188
98.30s call     tests/test_flobject.py::test_ansys_units_integration
68.71s call     tests/test_datamodel_service.py::test_add_on_affected
```
Reducing time usage for some of the above.

### DIFF
--- a/tests/test_solution_variables.py
+++ b/tests/test_solution_variables.py
@@ -15,14 +15,14 @@ def test_solution_variables(new_solver_session):
     import_file_name = examples.download_file(
         "mixing_elbow.cas.h5", "pyfluent/mixing_elbow"
     )
+    examples.download_file("mixing_elbow.dat.h5", "pyfluent/mixing_elbow")
 
     solution_variable_info = solver.fields.solution_variable_info
     solution_variable_data = solver.fields.solution_variable_data
+    assert solution_variable_info == solver.svar_info
+    assert solution_variable_data == solver.svar_data
 
-    solver.file.read(file_type="case", file_name=import_file_name)
-
-    solver.solution.initialization.hybrid_initialize()
-    solver.solution.run_calculation.iterate(iter_count=10)
+    solver.file.read_case_data(file_name=import_file_name)
 
     zones_info = solution_variable_info.get_zones_info()
 
@@ -54,13 +54,14 @@ def test_solution_variables(new_solver_session):
     )
 
     assert set(wall_fluid_info.solution_variables) == {
+        "SV_ADS_0",
         "SV_ADS_1",
         "SV_CENTROID",
         "SV_H",
         "SV_K",
         "SV_O",
-        "SV_T",
         "SV_P",
+        "SV_T",
         "SV_U",
         "SV_V",
         "SV_W",
@@ -131,6 +132,8 @@ def test_solution_variables_single_precision(new_solver_session_single_precision
 
     solution_variable_info = solver.fields.solution_variable_info
     solution_variable_data = solver.fields.solution_variable_data
+    assert solution_variable_info == solver.svar_info
+    assert solution_variable_data == solver.svar_data
 
     solver.file.read_case_data(file_name=import_file_name)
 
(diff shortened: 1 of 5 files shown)


tests/test_solution_variables.py
```diff
@@ -125,100 +126,19 @@ def test_solution_variables(new_solver_session):
...
-
-    assert sv_p_wall_fluid.zones == ["wall_tank", "tank"]
-
-    fluid_temp = sv_p_wall_fluid["tank"]
-    assert fluid_temp.size == 183424
-    assert str(fluid_temp.dtype) == "float32"
-
-
-@pytest.mark.fluent_version(">=23.2")
-def test_svars(new_solver_session):
```
mkundu1, line 206: Backward compatibility tests have been replaced with the object identity assertions:
```
assert solution_variable_info == solver.svar_info
assert solution_variable_data == solver.svar_data
```

tests/test_solution_variables.py
```diff
@@ -125,100 +126,19 @@ def test_solution_variables(new_solver_session):
 def test_solution_variables_single_precision(new_solver_session_single_precision):
     solver = new_solver_session_single_precision
     import_file_name = examples.download_file(
-        "vortex_init.cas.h5", "pyfluent/examples/Steady-Vortex-VOF"
+        "mixing_elbow.cas.h5", "pyfluent/mixing_elbow"
```
mkundu1, line 129: The mixing_elbow case consumes much less memory

tests/test_solution_variables.py
```diff
@@ -15,14 +15,14 @@ def test_solution_variables(new_solver_session):
...
 
     solution_variable_info = solver.fields.solution_variable_info
     solution_variable_data = solver.fields.solution_variable_data
+    assert solution_variable_info == solver.svar_info
+    assert solution_variable_data == solver.svar_data
 
-    solver.file.read(file_type="case", file_name=import_file_name)
-
-    solver.solution.initialization.hybrid_initialize()
-    solver.solution.run_calculation.iterate(iter_count=10)
```
raph-luc, line 25: Probably already considered, but are there other tests that still run some calculations? I don't think all tests need to run calculations, but certainly at least some of them should still do that, to cover this scenario

tests/test_solution_variables.py
```diff
@@ -278,116 +199,6 @@ def test_svars(new_solver_session):
...
-
-    assert updated_sv_p_data["elbow-fluid"].size == 17822
-    assert str(updated_sv_p_data["elbow-fluid"].dtype) == "float64"
-
-    assert updated_sv_p_data["wall-elbow"][0] == 500.0
-    assert updated_sv_p_data["elbow-fluid"][-1] == 600.0
-
-
-@pytest.mark.fluent_version(">=23.2")
-def test_svars_single_precision(new_solver_session_single_precision):
```
raph-luc, line 314: At first sight, it seems that this `test_svars_single_precision` and `test_solution_variables_single_precision` were basically duplicates of each other, no? Wonder why both tests were running

`test_svars` and `test_solution_variables` too

raph-luc (approved):

tests/test_solution_variables.py
mkundu1: Some tests like `test_field_data` compare results, so specified iterations are necessary (we can think of some alternate strategy like results change than results comparison, however the test is relative fast as of now). Some nightly/weekly tests runs iterations, I'm optimizing the PR tests only for now.

hpohekar (approved):

PR #2935, opened 2024-06-10
ci: Print and stop stale containers after run

mkundu1:


doc/source/conf.py
```diff
@@ -135,6 +135,7 @@ def _stop_fluent_container(gallery_conf, fname):
             .split()
         )
         for container_name in container_names:
+            print(f"Container still running for script {fname}")
```
mkundu1, line 138: Prints example script that leaves stale containers

Makefile
```diff
@@ -151,5 +151,8 @@ compare-flobject:
 	@python .ci/compare_flobject.py
 
 cleanup-previous-docker-containers:
-	@if [ -n "$(docker ps -a -q)" ]; then docker stop $(docker ps -a -q); fi
+	@if [ -n "$(docker ps -a -q)" ]; then \
+		docker inspect --format='{{.Config.Labels.test_name}}' $(docker ps -a -q); \
```
mkundu1, line 155: Prints unittest that leaves stale containers

hpohekar (approved):

doc/source/conf.py
hpohekar: I think in some test cases we are not exiting sessions explicitly by using `session.exit()`, because for similar kind of issue we have updated `test_fluent_containes` where we have called `exit()` explicitly.

Makefile
hpohekar: This will help to identify tests which require explicit `exit()` call as mentioned in above comment.

doc/source/conf.py
hpohekar: Sure.

prmukherj (approved):

.github/workflows/ci.yml
```diff
@@ -531,6 +525,9 @@ jobs:
         env:
           FLUENT_IMAGE_TAG: ${{ matrix.image-tag }}
 
+      - name: Cleanup previous docker containers
```
PipKat, line 528: Nitpick but "Clean up" is the verb (two words), while cleanup is the noun (1 word).

doc/source/conf.py
PipKat, line 138:
```suggestion
            print(f"Container still running for script {fname}".)
```
Message strings should have concluding punctuation.

Issue #2921, opened 2024-06-06
EventsManager examples needed in user guide


seanpearsonuk:
EventsManager examples needed in user guide

Note that there is one example here: https://fluent.docs.pyansys.com/version/stable/api/solver/events.html

PR #2946, opened 2024-06-14
docs: Add events manager example to user guide section and add comments in examples scripts

hpohekar:
closes https://github.com/ansys/pyfluent/issues/2921, https://github.com/ansys/pyfluent/issues/2949

Add events manager example to user guide section. We have added single example for now.

Add following comments - 
        1. Do not set pyfluent.pyfluent.CONTAINER_MOUNT_PATH and
         2. Execute <session>.upload() wherever necessary. (Thanks a lot @answillgm for reporting this.)

image

### DIFF
--- a/doc/source/user_guide/events_manager.rst
+++ b/doc/source/user_guide/events_manager.rst
@@ -1,0 +1,13 @@
+.. _ref_user_guide_events_manager:
+
+Use events manager
+==================
+
+The following code triggers a callback at the end of every iteration.
+
+.. code-block:: python
+
+    def callback_executed_at_end_of_iteration(session_id, event_info):
+        print("Iteration ended index", event_info.index)
+
+    cb_itr_id = session.events_manager.register_callback('IterationEndedEvent', callback_executed_at_end_of_iteration)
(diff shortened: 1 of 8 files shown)


prmukherj (approved):

seanpearsonuk:
This events manager documentation added in the user guide duplicates what is in the API reference. I am happy that it is _moved_ to the user guide as this is inherently user guide _guidance_ whereas the API reference should only contain information on classes, functions, etc. So, we would remove the origianl page from the API reference. (I have already been looking at this in current work.)

Since this PR would reduce to a _move_ that we would also apply to other services, and there is already significant user guide rework happening in a separate branch, I advocate that we create PRs relative to the other baseline branch instead. Please hold off on that until we have had more discussion.

The issue that was raised leading to this PR, requested _additional_ example, which we have not provided at this time. However, I would assert that the current example is perfectly adequate.

Please close this PR, pending further discussion and collaboration on the reorganization of the documentation. Thank you.

seanpearsonuk:
This PR also provoked me to create [a new issue](https://github.com/ansys/pyfluent/issues/2953).

PR #2963, opened 2024-06-21
feat: Create as command in Fluent 25.1

mkundu1:
`create()` is available as a command in Fluent 25.1. This PR contains PyFluent side change.

`CreatableNamedObjectMixinOld` - mixin with create method for Fluent < 25.1.
`CreatableNamedObjectMixin` - mixin without create method for Fluent >= 25.1 as create is available as a command.

Please see the changed test_wildcard_fnmatch for the currently supported usage (backward compatibility is provided).

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -1764,7 +1764,7 @@ class CreatableNamedObjectMixin(collections.abc.MutableMapping, Generic[ChildTyp
 
 
 class CreatableNamedObjectMixinOld(CreatableNamedObjectMixin):
-    """Provides creatable named objects for Fluent 24.2 and earlier."""
+    """Provides creatable named objects for Fluent 2024 R2 and earlier."""
 
     # In Fluent 2025 R1, the ``create()`` method is available as commands in the ``NamedObject`` class.
     def create(self, name: str = "") -> ChildTypeT:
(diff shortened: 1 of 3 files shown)


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1747,8 +1747,22 @@ def __len__(self):
 
 
 class CreatableNamedObjectMixin(collections.abc.MutableMapping, Generic[ChildTypeT]):
-    """Provides creatable named objects."""
+    """Provides creatable named objects for Fluent 25.1 and later."""
```
PipKat, line 1750: For our user-facing docs, we are supposed to use this version convention: Fluent 2025 R1.

seanpearsonuk (approved):

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1749,8 +1751,22 @@ def __len__(self):
...
+        child = self._objects.get(name)
+        if not child:
+            child = self._create_child_object(name)
+        child.set_state(value)
+
+
+class CreatableNamedObjectMixinOld(CreatableNamedObjectMixin):
+    """Provides creatable named objects for Fluent 24.2 and earlier."""
+
+    # In Fluent 25.1, create method is available as commands in NamedObject class.
```
PipKat, line 1769:
```suggestion
    """Provides creatable named objects for Fluent 2024 R2 and earlier."""

    # In Fluent 2025 R1, the ``create()`` method is available as commands in the ``NamedObject`` class.
```
User-facing doc should format Ansys versions like this: 2024 R2. In the comment, the edits to other than the release use the double backticks to surround code entites.

mkundu1: Actually, this is a code comment. It won't be visible in the user-facing doc.

PipKat: True, but the previous class description will be. It is best to keep the terminology consistent!

prmukherj (approved):

Issue #2949, opened 2024-06-14
Remove pyfluent.CONTAINER_MOUNT_PATH from example scripts and add `<session>.upload()` in comment.


hpohekar:
@answillgm 

1. Handle `host_mount_path` based on environment variable in CI.
2. Add comments specific to file transfer service. (Execute `<session>.upload()` wherever necessary.)
3. Remove `pyfluent.CONTAINER_MOUNT_PATH = pyfluent.EXAMPLES_PATH`

PR #2966, opened 2024-06-24
fix: Remove pyfluent.CONTAINER_MOUNT_PATH from example scripts and add <session>.upload().

hpohekar:
closes https://github.com/ansys/pyfluent/issues/2949

1. Handle `host_mount_path` based on the environment variable in CI.
2. Add comments specific to file transfer service.
3. Remove `pyfluent.CONTAINER_MOUNT_PATH = pyfluent.EXAMPLES_PATH`
4. Add `meshing.upload()` wherever it is necessary.


examples/00-fluent/exhaust_system.py
```diff
@@ -48,6 +48,8 @@
 import ansys.fluent.core as pyfluent
 from ansys.fluent.core import examples
 
+# Do not set pyfluent.CONTAINER_MOUNT_PATH. It has set to run this example in GitHub container only.
```
seanpearsonuk, line 51: @hpohekar these statements need to be written more clearly.

hpohekar: @seanpearsonuk 

We are going to make changes in CI script to remove this comment from example script. (discussed with @mkundu1)

examples/00-fluent/exhaust_system.py
```diff
@@ -82,6 +81,7 @@
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 # Import the CAD geometry file (``exhaust_system.fmd``) and selectively manage some
 # parts.
+# Execute meshing.upload(import_file_name) if Fluent is running in Ansys Lab.
```
seanpearsonuk, line 84: Is the precise condition for this and other similar scenarios not something like _if Fluent is running remotely and the file is not already available to it at the relevant path_?

The way we could structure it to make it unambiguous and easy for users is:
```Python
# Uncomment the following line of code if Fluent is running remotely and the file specified by 
# import_file_name is not available to it at the relevant path
# meshing.upload(import_file_name)
```

hpohekar: @seanpearsonuk 
No, it's case when the file exists on host machine or Ansys Lab's user directory but not uploaded to the server using PIM service. 

We are already handling this on the backend for most of the cases but `Execute()` or some `PartManagement` methods do not upload required files seamlessly therefore we are adding this comment.

hpohekar: We can update it as follows - 

```
# Uncomment the following line of code if Fluent is running in Ansys Lab 
# meshing.upload(import_file_name)
```

seanpearsonuk: In my original version I deliberately left out the space between `#` and `meshing`. That makes it "impossible" (within reason) to implement uncommenting incorrectly. Taken literally _uncomment the line_ means remove the comment character and leave everything else in tact, hence no space.  

I still don't understand the distinction between _Lab_ and any other client-server set-up. Let's discuss next week (scrum).

hpohekar: @seanpearsonuk 

`# meshing.upload(import_file_name)` and `#meshing.upload(import_file_name)` behaves same after uncommenting, if we use `Ctrl + ?` on selected line of code.

I had considered that user will use an editor shortcut to uncomment the above line of code.

Pycharm

image

VS code

image

hpohekar: Sure. We will discuss above point in next scrum.

seanpearsonuk: > @seanpearsonuk
> 
> `# meshing.upload(import_file_name)` and `#meshing.upload(import_file_name)` behave same after uncommenting, if we use `Ctrl + ?` on selected line of code.
> 
> I had considered that user will use an editor shortcut to uncomment the above line of code.
> 
> Pycharm
> 
> image
> 
> VS code
> 
> image

It's up to the user how he's going to do it. The point is to remove the space from the original script so that it's **definitely** not there on uncommenting. As mentioned before, uncommenting means remove only the comment character which leaves the Python invalid (`IndentationError`).

seanpearsonuk (approved):

prmukherj (approved):

PR #2970, opened 2024-06-25
refactor: update error message for Dimension enum.

prmukherj:
image

@seanpearsonuk, we can have a discussion whether we want to support "2" or "2d" as well. I think that dimension=2 (as an integer) or enum is clean and without any confusion. version="2d" is still supported, with a deprecation warning. dimension="2d" seems a bit redundant. dimension="2" as string can be supported but with a warning??
What do you think?

### DIFF
--- a/tests/test_cad_to_post_wtm.py
+++ b/tests/test_cad_to_post_wtm.py
@@ -5,7 +5,7 @@ This test covers the setup and solution of a three-dimensional
 turbulent fluid flow and heat transfer problem in a mixing elbow. The mixing
 elbow configuration is encountered in piping systems in power plants and
 process industries. It is often important to predict the flow field and
-temperature field in the area of the mixing region in order to properly design
+temperature field in the area of the mixing region to properly design
 the junction.
 
 This test queries the following using PyTest:
--- a/tests/test_cad_to_post_wtm.py
+++ b/tests/test_cad_to_post_wtm.py
@@ -4,8 +4,8 @@ End-to-end Fluent Solver Workflow using Watertight Meshing
 This test covers the setup and solution of a three-dimensional
 turbulent fluid flow and heat transfer problem in a mixing elbow. The mixing
 elbow configuration is encountered in piping systems in power plants and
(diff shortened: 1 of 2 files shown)


seanpearsonuk:
> image
> 
> @seanpearsonuk, we can have a discussion whether we want to support "2" or "2d" as well. I think that dimension=2 (as an integer) or enum is clean and without any confusion. version="2d" is still supported, with a deprecation warning. dimension="2d" seems a bit redundant. dimension="2" as string can be supported but with a warning?? What do you think?

@prmukherj I (already) agree with you. This came about following those internal discussions yesterday.

seanpearsonuk (approved):

tests/test_cad_to_post_wtm.py
```diff
@@ -4,8 +4,8 @@
 This test covers the setup and solution of a three-dimensional
 turbulent fluid flow and heat transfer problem in a mixing elbow. The mixing
 elbow configuration is encountered in piping systems in power plants and
-processindustries. It is often important to predict the flow field and
-temperature field in the area of the mixing regionin order to properly design
+process industries. It is often important to predict the flow field and
+temperature field in the area of the mixing region in order to properly design
```
PipKat, line 8:
```suggestion
temperature field in the area of the mixing region to properly design
```
The "in order" part is unnecessary. It just adds extra words.

seanpearsonuk: @prmukherj Yes, we can remove "in order" here.

mkundu1 (approved):

tests/test_cad_to_post_wtm.py
```diff
@@ -4,8 +4,8 @@
 This test covers the setup and solution of a three-dimensional
```
PipKat, line 4: Generally, we use 3D instead of writing out three-dimensional (given that our audience is well aware of the term).

prmukherj, line 4:
```suggestion
This test covers the setup and solution of a 3D
```

PR #2975, opened 2024-06-27
fix: name for compound child tasks

prmukherj:
Compound child task_names() should have a format of the parent name followed by child number. Please see the test for details.

### DIFF
--- a/tests/test_new_meshing_workflow.py
+++ b/tests/test_new_meshing_workflow.py
@@ -727,9 +727,7 @@ def test_watertight_workflow_children(mixing_elbow_geometry, new_mesh_session):
     describe_geometry = watertight.describe_geometry
     describe_geometry_children = describe_geometry.tasks()
     assert len(describe_geometry_children) == 2
-    describe_geometry_child_task_python_names = (
-        describe_geometry.child_task_python_names()
-    )
+    describe_geometry_child_task_python_names = describe_geometry.task_names()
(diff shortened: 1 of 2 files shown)


tests/test_new_meshing_workflow.py
```diff
@@ -1507,3 +1505,47 @@ def test_switching_workflow_interface(new_mesh_session):
...
+    watertight.add_local_sizing.add_child_and_update(
+        state={
+            "boi_control_name": "outlet",
+            "boi_face_label_list": ["outlet"],
+            "boi_size": 10,
+        }
+    )
+
+    assert {
+        "child_1_of_add_local_sizing",
```
seanpearsonuk, line 1539: Is there a reason for preferring child_n_of_ prefix over _child_n suffix?

prmukherj: No, is the latter better? We can update it.

seanpearsonuk: Does the _of_ part look a bit awkward in an object name? It's also a bit longer.

prmukherj: I agree. Will update the same.

seanpearsonuk (approved):

mkundu1 (approved):

hpohekar (approved):

PR #2977, opened 2024-06-28
refactor: field_data

prmukherj:
The field data interface is undergoing some changes to bring about uniformity. As "new_transaction" will be supported in PyConsole going forward we need to maintain uniformity of interfaces between directly getting and adding field data to transaction.

The documentation will also be updated with this PR.

The goal here is to support the earlier behavior as well with deprecation warnings so as not to break any existing scripts using this interface.


prmukherj:
@ajain-work, could you please have a look if anything more is required. Thank you.

prmukherj:
> @ajain-work, could you please have a look if anything more is required. Thank you.

One thing that needs to be done is updating tests as per the updated arguments. That will be done as part of a trailing PR as this PR will make sure that the earlier arguments work fine as well.

src/ansys/fluent/core/services/field_data.py
```diff
@@ -222,13 +224,23 @@ class FieldUnavailable(RuntimeError):
...
-    FacesConnectivity = 2
-    FacesNormal = 3
-    FacesCentroid = 4
+    Vertices = "vertices"
+    FacesConnectivity = "faces"
+    FacesNormal = "face-normal"
+    FacesCentroid = "centroid"
+
+    @classmethod
+    def _missing_(cls, value: str):
```
mkundu1, line 236: This method is same as `FluentEnum.__missing__`, so class can be derived from `FluentEnum` and the duplicate can be removed.

src/ansys/fluent/core/services/field_data.py
```diff
@@ -1221,12 +1234,56 @@ def get_scalar_field_data(
...
+        old_arg="surface_name",
+        new_arg="surface_names",
+        converter=lambda old_arg_val: [old_arg_val] if old_arg_val else None,
+        deprecation_class=PyFluentDeprecationWarning,
+    )
+    @deprecate_argument(
+        old_arg="data_type",
+        new_arg="provide_vertices",
+        converter=lambda old_arg_val: (
+            True if old_arg_val == SurfaceDataType.Vertices else False
```
mkundu1, line 1247: just `lambda old_arg_val: old_arg_val == SurfaceDataType.Vertices` will be sufficient

src/ansys/fluent/core/services/field_data.py
prmukherj: Thank you @mkundu1, I agree. But here we will not need the missing method, so I was planning to remove it altogether only.
Thank you.

tests/test_file_session.py
mkundu1, line 1: If it is not already tested elsewhere, you can add `assert file_session.fields.field_data == file_session.field_data` etc. to cover the backward compatibility of new properties.

prmukherj: ok, I'll add these tests. Thank you.

src/ansys/fluent/core/services/field_data.py
```diff
@@ -1221,12 +1234,56 @@ def get_scalar_field_data(
...
+        old_arg="data_type",
+        new_arg="provide_faces_normal",
+        converter=lambda old_arg_val: (
+            True if old_arg_val == SurfaceDataType.FacesNormal else False
+        ),
+        deprecation_class=PyFluentDeprecationWarning,
+    )
     def get_surface_data(
         self,
-        data_type: SurfaceDataType,
```
mkundu1, line 1226: IMO the older interface looks better. Why do we need to change it?

mkundu1: Note that, there is a [Flag](https://docs.python.org/3/library/enum.html#enum.Flag) class to compose Enum values if that is what we are looking for here.

prmukherj: If this interface is better, then we need to have "data_types" here as well as in add_suface_request. I'll  do that change then.
Thank you.

prmukherj: I'll discuss with you this point.

src/ansys/fluent/core/utils/deprecate_args.py
```diff
@@ -58,3 +63,53 @@ def wrapper(*args, **kwargs):
...
+    deprecated arguments with the appropriate new arguments."""
+
+    def decorator(func):
+        """Holds the original method to perform operations on it."""
+
+        def _convert_list_to_str(args):
+            arg_list = ""
+            for arg in args:
+                arg_list += arg + ", "
+            return arg_list[:-2]
```
seanpearsonuk, line 84: `', '.join(args)` is equivalent.

src/ansys/fluent/core/utils/deprecate_args.py
```diff
@@ -58,3 +63,53 @@ def wrapper(*args, **kwargs):
...
+
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            """Warns about the deprecated arguments and replaces them with the new
+            arguments."""
+            old_arg_list = _convert_list_to_str(old_args)
+            new_arg_list = _convert_list_to_str(new_args)
+            warnings.warn(
+                f"'{old_arg_list}' are deprecated. Use '{new_arg_list}' instead.",
```
seanpearsonuk, line 93:
```suggestion
                f"The argument list, '{old_arg_list}' is deprecated. Use '{new_arg_list}' instead.",
```

src/ansys/fluent/core/services/field_data.py
```diff
@@ -354,6 +356,19 @@ def __call__(self, *args, **kwargs):
         return self._field_data_accessor(*args, **kwargs)
 
 
+def _surface_converter(old_args_dict, new_args):
+    surfaces_list = []
+    if "surface_name" in old_args_dict:
+        surfaces_list.append(old_args_dict["surface_name"])
```
seanpearsonuk, line 362: `surface_name` is mentioned here but not in the old interface unless this function is being used in different contexts (but don't do that).

src/ansys/fluent/core/services/field_data.py
```diff
@@ -415,35 +430,53 @@ def __init__(
...
+            d_type_list.append(SurfaceDataType.FacesConnectivity)
+        if old_args_dict.get("provide_faces_centroid"):
+            d_type_list.append(SurfaceDataType.FacesCentroid)
+        if old_args_dict.get("provide_faces_normal"):
+            d_type_list.append(SurfaceDataType.FacesNormal)
+        return {new_args[0]: d_type_list}
+
+    @deprecate_arguments(
+        old_args=["surface_names", "surface_ids"],
+        new_args=["surfaces"],
```
seanpearsonuk, line 448: If the converters are acting on the entire dictionaries then specifying these lists is redundant, which is evident in the converter implementations.

src/ansys/fluent/core/utils/deprecate_args.py
```diff
@@ -2,22 +2,27 @@
...
+    new_arg,
+    converter,
+    deprecation_class=PyFluentDeprecationWarning,
 ):
     """Warns that the argument provided is deprecated and automatically replaces the
     deprecated argument with the appropriate new argument."""
 
     def _str_repr(var):
         """Converts a string or FluentEnum variable to quoted string representation."""
+        from ansys.fluent.core.launcher.pyfluent_enums import Dimension, FluentEnum
```
seanpearsonuk, line 24: Please can we have a solution without mentioning these concrete types. Provide a way to use those enums generically if necessary.

src/ansys/fluent/core/utils/deprecate_args.py
```diff
@@ -58,3 +63,53 @@ def wrapper(*args, **kwargs):
...
+                        f" only '{key} = {kwargs[key]}' applies."
+                    )
+
+            kwargs = {**new_args_dict, **kwargs}
+
+            return func(*args, **kwargs)
+
+        return wrapper
+
+    return decorator
```
seanpearsonuk, line 115: Perhaps there are some design options here, but this one seems relatively complicated. Essentially, you want to pass the provided argument dictionary to the converter to process. If anything changes between the input and ouput, then warn.

src/ansys/fluent/core/services/field_data.py
```diff
@@ -9,17 +9,19 @@
...
-)
+from ansys.fluent.core.exceptions import DisallowedValuesError
 from ansys.fluent.core.services.interceptors import (
     BatchInterceptor,
     ErrorStateInterceptor,
     GrpcErrorInterceptor,
     TracingInterceptor,
 )
 from ansys.fluent.core.services.streaming import StreamingService
+from ansys.fluent.core.utils.deprecate_args import (
```
seanpearsonuk, line 20: We can drop the args suffix from the module name.

src/ansys/fluent/core/utils/deprecate_args.py
prmukherj: Thank you, Done.

src/ansys/fluent/core/utils/deprecate_args.py
```diff
@@ -58,3 +63,53 @@ def wrapper(*args, **kwargs):
         return wrapper
 
     return decorator
+
+
+def deprecate_arguments(
+    old_args: List,
+    new_args: List,
+    converter,
+    deprecation_class=PyFluentDeprecationWarning,
```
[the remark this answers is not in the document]
prmukherj: Done, Thank you.

src/ansys/fluent/core/services/field_data.py
prmukherj: This is actually being used by both field_data and field_transaction. That's why a common method.

src/ansys/fluent/core/utils/deprecate_args.py
prmukherj: Done, Thank you.

src/ansys/fluent/core/utils/deprecate_args.py
prmukherj: Done. Thank you.

seanpearsonuk (approved):

hpohekar (approved):

mkundu1 (approved):

PR #3009, opened 2024-07-04
fix: update docker requests versions

hpohekar:


mkundu1:
Abandoning this as we have the other PR

PR #3034, opened 2024-07-09
fix: Classic Meshing Workflow.

prmukherj:
Tested both in PyFluent and PyConsole.

### DIFF
--- a/src/ansys/fluent/core/workflow.py
+++ b/src/ansys/fluent/core/workflow.py
@@ -380,17 +380,20 @@ class BaseTask:
         return self._python_name
 
     def _set_python_name(self, compound=False):
-        this_command = self._command()
-        if compound:
-            p_name = (
-                self._command_source._compound_parent_task_python_name_id[0]
-                + f"_child_{self._command_source._compound_parent_task_python_name_id[1]}"
-            )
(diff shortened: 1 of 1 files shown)


hpohekar (approved):

mkundu1 (approved):

src/ansys/fluent/core/workflow.py
```diff
@@ -380,17 +380,20 @@ def python_name(self) -> str:
...
+                    self._command_source._compound_parent_task_python_name_id[0]
+                    + f"_child_{self._command_source._compound_parent_task_python_name_id[1]}"
+                )
+                self._python_name = p_name
+                self._command_source._compound_task_map[self.name()] = p_name
+            else:
+                self._python_name = camel_to_snake_case(
+                    this_command.get_attr("helpString")
+                )
+            self._cache_data(this_command)
```
seanpearsonuk, line 396: @prmukherj:

Subsequently, we shall eliminate this code also:

image

because it is a bug.

prmukherj: ok, sure. I'll investigate into this.

seanpearsonuk (approved):

PR #3041, opened 2024-07-09
feat: Update datamodel cache from all mutating rpc response

mkundu1:


prmukherj (approved):

hpohekar (approved):

seanpearsonuk (approved):

tests/test_data_model_cache.py
```diff
@@ -45,7 +45,7 @@ def test_data_model_cache():
         ({"r1": {}}, "r1", {"A": [3.0, 6.0]}, [], {"r1": {"A": [3.0, 6.0]}}),
         ({"r1": {}}, "r1", {"A": ["ab", "cd"]}, [], {"r1": {"A": ["ab", "cd"]}}),
         ({"r1": {"A": {}}}, "r1", {"A": {"B": 5}}, [], {"r1": {"A": {"B": 5}}}),
-        ({"r1": {"A": 5}}, "r1", {"A": {}}, [], {"r1": {"A": 5}}),
+        ({"r1": {"A": 5}}, "r1", {"A": {}}, [], {"r1": {"A": {}}}),
```
mkundu1, line 48: I think this is the correct behaviour and now required when named objects are deleted and re-created during a command execution. The state difference of those new named objects contains empty dicts which we need to clear in the client's cache.

seanpearsonuk: @mkundu1 

A motivating use case is as follows.
```
watertight.describe_geometry() # execution causes creation of downstream add_multizone_controls
# no sleep between the two steps
add_multizone_controls = watertight.add_multizone_controls
# use add_multizone_controls 
```
Without these changes, a sleep is needed between the two steps. The required sleep time is significantly greater if cacheing is disabled (this seems to indicate a slight flaw in the (enhanced) workflow: probably, the parent should refresh itself if it cannot find the child). With these changes, no sleep will be required at all; that's true in principle though I will be interested to try out the (enhanced) meshing workflow with the cache deactivated.

A more complete version of the test I have is on another branch (client-side branch to activate and test the new data model service):

```
wt = meshing.watertight()

import_file_name = examples.download_file(
    "mixing_elbow.pmdb", "pyfluent/mixing_elbow"
)

wt.import_geometry.file_name.set_state(import_file_name)

# trigger creation of downstream task when this task is updated:
wt.describe_geometry.multizone.set_state(True)
wt.describe_geometry()

# wait for the downstream to be "created":
add_multizone_controls = None
attr_err_count = 0

while add_multizone_controls is None and attr_err_count <= 10:
    try:
        add_multizone_controls = wt.add_multizone_controls
    except AttributeError:
        attr_err_count += 1
        time.sleep(1)

# following assertion commented out because we had to wait for the 
# stream to update the cache:
# assert attr_err_count == 0

assert add_multizone_controls
```
cc: @prmukherj

mkundu1: There is still an issue with updating cache for dict-type parameters. In the test `test_attrs_in_fault_tolerant_meshing_workflow1`, before re-initialization the client's cache contains the following state for dict-type `Arguments` of `fault_tolerant.import_cad_and_part_management`:
```
{'Arguments': {'CreateObjectPer': 'Custom', 'Context': 0, 'ObjectSetting': 'DefaultObjectSetting', 'FMDFileName': '<path-to-dir>\\exhaust_system.fmd'}}
```
On re-initialization, the following state-difference is returned:
```
{'Arguments': {'Context': 0, 'ObjectSetting': 'DefaultObjectSetting'}}
```
As the cache-updater code is unaware of the parameter type, the existing `FMDFileName` key is not cleared during the cache update.

I think to resolve this kind of issue, we need to make caching code aware of dict-type parameters. When a different state is received for dict-type parameters, it will clear existing state in the cache before applying the incoming state.

@seanpearsonuk @prmukherj

mkundu1: I've added the test suggested by Sean.

src/ansys/fluent/core/workflow.py
mkundu1, line 1: @prmukherj @seanpearsonuk Please note the changes in this file. I'm now manually refreshing the client-side tasklists after a couple of calls that modify the tasklists on the server. Listening to the on-affected event doesn't work anymore as the on-affected event for the task modification seems to be now received at a later time. We could generally remove the on-affected mechanism and immediately refresh the client-side tasklists as the modified tasklists is now immediately available in the client's cache.

Issue #3014, opened 2024-07-04
Remove TUI calls from scripts in pyfluent/example


seanpearsonuk:
A couple of examples there have (solver) TUI and non-TUI counterpart scripts. The TUI version of each script can be completely removed. Some other scripts, where there is only one version, need to be updated to use settings API only. Check if there are non-essential meshing TUI calls that we can simply delete.

PR #3048, opened 2024-07-10
fix: Remove non-settings version of example scripts

hpohekar:
closes https://github.com/ansys/pyfluent/issues/3014

1. Remove non-settings version of example scripts.
2. Replace TUI calls with settings wherever it is possible.


examples/00-fluent/exhaust_system_settings_api.py
```diff
@@ -535,13 +535,6 @@
 )
 volume_mesh_gen.Execute()
 
-###############################################################################
-# Check mesh
-# ~~~~~~~~~~
-# Check the mesh.
-
-meshing.tui.mesh.check_mesh()
```
prmukherj, line 543: If we are removing meshing.tui calls as well, then we shoul provide an alternative for mesh check, etc. Perhaps a scheme call?

seanpearsonuk: @hpohekar  We have to retain TUI calls on the meshing side where there is no alternative. Scheme calls are an inferior alternative.

examples/00-fluent/external_compressible_flow.py
```diff
@@ -199,19 +199,12 @@
...
-
-meshing.tui.mesh.check_mesh()
-
 ###############################################################################
 # Save mesh file
 # ~~~~~~~~~~~~~~
 # Save the mesh file (``wing.msh.h5``).
 
-meshing.tui.file.write_mesh("wing.msh.h5")
+meshing.meshing.File.WriteMesh(FileName="wing.msh.h5")
```
seanpearsonuk, line 207: :+1:

examples/00-fluent/modeling_cavitation.py
```diff
@@ -306,7 +303,7 @@
 ###############################################################################
 # Mirror the display around the symmetry plane to show the full model.
 
-solver.tui.display.set.mirror_zones(["symm_2", "symm_1"])
+solver.results.graphics.views.mirror_zones = ["symm_2", "symm_1"]
```
seanpearsonuk, line 306: :+1: Caveat: we ought to start to use `solver.settings` (also below).

hpohekar: Done. Thank you.

examples/00-fluent/exhaust_system_settings_api.py
hpohekar: Done. Thank you.

prmukherj (approved):

seanpearsonuk (approved):

Issue #3051, opened 2024-07-10
Move LocalParametricStudy from parametric to pyfluent repo.


mkundu1:
Move https://github.com/ansys/pyfluent-parametric/tree/main/src/ansys/fluent/parametric/local from parametric to pyfluent.

Check if any other important code is remaining to move.

PR #3058, opened 2024-07-11
feat: move `LocalParametricStudy` from parametric to PyFluent

hpohekar:
closes https://github.com/ansys/pyfluent/issues/3051, https://github.com/ansys/pyfluent/issues/3063

1. `LocalParametrictStudy` has `run_in_fluent` method which calls  `_run_local_study_in_fluent`  therefore we need to copy all other dependent code of `ParametricStudy*`.
2. We can change name and location of `parametric.py` if required.
3. Add tests.
4. Moved search.py under core.
5. Resolved Run API codegen failure by installing `reader` package.


prmukherj (approved):

src/ansys/fluent/core/utils/parametric.py
seanpearsonuk, line 1: My view is that utils would only contain quite generic _library_ code (i.e. code that's separate from the specific business logic of this library). Most of what's in utils fits that pattern. This code can go in either a package or module under core.

hpohekar: Done. Thank you.

.github/workflows/ci.yml
```diff
@@ -165,7 +165,9 @@ jobs:
 
       - name: Run API codegen
         if: steps.cache-api-code.outputs.cache-hit != 'true'
-        run: make api-codegen
+        run: |
+          poetry install --all-extras
+          make api-codegen
```
hpohekar, line 170: Need to install reader package.

image

https://github.com/ansys/pyfluent/actions/runs/9889959134/job/27317176855?pr=3058

Makefile
```diff
@@ -128,6 +128,7 @@ api-codegen:
 	@python -m venv env
 	@. env/bin/activate
 	@pip install -q -e .
+	@pip install -q -e .[reader]
```
hpohekar, line 131: Need to install reader package.

image

https://github.com/ansys/pyfluent/actions/runs/9889959134/job/27317176855?pr=3058

mkundu1: This is because CaseReader is now a top-level import. I think we should avoid that as the reader package is optional.

Or, maybe we can include the h5py dependency by default. We had separated it due to some temporary issue in a previous h5py version. TBD.

mkundu1:
I don't think we shall include DesignPoint/ParametricStudy/ParametricProject classes, those can be replaced by settings API.

Makefile
hpohekar: Ok. We will add `h5py` dependency by default.

hpohekar:
https://github.com/ansys/pyfluent/pull/3070

Issue #2631, opened 2024-03-28
wildacard + alias not working


mkundu1:
E.g., the following is not working:
```
solver.setup.boundary_conditions.wall["*"].thermal.q()
```

seanpearsonuk:
@mkundu1 is the issue in flobject.py, in the back-end, or yet unknown? Please consider re-assigning within the team if appropriate.

PR #3108, opened 2024-07-18
fix: wildcard level set_state

mkundu1:
This will require Fluent image update + passing nightly dev build

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -1165,13 +1165,23 @@ class WildcardPath(Group):
             if fnmatch.fnmatch(item, self._path.rsplit(sep="/", maxsplit=1)[-1]):
                 yield item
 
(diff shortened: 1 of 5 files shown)


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1164,14 +1164,24 @@ def __iter__(self):
...
-        return self._state_cls.to_scheme_keys(value)
+        return self._settings_cls.to_scheme_keys(value)
 
     def to_python_keys(self, value):
         """Convert value to have keys with Python names."""
         return self._state_cls.to_python_keys(value)
 
+    @classmethod
+    def _unalias(cls, value):
+        # Not yet implemented
```
mkundu1, line 1182: will look into this in #2631

hpohekar (approved):

prmukherj (approved):

PR #3117, opened 2024-07-22
refactor: Refactoring and renaming.

prmukherj:


src/ansys/fluent/core/file_session.py
```diff
@@ -810,7 +810,7 @@ def __init__(self):
         """__init__ method of FileSession class."""
         self._case_file = None
         self._data_file = None
-        self.monitors_manager = lambda: None
+        self.monitors = lambda: None
```
seanpearsonuk, line 813: This would still be accessed like a method call so still different from live sessions.

src/ansys/fluent/core/post_objects/post_object_definitions.py
```diff
@@ -85,7 +85,7 @@ class monitor_set_name(metaclass=PyLocalPropertyMeta):
         @Attribute
         def allowed_values(self):
             """Monitor set allowed values."""
-            return self._api_helper.monitors_manager().get_monitor_set_names()
+            return self._api_helper.monitors().get_monitor_set_names()
```
prmukherj, line 88:
```suggestion
            return self._api_helper.monitors.get_monitor_set_names()
```

src/ansys/fluent/core/post_objects/post_helper.py
```diff
@@ -103,7 +103,7 @@ def __init__(self, obj):
         self.obj = obj
         self.field_info = lambda: obj.get_root().session.fields.field_info
         self.field_data = lambda: obj.get_root().session.fields.field_data
-        self.monitors_manager = lambda: obj.get_root().session.monitors
+        self.monitors = lambda: obj.get_root().session.monitors
```
prmukherj, line 106:
```suggestion
        self.monitors = obj.get_root().session.monitors
```

src/ansys/fluent/core/file_session.py
prmukherj, line 813:
```suggestion
        self.monitors = None
```

seanpearsonuk (approved):

hpohekar (approved):

src/ansys/fluent/core/services/field_data.py
```diff
@@ -434,13 +434,13 @@ def __init__(
     @deprecate_argument(
         old_arg="surface_names",
         new_arg="surfaces",
-        converter=lambda old_arg_val: old_arg_val,
+        converter=lambda old_arg_val: old_arg_val if old_arg_val else None,
```
mkundu1, line 437: This will pass `None` to the new argument but `None` is not allowed from its typehint. The correct lambda is `lambda old_arg_val: old_arg_val or []`.

Also, `PyFluentDeprecationWarning` is the default value for `warning_cls`.

prmukherj: Done, Thank you.

Issue #3085, opened 2024-07-15
settings objects contains a public file_transfer_service


seanpearsonuk:
I don't believe the naming was as intended. It's needed within the implementation only so it should be private. As such, it can straightforwardly be a method (instead of using `@property`). With it being private, we can also drop the docstring.

hpohekar:
@seanpearsonuk 

We are already setting `_file_transfer_service` as a private attribute to base / root object therefore we can not make `file_transfer_service` as a private method or attribute in `Base` class.

![image](https://github.com/user-attachments/assets/c61b8703-2079-4480-a11d-2f5979614216)

seanpearsonuk:
@hpohekar This issue is fixed now, I believe. Is that right?

PR #3129, opened 2024-07-24
fix: Make file_transfer_service private method

hpohekar:
closes https://github.com/ansys/pyfluent/issues/3085

Rename `file_transfer_service`  to `_file_transfer_handler`.


prmukherj (approved):

seanpearsonuk (approved):

mkundu1 (approved):

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -227,18 +227,13 @@ def flproxy(self):
             return self._parent.flproxy
         return self._flproxy
 
-    @property
-    def file_transfer_service(self):
-        """Remote file handler.
-
-        Supports file upload and download.
-        """
+    def _file_transfer_service(self):
```
hpohekar, line 230:
```suggestion
    @property
    def _file_transfer_service(self):
```

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -228,7 +228,7 @@ def flproxy(self):
         return self._flproxy
 
     @property
-    def file_transfer_service(self):
+    def _file_transfer_handler(self):
```
hpohekar, line 231: @seanpearsonuk 

We need to rename `file_transfer_service` to `__file_transfer_handler` as I already mentioned in the issue description that we are setting `_file_transfer_service` on root object therefore we can not use same name. Tests are passing now.

PR #3144, opened 2024-07-30
fix: Nightly meshing tests for 251 Fluent image.

prmukherj:


tests/test_new_meshing_workflow.py
```diff
@@ -29,13 +29,6 @@ def test_new_watertight_workflow(new_meshing_session):
     watertight.add_local_sizing()
 
     # Generate surface mesh
-    assert (
```
prmukherj, line 32: This is redundant check for 'get_state' and can just be got rid of.

mkundu1 (approved):

hpohekar (approved):

Issue #2791, opened 2024-05-07
product_version argument overridden by environment


seanpearsonuk:
In `solver = pf.launch_fluent(product_version=pf.FluentVersion.v241)` the `product_version` is overridden by the `PYFLUENT_FLUENT_ROOT` environment variable, which should only hold in the absence of a specific version choice for the specific launch. In my opinion.

Please comment, @mkundu1 @prmukherj @hpohekar @raph-luc

hpohekar:
@seanpearsonuk 

`PYFLUENT_FLUENT_ROOT` is a dev environment variable, user will not set it. So there will be only `product_version` if user specifies it otherwise `FluentVersion.get_latest_installed()` will be used.

![image](https://github.com/ansys/pyfluent/assets/106588300/bf56b6c4-7205-440c-869a-0a8f53f1efcd)

mkundu1:
Right, the dev environments (like `PYFLUENT_FLUENT_ROOT`, `PYFLUENT_LAUNCH_CONTAINER` etc.) are not documented for users. They were added to run the same journal in different versions/platforms without any modification. These are useful in local development or CI.

seanpearsonuk:
I realised that the environment variable is undocumented. I'm still querying whether the current behaviour is the most appropriate. If I set `product_version` in the context of a single call, it is inconvenient if it is not respected. I would not set the argument if I did not want to launch that version. My environment is permanently set up to point to a build but occasionally I want to run a different, specific version. I would have to override the environment as well as specifying `product_version`.

raph-luc:
I think @seanpearsonuk raises a good point that the implementation of this environment variable may not be appropriate (even if it is only to be used for dev/CI). The environment variable should probably not override a `launch_fluent` argument specifying `product_version`. The env var should only come into play when `product_version` is unspecified (`None`).

On this topic, I would also vote for actually documenting all these environment variables somewhere in the docs, even if we don't expect users to use them. There will certainly always be niche use cases for these.

mkundu1:
Maybe I'm using the env var `PYFLUENT_FLUENT_ROOT` differently. This is what I usually do from a cmd. I also have `PYFLUENT_FLUENT_ROOT` pointed to a local build in system settings.
```
python -i <journal> # to run with the local build
set PYFLUENT_FLUENT_ROOT=
python -i <journal> # to run with the latest installed build (24.2)
set PYFLUENT_FLUENT_ROOT=%AWP_ROOTXXX%\fluent
[...]

PR #3157, opened 2024-08-02
fix: Update launch_args env.

prmukherj:
s = pyfluent.launch_fluent(ui_mode="gui", custom_fluent_path=r"...\vNNN\fluent\ntbin\win64\fluent.exe")


hpohekar (approved):

tejalprabhu01 (approved):
Tested the changes! works as expected!

hpohekar:
@prmukherj @seanpearsonuk @mkundu1 

As per [the previous discussion](https://github.com/ansys/pyfluent/issues/2791) `PYFLUENT_FLUENT_ROOT` is for local development. It is not documented for users. User should specify `product_version` to launch specific Fluent version not the `PYFLUENT_FLUENT_ROOT`.

tejalprabhu01:
> @prmukherj @seanpearsonuk @mkundu1
> 
> As per [the previous discussion](https://github.com/ansys/pyfluent/issues/2791) `PYFLUENT_FLUENT_ROOT` is for local development. It is not documented for users. User should specify `product_version` to launch specific Fluent version not the `PYFLUENT_FLUENT_ROOT`.

@hpohekar Arent users allowed to provide any custom paths to Fluent installation (other than AWP_ROOT)? I am asking because in optiSLang we provide users setting where they can provide any path to fluent installation. If this should not be the case; then we might have to also do the changes on optiSLang side.

prmukherj:
> > @prmukherj @seanpearsonuk @mkundu1
> > As per [the previous discussion](https://github.com/ansys/pyfluent/issues/2791) `PYFLUENT_FLUENT_ROOT` is for local development. It is not documented for users. User should specify `product_version` to launch specific Fluent version not the `PYFLUENT_FLUENT_ROOT`.
> 
> @hpohekar Arent users allowed to provide any custom paths to Fluent installation (other than AWP_ROOT)? I am asking because in optiSLang we provide users setting where they can provide any path to fluent installation. If this should not be the case; then we might have to also do the changes on optiSLang side. Based on the discussion link you provided, it still isnt clear to me.

Ideally providing a custom path should be allowed. We'll discuss this within the team and get back.
Thanks

hpohekar:
@tejalprabhu01 

Aren't users allowed to provide any custom paths to Fluent installation (other than AWP_ROOT)? - No.

User don't need to specify any custom paths on Windows as `AWP_ROOT<version>` is already set for each Fluent release. So user just need to specify product version.

On Linux, user need to specify custom Fluent version path using AWP_ROOT<version> only.

Please have a look at the following specification.

image

https://fluent.docs.pyansys.com/version/stable/getting_started/installation.html

tejalprabhu01:
@hpohekar @seanpearsonuk @mkundu1 
There will be no AWP_ROOT when using Fluent installation from shared location and also we have previously encountered that users sometimes do not have administrative rights and hence cannot set the environment variables like AWP_ROOT. I understand having to set only the version makes things easier for the pyfluent consumers. But it is also important to have a bit of flexibility when using pyfluent from other tools like optiSLang.

hpohekar:
> @hpohekar @seanpearsonuk @mkundu1 There will be no AWP_ROOT when using Fluent installation from shared location and also we have previously encountered that users sometimes do not have administrative rights and hence cannot set the environment variables like AWP_ROOT. I understand having to set only the version makes things easier for the pyfluent consumers. But it is also important to have a bit of flexibility when using pyfluent from other tools like optiSLang.

@tejalprabhu01 @seanpearsonuk @mkundu1 

Ok. I see. We should document `PYFLUENT_FLUENT_ROOT` usage considering cases like use of Fluent installation from shared location and lack of administrative rights to set environment variables. 

It is worth of considering the point of lack of administrative rights to set environment variable, specifically on Linux system because current PyFluent document suggests to set AWP_ROOT compulsory.

Thanks for the clarification.

mkundu1:
Note that, the `env` argument in `launch_fluent` is for setting environment variables in the Fluent process. It is better not to mix PyFluent process's env vars within it. For PyFluent, `os.environ` or a new `launch_fluent` argument for setting custom Fluent path is a better solution.
@prmukherj @hpohekar

prmukherj:
> Note that, the `env` argument in `launch_fluent` is for setting environment variables in the Fluent process. It is better not to mix PyFluent process's env vars within it. For PyFluent, `os.environ` or a new `launch_fluent` argument for setting custom Fluent path is a better solution. @prmukherj @hpohekar

I agree with this point.
Thank you Mainak. Will just confirm from @seanpearsonuk and @hpohekar about adding this new argument -> 'custom_fluent_path'?

Thank you.

hpohekar:
> Note that, the `env` argument in `launch_fluent` is for setting environment variables in the Fluent process. It is better not to mix PyFluent process's env vars within it. For PyFluent, `os.environ` or a new `launch_fluent` argument for setting custom Fluent path is a better solution. @prmukherj @hpohekar

@mkundu1 Yes, I agree. Thank you.

src/ansys/fluent/core/launcher/slurm_launcher.py
```diff
@@ -281,6 +281,7 @@ def __init__(
         py: Optional[bool] = None,
         gpu: Optional[bool] = None,
         cwd: Optional[str] = None,
+        custom_fluent_path: Optional[str] = None,
```
seanpearsonuk, line 284: @prmukherj Is this argument appropriate here?

prmukherj: @mkundu1, could you please confirm if this is required for Slurm launcher?

seanpearsonuk: @prmukherj we could go ahead and merge meanwhile.

hpohekar (approved):

seanpearsonuk (approved):

tejalprabhu01 (approved):
Thanks for the implementation.

.github/workflows/ci.yml
```diff
@@ -452,7 +452,7 @@ jobs:
     name: Unit Testing
     if: ${{ always() }}
     needs: build
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, pyfluent]
```
prmukherj, line 455:
```suggestion
    runs-on: [public-ubuntu-latest-8-cores]
```

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -117,6 +117,7 @@ def launch_fluent(
     py: Optional[bool] = None,
     gpu: Union[bool, list[int], None] = None,
     cwd: Optional[str] = None,
+    custom_fluent_path: Optional[str] = None,
```
mkundu1, line 120: Can be replaced with a shorter name `fluent_path`

prmukherj, line 120:
```suggestion
    fluent_path: Optional[str] = None,
```

src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -217,6 +218,8 @@ def launch_fluent(
         information like how to determine the GPU IDs.
     cwd : str, Optional
         Working directory for the Fluent client.
+    custom_fluent_path: str, Optional
```
prmukherj, line 221:
```suggestion
    fluent_path: str, Optional
```

src/ansys/fluent/core/launcher/process_launch_string.py
```diff
@@ -150,5 +150,10 @@ def get_exe_path(fluent_root: Path) -> Path:
     if fluent_root:
         return get_exe_path(Path(fluent_root))
 
-    # 2. the latest ANSYS version from AWP_ROOT environment variables
+    # 2. Custom Path provided by the user in launch_fluent
+    fluent_path = launch_argvals.get("custom_fluent_path")
```
prmukherj, line 154:
```suggestion
    fluent_path = launch_argvals.get("fluent_path")
```

src/ansys/fluent/core/launcher/slurm_launcher.py
prmukherj, line 284:
```suggestion
        fluent_path: Optional[str] = None,
```

src/ansys/fluent/core/launcher/slurm_launcher.py
```diff
@@ -353,6 +354,8 @@ def __init__(
             If True, Fluent will start with GPU Solver.
         cwd : str, Optional
             Working directory for the Fluent client.
+        custom_fluent_path: str, Optional
```
prmukherj, line 357:
```suggestion
        fluent_path: str, Optional
```

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -78,6 +78,7 @@ def __init__(
         py: Optional[bool] = None,
         gpu: Optional[bool] = None,
         cwd: Optional[str] = None,
+        custom_fluent_path: Optional[str] = None,
```
prmukherj, line 81:
```suggestion
        fluent_path: Optional[str] = None,
```

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -149,6 +150,8 @@ def __init__(
             If True, Fluent will start with GPU Solver.
         cwd : str, Optional
             Working directory for the Fluent client.
+        custom_fluent_path: str, Optional
```
prmukherj, line 153:
```suggestion
        fluent_path: str, Optional
```

PR #3166, opened 2024-08-06
feat: Add 'current_workflow()' to enhanced meshing workflow.

prmukherj:
```Python
>>> current_workflow = meshing.current_workflow()
```
For accessing a workflow while connecting to an already running session.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -452,7 +452,7 @@ jobs:
     name: Unit Testing
     if: ${{ always() }}
     needs: build
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, pyfluent]
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -452,7 +452,7 @@ jobs:
     name: Unit Testing
     if: ${{ always() }}
     needs: build
-    runs-on: [self-hosted, pyfluent]
+    runs-on: [public-ubuntu-latest-8-cores]
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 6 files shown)


src/ansys/fluent/core/meshing/meshing_workflow.py
```diff
@@ -297,3 +297,33 @@ def __init__(
         self._meshing = meshing
         self._unsubscribe_root_affected_callback()
         self._create_workflow()
+
+
+class CurrentWorkflow(Workflow):
```
seanpearsonuk, line 302: The problem is that _current_ is not a workflow type. An instantiated workflow object of type `CurrentWorkflow` has none of the specific characteristeristics of the true workflow type.

src/ansys/fluent/core/session_base_meshing.py
```diff
@@ -184,6 +185,21 @@ def create_workflow(self):
             self.get_fluent_version(),
         )
 
+    @property
+    def current_workflow(self):
+        """Datamodel root of the workflow exposed in an object-oriented manner."""
```
seanpearsonuk, line 190: Given that this is a Python class, it ought to be redundant to say that this is _in an object-oriented manner_ - we should remove those comments.

seanpearsonuk (approved):

hpohekar (approved):

.github/workflows/ci.yml
```diff
@@ -452,7 +452,7 @@ jobs:
     name: Unit Testing
     if: ${{ always() }}
     needs: build
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, pyfluent]
```
prmukherj, line 455:
```suggestion
    runs-on: [public-ubuntu-latest-8-cores]
```

PR #3167, opened 2024-08-07
fix: CI hang issue

hpohekar:
[This](https://github.com/ansys/pyfluent/pull/3153) is latest merged PR, just trying to verify whether it is causing an issue for CI.


prmukherj:
Made this change as part of other PRs.

Issue #2847, opened 2024-05-21
Python 3.9.0 compatibility with PyFluent 0.20


mhoeijm:
### 🔍 Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### 🐞 Description of the bug

I am getting a `TypeError: unhashable type: 'list'` when trying to import pyfluent in a clean Python 3.9.0 virtual environment (managed through conda): 

```
Python 3.9.0 | packaged by conda-forge | (default, Nov 26 2020, 07:53:15) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import ansys.fluent.core as pyfluent
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\ProgramData\miniforge3\envs\pyfluent-0.20-py39\lib\site-packages\ansys\fluent\core\__init__.py", line 19, in <module>
    from ansys.fluent.core.launcher.launcher import (  # noqa: F401
  File "C:\ProgramData\miniforge3\envs\pyfluent-0.20-py39\lib\site-packages\ansys\fluent\core\launcher\launcher.py", line 14, in <module>
    from ansys.fluent.core.launcher.container_launcher import DockerLauncher
  File "C:\ProgramData\miniforge3\envs\pyfluent-0.20-py39\lib\site-packages\ansys\fluent\core\launcher\container_launcher.py", line 21, in <module>
    from ansys.fluent.core.launcher.error_handler import _process_invalid_args
  File "C:\ProgramData\miniforge3\envs\pyfluent-0.20-py39\lib\site-packages\ansys\fluent\core\launcher\error_handler.py", line 4, in <module>
    from ansys.fluent.core.launcher import launcher_utils
  File "C:\ProgramData\miniforge3\envs\pyfluent-0.20-py39\lib\site-packages\ansys\fluent\core\launcher\launcher_utils.py", line 12, in <module>
    from beartype import BeartypeConf, beartype
  File "C:\ProgramData\miniforge3\envs\pyfluent-0.20-py39\lib\site-packages\beartype\__init__.py", line 58, in <module>
    from beartype._decor.decormain import (
  File "C:\ProgramData\miniforge3\envs\pyfluent-0.20-py39\lib\site-packages\beartype\_decor\decormain.py", line 23, in <module>
    from beartype._conf.confcls import (
[...]

PR #3168, opened 2024-08-07
fix: Drop support for Python 3.9

hpohekar:
closes https://github.com/ansys/pyfluent/issues/2847, https://github.com/ansys/pyfluent/issues/3169

1. Drop support for Python 3.9.
2. Update doc.
3. Set main Python version to 3.10.


seanpearsonuk (approved):

prmukherj (approved):

mkundu1 (approved):

iboyd-ansys:
I'm looking to drop support for 3.9 in PySystemCoupling and was wondering exactly how other repos had handled it. I notice here that `pyproject.toml` hasn't been changed, so as far as I can tell it still allows Python 3.9. Was that an oversight or is there a reason for this? I'd like to be consistent with what others are doing. @hpohekar @seanpearsonuk

hpohekar:
> I'm looking to drop support for 3.9 in PySystemCoupling and was wondering exactly how other repos had handled it. I notice here that `pyproject.toml` hasn't been changed, so as far as I can tell it still allows Python 3.9. Was that an oversight or is there a reason for this? I'd like to be consistent with what others are doing. @hpohekar @seanpearsonuk

@iboyd-ansys Yes, we have updated the `pyproject.toml` now. It should be 3.10. Thank you.

PR #3182, opened 2024-08-12
fix: Command Argument set_state.

prmukherj:
**Earlier behavior:**
Setting the state of command arguments was overwriting the entire parent state dictionary. In this case, the other parameters were being overwritten by their default values. If there is any dynamic dependency, then that was getting overwritten by the default value.

**Present behavior**
Only the required parameter is updated as a dictionary state.

Please refer to the test for more information.


tests/test_datamodel_service.py
```diff
@@ -782,3 +782,18 @@ def test_on_affected_lifetime_with_delete_all_child_objects(new_solver_session):
...
     assert "/test/affected/A:A1-1" not in solver._se_service.subscriptions
+
+
+@pytest.mark.fluent_version(">=24.1")
+def test_dynamic_dependency(new_meshing_session):
+    meshing = new_meshing_session
+    ic = meshing.meshing.LoadCADGeometry.create_instance()
+
+    d = ic.Refaceting.Deviation.get_state()
+    cd = ic.Refaceting.CustomDeviation.get_state()
```
prmukherj, line 793: **CustomDeviation** is dynamically dependent on **Deviation**.

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1808,9 +1808,7 @@ def get_state(self) -> Any:
 
     def set_state(self, state) -> Any:
         """Set state of the command argument."""
-        parent_state = self.parent.get_state()
-        parent_state[self.name] = state
-        self.parent.set_state(parent_state)
+        self.parent.set_state({self.name: state})
```
seanpearsonuk, line 1811: :+1: looks correct now.

seanpearsonuk (approved):

hpohekar (approved):

PR #3192, opened 2024-08-14
fix: None type tasks in workflow.

prmukherj:
```Python
>>> workflow.setState({r'TaskObject:Describe Overset Features': {},})
```

The above was not working earlier as the task type of the above task object is None.

Presently we have set the None type task objects to "BaseTask" to resolve this issue where one wants to set None type Task Objects.


seanpearsonuk:
The example you give is for an actual command that has arguments. Is its `TaskType` always `None`? On the surface, making it a `BaseTask` does not appear to be a comprehensive solution given that those are not for commands and are generally not used concretely. Can you provide more information to back this change up?

seanpearsonuk (approved):

src/ansys/fluent/core/workflow.py
```diff
@@ -1258,20 +1258,21 @@ def compound_child(self, name: str):
 
 def _makeTask(command_source, name: str) -> BaseTask:
     task = command_source._workflow.TaskObject[name]
-    task_type = task.TaskType()
     kinds = {
+        None: BaseTask,
```
prmukherj, line 1262: Only issue here is that even when there might be a possibility that we can know the task types, we are going ahead with BaseTask. However, this is done to support old journals, so this should be fine.

mkundu1 (approved):

hpohekar (approved):

PR #3206, opened 2024-08-16
refactor: Skip meshing utilities test.

prmukherj:
Skip failing test for meshing utilities.


hpohekar (approved):

hpohekar:
@prmukherj We don't need this PR. There is GitHub side issue in that docker image as I have confirmed that correct results are returned in both Fluent and PyFluent.

image

image

image

Issue #3116, opened 2024-07-19
Suspicious error message


seanpearsonuk:
Test fails for 22.2 but the error message seems to be picking up the API Info for the wrong version.

https://github.com/ansys/pyfluent/actions/runs/10012396726/job/27678968477?pr=3115

```
FAILED tests/test_solver_monitors.py::test_solver_monitors - AttributeError: 'file' object has no attribute 'read_case'.

The most similar API names are:
file.read_case (Command)
file.read_case_data (Command)
file.read_case_setting (Command)
```

mkundu1:
`read_case` was added in 23.1. In 22.2, we had `file.read_case(file_type=...`.

hpohekar:
@seanpearsonuk 

It's correct that `'file' object has no attribute 'read_case'.` for 22.2 version. It has only `read` method.

![image](https://github.com/user-attachments/assets/690cabe2-5316-4c3c-8025-0c5367cf4579)

We are not generating `api_objects.json` per version to reduce packaging size in Fluent. 

Therefore it is showing search results for latest release Fluent version.

PR #3229, opened 2024-08-22
fix: Correct error message for Fluent versions

hpohekar:
closes https://github.com/ansys/pyfluent/issues/3116

Handle the error message according to Fluent version.

1. Use `_search` if Fluent version and latest api_tree pickle file version is same.

Earlier -

```
>>> import os
>>> os.environ["PYFLUENT_FLUENT_ROOT"] = r"D:\Installations\Ansys\v222_07082024\ANSYS Inc\v222\fluent"
>>> solver = pf.launch_fluent(ui_mode="gui")
>>> solver.file.read_case
 AttributeError: 'file' object has no attribute 'read_case'.

The most similar API names are:
file.read_case (Command)
file.read_case_data (Command)
file.read_case_setting (Command)
```

Now - 

image

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -24,6 +24,7 @@ import fnmatch
 import hashlib
 import keyword
 import logging
+import os
 import os.path
 import pickle
 import string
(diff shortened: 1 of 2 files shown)


prmukherj:
Shall we add a test here to check the difference in behavior @hpohekar ?

hpohekar:
> Shall we add a test here to check the difference in behavior @hpohekar ?

@prmukherj What type behavior we want to test here ?

prmukherj (approved):

PR #3230, opened 2024-08-23
refactor: PyFluent enums.

prmukherj:
The FluentEnums now follow a standard Enum: "corresponding string in lower case". There is a map for fluent values inside it.

Further optimization to be done as part of this PR:
1. Have a common implementation of _missing_
2. Have default in Fluent Enums.


src/ansys/fluent/core/launcher/pyfluent_enums.py
```diff
@@ -17,48 +17,99 @@
...
+        def is_int():
+            for m in cls:
+                return True if type(m.value) == int else False
 
-        Returns
-        -------
-        FluentMode
-            Fluent mode.
+        msg = ", " if is_int() else "', '"
+        msg = (
```
mkundu1, line 51: Using `repr` will probably simplify this code:
```
>>> repr(5)
'5'
>>> repr("5")
"'5'"
>>> str(5)
'5'
>>> str("5")
'5'
```

mkundu1 (approved):

src/ansys/fluent/core/launcher/pyfluent_enums.py
```diff
@@ -17,48 +17,99 @@
...
+    """
 
+    def _get_enum_map(self):
+        _fl_val_map = {}
+        return _fl_val_map
 
-class FluentMode(Enum):
-    """Enumerates over supported Fluent modes."""
+    def get_fl_value(self):
+        """Returns the fluent value of the enum."""
```
raph-luc, line 32: I would suggest being more verbose than just using "fl" for "fluent", makes the code more readable and avoids confusion in the future:
```suggestion
    def get_fluent_value(self):
        """Returns the Fluent value of the enum."""
```
But what do we mean by this? This value isn't anything that is being sent to or used by Fluent. These are being used by PyFluent, for example `Meshing` session object for `FluentMode.MESHING`

src/ansys/fluent/core/launcher/pyfluent_enums.py
```diff
@@ -17,48 +17,99 @@
...
-    STANDALONE = 1
-    PIM = 2
-    CONTAINER = 3
-    SLURM = 4
+    Accepts lowercase member names as values and supports comparison operators.
+    """
 
+    def _get_enum_map(self):
+        _fl_val_map = {}
+        return _fl_val_map
```
raph-luc, line 29:
```suggestion
        return {}
```

src/ansys/fluent/core/launcher/pyfluent_enums.py
```diff
@@ -17,48 +17,99 @@
...
+    SLURM = "slurm"
+
+    def _get_enum_map(self):
+        _fl_val_map = {
+            self.STANDALONE: 1,
+            self.PIM: 2,
+            self.CONTAINER: 3,
+            self.SLURM: 4,
+        }
+        return _fl_val_map
```
raph-luc, line 91:
```suggestion
        return {
            self.STANDALONE: 1,
            self.PIM: 2,
            self.CONTAINER: 3,
            self.SLURM: 4,
        }
```

raph-luc (approved):
Looks good to me, just a few suggestions for you to consider. And I think the definition of this `_fl_val_map` is unnecessary, maybe just return the dict right away as in a few examples I commented

prmukherj:
> Looks good to me, just a few suggestions for you to consider. And I think the definition of this `_fl_val_map` is unnecessary, maybe just return the dict right away as in a few examples I commented

Thank you @raph-luc for your suggestions, they improve the readability a lot. The changes have been implemented.

PR #3236, opened 2024-08-24
refactor: Clean-up post segregation of classic and enhanced meshing w…

prmukherj:
Clean-up the Enhanced meshing workflow post segregation from the Classic one. The "_dynamic_interface" flag is no longer to be required.

The "time.sleep(1)" is no longer needed as well now.

### DIFF
--- a/src/ansys/fluent/core/workflow.py
+++ b/src/ansys/fluent/core/workflow.py
@@ -1184,7 +1184,7 @@ class CompoundTask(CommandTask):
         return self._last_child()
 
     def _last_child(self) -> BaseTask:
-        """Get the last child of this CompoundTask and formulate their python name.
+        """Get the last child of this CompoundTask and set their Python name.
 
         Returns
         -------
--- a/src/ansys/fluent/core/workflow.py
+++ b/src/ansys/fluent/core/workflow.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 import logging
 import re
 import threading
-import time
 from typing import Any, Iterable, Iterator, Optional, Tuple, Union
 import warnings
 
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/workflow.py
```diff
@@ -477,44 +468,43 @@ def delete(self) -> None:
...
-                        self.python_name(), None
-                    )
-                )
-                self._command_source._python_name_display_text_map.pop(
-                    self.python_name(), None
-                )
-                self._command_source._repeated_task_python_name_display_text_map.pop(
+        self._command_source._dynamic_python_names = True
+        if (
+            self.python_name()
```
mkundu1, line 473: Shall we capture python_name in a local variable and use it throughout this function?

src/ansys/fluent/core/workflow.py
```diff
@@ -1242,12 +1181,10 @@ def add_child_and_update(self, state=None, defer_update=None):
...
             self._task.AddChildAndUpdate()
-        time.sleep(1)
-        self.tasks()[-1]._set_python_name(compound=True)
-        return self.last_child()
+        return self._last_child()
 
-    def last_child(self) -> BaseTask:
-        """Get the last child of this CompoundTask.
+    def _last_child(self) -> BaseTask:
+        """Get the last child of this CompoundTask and formulate their python name.
```
raph-luc, line 1187:
```suggestion
        """Get the last child of this CompoundTask and set their Python name.
```

src/ansys/fluent/core/workflow.py
```diff
@@ -1256,8 +1193,23 @@ def last_child(self) -> BaseTask:
         """
         children = self.tasks()
         if children:
+            py_name = self._get_python_names_for_compound_child()
```
raph-luc, line 1196: Does `_get_python_names_for_compound_child` only work for the last child of a compound task?

raph-luc (approved):

src/ansys/fluent/core/workflow.py
prmukherj: @raph-luc, yes as this step always follows the insertion of a Compound child task.

src/ansys/fluent/core/workflow.py

PR #3239, opened 2024-08-27
fix: Convert protobuf Point to python Vector.

prmukherj:
Earlier solver.fields.reduction.<any Point type> was returning a protobuf message. Currently they are wrapped as a python class.

For instance:

Earlier behavior:
```Python
>>> x = solver.fields.reduction.force(locations=["wall"])
>>> x
x: 2.862068334485282
y: -2.5299803741481259
z: -0.0014375438414745349

>>> type(x)
<class 'ansys.api.fluent.v0.common_api_pb2.Point'>
```

Current behavior:

```Python
>>> x = s.fields.reduction.force(locations=["wall"])
>>> x                                                
(2.600005424922768e-31, -1.080374020897745e-24, 3.302743550728087e-17)
>>> type(x) 
<class 'tuple'>
>>> type(x[0]) 
<class 'float'>
```

### DIFF
--- a/src/ansys/fluent/core/services/reduction.py
+++ b/src/ansys/fluent/core/services/reduction.py
@@ -7,7 +7,7 @@ import grpc
 
 from ansys.api.fluent.v0 import reduction_pb2 as ReductionProtoModule
 from ansys.api.fluent.v0 import reduction_pb2_grpc as ReductionGrpcModule
-from ansys.fluent.core.services.datamodel_se import _convert_variant_to_value
+from ansys.fluent.core.services.datamodel_se import Vector, _convert_variant_to_value
 from ansys.fluent.core.services.interceptors import (
     BatchInterceptor,
     ErrorStateInterceptor,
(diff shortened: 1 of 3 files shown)


ajain-work (approved):

src/ansys/fluent/core/services/reduction.py
```diff
@@ -300,7 +300,7 @@ def centroid(self, locations, ctxt=None) -> Any:
         request = ReductionProtoModule.CentroidRequest()
         request.locations.extend(self._get_location_string(locations, ctxt))
         response = self.service.centroid(request)
-        return response.value
+        return Vector(response.value)
```
raph-luc, line 303: Question: has returning a numpy array or a simple list been considered? Why return a custom class `Vector` instead of one of the standard types?

raph-luc (approved):

src/ansys/fluent/core/services/reduction.py
prmukherj: Good point @raph-luc. Will surely look into it.

mkundu1: I think returning a tuple of reals is better, which will be consistent with settings API also.

prmukherj: Done, Thank you.

hpohekar (approved):

seanpearsonuk (approved):

PR #3241, opened 2024-08-27
refactor: Deprecate 'task' from enhanced meshing workflow.

prmukherj:
Deprecate the way we used to access tasks in the enhanced meshing workflow using display names.

For example:

>>> watertight.task("Import Geometry")  -->> is now deprecated

Alternate usage:

>>> watertight.import_geometry

Further:
Provide verbose deprecation warning suggesting the alternate usages.

Still ongoing:
Update tests.

### DIFF
--- a/src/ansys/fluent/core/workflow.py
+++ b/src/ansys/fluent/core/workflow.py
@@ -14,6 +14,7 @@ from ansys.fluent.core.services.datamodel_se import (
     PyMenuGeneric,
     PySingletonCommandArgumentsSubItem,
 )
+from ansys.fluent.core.solver.flobject import DeprecatedSettingWarning
 from ansys.fluent.core.utils.dictionary_operations import get_first_dict_key_for_value
 from ansys.fluent.core.utils.fluent_version import FluentVersion
 from ansys.fluent.core.warnings import PyFluentDeprecationWarning, PyFluentUserWarning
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/workflow.py
```diff
@@ -1183,18 +1189,20 @@ def add_child_and_update(self, state=None, defer_update=None):
...
-        self._command_source._compound_child = False
+        try:
+            if self._fluent_version >= FluentVersion.v241:
+                if defer_update is None:
+                    defer_update = False
+                self._task.AddChildAndUpdate(DeferUpdate=defer_update)
+            else:
+                if defer_update is not None:
+                    warnings.warn(
+                        " The 'defer_update()' method is supported in Fluent 2024 R1 and later.",
```
seanpearsonuk, line 1200: It's an argument, rather than a method, and it's supported _with_ that version of Fluent rather than in it. I also think it would be useful if the string conversion of the given `FluentVersion` had the required format to be used everywhere we want to output it as a string.

seanpearsonuk: Is the leading space required?

prmukherj: > Is the leading space in the warning message required?

No, will remove it.

prmukherj: Is this fine:
"The 'defer_update' argument is supported with 'FluentVersion.v242' and later."

or

"The 'defer_update' argument is supported with Fluent version '24.2.0' and later."

seanpearsonuk: The way you wrote the Fluent version was correct in the first place, so don't change that. I was only suggesting that we could in general be automating that based on the enum so that we would always give the output in a consistent format. Otherwise this change is fine.

prmukherj, line 1200:
```suggestion
                        "The 'defer_update()' method is supported in Fluent 2024 R1 and later.",
```

seanpearsonuk (approved):

hpohekar (approved):

PR #3254, opened 2024-09-03
refactor: Clean-up PyFluent Enums.

prmukherj:
We don't need to compare 'FluentEnums' for greater than or less than the other now. Only '==' should be supported and that is there by default.

~For UIMode, the other operations should be supported as well.~

### DIFF
--- a/src/ansys/fluent/core/launcher/pyfluent_enums.py
+++ b/src/ansys/fluent/core/launcher/pyfluent_enums.py
@@ -1,7 +1,6 @@
 """Provides a module for enums used in the PyFluent."""
 
 from enum import Enum
-from functools import total_ordering
 import os
 from typing import Optional, Union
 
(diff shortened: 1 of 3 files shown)


src/ansys/fluent/core/launcher/pyfluent_enums.py
```diff
@@ -149,6 +134,19 @@ def _get_enum_map(self):
             self.GUI: ("",),
         }
 
+    def __lt__(self, other):
```
seanpearsonuk, line 137: It's used for `UIMode` then? Please could you explain it in the description?

prmukherj: @seanpearsonuk, it is not required. The earlier usage is modified now.

seanpearsonuk (approved):

hpohekar (approved):

PR #3261, opened 2024-09-04
feat: Add mesh reading capability to case reader.

prmukherj:
1. Case reader to support operations for reading mesh files.
2. The attributes not relevant to a mesh file returns None type container.
3. The below usage is supported now:

>>> mesh_reader_2d.get_mesh().get_mesh_type() == MeshType.SURFACE

### DIFF
--- a/src/ansys/fluent/core/filereader/case_file.py
+++ b/src/ansys/fluent/core/filereader/case_file.py
@@ -221,6 +221,7 @@ class MeshType(Enum):
 
     SURFACE = "surface"
     VOLUME = "volume"
+    UNKNOWN = "unknown"
 
 
 class Mesh:
@@ -246,10 +247,13 @@ class Mesh:
 
     def get_mesh_type(self) -> MeshType:
         """Returns the type of the mesh."""
-        if "cells" in self._file_handle["meshes"]["1"].keys():
-            return MeshType.VOLUME
-        else:
-            return MeshType.SURFACE
+        try:
+            if "cells" in self._file_handle["meshes"]["1"].keys():
+                return MeshType.VOLUME
+            else:
+                return MeshType.SURFACE
+        except Exception:
+            return MeshType.UNKNOWN
 
     def get_surface_ids(self) -> list:
         """Returns list of ids of all available surfaces."""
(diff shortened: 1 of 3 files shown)


src/ansys/fluent/core/filereader/case_file.py
```diff
@@ -236,6 +244,13 @@ def __init__(self, file_handle):
         """Initialize the object."""
         self._file_handle = file_handle
 
+    def get_mesh_type(self) -> MeshType:
```
seanpearsonuk, line 247: It's significant that the return value can be None.

Aside:: since we've dropped 3.9, it can now be written `MeshType | None`, which we _could_ apply everywhere over `Optional[]`.

src/ansys/fluent/core/filereader/case_file.py
```diff
@@ -236,6 +244,13 @@ def __init__(self, file_handle):
...
+        """Returns the type of the mesh."""
+        if "cells" in self._file_handle["meshes"]["1"].keys():
```
seanpearsonuk, line 249: I wonder if we should try/except this and return `MeshType.UNKNOWN` in case of an exception.

src/ansys/fluent/core/filereader/case_file.py
```diff
@@ -603,13 +641,22 @@ def __init__(
...
+            super().__init__(rp_vars_str=rp_vars_str)
         self._mesh = Mesh(_file)
 
     def get_mesh(self):
         """Get the mesh data."""
         return self._mesh
 
+    def __getattribute__(self, item):
+        if item != "_is_case_file" and not self._is_case_file:
+            if item in set(
```
seanpearsonuk, line 654: These can be combined together.

src/ansys/fluent/core/filereader/case_file.py
prmukherj: This particular case always returns "MeshType", right @seanpearsonuk ?

src/ansys/fluent/core/filereader/case_file.py
prmukherj: Done, Thank you.

src/ansys/fluent/core/filereader/case_file.py
prmukherj: Done, Thank you.

src/ansys/fluent/core/filereader/case_file.py
seanpearsonuk: Seems correct

seanpearsonuk (approved):

hpohekar (approved):

PR #3264, opened 2024-09-04
test: Update enhanced meshing workflow tests.

prmukherj:
Update tests to run in nightly CI.

### DIFF
--- a/src/ansys/fluent/core/workflow.py
+++ b/src/ansys/fluent/core/workflow.py
@@ -14,6 +14,7 @@ from ansys.fluent.core.services.datamodel_se import (
     PyMenuGeneric,
     PySingletonCommandArgumentsSubItem,
 )
+from ansys.fluent.core.solver.flobject import DeprecatedSettingWarning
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/workflow.py
```diff
@@ -756,6 +756,8 @@ def _assign(self, args: dict, fn) -> None:
         except Exception:
             old_state = None
         camel_args = {}
+        if "add_child" in args:
+            self._snake_to_camel_map["add_child"] = "AddChild"
```
seanpearsonuk, line 760: This is good for now if it resolves a test issue. Given that this seems to be specifying something thath looks like a generic mapping (surprisingly), would there be a different solution for the longer term?

prmukherj: yes, @seanpearsonuk. I'll investigate further. Actually we should update the map during "_add_child()" call in compound task, but that was somehow not updating it. I'll look into it further.

prmukherj: Added a TODO.

seanpearsonuk (approved):

PR #3300, opened 2024-09-13
fix: Remove exit from settings.

prmukherj:
Remove 'exit' from solver.settings in PyFluent exposure since we have a well documented way of exiting in PyFluent: solver.exit(). It solves a bug arising from solver.settings.exit() as well.

Earlier:

>>> solver.settings.exit()

Now:

"exit" not present in dir(solver.settings) and using it raises an attribute error.


hpohekar (approved):

seanpearsonuk (approved):

mkundu1 (approved):
We could also block the exit command at a more upstream level: flobject.py::get_cls.

mkundu1 (approved):

PR #3306, opened 2024-09-19
refactor: Access named objects via internal names in datamodel_se

prmukherj:
Earlier, only access via display names was allowed:
```Python
>>> assert meshing.workflow.TaskObject["Import Geometry"]
```
Now, internal names are supported as well:

```Python
>>> assert meshing.workflow.TaskObject["TaskObject1"]
>>> assert meshing.workflow.TaskObject["TaskObject:TaskObject1"]
```

### DIFF
--- a/src/ansys/fluent/core/services/datamodel_se.py
+++ b/src/ansys/fluent/core/services/datamodel_se.py
@@ -1480,6 +1480,23 @@ class PyNamedObjectContainer:
             )
         return child_object_display_names
 
+    def _get_id_name(self):
+        child_object_id_name_list = []
+        obj_type = self.path[-1][0]
+        for name in self._get_child_object_names():
+            child_object_id_name_list.append(obj_type + ":" + name)
+        return child_object_id_name_list
+
+    def _check_via_internal_names(self, name) -> str:
+        if name in self._get_id_name() or name in self._get_child_object_names():
+            _name = name.split(":")[-1] if ":" in name else name
+            name_path = self.path[0:-1]
+            name_path.append((self.path[-1][0], _name))
+            name_path.append(("_name_", ""))
+            return PyMenu(self.service, self.rules, name_path).get_state()
+        else:
+            return name
+
     def get_object_names(self) -> Any:
         """Displays the name of objects within a container."""
         return self.service.get_object_names(
@@ -1514,6 +1531,7 @@ class PyNamedObjectContainer:
             )
 
     def _get_item(self, key: str) -> PyMenu:
+        key = self._check_via_internal_names(key)
         if key in self._get_child_object_display_names():
             child_path = self.path[:-1]
             child_path.append((self.path[-1][0], key))
@@ -1526,6 +1544,7 @@ class PyNamedObjectContainer:
             )
 
     def _del_item(self, key: str) -> None:
+        key = self._check_via_internal_names(key)
         if key in self._get_child_object_display_names():
             child_path = self.path[:-1]
             child_path.append((self.path[-1][0], key))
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1514,6 +1531,7 @@ def __iter__(self) -> Iterator[PyMenu]:
             )
 
     def _get_item(self, key: str) -> PyMenu:
+        key = self._check_via_internal_names(key)
```
seanpearsonuk, line 1534: `_get_child_object_display_names()` is being called twice in each of these methods. To deal with this, consolidate the code that handles checking the provided key. You should also consolidate the code that evaluates the `child_path` which is currently duplicated across these methods.

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1480,6 +1480,23 @@ def _get_child_object_display_names(self) -> list[str]:
             )
         return child_object_display_names
 
+    def _get_id_name(self):
+        child_object_id_name_list = []
+        obj_type = self.path[-1][0]
+        for name in self._get_child_object_names():
```
seanpearsonuk, line 1486: @prmukherj this is a third call to `_get_child_object_names()`.

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1480,6 +1480,23 @@ def _get_child_object_display_names(self) -> list[str]:
             )
         return child_object_display_names
 
+    def _get_id_name(self):
```
seanpearsonuk, line 1483: @prmukherj This method is called `_get_id_name()` but it returns a list of names, which contradicts the method name. Even still, I don't know what the method name means.

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1480,6 +1480,23 @@ def _get_child_object_display_names(self) -> list[str]:
...
+        child_object_id_name_list = []
+        obj_type = self.path[-1][0]
+        for name in self._get_child_object_names():
+            child_object_id_name_list.append(obj_type + ":" + name)
+        return child_object_id_name_list
```
seanpearsonuk, line 1488: @prmukherj So this is returning a list like `["TaskObject:Import Geometry",  "TaskObject:Add Local Sizing"]`?

src/ansys/fluent/core/services/datamodel_se.py
seanpearsonuk: @prmukherj If this is complicated, we should through the name at the server and let it raise any exceptions. Clients should be simple (we want more of them).

tests/test_meshing_workflow.py
```diff
@@ -424,3 +424,16 @@ def test_setting_none_type_tasks(new_meshing_session):
...
+def test_named_object_access(new_meshing_session):
+    meshing = new_meshing_session
+    meshing.workflow.InitializeWorkflow(WorkflowType=r"Fault-tolerant Meshing")
+    assert meshing.workflow.TaskObject["Import CAD and Part Management"]
+    assert meshing.workflow.TaskObject["TaskObject1"]
+    assert meshing.workflow.TaskObject["TaskObject:TaskObject1"]
+
+    assert meshing.PartManagement.Node["Meshing Model"]
+    assert meshing.PartManagement.Node["Node1"]
+    assert meshing.PartManagement.Node["Node:Node1"]
```
mkundu1, line 439: If user passes something like `XYZ:Node1`, will that also return the object?

PR #3307, opened 2024-09-19
refactor: Remove refreshing of task_accessors.

prmukherj:
https://github.com/ansys/pyfluent/pull/3041#discussion_r1765348183

As per this comment we have now removed the task-list refreshing based on add-on-affected. Some manual refreshing from the client side has been done instead on functions which can update the tasks, like -"delete", "rename" and "add_child_and_update".

### DIFF
--- a/src/ansys/fluent/core/workflow.py
+++ b/src/ansys/fluent/core/workflow.py
@@ -458,7 +458,6 @@ class BaseTask:
     def delete(self) -> None:
         """Delete this task from the workflow."""
         self._command_source.delete_tasks(list_of_tasks=[self.python_name()])
-        _call_refresh_task_accessors(self._command_source)
 
     def rename(self, new_name: str):
         """Rename the current task to a given name."""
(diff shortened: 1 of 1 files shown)


src/ansys/fluent/core/workflow.py
```diff
@@ -1583,22 +1586,6 @@ def _populate_first_tasks_python_name_command_id_map(self):
...
-            def refresh_after_sleep(_):
-                while self._refreshing:
-                    logger.debug("Already _refreshing, ...")
-                self._refreshing = True
-                logger.debug("Call _refresh_task_accessors")
-                _call_refresh_task_accessors(self)
-                self._refresh_count += 1
-                self._refreshing = False
-
-            self._root_affected_cb_by_server[self._workflow.service] = (
```
mkundu1, line 1599: Please also clean up the attribute `self._root_affected_cb_by_server`.

seanpearsonuk:
How well does this work for multiple clients - single server?

prmukherj:
> How well does this work for multiple clients - single server?

Good point Sean. Will verify it. Thank you.

PR #3315, opened 2024-09-25
feat: Support delayed settings_source assignment for built-in settings objects

mkundu1:
The following 2 syntaxes are now supported:
1.
```
x = MyBuiltinSettings(settings_source=<solver_session>/<settings_root>, ...)
assert isinstance(x, <conventional settings class>)
```
2.
```
x = MyBuiltinSettings(...)
assert isinstance(x, MyBuiltinSettings)
x.settings_source = <solver_session>/<settings_root>
assert isinstance(x, <conventional settings class>)
x.settings_source = <solver_session>/<settings_root>  # RuntimeError thrown from <conventional settings class>, re-assignment is not allowed. (Ideally, it should be AttributeError, need to fix from flobject.py.)
```


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -437,6 +437,9 @@ def _while_executing_command(self):
         """Avoid additional processing while executing a command."""
         return nullcontext()
 
+    def __eq__(self, other):
+        return self.flproxy == other.flproxy and self.path == other.path
```
mkundu1, line 441: Object identity no longer holds between built setting object and corresponding conventional setting object as the address of the built setting object is now different although it is transformed into a conventional setting object.

tests/test_builtin_settings.py
```diff
@@ -0,0 +1,393 @@
...
+    VirtualBladeModel,
+    Viscous,
+    WallBoundaries,
+    WallBoundary,
+)
+from ansys.fluent.core.examples import download_file
+from ansys.fluent.core.utils.fluent_version import FluentVersion
+
+
+def test_builtin_settings(static_mixer_case_session):
```
mkundu1, line 70: This test is not changed.

tests/test_builtin_settings.py
```diff
@@ -0,0 +1,393 @@
...
+    solver = new_solver_session
+
+    models = Models()
+    assert isinstance(models, Models)
+    with pytest.raises(TypeError):
+        models.solver = meshing
+    models.solver = solver
+    assert not isinstance(models, Models)
+    assert models.path == "setup/models"
+    assert not models.is_active()
```
mkundu1, line 339: This doesn't hold in general. E.g., `models.viscous` is active, even though `models` is inactive. This needs to be fixed in the server-side.

src/ansys/fluent/core/solver/settings_builtin_bases.py
```diff
@@ -3,44 +3,86 @@
...
+    def __init__(self, solver: Optional[Solver] = None):
+        self.__dict__.update(dict(solver=None))
+        if solver is not None:
+            self.solver = solver
+
+    def __setattr__(self, name, value):
+        if name == "solver":
+            obj = _get_obj(value, self.__class__.__name__)
+            self.__class__ = obj.__class__
+            self.__dict__.clear()
```
mkundu1, line 37: Clearing existing attributes (`solver`/`name`) as they may conflict with settings child of same name.

seanpearsonuk: Whenever user provides a `solver` object, should we be flexible and allow this to be either:
(a) a `settings` object (`<solver-session>.settings`), or
(b) a `<solver-session>` object

We would check for **(a)** first via a type check. Then we check for **(b)** by looking for a `settings` attribute. In both cases, we would store the `settings` object. Presumably, there's no broader dependency on the solver.

mkundu1: @seanpearsonuk I'll add that. One problem with storing the attribute is that some settings classes have child named `solver` or `settings`. So the attribute name will conflict with those. In the current code, I've cleared the existing attributes before casting it to the settings object.

Agree that the following behaviour is non-intuitive:
```
x.solver = solver
x.solver  # AttributeError
```

mkundu1: I've changed the attribute name to `settings_source`. The settings API doesn't have any settings class with that name, so that attribute can now be stored in the transformed settings object under that name.

Also, the dependency of solver/session is removed from builtin class implementation.

seanpearsonuk (approved):

hpohekar (approved):

prmukherj (approved):

PR #3318, opened 2024-09-25
refactor: Remove 'exit' from tui.

prmukherj:
Remove 'exit' from tui as we have a well documented exit method from session.

### DIFF
--- a/src/ansys/fluent/core/services/datamodel_tui.py
+++ b/src/ansys/fluent/core/services/datamodel_tui.py
@@ -325,14 +325,13 @@ class TUIMenu:
         self._path = path
 
     def __dir__(self) -> list[str]:
-        ret_val = [
+        return [
             convert_tui_menu_to_func_name(x)
             for x in PyMenu(
                 self._service, self._version, self._mode, self._path
             ).get_child_names()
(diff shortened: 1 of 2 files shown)


hpohekar (approved):

src/ansys/fluent/core/services/datamodel_tui.py
```diff
@@ -325,14 +325,20 @@ def __init__(self, service, version, mode, path) -> None:
...
 
     def __dir__(self) -> list[str]:
-        return [
+        ret_val = [
             convert_tui_menu_to_func_name(x)
             for x in PyMenu(
                 self._service, self._version, self._mode, self._path
             ).get_child_names()
         ]
+        ret_val.remove("exit") if "exit" in ret_val else None
```
seanpearsonuk, line 334: "exit" could be filtered out in the original expression. If using a separate statement (based on `if` rather than `try`-`except`), it should be in the simple form:
```Python
     if "exit" in ret_val:
         ret_val.remove("exit")
```

prmukherj: Done, Thank you.

mkundu1 (approved):

PR #3321, opened 2024-09-25
fix: dir() over ArgumentWrapper.

prmukherj:
In enhanced meshing it is possible to call `fault_tolerant.describe_geometry_and_flow.flow_type.allowed_values()`, but if `dir(fault_tolerant.describe_geometry_and_flow.flow_type)` is run 'allowed_values' isn't listed (nor are 'default_value', 'is_active', etc).

The dir() method over ArgumentWrapper is corrected.


src/ansys/fluent/core/workflow.py
```diff
@@ -902,10 +902,15 @@ def __setattr__(self, attr, value):
             self.set_state({attr: value})
 
     def __dir__(self):
-        arg_list = []
-        for arg in self():
-            arg_list.append(camel_to_snake_case(arg))
-        return sorted(set(list(self.__dict__.keys()) + dir(type(self)) + arg_list))
+        arg_list = (
+            [arg for arg in self() if isinstance(self(), dict)]
```
seanpearsonuk, line 906: Calling `get_state()` is much more readable here.

src/ansys/fluent/core/workflow.py
```diff
@@ -1253,8 +1259,11 @@ def _makeTask(command_source, name: str) -> BaseTask:
         "Conditional": ConditionalTask,
     }
     task_type = task.TaskType()
-    if task_type is None and command_source._compound_child:
-        kind = CompoundChild
+    if task_type is None:
+        if command_source._compound_child:
+            kind = CompoundChild
+        else:
+            kind = SimpleTask
```
prmukherj, line 1266: This is done currently to fix a fluent side bug while inserting new tasks. Later this fix can be replaced by a more accurate check of state difference.

src/ansys/fluent/core/workflow.py
```diff
@@ -902,10 +902,16 @@ def __setattr__(self, attr, value):
...
 
     def __dir__(self):
-        arg_list = []
-        for arg in self():
-            arg_list.append(camel_to_snake_case(arg))
-        return sorted(set(list(self.__dict__.keys()) + dir(type(self)) + arg_list))
+        arg_state = self.get_state()
+        arg_list = (
+            [arg for arg in arg_state if isinstance(arg_state, dict)]
+            if self() is not None
```
mkundu1, line 908: Is `self()` same as `arg_state`? Instead of calling get_state again can check `arg_state is not None`

prmukherj, line 908:
```suggestion
            if arg_state is not None
```

prmukherj: Good catch, Thanks.
Done.

mkundu1: `arg_state = list(args_state) if isinstance(arg_state, dict) else []`

mkundu1 (approved):

hpohekar (approved):

seanpearsonuk (approved):

PR #3326, opened 2024-09-27
feat: Implement IterationEndedSyncEvent and TimestepEndedSyncEvent

mkundu1:
Implemented sync events for IterationEnded and TimestepEnded to execute PyFluent-side callbacks synchronously. Fluent's solution monitoring mechanism has been used for the implementation. On every solution event, Fluent streams an auto-pause event and then goes to a paused state. Upon receiving the auto-pause event, the client executes the PyFluent-side callback and then remotely resumes Fluent.

Any exception thrown in the callback is converted to a logging error to ensure Fluent resumes at the end of callback execution. Those exceptions are raised in a secondary thread, the user's script cannot react in any way to those exceptions from the primary thread. There could still be some edge cases like calling `solver.exit()` within the callback which will indefinitely hang Fluent.

### DIFF
--- a/src/ansys/fluent/core/streaming_services/events_streaming.py
+++ b/src/ansys/fluent/core/streaming_services/events_streaming.py
@@ -4,7 +4,7 @@ from enum import Enum
 from functools import partial
 import inspect
 import logging
-from typing import Callable, Generic, Type, TypeVar
+from typing import Callable, Generic, Literal, Type, TypeVar
 import warnings
 
 from ansys.api.fluent.v0 import events_pb2 as EventsProtoModule
@@ -27,8 +27,10 @@ class SolverEvent(Enum):
 
     TIMESTEP_STARTED = "TimestepStartedEvent"
     TIMESTEP_ENDED = "TimestepEndedEvent"
+    TIMESTEP_ENDED_SYNC = "TimestepEndedSyncEvent"
     ITERATION_STARTED = "IterationStartedEvent"
     ITERATION_ENDED = "IterationEndedEvent"
+    ITERATION_ENDED_SYNC = "IterationEndedSyncEvent"
     CALCULATIONS_STARTED = "CalculationsStartedEvent"
     CALCULATIONS_ENDED = "CalculationsEndedEvent"
     CALCULATIONS_PAUSED = "CalculationsPausedEvent"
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/streaming_services/events_streaming.py
```diff
@@ -225,3 +243,57 @@ def start(self, *args, **kwargs) -> None:
...
+
+    def _register_solution_event_sync_callback(
+        self,
+        event_type: Literal[
+            SolverEvent.ITERATION_ENDED_SYNC, SolverEvent.TIMESTEP_ENDED_SYNC
+        ],
+        callback_id: str,
+        callback: Callable,
+    ):
+        unique_id = self._session.scheme_eval.scheme_eval(
```
mkundu1, line 255: The unique_id is streamed through the auto-pause event which is checked within the client-side callback.

prmukherj:
Probably need to mark these tests to skip 22R2.

src/ansys/fluent/core/streaming_services/events_streaming.py
```diff
@@ -27,8 +27,10 @@ class SolverEvent(Enum):
 
     TIMESTEP_STARTED = "TimestepStartedEvent"
     TIMESTEP_ENDED = "TimestepEndedEvent"
+    TIMESTEP_ENDED_SYNC = "TimestepEndedSyncEvent"
```
seanpearsonuk, line 30: @mkundu1 Do we need to create a documentation issue?  The user will need to know the difference between the new and existing events. Event type details are in general missing in the API reference, so could be added here (in a separate task).

src/ansys/fluent/core/streaming_services/events_streaming.py
```diff
@@ -225,3 +243,64 @@ def start(self, *args, **kwargs) -> None:
...
+                                )
+                            )
+                        ()
+                        )
+                    {'#t' if event_type == SolverEvent.TIMESTEP_ENDED_SYNC else '#f'}
+                    )
+                (car ids)
+                )
+        """
+        )
```
seanpearsonuk, line 283: 😮

seanpearsonuk (approved):

src/ansys/fluent/core/streaming_services/events_streaming.py
mkundu1: An alternative simpler strategy could be to send `uuid1` from clients, but the size of `uuid1` won't fit in the auto-pause response.

src/ansys/fluent/core/streaming_services/events_streaming.py
mkundu1: As discussed, made the original `ITERATION_ENDED` and `TIMESTEP_ENDED` as sync events, no doc change is required

hpohekar (approved):

PR #3340, opened 2024-10-01
docs: mesh

seanpearsonuk:
Update Mesh class docstring to clarify its scope of usability.

### DIFF
--- a/src/ansys/fluent/core/filereader/case_file.py
+++ b/src/ansys/fluent/core/filereader/case_file.py
@@ -225,7 +225,7 @@ class MeshType(Enum):
 
 
 class Mesh:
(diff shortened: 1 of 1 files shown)


src/ansys/fluent/core/filereader/case_file.py
```diff
@@ -225,10 +225,15 @@ class MeshType(Enum):
 
 
 class Mesh:
-    """Class to provide mesh data.
+    """Class to provide information about Fluent mesh files.
```
seanpearsonuk, line 228:
```suggestion
    """Class to provide data from and information about Fluent mesh files.
```

mkundu1 (approved):

prmukherj (approved):

PR #3345, opened 2024-10-01
refactor: small refactoring of command classes

mkundu1:
Small refactoring of command classes, so `Command` and `CommandWithPositionalArgs` behave almost similarly. The only difference between these 2 classes is the call signature.

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -1682,14 +1682,15 @@ class BaseCommand(Action):
                 self._parent.path, self.obj_name, **newkwds
             )
             if prompt:
+                valid_responses = {"y": True, "yes": True, "n": False, "no": False}
                 while True:
-                    response = input(prompt + ": y[es]/n[o] ")
-                    if response in ["y", "Y", "n", "N", "yes", "no"]:
+                    response = input(prompt + ": y[es]/n[o] ").strip().lower()
+                    if response in valid_responses:
+                        if not valid_responses[response]:
+                            return
                         break
                     else:
-                        print("Enter y[es]/n[o]")
-                if response in ["n", "N", "no"]:
-                    return
+                        print("Please enter 'y[es]' or 'n[o]'.")
         with self._while_executing_command():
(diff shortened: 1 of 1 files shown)


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1674,6 +1674,32 @@ def __getattr__(self, name: str):
 class BaseCommand(Action):
     """Executes command."""
 
+    def _execute_command(self, *args, **kwds):
```
mkundu1, line 1677: Moved from derived classes as they would be identical

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1799,7 +1784,7 @@ class Query(Action):
 
     def __call__(self, **kwds):
         """Call a query with the specified keyword arguments."""
-        newkwds = _get_new_keywords(self, [], kwds)
+        newkwds = _get_new_keywords(self, **kwds)
```
mkundu1, line 1787: The only difference between `Command` and `CommandWithPositionalArgs` is the call signature.

prmukherj (approved):

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1674,6 +1674,32 @@ def __getattr__(self, name: str):
...
+            )
+            if prompt:
+                while True:
+                    response = input(prompt + ": y[es]/n[o] ")
+                    if response in ["y", "Y", "n", "N", "yes", "no"]:
+                        break
+                    else:
+                        print("Enter y[es]/n[o]")
+                if response in ["n", "N", "no"]:
+                    return
```
seanpearsonuk, line 1692:
```suggestion
            if prompt:
                valid_responses = {"y": True, "yes": True, "n": False, "no": False}
                while True:
                    response = input(prompt + ": y[es]/n[o] ").strip().lower()
                    if response in valid_responses:
                        if not valid_responses[response]:
                            return
                        break
                    else:
                        print("Please enter 'y[es]' or 'n[o]'.")
```

hpohekar (approved):

seanpearsonuk (approved):

PR #3354, opened 2024-10-03
fix: Update test_meshing_utilities

hpohekar:
closes https://github.com/ansys/pyfluent/issues/3348

Update `test_meshing_utilities`

### DIFF
--- a/tests/test_meshing_utilities.py
+++ b/tests/test_meshing_utilities.py
@@ -2247,9 +2247,9 @@ def test_meshing_utilities(new_meshing_session):
         is None
     )
 
-    assert meshing_session.meshing_utilities.delete_marked_faces_in_zones(
-        face_zone_name_pattern="*"
-    ) == [0, 0, 0, 0, 3446]
+    # assert meshing_session.meshing_utilities.delete_marked_faces_in_zones(
+    #     face_zone_name_pattern="*"
+    # ) == [0, 0, 0, 0, 3446]
 
     assert (
         meshing_session.meshing_utilities.delete_empty_face_zones(


hpohekar:
`test_meshing_utilities` is passing in nightly dev test now.

https://github.com/ansys/pyfluent/actions/runs/11157941978/job/31019269357

tests/test_meshing_utilities.py
```diff
@@ -2247,9 +2247,9 @@ def test_meshing_utilities(new_meshing_session):
         is None
     )
 
-    assert meshing_session.meshing_utilities.delete_marked_faces_in_zones(
-        face_zone_name_pattern="*"
-    ) == [0, 0, 0, 0, 3446]
+    # assert meshing_session.meshing_utilities.delete_marked_faces_in_zones(
```
prmukherj, line 2250: @hpohekar, is this a temporary fix? Or shall we get rid of the commented code altogether?

hpohekar: @prmukherj We have kept it for reference even thought it is failing in assertions. You can see that we have already commented other code apart from this.

seanpearsonuk: @hpohekar Could you please expand on what you mean by "reference usage" in this context? Thanks.

prmukherj: In that case can we update this with a correct code? Or it does not work here? Then anyways the comment will not make sense, right?

hpohekar: We had already discussed that we will keep commenting code that gives varying output (in future also). So we are following same rule here.

We should also consider point that same test is passing in each CI run but why it is failing in nightly only.  If we can fix this behavior then we don't need to comment current code.

As already explained above we are keeping commented code for reference that makes sense right ?

hpohekar: Exactly that's why to avoid this varying output update every time (as image gets updated on daily basis) we are commenting it.

prmukherj: so, later if it is fixed again we will uncomment, right?

mkundu1 (approved):

prmukherj (approved):

tests/test_meshing_utilities.py
seanpearsonuk: @hpohekar @mkundu1 @prmukherj 

Some rationale has been given in the above discussion but the rationale is missing from the file itself.

hpohekar: @seanpearsonuk Do we want to add reason why we have commented this out ?

image

mkundu1: @seanpearsonuk @hpohekar @prmukherj Shall we track this in a github issue? Instead of comparing against fixed values which can change with Fluent version, should we compare with the Scheme call results? The Scheme API should stable as it is documented in Fluent.

hpohekar: @mkundu1 Yes, we can do this to avoid this kind of update every time.

hpohekar: @mkundu1 I think only concern is about output as it will be unknown if we directly compare with scheme output.

PR #3356, opened 2024-10-03
fix: meshng wf update

seanpearsonuk:


src/ansys/fluent/core/session_base_meshing.py
```diff
@@ -122,12 +122,8 @@ def meshing_utilities(self):
...
     def workflow(self):
         """Datamodel root of workflow."""
-        if not self._old_workflow:
-            self._old_workflow = WorkflowMode.CLASSIC_MESHING_MODE.value(
-                _make_datamodel_module(self, "workflow"),
-                self.meshing,
-                self.get_fluent_version(),
-            )
+        if self._old_workflow is None:
+            self._old_workflow = _make_datamodel_module(self, "workflow")
```
prmukherj, line 126: We can get rid of the classic meshing workflow altogether only, right?

prmukherj: Shall we do the clean-up in a separate PR?

seanpearsonuk: This is to test it at this time. If this looks good, we should check the console regressions.

prmukherj: got it. Thanks.

tests/test_meshing_workflow.py
```diff
@@ -231,7 +231,6 @@ def test_nonexistent_attrs(new_meshing_session):
     assert not hasattr(meshing.workflow, "xyz")
     with pytest.raises(AttributeError) as msg:
         meshing.workflow.xyz
-    assert msg.value.args[0] == "'ClassicMeshingWorkflow' object has no attribute 'xyz'"
```
seanpearsonuk, line 234: The previous assertion was enough -- the rest is just implementation.

seanpearsonuk:
@mkundu1 @prmukherj @hpohekar 

If tests pass here and the regressions are fixed by this then we should include. It's a reversion to a simpler set-up after all.

mkundu1 (approved):

prmukherj (approved):

PR #3381, opened 2024-10-10
fix: Hiding switch-to-meshing-mode command from tui and settings

raph-luc:
ADO Fluent PR 530389
GitHub fluids-webui PR 398

Changes
- Hiding switch-to-meshing-mode command from solver.tui and solver.settings
- Adding tests to make sure this command does not appear in either location as we will most likely never want it in those locations, in my understanding

### DIFF
--- a/src/ansys/fluent/core/services/datamodel_tui.py
+++ b/src/ansys/fluent/core/services/datamodel_tui.py
@@ -330,13 +330,13 @@ class TUIMenu:
             for x in PyMenu(
                 self._service, self._version, self._mode, self._path
             ).get_child_names()
-            if x != "exit"
+            if x not in ["exit", "switch_to_meshing_mode"]
         ]
 
     def __getattribute__(self, name) -> Any:
-        if name == "exit" and not self._path:
+        if name in ["exit", "switch_to_meshing_mode"] and not self._path:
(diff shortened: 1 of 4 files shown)


src/ansys/fluent/core/services/datamodel_tui.py
```diff
@@ -330,13 +330,13 @@ def __dir__(self) -> list[str]:
             for x in PyMenu(
                 self._service, self._version, self._mode, self._path
             ).get_child_names()
-            if x != "exit"
+            if x not in ["exit", "switch_to_meshing_mode"]
         ]
 
     def __getattribute__(self, name) -> Any:
-        if name == "exit" and not self._path:
+        if name in ["exit", "switch_to_meshing_mode"] and not self._path:
```
prmukherj, line 337: Tui is better handled from server-side. Please refer PR by @mkundu1

raph-luc: OK, I had already pinged him

mkundu1: This change is fine for PyFluent perspective. We'll consolidate the server-side changes in 25.2.

prmukherj (approved):

PR #3393, opened 2024-10-15
test: Example of unittests for search-related functions

mkundu1:
@hpohekar I've added comments in the test function for the existing issues. Please try to add similar tests for other functions and keyword combinations.

### DIFF
--- a/src/ansys/fluent/core/search.py
+++ b/src/ansys/fluent/core/search.py
@@ -684,7 +684,10 @@ def search(
             )
         else:
             _search_whole_word(
-                search_string, match_case=True, api_tree_data=api_tree_data
+                search_string,
+                match_case=True,
+                match_whole_word=True,
+                api_tree_data=api_tree_data,
             )
     else:
         _search_whole_word(
--- a/tests/test_search.py
+++ b/tests/test_search.py
@@ -509,14 +509,14 @@ def test_search_whole_word_parent_child(monkeypatch, capsys):
     search_module = sys.modules["ansys.fluent.core.search"]
     monkeypatch.setattr(search_module, "_get_api_tree_data", lambda: api_tree_data)
     # _search_whole_word() should return a list. Print should happen in the top-level search() function.
-    _search_whole_word("parent", match_whole_word=True)
+    _search_whole_word("parent", match_whole_word=True, api_tree_data=api_tree_data)
     lines = capsys.readouterr().out.splitlines()
     # Currently, we get both paths
     with pytest.raises(AssertionError):
         assert lines == ["<solver_session>.parent (Object)"]
     # why is there a keyword named match_whole_word in a function named _search_whole_word()?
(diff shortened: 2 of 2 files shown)


src/ansys/fluent/core/search.py
```diff
@@ -684,7 +684,10 @@ def search(
             )
         else:
             _search_whole_word(
-                search_string, match_case=True, api_tree_data=api_tree_data
+                search_string,
+                match_case=True,
+                match_whole_word=True,
```
mkundu1, line 689: `match_whole_word=True` must be passed True.

hpohekar: Done. Thank you.

tests/test_search.py
```diff
@@ -489,3 +492,35 @@ def test_search_settings_from_root(capsys, static_mixer_settings_session):
...
+
+
+def test_search_whole_word_parent_child(monkeypatch, capsys):
+    api_tree_data = {
+        "api_objects": [
+            "<solver_session>.parent (Object)",
+            "<solver_session>.parent.child (Parameter)",
+        ],
+        "api_tui_objects": [],
+        "all_api_object_name_synsets": {  # why doesn't _search_whole_word() work without synsets?
```
hpohekar, line 504: `why doesn't _search_whole_word() work without synsets?` - It does. We were not using synsets for `_search_whole_word()`. We were using `api_tree_data["all_api_object_synsets"].keys()` as it was a list of all API object names. 

Now we have added `all_api_object_names` separately therefore this is not relevant now.

tests/test_search.py
```diff
@@ -489,3 +492,35 @@ def test_search_settings_from_root(capsys, static_mixer_settings_session):
...
+            "parent": ["parent"],
+            "child": ["child"],
+        },
+    }
+    search_module = sys.modules["ansys.fluent.core.search"]
+    monkeypatch.setattr(search_module, "_get_api_tree_data", lambda: api_tree_data)
+    # _search_whole_word() should return a list. Print should happen in the top-level search() function.
```
hpohekar, line 511: We have added a module level variable by using that we can either print or return results as per the need. Thank you.

tests/test_search.py
```diff
@@ -489,3 +492,35 @@ def test_search_settings_from_root(capsys, static_mixer_settings_session):
...
+    _search_whole_word("parent", match_whole_word=True, api_tree_data=api_tree_data)
+    lines = capsys.readouterr().out.splitlines()
+    # Currently, we get both paths
+    with pytest.raises(AssertionError):
+        assert lines == ["<solver_session>.parent (Object)"]
+    # why is there a keyword named match_whole_word in a function named _search_whole_word()?
+    # and why is it default to False?
```
hpohekar, line 518: Because it does contain `match_case` as well which is a sub condition of `_search_whole_word()` therefore it is default to False.

PR #3399, opened 2024-10-16
fix: Remove settings API hash comparison and runtime generation of settings API classes

mkundu1:
This PR removes settings API hash comparison and runtime generation of settings API classes. This is not required anymore because:
- In PyConsole, we don't use this mechanism. The latest settings API classes are always generated while building Fluent.
- In released versions of PyFluent, the settings API classes generated from the latest Fluent images are up-to-date.
- This was useful only during the local development of PyFluent to avoid running settingsgen after updating the Fluent code. However, it is better to run settingsgen in this case so the generated files are used and tested during local development. Note that, we can run just settingsgen as `python src\ansys\fluent\core\codegen\settingsgen.py` which takes much less time than full codegen.

This PR saves 4s during the first settings API access at runtime.

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -2207,42 +2207,28 @@ def get_root(
     """
     from ansys.fluent.core import CODEGEN_OUTDIR, CODEGEN_ZIP_SETTINGS, utils
 
-    obj_info = flproxy.get_static_info()
-    try:
-        if os.getenv("PYFLUENT_USE_OLD_SETTINGSGEN") != "1":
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -2184,42 +2184,28 @@ def get_root(
     """
     from ansys.fluent.core import CODEGEN_OUTDIR, CODEGEN_ZIP_SETTINGS, utils
 
-    obj_info = flproxy.get_static_info()
```
mkundu1, line 2187: Most of the time is taken in this call.

prmukherj (approved):

hpohekar (approved):

seanpearsonuk (approved):

PR #3417, opened 2024-10-22
fix: Fix recent nightly failure [use latest image]

mkundu1:
`[use latest image]` is a new directive for PR titles to run the PR CI with the latest 25.1 image instead of the stable image updated by nightly CI. The fixed test passes with the latest image, but will fail with the old stable image.


hpohekar (approved):

mkundu1:
I'm abandoning this as we need to disable the test until the stable image can be updated; otherwise, it will keep failing in other PRs.

Issue #3423, opened 2024-10-24
Reduce thirdparty dependecies


hpohekar:
Some thirdparty dependencies have been identified which can be replaced by native Python code:

- [ ] alive-progress
- [ ] beartype
- [ ] lxml
- [ ] platformdirs
- [ ] psutil
- [ ] requests

`pyyaml` is a dependency of `pyansys-units` too, so not planning to remove it now. TOML format is supported since Python 3.11, we can consider switching to that format when Python 3.11 becomes the minimum supported version.

PR #3424, opened 2024-10-24
feat: Remove alive-progress dependency

hpohekar:
closes https://github.com/ansys/pyfluent/issues/3423

Remove alive-progress dependency.

image


src/ansys/fluent/core/utils/file_transfer_service.py
```diff
@@ -369,6 +370,25 @@ def download(self, file_name: list[str] | str, local_directory: str | None = Non
                     )
 
 
+def _progress_bar(file_name: str, upload: bool):
+    progress_bar_size = 40
+    for file in range(2):
+        progress = file / 1
```
mkundu1, line 376: what do we get from this line?

hpohekar: We get the percentage of completion.

hpohekar: Ok. Right. Here we are handling single file at a time only. We don't need this. Thank you.

hpohekar: Earlier it was based on total number of files then changed to single file only.

mkundu1: OK, it will just show 100% after uploading/downloading each file? I'm not sure if the progress-bar is useful like that.

I was under the impression that we get some progresss percentage or something from the file-transfer service which we are showing in the progress-bar. It is okay if the service does't provide anything like that, we can cleanup the progress-bar code.

hpohekar: it will just show 100% after uploading/downloading each file? - yes, after each successful uploading/downloading. 

We just want to show that file is successfully uploaded/downloaded by showing 100% completion of progress bar.

Earlier code was also working in same manner.

The user requirement was there should be some indication that file has uploaded or downloaded. 

So instead of printing some text only we have integrated progress bar to highlight this because Fluent transcript is also appears immediately after reading or writing a file. 

To have file size based real time information to update the progress bar we will need to use another modules like `requests` and we don't want to create dependency on such modules or even using native python code we will have to get file size first then again we will have to write new file to get real time progress and we do not want to add extra code for this.

hpohekar: We have removed progress bar code. Thank you.

seanpearsonuk (approved):

mkundu1 (approved):

PR #3426, opened 2024-10-24
feat: remove beartype, platformdirs and requests 

mkundu1:


src/ansys/fluent/core/utils/__init__.py
```diff
@@ -17,3 +18,19 @@ def load_module(module_name, file_path):
     spec.loader.exec_module(module)
     logger.info(f"Loaded module {module_name} from {file_path}")
     return module
+
+
+def get_examples_download_dir():
+    """Return the path to the examples download directory."""
+    parent_path = Path.home() / "Downloads"
```
mkundu1, line 25: I've changed this from Documents to Downloads folder. This will no longer upload files to OneDrive, let me know if that is an issue.

hpohekar: Yes, this is better now. Earlier it was going to OneDrive that was an issue.

hpohekar (approved):

prmukherj (approved):

PR #3439, opened 2024-10-31
fix: Remove psutil dependency

hpohekar:
closes https://github.com/ansys/pyfluent/issues/3423

1. Remove psutil dependency
2. Implemented `_pid_exists` in replacement of `psutil.pid_exists()`


prmukherj (approved):

hpohekar:
> @hpohekar I've also tested this approach. Does it work in Windows?

@mkundu1 Right, it was not working in Windows but we have got 2 solutions, both of them are working fine.

Personally I think 1st is better.

Solution 1:

```
elif platform.system() == "Windows":
        process_query_information = 0x1000
        process_handle = ctypes.windll.kernel32.OpenProcess(
            process_query_information, 0, pid
        )
        if process_handle == 0:
            return False
        else:
            ctypes.windll.kernel32.CloseHandle(process_handle)
            return True
```

Solution 2:

```
elif platform.system() == "Windows":
	try:
	command = f'tasklist /FI "PID eq {pid}"'
	result = subprocess.run(command, shell=True, text=True, capture_output=True)
	if result.returncode == 0:
			output = result.stdout.strip()
			lines = output.splitlines()
			if len(lines) == 3 and str(pid) in lines[2]:
			return True
			else:
			return False
	except Exception as e:
			return False
```

I have checked it locally, tests are passing locally in Windows now.

image

mkundu1:
> > @hpohekar I've also tested this approach. Does it work in Windows?
> 
> @mkundu1 Right, it was not working in Windows but we have got 2 solutions, both of them are working fine.
> 
> Personally I think 1st is better.
> 
> Solution 1:
> 
> ```
> elif platform.system() == "Windows":
>         process_query_information = 0x1000
>         process_handle = ctypes.windll.kernel32.OpenProcess(
>             process_query_information, 0, pid
>         )
>         if process_handle == 0:
>             return False
>         else:
>             ctypes.windll.kernel32.CloseHandle(process_handle)
>             return True
> ```
> 
> Solution 2:
> 
> ```
> elif platform.system() == "Windows":
> 	try:
> 		command = f'tasklist /FI "PID eq {pid}"'
> 		result = subprocess.run(command, shell=True, text=True, capture_output=True)
> 		if result.returncode == 0:
> 			output = result.stdout.strip()
> 			lines = output.splitlines()
> 			if len(lines) == 3 and str(pid) in lines[2]:
> 				return True
> 			else:
> 				return False
> 	except Exception as e:
> 		return False
> ```
> 
> I have checked it locally, tests are passing in Windows now. We have also added a separate test for standalone mode.
> 
> image

Tested it, looks fine. Thanks.

tests/test_fluent_session.py
```diff
@@ -63,27 +66,44 @@ def print_transcript(transcript):
...
+    )
+
+    if inside_container:
+        assert not get_container(cortex_host)
+    else:
+        assert not _pid_exists(fluent_host_pid)
+
 
+@pytest.mark.standalone
+def test_server_exits_when_session_goes_out_of_scope_on_windows() -> None:
```
mkundu1, line 94: What is the difference between this and above? The original test can be run in windows locally.

hpohekar: Right. We will keep the original test.

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -292,6 +293,26 @@ def exit_server(self):
...
+def _pid_exists(pid):
+    if platform.system() == "Linux":
+        try:
+            os.kill(pid, 0)
+        except OSError:
+            return False
+        else:
+            return True
+    elif platform.system() == "Windows":
+        process_query_information = 0x1000
```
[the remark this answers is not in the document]
hpohekar: Done. Thank you.

hpohekar:
> watchdog_exec

Right. Thank you.

hpohekar:
> @hpohekar watchdog_exec also uses psutil.

@mkundu1 Yes. Thank you.

prmukherj (approved):

PR #3441, opened 2024-10-31
fix: Option to clear Fluent parallel env vars

mkundu1:
Related Fluent bug - 1160401

Adding option to clear some environment variables related to parallel Fluent before launch_fluent(). These env vars control number of processors of Fluent process launched from meshing PyConsole.

### DIFF
--- a/src/ansys/fluent/core/__init__.py
+++ b/src/ansys/fluent/core/__init__.py
@@ -124,3 +124,6 @@ START_WATCHDOG = None
 
 # Whether to skip health check
 CHECK_HEALTH = True
+
+# Whether to clear environment variables related to Fluent parallel mode
+CLEAR_FLUENT_PARA_ENVS = False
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/__init__.py
```diff
@@ -124,3 +124,6 @@ def version_info() -> str:
 
 # Whether to skip health check
 CHECK_HEALTH = True
+
+# Whether to clear environment variables related to Fluent parallel mode
+CLEAR_FLUENT_PARA_ENVS = False
```
mkundu1, line 129: We'll set this in PyConsole

hpohekar (approved):

prmukherj (approved):

src/ansys/fluent/core/__init__.py
prmukherj: This fix will be part of 25R2?

mkundu1: that's right

Issue #3456, opened 2024-11-07
Documentation updates


seanpearsonuk:

mkundu1:
- [ ] [@prmukherj] https://github.com/ansys/pyfluent/pull/3472#discussion_r1837758398
- [ ] [@prmukherj] https://github.com/ansys/pyfluent/pull/3472#discussion_r1837805805

PR #3472, opened 2024-11-11
docs: Document set_batch_options limitation and fix more doc warnings

mkundu1:
1. Add `set_batch_options` limitation to TUI user guide.
2. Fix some more warnings as found during doc build.


doc/source/user_guide/session/session.rst
```diff
@@ -18,9 +18,6 @@ You can obtain a PyFluent session object by calling either of the functions, :fu
   >>> solver = pyfluent.launch_fluent(case_data_file_name=case_file_name)
 
 
-Note: You can find out more about using :ref:`ref_session_guide`, and :ref:`ref_launch_guide`.
```
mkundu1, line 21: The first link is self-referencing. I think this sentence got copied from the parent section.

hpohekar (approved):

prmukherj (approved):

src/ansys/fluent/core/filereader/case_file.py
```diff
@@ -234,15 +234,15 @@ class Mesh:
...
     -------
 
-    get_surface_ids
+    get_surface_ids()
         Get a list of surface ids.
-    get_surface_names
+    get_surface_names()
         Get a list of surface names.
-    get_surface_locs
+    get_surface_locs(surface_id)
```
seanpearsonuk, line 241: I noticed that `get_surface_locs()` is abbreviated differently from the other methods, which impacts the interface's consistency. Additionally, its name doesn’t fully capture the purpose described in the documentation. Perhaps `get_surface_location_bounds()` or `get_surface_location_indices()` could better convey its function, depending on what it's meant to return.

@prmukherj

src/ansys/fluent/core/filereader/case_file.py
```diff
@@ -567,7 +567,7 @@ class CaseFile(RPVarProcessor):
 
     Methods
     -------
-    get_mesh
+    get_mesh()
```
seanpearsonuk, line 570: `get_mesh()` actually returns an object of type `Mesh`, so the docstring can be improved here. We can also add a type hint for the return type.

src/ansys/fluent/core/filereader/case_file.py
prmukherj: I agree @seanpearsonuk. I think the former makes better sense: ```get_surface_location_bounds()```

mkundu1: For now, I'm listing these tasks in the documentation [super-issue](https://github.com/ansys/pyfluent/issues/3456).

src/ansys/fluent/core/filereader/case_file.py
mkundu1: listed in documentation [super-issue](https://github.com/ansys/pyfluent/issues/3456).

PR #3497, opened 2024-11-19
feat: search module improvements

hpohekar:
closes https://github.com/ansys/pyfluent/issues/3496, https://github.com/ansys/pyfluent/issues/3498, https://github.com/ansys/pyfluent/issues/3500https://github.com/ansys/pyfluent/issues/3499

**Improve synset handling**

1. Removed empty synset lists.
2. Removed duplicate values from synset list.
3. Store synset names.

**Remove duplicate data and dependency on `api_tree_version.pickle` file**

1. Removed need of the  `api_tree_<version>.pickle` file.

**Refactor flobject.py for search function**

1. Refactored `_search`.
2. Added module level variable to return or print results as per the requirement.
4. Refactored flobject.py
5. Refactored `test_search.py` and removed tests related to legacy search function.
6. Refactored `settings_extrenal.py`

**Reorder search function**

1. Changed the order of search based on the arguments preferences

**Add more tests**

1. Added test from https://github.com/ansys/pyfluent/pull/3393

### DIFF
--- a/src/ansys/fluent/core/search.py
+++ b/src/ansys/fluent/core/search.py
@@ -8,7 +8,6 @@ import os
 from pathlib import Path
 import pickle
 import re
-import sys
 from typing import Any
 import warnings
 
@@ -223,11 +222,7 @@ def _generate_api_data(
 
         all_api_object_name_synsets = dict()
         for name in api_object_names:
-            api_object_name_synsets = (
-                wn.synsets(name.decode("utf-8"), lang="eng")
-                if sys.version_info[0] < 3
-                else wn.synsets(name, lang="eng")
-            )
+            api_object_name_synsets = wn.synsets(name, lang="eng")
             synset_names = set()
             for api_object_name_synset in api_object_name_synsets:
                 synset_names.add(api_object_name_synset.name().split(".")[0])
--- a/src/ansys/fluent/core/search.py
+++ b/src/ansys/fluent/core/search.py
@@ -286,8 +286,7 @@ def _search(
             for api_object_name_synset in api_object_name_synsets:
                 synset_names.add(api_object_name_synset.name().split(".")[0])
             if synset_names:
-                if name in synset_names:
-                    synset_names.remove(name)
+                synset_names.discard(name)
                 all_api_object_name_synsets[name] = synset_names
         api_tree_data["all_api_object_name_synsets"] = all_api_object_name_synsets
 
--- a/src/ansys/fluent/core/search.py
+++ b/src/ansys/fluent/core/search.py
@@ -228,11 +228,11 @@ def _generate_api_data(
                 if sys.version_info[0] < 3
                 else wn.synsets(name, lang="eng")
             )
-            synset_names = []
+            synset_names = set()
             for api_object_name_synset in api_object_name_synsets:
-                synset_names.append(api_object_name_synset.name().split(".")[0])
+                synset_names.add(api_object_name_synset.name().split(".")[0])
             if synset_names:
-                all_api_object_name_synsets[name] = synset_names
+                all_api_object_name_synsets[name] = sorted(list(synset_names))
         api_tree_data["all_api_object_name_synsets"] = all_api_object_name_synsets
 
         api_tree_file_path = _get_api_tree_data_file_path()
--- a/src/ansys/fluent/core/search.py
+++ b/src/ansys/fluent/core/search.py
@@ -282,9 +282,9 @@ def _search(
                 if sys.version_info[0] < 3
                 else wn.synsets(name, lang="eng")
             )
-            synset_names = []
+            synset_names = set()
             for api_object_name_synset in api_object_name_synsets:
-                synset_names.append(api_object_name_synset.name().split(".")[0])
+                synset_names.add(api_object_name_synset.name().split(".")[0])
             if synset_names:
                 all_api_object_name_synsets[name] = synset_names
         api_tree_data["all_api_object_name_synsets"] = all_api_object_name_synsets
--- a/src/ansys/fluent/core/search.py
+++ b/src/ansys/fluent/core/search.py
@@ -12,6 +12,7 @@ import sys
 from typing import Any
 import warnings
 
+import ansys.fluent.core as pyfluent
 from ansys.fluent.core.solver import flobject
 from ansys.fluent.core.solver.error_message import closest_allowed_names
 from ansys.fluent.core.utils.fluent_version import (
@@ -142,60 +143,26 @@ def _get_version_path_prefix_from_obj(obj: Any):
     return version, path, prefix
 
 
-def _search(
-    word: str,
-    match_whole_word: bool = False,
-    match_case: bool = False,
+def _generate_api_data(
     version: str | None = None,
-    search_root: Any | None = None,
-    write_api_tree_data: bool | None = False,
 ):
-    """Search for a word through the Fluent's object hierarchy.
+    """Generate API tree data.
 
     Parameters
     ----------
-    word : str
-        Word to search for.
-    match_whole_word : bool, optional
-        Whether to match whole word, by default False
-    match_case : bool, optional
-        Whether to match case, by default False
     version : str, optional
         Fluent version to search in. The default is ``None``. If ``None``,
         it searches in the latest version for which codegen was run.
-    search_root : Any, optional
-        The root object within which the search is performed.
-        It can be a session object or any API object within a session.
-        The default is ``None``. If ``None``, it searches everything.
     write_api_tree_data: bool, optional
         Whether to write the API tree data.
-
-    Examples
-    --------
-    >>> import ansys.fluent.core as pyfluent
-    >>> pyfluent.search("geometry")
-    <meshing_session>.tui.file.import_.cad_geometry (Command)
-    <meshing_session>.tui.display.update_scene.select_geometry (Command)
-    <meshing_session>.meshing.ImportGeometry (Command)
-    <meshing_session>.meshing.LoadCADGeometry (Command)
-    <solver_session>.tui.solve.initialize.compute_defaults.geometry (Command)
-    <solver_session>.tui.report.reference_values.compute.geometry (Command)
-    <solver_session>.tui.define.geometry (Command)
-    <solver_session>.tui.mesh.geometry (Object)
-    <solver_session>.setup.boundary_conditions.geometry["<name>"] (Object)
-    <solver_session>.setup.geometry (Object)
-    <solver_session>.solution.report_definitions.surface["<name>"].geometry (Parameter)
-    <solver_session>.solution.report_definitions.volume["<name>"].geometry (Parameter)
-    <solver_session>.results.graphics.mesh["<name>"].geometry (Parameter)
-    <solver_session>.results.graphics.contour["<name>"].geometry (Parameter)
     """
-    api_objects = []
-    api_tui_objects = []
-    api_object_names = []
-    results = []
+    api_objects = set()
+    api_tui_objects = set()
+    api_object_names = set()
+    search_root = None
     if version:
         version = get_version_for_file_name(version)
-    root_version, root_path, prefix = _get_version_path_prefix_from_obj(search_root)
+    _, root_path, prefix = _get_version_path_prefix_from_obj(search_root)
     if search_root and not prefix:
         return
     if not version:
@@ -206,25 +173,6 @@ def _search(
     api_tree_file = get_api_tree_file_name(version)
     with open(api_tree_file, "rb") as f:
         api_tree = pickle.load(f)
-
-    if isinstance(search_root, (flobject.Group, flobject.NamedObject)):
-        path = root_path + [
-            flobject.to_python_name(x) for x in search_root.path.split("/")
-        ]
-        root_path = []
-        tree = api_tree
-        while path:
-            p = path.pop(0)
-            if p in tree:
-                tree = tree[p]
-                root_path.append(p)
-            elif f"{p}:<name>" in tree:
-                tree = tree[f"{p}:<name>"]
-                root_path.append(f"{p}:<name>")
-                if path:
-                    path.pop(0)
-            else:
-                return
 
     def inner(tree, path, root_path):
         if root_path:
@@ -299,12 +247,10 @@ def _search(
         with open(api_tree_file, "w") as json_file:
             json.dump(api_tree_data, json_file)
 
-    if write_api_tree_data:
-        _write_api_tree_file(
-            api_tree_data=api_tree_data, api_object_names=list(api_object_names)
-        )
-        api_tree_file.unlink()
-    return results
+    _write_api_tree_file(
+        api_tree_data=api_tree_data, api_object_names=list(api_object_names)
+    )
+    api_tree_file.unlink()
 
 
 @functools.cache
@@ -327,13 +273,19 @@ def _print_search_results(queries: list, api_tree_data: dict):
     api_tree_data: dict
         All API object data.
     """
+    results = []
     api_tree_data = api_tree_data if api_tree_data else _get_api_tree_data()
     api_tree_datas = [api_tree_data["api_objects"], api_tree_data["api_tui_objects"]]
     for api_tree_data in api_tree_datas:
         for query in queries:
             for api_object in api_tree_data:
                 if query in api_object:
-                    print(api_object)
+                    results.append(api_object)
+    if pyfluent.PRINT_SEARCH_RESULTS:
+        for result in results:
+            print(result)
+    else:
+        return results
 
 
 def _get_wildcard_matches_for_word_from_names(word: str, names: list):
@@ -375,7 +327,7 @@ def _search_wildcard(search_string: str, api_tree_data: dict):
         search_string, names=api_tree_data["all_api_object_names"]
     )
     if queries:
-        _print_search_results(queries, api_tree_data=api_tree_data)
+        return _print_search_results(queries, api_tree_data=api_tree_data)
 
 
 def _get_exact_match_for_word_from_names(
@@ -527,7 +479,7 @@ def _search_whole_word(
             )
         )
     if queries:
-        _print_search_results(queries, api_tree_data=api_tree_data)
+        return _print_search_results(queries, api_tree_data=api_tree_data)
 
 
 def _download_nltk_data():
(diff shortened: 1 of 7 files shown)


src/ansys/fluent/core/search.py
```diff
@@ -281,10 +282,13 @@ def _write_api_tree_file(api_tree_data: dict, api_object_names: list):
...
             )
-            synset_names = []
+            synset_names = set()
             for api_object_name_synset in api_object_name_synsets:
-                synset_names.append(api_object_name_synset.name().split(".")[0])
-            all_api_object_name_synsets[name] = synset_names
+                synset_names.add(api_object_name_synset.name().split(".")[0])
+            if synset_names:
+                if name in synset_names:
+                    synset_names.remove(name)
```
seanpearsonuk, line 290:
```suggestion
                synset_names.discard(name)
```

hpohekar: Right. It  does not raise an exception. Done. Thank you.

src/ansys/fluent/core/search.py
```diff
@@ -281,10 +280,18 @@ def _write_api_tree_file(api_tree_data: dict, api_object_names: list):
...
+                synset_names.discard(name)
+                for synset_name in synset_names:
+                    if synset_name not in api_object_names:
+                        unrelated_synsets.add(synset_name)
+                all_api_object_name_synsets[name] = list(
+                    synset_names.difference(unrelated_synsets)
+                )
         api_tree_data["all_api_object_name_synsets"] = all_api_object_name_synsets
 
         api_tree_file = _get_api_tree_data_file()
```
seanpearsonuk, line 297: From both the names `api_tree_file` and `_get_api_tree_data_file()`, I get the impression that `api_tree_file` is going to be a file object, but it's a path object.

src/ansys/fluent/core/search.py
```diff
@@ -281,10 +280,18 @@ def _write_api_tree_file(api_tree_data: dict, api_object_names: list):
...
+                synset_names.add(api_object_name_synset.name().split(".")[0])
+            if synset_names:
+                unrelated_synsets = set()
+                synset_names.discard(name)
+                for synset_name in synset_names:
+                    if synset_name not in api_object_names:
+                        unrelated_synsets.add(synset_name)
+                all_api_object_name_synsets[name] = list(
+                    synset_names.difference(unrelated_synsets)
+                )
```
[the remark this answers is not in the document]
hpohekar: No.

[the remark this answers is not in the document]
seanpearsonuk: I mean everything inside the if block only.

src/ansys/fluent/core/search.py
hpohekar: It's path object.

src/ansys/fluent/core/search.py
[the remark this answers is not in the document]
seanpearsonuk: @hpohekar 
```Python
>>> def intersect_1(synset_names, api_object_names):
...     unrelated_synsets = set()
...     for synset_name in synset_names:
...             if synset_name not in api_object_names:
...                     unrelated_synsets.add(synset_name)
...     return list(
...             synset_names.difference(unrelated_synsets)
...     )
...
...
>>> def intersect_2(synset_names, api_object_names):
...     return list(
...             synset_names.intersection(api_object_names)
...     )
...
...
>>> data = [
... [{"a", "b", "c"}, {"b", "c"}],
... [{"a", "b", "c"}, {"b", "c", "d", "e"}],
... [{"b", "c", "d", "e"}, {"a", "b", "c"}],
... [{"a", "b", "c"}, {"a", "b", "c"}],
...  ]
>>>
>>> for d in data:
...   assert intersect_1(d[0], d[1]) == intersect_2(d[0], d[1])
...
```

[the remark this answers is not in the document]
hpohekar: Thanks for the confirmation.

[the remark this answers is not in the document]
seanpearsonuk: Try to be more flexible in the way you think about review comments. I'm not making random comments here.

src/ansys/fluent/core/search.py
```diff
@@ -281,10 +280,18 @@ def _write_api_tree_file(api_tree_data: dict, api_object_names: list):
...
+            synset_names = set()
             for api_object_name_synset in api_object_name_synsets:
-                synset_names.append(api_object_name_synset.name().split(".")[0])
-            all_api_object_name_synsets[name] = synset_names
+                synset_names.add(api_object_name_synset.name().split(".")[0])
+            if synset_names:
+                unrelated_synsets = set()
+                synset_names.discard(name)
+                for synset_name in synset_names:
+                    if synset_name not in api_object_names:
```
mkundu1, line 290: Not sure if I understand this logic. If `A` and `B` are api names at 2 different paths, `S1` and `S2` are synsets of `A`. If `B == S1`, will this code remove `S1` from synsets of A?

src/ansys/fluent/core/search.py
```diff
@@ -296,6 +303,7 @@ def _write_api_tree_file(api_tree_data: dict, api_object_names: list):
         _write_api_tree_file(
             api_tree_data=api_tree_data, api_object_names=list(api_object_names)
         )
+        api_tree_file.unlink()
```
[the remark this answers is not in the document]
seanpearsonuk: Can you please confirm if I am correctly following the algorithm? These seem to be possible steps if `write_api_tree_data` is truthy:
1. Write file `X` (don't read it again).
2. Write file `Y`.
3. Remove file `X`.

src/ansys/fluent/core/search.py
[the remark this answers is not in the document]
hpohekar: @seanpearsonuk 

I have tested it as follows but there was a typo in following code snippet. This should contains `x.intersection(api_object_name)` instead of `x.intersection(y)`. Therefore I said it's not equivalent, but I realized it later.

image

tests/test_search.py
seanpearsonuk, line 1: This is a client-side feature where we could do detailed unit testing.

src/ansys/fluent/core/search.py
hpohekar: Ok. I see. We will add `_path` suffix at both places. Thank you.

src/ansys/fluent/core/search.py
mkundu1: This is the fragment from the original .json:
```
"feature": [
    "feature",
    "feature",
    "feature",
    "feature",
    "feature_of_speech",
    "feature",
    "have",
    "sport"
],
```
I thought that the algotithm will still work if we change the above fragment to:
```
"feature": [
    "feature_of_speech",
    "have",
    "sport"
],
```
It is upto you to decide the final requirement. I now have some doubts that we may actually require `feature` among the synsets because at runtime we compute the closeness between 2 sets of synsets.
E.g.,
`A` has the synsets `[A, B, S1, S2]`
`B` has the synsets `[A, B, S3, S4]`
If we remove `A` from synsets of `A`, them at runtime we cannot match `B` with `A`. Please verify this.

hpohekar: @mkundu1 

1. Actually we have done this remove unrelated synsets like `mercantile_establishment` which is also not present in our API object names list.

2. In our algorithm, we are creating a list of synsets for an API object name. Then we are iterating over that list to print the search results.

Now I realized that we should not discard API name from the synsets list due to point 2. I will update this.

hpohekar: Please let me know if you have still any doubts. Thank you.

hpohekar: I agree with your point that "we may actually require feature among the synsets" due to point 2.

mkundu1: OK, I understand the runtime behaviour now, thanks.

> We want synsets that are also present in our API object name list.

Is this a valid requirement? E.g. `shade` is not present in api names, but it is a synonym of color. If do a semantic search with `shade`, can I expect a match with the api paths for color options (e.g. contour coloring)?

hpohekar: @mkundu1 Yes, this is a valid requirement. If user gives `search_string` as `shade` then it's synsets list should contains `colour`, `colouring` then we can expect a match with the API paths for colour options.

tests/test_search.py
hpohekar: Yes, we are going to refactor first then we will add more tests. Thank you.

src/ansys/fluent/core/search.py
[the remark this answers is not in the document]
hpohekar: Done. Thank you.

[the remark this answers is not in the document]
hpohekar: as per the latest discussion we don't need this code.

src/ansys/fluent/core/search.py
```diff
@@ -537,8 +537,11 @@ def _search_semantic(search_string: str, language: str, api_tree_data: dict):
                 ):
                     similar_keys.add(api_object_synset_name + "*")
     if similar_keys:
+        results = []
         for key in similar_keys:
-            return _search_wildcard(key, api_tree_data)
+            if _search_wildcard(key, api_tree_data):
```
seanpearsonuk, line 542: @hpohekar One could use a local variable to avoid two identical calls.

hpohekar: We will update this now. Thank you.

src/ansys/fluent/core/search.py
seanpearsonuk: We need to make sure that the unit testing is comprehensive so that we don't need to debate what the code might be doing -- or not doing.

src/ansys/fluent/core/search.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/search.py
mkundu1: @hpohekar There can be words which are not synsets of each other but have one or more common synsets.

However, I checked the latest generated json in your branch, it contains synsets which are non-api names. So it seems to agree with what I expect.

hpohekar: @mkundu1 Yes, I have updated it to contain non-api names as well, as per our above discussion.

src/ansys/fluent/core/search.py
```diff
@@ -281,30 +228,30 @@ def _write_api_tree_file(api_tree_data: dict, api_object_names: list):
                 if sys.version_info[0] < 3
```
mkundu1, line 228: this check is not required

src/ansys/fluent/core/search.py
```diff
@@ -584,15 +537,19 @@ def _search_semantic(search_string: str, language: str, api_tree_data: dict):
                 ):
                     similar_keys.add(api_object_synset_name + "*")
     if similar_keys:
+        results = []
         for key in similar_keys:
-            _search_wildcard(key, api_tree_data)
+            result = _search_wildcard(key, api_tree_data)
```
mkundu1, line 542: We are searching by api-names. So, the exact string-search should work here?

Can the result contain duplicate? Multiple api-name can belong to same api-path.

mkundu1:
@hpohekar  We don't support search_root in the new search function. Feel free to clean up the function `_get_version_path_prefix_from_obj` and relevant tests (in same or different PR).

mkundu1:
@hpohekar we'll check if any change in required for PyConsole after merging this PR

mkundu1:
@hpohekar Can you please add this test?
```
api_tree_data = {
    "api_objects": [
        "<solver_session>.first_last (Object)",
        "<solver_session>.none (Object)",
    ],
    ...
}
assert pyfluent.search("first", match_whole_word=True) == ["<solver_session>.first_last (Object)"]
assert pyfluent.search("last", match_whole_word=True) == ["<solver_session>.first_last (Object)"]
assert pyfluent.search("first_last", match_whole_word=True) == ["<solver_session>.first_last (Object)"]
```

hpohekar:
> @hpohekar we'll check if any change in required for PyConsole after merging this PR

@mkundu1 Sure.

src/ansys/fluent/core/search.py
hpohekar: Removed. Thank you.

hpohekar:
> @hpohekar Can you please add this test?
> 
> ```
> api_tree_data = {
>     "api_objects": [
>         "<solver_session>.first_last (Object)",
>         "<solver_session>.none (Object)",
>     ],
>     ...
> }
> assert pyfluent.search("first", match_whole_word=True) == ["<solver_session>.first_last (Object)"]
> assert pyfluent.search("last", match_whole_word=True) == ["<solver_session>.first_last (Object)"]
> # It is OK if we don't support the following - the above 2 are more important
> assert pyfluent.search("first_last", match_whole_word=True) == ["<solver_session>.first_last (Object)"]
> ```

@mkundu1 

We have `match_case=True` by default and by passing `match_whole_word=True` then it becomes exact search, according to which following 2 tests are not valid.

assert pyfluent.search("first", match_whole_word=True) == ["<solver_session>.first_last (Object)"]
assert pyfluent.search("last", match_whole_word=True) == ["<solver_session>.first_last (Object)"]

Following tests are valid:

assert pyfluent.search("first") == ["<solver_session>.first_last (Object)"]
assert pyfluent.search("last") == ["<solver_session>.first_last (Object)"]

I have updated the tests accordingly.

src/ansys/fluent/core/search.py
hpohekar: exact string-search will work here but it will loose semantic search meaning as wildcard search shows all possibilities but this is not the case with exact-search therefore we have used wildcard search here.

 Can the result contain duplicate? - No. Now we are showing results that ends with the given search string.

src/ansys/fluent/core/search.py
```diff
@@ -281,30 +228,30 @@ def _write_api_tree_file(api_tree_data: dict, api_object_names: list):
                 if sys.version_info[0] < 3
                 else wn.synsets(name, lang="eng")
             )
-            synset_names = []
+            synset_names = set()
             for api_object_name_synset in api_object_name_synsets:
-                synset_names.append(api_object_name_synset.name().split(".")[0])
-            all_api_object_name_synsets[name] = synset_names
+                synset_names.add(api_object_name_synset.name().split(".")[0])
```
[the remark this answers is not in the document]
hpohekar: @mkundu1 

We can not store Synset objects as it is not JSON serializable

```
  File "d:\pyfluent\codegen\allapigen.py", line 69, in <module>
    _generate_api_data(version=version)
  File "D:\pyfluent\src\ansys\fluent\core\search.py", line 239, in _generate_api_data
    _write_api_tree_file(
  File "D:\pyfluent\src\ansys\fluent\core\search.py", line 237, in _write_api_tree_file
    json.dump(api_tree_data, json_file)
  File "C:\Users\hpohekar\AppData\Local\Programs\Python\Python310\lib\json\__init__.py", line 179, in dump
    for chunk in iterable:
  File "C:\Users\hpohekar\AppData\Local\Programs\Python\Python310\lib\json\encoder.py", line 431, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "C:\Users\hpohekar\AppData\Local\Programs\Python\Python310\lib\json\encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "C:\Users\hpohekar\AppData\Local\Programs\Python\Python310\lib\json\encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "C:\Users\hpohekar\AppData\Local\Programs\Python\Python310\lib\json\encoder.py", line 325, in _iterencode_list
    yield from chunks
  File "C:\Users\hpohekar\AppData\Local\Programs\Python\Python310\lib\json\encoder.py", line 438, in _iterencode
    o = _default(o)
  File "C:\Users\hpohekar\AppData\Local\Programs\Python\Python310\lib\json\encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type Synset is not JSON serializable
PS D:\pyfluent>
```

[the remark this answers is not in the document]
hpohekar: @mkundu1 

We have tried pickle file format as well but it gives following error when we try to store Synset object.

```
  File "d:\pyfluent\codegen\allapigen.py", line 69, in <module>
    _generate_api_data(version=version)
  File "D:\pyfluent\src\ansys\fluent\core\search.py", line 238, in _generate_api_data
    _write_api_tree_file(
  File "D:\pyfluent\src\ansys\fluent\core\search.py", line 236, in _write_api_tree_file
    pickle.dump(api_tree_data, pickle_file)
TypeError: cannot pickle '_thread.RLock' object
```

src/ansys/fluent/core/search.py
```diff
@@ -584,15 +537,19 @@ def _search_semantic(search_string: str, language: str, api_tree_data: dict):
                 ):
```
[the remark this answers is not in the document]
hpohekar: We need to update above pseudo-code as we can not store Synset objects directly in either json or pickle format.

hpohekar:
> @hpohekar We don't support search_root in the new search function. Feel free to clean up the function `_get_version_path_prefix_from_obj` and relevant tests (in same or different PR).

@mkundu1 Removed. Thank you.

src/ansys/fluent/core/search.py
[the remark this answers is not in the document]
hpohekar: Done. Thank you.

src/ansys/fluent/core/search.py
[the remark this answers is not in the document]
hpohekar: @mkundu1 We have updated it as follows:

```
similar_keys = set()
    search_string_synsets = set(wn.synsets(search_string, lang=language))
    for api_object_name, api_object_synset_names in list(
        api_tree_data["all_api_object_name_synsets"].items()
    ):
        api_object_synsets = {
            wn.synset(api_object_synset_name)
            for api_object_synset_name in api_object_synset_names
        }
        if search_string_synsets & api_object_synsets:
            similar_keys.add(api_object_name + "*")
```

mkundu1 (approved):
The present changes look good to me, thanks. There is one issue related to the presentationof the results, will file a separate defect.

hpohekar:
> The present changes look good to me, thanks. There is one issue related to the presentationof the results, will file a separate defect.

Sure. Please go ahead. Thank you.

prmukherj (approved):

Issue #3436, opened 2024-10-29
Extract connection-info from server's stdout


mkundu1:
Some work is done in the branch [research/serverinfo-stdout](https://github.com/ansys/pyfluent/tree/research/serverinfo-stdout).

Limitations of this approach:
1. stdout is not piped from Fluent if Fluent is launched with GUI.
2. On some systems, enabling stdout piping exhausts buffer memory of pipes during long-running simulation - https://github.com/ansys/pyfluent/issues/1628. Currently, stdout piping is enabled only for slurm launcher in the main branch where the slurm job id is read from stdout.

mkundu1:
Closed as not planned

PR #3504, opened 2024-11-22
feat: Add a hidden way to access Fluent's stdout/stderr

mkundu1:
Requested by @PyPablo 

We cannot expose this as a full feature due to the limitations mentioned in #3436. Exposing a hidden way to access Fluent's stdout/stderr.
```
import subprocess
pyfluent.LAUNCH_FLUENT_STDOUT = subprocess.PIPE
solver = pyfluent.launch_fluent()

for line in iter(solver._process.stdout.readline, b''):
    decoded_line = line.decode('utf-8')
    print(decoded_line, end='')
    if decoded_line.startswith("Cleanup script file"):  # Any other termination condition, e.g. break after 10 lines
        break
```
Note that, this doesn't work in `ui_mode="gui"` as mentioned in #3436.

### DIFF
--- a/src/ansys/fluent/core/__init__.py
+++ b/src/ansys/fluent/core/__init__.py
@@ -133,3 +133,11 @@ PRINT_SEARCH_RESULTS = True
 
 # Whether to clear environment variables related to Fluent parallel mode
 CLEAR_FLUENT_PARA_ENVS = False
+
+# Set stdout of the launched Fluent process
+# Valid values are same as subprocess.Popen's stdout argument
+LAUNCH_FLUENT_STDOUT = None
+
+# Set stderr of the launched Fluent process
+# Valid values are same as subprocess.Popen's stderr argument
+LAUNCH_FLUENT_STDERR = None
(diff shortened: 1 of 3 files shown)


prmukherj (approved):

PyPablo:
Hi @mkundu1 thanks a lot! If this something the user can access through a protected method in launch_fluent?
i.e.
session.__stdout

returns the str or whatever?

Also, is there any possibility to capture the return from the very beginning, i.e. what is printed out in the .trn file? Obviously one can read and parse the .trn file, but it is not as clean (and maybe not so obvious) than a built-in method

seanpearsonuk (approved):

mkundu1:
> Hi @mkundu1 thanks a lot! If this something the user can access through a protected method in launch_fluent? i.e. session.__stdout
> 
> returns the str or whatever?
> 
> Also, is there any possibility to capture the return from the very beginning, i.e. what is printed out in the .trn file? Obviously one can read and parse the .trn file, but it is not as clean (and maybe not so obvious) than a built-in method

Yes, the example code in the PR description above will print from the begining of the Fluent output till the line "Cleanup script file ..". The stdout is a buffer which will keep being populated as long as Fluent is running. The above code is a standard pattern to read from a buffer.

hpohekar (approved):

hpohekar:
@mkundu1 We can track the license error at least in non GUI mode right ?

Issue #3508, opened 2024-11-23
Move backward compatibility code for dict set-state from PyFluent to settings API


mkundu1:
The backward compatibilty while setting state with a dictionary is handled in the following way in PyFluent:

Suppose, `a` is group, `x` is an child-alias of a pointing to some relative path `b` which may or may not be inside `a`. We want to do a set-state at `a` with the dictionary `{"x": .., ..}`. The dictionary is transformed to convert key `"x"` to key `"b"` (which can be at a different level) before sending it to settings API. If `b` is outside `a`, the whole this is resolved to a set-state outside `a`. We want to move this state-transformation code to settings API. PyFluent will just pass `{"x": .., ..}` to the settings API with `"x"` converted to its original scheme-name.
 
PyFluent doesn't yet handle the scenario when `a` is a named object. Resolving `"x"` within PyFluent is problematic in that scenario as `"x"` can be a part of a state which is not created yet which means we cannot traverse the object-hierarchy. Note that, PyFluent class-hierarchy cannot provide child -> parent link.

PR #3509, opened 2024-11-23
feat: Store the original scheme name of aliases

mkundu1:
This will be required after #3508 is realized in settings API. We do not have any other way to recover the original scheme names of aliases as they are not real classes in settings API.

I've avoided generation of explicit classes for aliases as it results into too much duplication of class hierarchy at various levels. We need to support a different journaling behaviour for alias classes and their descendents.


mkundu1:
@prmukherj It is better if this PR goes in before the release.

prmukherj (approved):

hpohekar (approved):

Issue #3512, opened 2024-11-25
Replace some scheme_eval service calls


hpohekar:
Point 1:

`(map car (apply append (map client-inquire-cell-vector-functions (inquire-domain-for-cell-functions))))
(velocity relative-velocity)` can be replaced by `GetVectorFieldsInfo`

seanpearsonuk:
@hpohekar What details do you have in mind here?

hpohekar:
@seanpearsonuk I have already discussed few things with @mkundu1 as follows:

Point 1:

Regarding (units/get-pretty-wb-units-from-dimension (units/inquire-dimension '{quantity})) we always get field-data in SI units. The SI unit can be queried from the ansys-units library.

Point 2:

`(map car (apply append (map client-inquire-cell-vector-functions (inquire-domain-for-cell-functions))))
(velocity relative-velocity)` can be replaced by `GetVectorFieldsInfo`

PR #3514, opened 2024-11-25
feat: Replace scheme call in get_vector_fields

hpohekar:
closes https://github.com/ansys/pyfluent/issues/3512

Replaced scheme call in `get_vector_fields()`

### DIFF
--- a/src/ansys/fluent/core/post_objects/post_helper.py
+++ b/src/ansys/fluent/core/post_objects/post_helper.py
@@ -2,8 +2,6 @@
 
 import re
 
-from ansys.api.fluent.v0 import field_data_pb2 as FieldDataProtoModule
-
 
 class IncompleteISOSurfaceDefinition(RuntimeError):
     """Raised when iso-surface definition is incomplete."""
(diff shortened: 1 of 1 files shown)


prmukherj:
@hpohekar, I think we can address all the scheme calls together in a PR? We can update the proto files as required for this. What do you think?

seanpearsonuk:
> @hpohekar, I think we can address all the scheme calls together in a PR? We can update the proto files as required for this. What do you think?

@prmukherj code change so far looks potentially OK (as far as I can tell). Isn't this approach OK, gradually removing the Scheme dependencies?

prmukherj:
> > @hpohekar, I think we can address all the scheme calls together in a PR? We can update the proto files as required for this. What do you think?
> 
> @prmukherj code change so far looks potentially OK (as far as I can tell). Isn't this approach OK, gradually removing the Scheme dependencies?

Yes @seanpearsonuk, should be fine. Just wanted to take a look into all the calls once and decide, that's all.

mkundu1 (approved):

prmukherj (approved):

prmukherj:
Tested visualization with this update and seems to run correctly. So no effect.

PR #3523, opened 2024-11-26
feat: remove unalias function

mkundu1:
Remove the `_unalias` function as it is already supported in the server. We still need to traverse to the alias class to convert disctionary state keys to scheme name.


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -696,7 +686,7 @@ class SettingsBase(Base, Generic[StateT]):
     """
 
     @classmethod
-    def to_scheme_keys(cls, value: StateT) -> StateT:
+    def to_scheme_keys(cls, value: StateT, root_cls, path: list[str]) -> StateT:
```
mkundu1, line 689: The alias class can be at a parent level, but we cannot traverse directly to a parent in the class hierarchy (we are inside a classmethod). So, we are traversing down from the root class.

hpohekar (approved):

PR #3526, opened 2024-11-27
fix: Empty locations list as input for reduction.

prmukherj:
Raise appropriate error for reduction with empty list as input for locations.

The test will pass after server-side updates are merged.


src/ansys/fluent/core/services/reduction.py
```diff
@@ -263,6 +263,8 @@ def _validate_str_location(self, loc: str):
             raise ValueError(f"Invalid location input: '{loc}'")
 
     def _get_location_string(self, locations, ctxt) -> List[str]:
+        if locations == []:
+            return []
```
seanpearsonuk, line 267: This is a bit odd, but OK

seanpearsonuk (approved):

hpohekar (approved):

PR #3531, opened 2024-11-27
feat: Classes for event info

mkundu1:
Defining explicit classes for various types of event info. These class instances will be passed to the event-streaming callback instead of the protobuf message instances that are currently passed.

TODO: Update user guide, PyConsole change

Covered by exisitng unittests.

### DIFF
--- a/tests/test_events_manager.py
+++ b/tests/test_events_manager.py
@@ -3,7 +3,7 @@ from pathlib import Path
 import pytest
 
 import ansys.fluent.core as pyfluent
-from ansys.fluent.core import MeshingEvent, SolverEvent, examples
+from ansys.fluent.core import FluentVersion, MeshingEvent, SolverEvent, examples
 from ansys.fluent.core.warnings import PyFluentDeprecationWarning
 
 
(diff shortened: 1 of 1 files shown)


src/ansys/fluent/core/streaming_services/events_streaming.py
```diff
@@ -277,14 +552,16 @@ def _register_solution_event_sync_callback(
...
-                    else EventsProtoModule.IterationEndedEvent
-                )
-                event_info = event_info_cls(index=event_info.index)
+        def on_pause(session, event_info: SolutionPausedEventInfo):
+            if unique_id == int(event_info.level):
+                if event_type == SolverEvent.ITERATION_ENDED:
+                    event_info = IterationEndedEventInfo(index=event_info.index)
+                else:
+                    event_info = TimestepEndedEventInfo(
+                        # TODO: Timestep size is currently not available
```
mkundu1, line 561: We are casting a `SolutionPausedEventInfo` into `TimestepEndedEventInfo`. Currenlty there is no way to serialize time-step size to `EventsProtoModule.AutoPauseEvent`. We need to change the proto - will look into separately.

seanpearsonuk (approved):

prmukherj (approved):

PR #3535, opened 2024-11-29
fix: Update return-type of operations on named-objects.

prmukherj:
Current behavior:
image

Test will pass after the server-side changes are in place.


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1672,7 +1672,7 @@ def execute_command(self, *args, **kwds):
                 command_name=self.python_name, value=value, kwargs=kwds
             )
         if (
-            self.obj_name == "create"
+            self.obj_name in ["create", "make-a-copy"]
```
seanpearsonuk, line 1675: :+1:

seanpearsonuk (approved):

tests/test_settings_api.py
```diff
@@ -584,3 +585,23 @@ def test_deprecated_command_arguments(mixing_elbow_case_data_session):
         "m2",
         "m3",
     }
+
+
+@pytest.mark.fluent_version(">=25.2")
+def test_return_types_of_operations_on_named_objects(mixing_elbow_settings_session):
+    solver = mixing_elbow_settings_session
+
+    var1 = solver.settings.setup.materials.fluid.create("air-created")
```
mkundu1, line 594: We could also directly compare the objects:
```
assert solver.settings.setup.materials.fluid.create("air-created") == solver.settings.setup.materials.fluid["air-created"]

mkundu1 (approved):

tests/test_settings_api.py
prmukherj: Done.

PR #3546, opened 2024-12-03
docs: event streaming callback

mkundu1:
Add documentation of event streaming callback.

I think the signature
```
def cb(session, event_info, x, y):  # additional arguments at end
```
would be preferred over 
```
def cb(x, y, session, event_info)  # additional arguments at start
```
The current implementation supports both.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -42,10 +42,7 @@ extensions = [
 
 toggleprompt_offset_right = 35
 
-skip_examples = int(os.getenv("PYFLUENT_SKIP_EXAMPLES_DOC", 0))
-if skip_examples:
-    pass
-else:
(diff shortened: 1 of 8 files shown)


hpohekar (approved):

prmukherj (approved):

doc/source/conf.py
```diff
@@ -42,10 +42,7 @@
 
 toggleprompt_offset_right = 35
 
-skip_examples = int(os.getenv("PYFLUENT_SKIP_EXAMPLES_DOC", 0))
-if skip_examples:
-    pass
-else:
+if os.getenv("PYFLUENT_DOC_SKIP_EXAMPLES") != "1":
```
seanpearsonuk, line 45: :+1:

seanpearsonuk (approved):

PR #3548, opened 2024-12-03
ci: fix fluent test run

mkundu1:


.ci/fluent_test_runner.py
```diff
@@ -95,14 +95,14 @@ def _run_single_test(
...
+                for line in stderr.splitlines():
                     if line.strip().startswith("Error:"):
                         if "Expected exception" in line:  # for check_assert.py
                             container.stop()
                         else:
-                            raise FluentRuntimeError(line)
+                            raise FluentRuntimeError(stderr)
             sleep(1)
-        logging.debug(container.logs(stderr=True).decode())
+        print(container.logs(stderr=True).decode())
```
prmukherj, line 105: @mkundu1, need to print?

prmukherj (approved):

.ci/fluent_test_runner.py
mkundu1: yeah, as it is an error, better to show it by default

millerj97 (approved):

PR #3552, opened 2024-12-05
feat: App utilities gRPC services

hpohekar:
closes https://github.com/ansys/pyfluent/issues/3553


src/ansys/fluent/core/session.py
```diff
@@ -212,6 +210,11 @@ def __init__(self, _session):
         for obj in filter(None, (self._datamodel_events, self.transcript, self.events)):
             self._fluent_connection.register_finalizer_cb(obj.stop)
 
+    @property
+    def app_utilities(self):
```
mkundu1, line 214: Maybe we don't need to expose this for users as this is used only in the implemntation code.

src/ansys/fluent/core/services/app_utilities.py
```diff
@@ -0,0 +1,241 @@
...
+            "working_directory": response.working_directory,
+        }
+
+    def get_app_mode(self) -> Any:
+        """Get app mode."""
+        request = AppUtilitiesProtoModule.GetAppModeRequest()
+        response = self.service.get_app_mode(request)
+        if response.app_mode == 0:
+            return "APP_MODE_UNKNOWN"
+        elif response.app_mode == 1:
```
mkundu1, line 170: Can check against enum instead of integer. There should be an enum available in app_utilities_pb2.

src/ansys/fluent/core/services/app_utilities.py
```diff
@@ -0,0 +1,241 @@
...
+            "process_id": response.process_id,
+            "working_directory": response.working_directory,
+        }
+
+    def get_app_mode(self) -> Any:
+        """Get app mode."""
+        request = AppUtilitiesProtoModule.GetAppModeRequest()
+        response = self.service.get_app_mode(request)
+        if response.app_mode == 0:
+            return "APP_MODE_UNKNOWN"
```
mkundu1, line 169: maybe just throw from here

src/ansys/fluent/core/streaming_services/events_streaming.py
```diff
@@ -556,34 +556,8 @@ def _register_solution_event_sync_callback(
...
-                            )
-                        ()
-                        )
-                    {'#t' if event_type == SolverEvent.TIMESTEP_ENDED else '#f'}
-                    )
-                (car ids)
-                )
-        """
+        unique_id: str = self._session.app_utilitites.register_pause_on_solution_events(
+            solution_event=SolverEvent.TIMESTEP_ENDED
```
mkundu1, line 560: what about `ITERATION_ENDED`?

src/ansys/fluent/core/session.py
hpohekar: Yes. Done. Thank you.

src/ansys/fluent/core/services/app_utilities.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/services/app_utilities.py
hpohekar: Now we are using an enum.

src/ansys/fluent/core/streaming_services/events_streaming.py
hpohekar: Right. Updated. Thank you.

src/ansys/fluent/core/codegen/print_fluent_version.py
```diff
@@ -4,23 +4,20 @@
 from ansys.fluent.core.utils.fluent_version import get_version_for_file_name
 
 
-def print_fluent_version(version: str, scheme_eval):
+def print_fluent_version(version: str, app_utilities):
```
seanpearsonuk, line 7: version is via scheme eval

hpohekar: @seanpearsonuk  Didn't get you.

src/ansys/fluent/core/codegen/print_fluent_version.py
```diff
@@ -4,23 +4,20 @@
...
     """Write Fluent version information to file."""
     version_file = (CODEGEN_OUTDIR / f"fluent_version_{version}.py").resolve()
     with open(version_file, "w", encoding="utf8") as f:
         f.write(f'FLUENT_VERSION = "{version}"\n')
-        f.write(f'FLUENT_BUILD_TIME = "{scheme_eval("(inquire-build-time)")}"\n')
-        f.write(f'FLUENT_BUILD_ID = "{scheme_eval("(inquire-build-id)")}"\n')
-        f.write(f'FLUENT_REVISION = "{scheme_eval("(inquire-src-vcs-id)")}"\n')
-        f.write(f'FLUENT_BRANCH = "{scheme_eval("(inquire-src-vcs-branch)")}"\n')
+        f.write(f'"{app_utilities.get_build_info()}"\n')
```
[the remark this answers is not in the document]
hpohekar: @mkundu1 We are now printing a single line with all details. Do we want to print it on each separately ?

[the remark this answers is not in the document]
mkundu1: @hpohekar Those details were global variables in `fluent_version_<version>.py` in the old version.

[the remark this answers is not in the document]
hpohekar: Right. We will update this according to it. Thank you.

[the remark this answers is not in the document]
hpohekar: @mkundu1 Where are we using those global variables ?

src/ansys/fluent/core/services/app_utilities.py
```diff
@@ -0,0 +1,247 @@
...
+        return {
+            "hostname": response.hostname,
+            "process_id": response.process_id,
+            "working_directory": response.working_directory,
+        }
+
+    def get_solver_process_info(self) -> Any:
+        """Get solver process info."""
+        request = AppUtilitiesProtoModule.GetSolverProcessInfoRequest()
+        response = self.service.get_solver_process_info(request)
```
mkundu1, line 157: Following code will avoid manually building the dictionary (see usage in exisiting code)
```
from google.protobuf.json_format import MessageToDict
return MessageToDict(response)
```

src/ansys/fluent/core/services/app_utilities.py
```diff
@@ -0,0 +1,247 @@
...
+        """Get app mode.
+
+        Raises
+        ------
+        ValueError
+            If app mode is unknown.
+        """
+        request = AppUtilitiesProtoModule.GetAppModeRequest()
+        response = self.service.get_app_mode(request)
+        if response.app_mode == AppUtilitiesProtoModule.APP_MODE_UNKNOWN:
```
mkundu1, line 174: Just FYI, python now supports match..case as a replacement of multiple if checks like this

src/ansys/fluent/core/codegen/print_fluent_version.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/services/app_utilities.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/services/app_utilities.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/services/app_utilities.py
hpohekar: @mkundu1 MessageToDict converts all values to strings. It converted integer `process_id` to string as follows and therefore Fluent connection was terminating without proper error message. It has taken some time to investigate the exact cause of error but we have identified it now.

image

image

We are creating dictionary manually now.

codegen/allapigen.py
```diff
@@ -60,7 +61,7 @@
     t1 = time()
     print(f"Time to fetch static info: {t1 - t0:.2f} seconds")
     CODEGEN_OUTDIR.mkdir(parents=True, exist_ok=True)
-    print_fluent_version.generate(version, solver.scheme_eval.scheme_eval)
+    print_fluent_version(solver._app_utilities)
```
seanpearsonuk, line 64: :+1:

seanpearsonuk:
Is the AppUtilities interface implemented differently for older (pre 252) versions of Fluent?

hpohekar:
> @hpohekar Is the AppUtilities interface implemented differently for older (pre 252) versions of Fluent?

@seanpearsonuk No. We don't have such cases for app utilities.

seanpearsonuk:
> > @hpohekar Is the AppUtilities interface implemented differently for older (pre 252) versions of Fluent?
> 
> @seanpearsonuk No. We don't have such cases for app utilities.
@hpohekar I need some help to understand the design. How are we still supporting the older Fluent versions with these changes? Aren't we switching code that depended on scheme_eval etc to now depend on app_utilities?

hpohekar:
> > > @hpohekar Is the AppUtilities interface implemented differently for older (pre 252) versions of Fluent?
> > 
> > 
> > @seanpearsonuk No. We don't have such cases for app utilities.
> > @hpohekar I need some help to understand the design. How are we still supporting the older Fluent versions with these changes? Aren't we switching code that depended on scheme_eval etc to now depend on app_utilities?

@seanpearsonuk The changes we have made in this PR are independent of Fluent versions.

mkundu1:
> @hpohekar Is the AppUtilities interface implemented differently for older (pre 252) versions of Fluent?

Good point, eventually we'll need another implementation of AppUtilities using scheme-eval to get it working with Fluent version < 25.2. We'll keep the usage code of AppUtilities unchanged, abstracting the version-dependenent implementation inside the services module.

@hpohekar

hpohekar:
> > @hpohekar Is the AppUtilities interface implemented differently for older (pre 252) versions of Fluent?
> 
> Good point, eventually we'll need another implementation of AppUtilities using scheme-eval to get it working with Fluent version < 25.2. We'll keep the usage code of AppUtilities unchanged, abstracting the version-dependenent implementation inside the services module.
> 
> @hpohekar

@seanpearsonuk @mkundu1 

Yes, got it now. `app_utilities.proto` will be available in the 25R2 only. Yes, I agree with both of you.

src/ansys/fluent/core/services/app_utilities.py
```diff
@@ -0,0 +1,252 @@
...
+        self.service = service
+
+    def get_product_version(self, with_patch: bool = True) -> Any:
+        """Get product version."""
+        request = AppUtilitiesProtoModule.GetProductVersionRequest()
+        response = self.service.get_product_version(request)
+        if with_patch:
+            return f"{response.major}.{response.minor}.{response.patch}"
+        else:
+            return f"{response.major}{response.minor}"
```
seanpearsonuk, line 140: @hpohekar The formatting seems to be wrong here.

src/ansys/fluent/core/services/app_utilities.py
```diff
@@ -0,0 +1,252 @@
...
+
+
+class AppUtilities:
+    """AppUtilities."""
+
+    def __init__(self, service: AppUtilitiesService):
+        """__init__ method of AppUtilities class."""
+        self.service = service
+
+    def get_product_version(self, with_patch: bool = True) -> Any:
```
seanpearsonuk, line 133: `with_patch` obfuscates this interface. Why not simply return a `FluentVersion` object and allow the caller to take the information they want.

src/ansys/fluent/core/services/app_utilities.py
```diff
@@ -0,0 +1,252 @@
...
+        request = AppUtilitiesProtoModule.StartPythonJournalRequest()
+        request.journal_name = journal_name
+        response = self.service.start_python_journal(request)
+        return response
+
+    def stop_python_journal(self) -> Any:
+        """Stop python journal."""
+        request = AppUtilitiesProtoModule.StopPythonJournalRequest()
+        response = self.service.stop_python_journal(request)
+        return response
```
seanpearsonuk, line 201: @hpohekar These two methods return gRPC responses, which should not be the case.

src/ansys/fluent/core/services/app_utilities.py
```diff
@@ -0,0 +1,252 @@
...
+        request = AppUtilitiesProtoModule.UnregisterPauseOnSolutionEventsRequest()
+        request.registration_id = registration_id
+        response = self.service.unregister_pause_on_solution_events(request)
+        return response
+
+    def exit(self) -> Any:
+        """Exit."""
+        request = AppUtilitiesProtoModule.ExitRequest()
+        response = self.service.exit(request)
+        return response
```
seanpearsonuk, line 252: @hpohekar These three methods return gRPC responses.

src/ansys/fluent/core/services/settings.py
```diff
@@ -367,9 +371,13 @@ def get_attrs(self, path: str, attrs: list[str], recursive: bool = False) -> Any
...
-        return self._scheme_eval.is_defined(
-            "has-fnmatch-wild-card?"
-        ) and self._scheme_eval.scheme_eval(f'(has-fnmatch-wild-card? "{name}")')
+        if FluentVersion(self.scheme_eval.version) < FluentVersion.v252:
+            has_wildcard = self._scheme_eval.scheme_eval(
+                f'(has-fnmatch-wild-card? "{name}")'
+            )
+        else:
+            has_wildcard = self._app_utilities.is_wildcard(name)
+        return self._scheme_eval.is_defined("has-fnmatch-wild-card?") and has_wildcard
```
seanpearsonuk, line 380: @hpohekar It looks like it shouldn't be too difficult to encapsulate this in the `AppUtilities` object.

seanpearsonuk:
> > > @hpohekar Is the AppUtilities interface implemented differently for older (pre 252) versions of Fluent?
> > 
> > 
> > Good point, eventually we'll need another implementation of AppUtilities using scheme-eval to get it working with Fluent version < 25.2. We'll keep the usage code of AppUtilities unchanged, abstracting the version-dependenent implementation inside the services module.
> > @hpohekar
> 
> @seanpearsonuk @mkundu1
> 
> Yes, got it now. `app_utilities.proto` will be available in the 25R2 only not in the prior versions. Yes, I agree with both of you. I considered ansys-api-fluent only, for the availability of the `app_utilities.proto` but it is present in Fluent as well.

Finally, to answer my own question, the following is the kind of change we are putting in all the calling code. This was my assumption, but I don't think that tis question is being addressed anywhere.

image

src/ansys/fluent/core/services/app_utilities.py
hpohekar: The attempt to return `FluentVersion` gives circular import issue we have removed `with_patch` and we are returning the string now.

src/ansys/fluent/core/services/settings.py
hpohekar: @seanpearsonuk What do we want to encapsulate in the `AppUtilities` object ?

hpohekar:
> > > > @hpohekar Is the AppUtilities interface implemented differently for older (pre 252) versions of Fluent?
> > > 
> > > 
> > > Good point, eventually we'll need another implementation of AppUtilities using scheme-eval to get it working with Fluent version < 25.2. We'll keep the usage code of AppUtilities unchanged, abstracting the version-dependenent implementation inside the services module.
> > > @hpohekar
> > 
> > 
> > @seanpearsonuk @mkundu1
> > Yes, got it now. `app_utilities.proto` will be available in the 25R2 only not in the prior versions. Yes, I agree with both of you. I considered ansys-api-fluent only, for the availability of the `app_utilities.proto` but it is present in Fluent as well.
> 
> Finally, to answer my own question, the following (a commit from today) is the kind of change we are putting in all the calling code. This was my assumption, but I don't think that this question is being addressed anywhere.
> 
> image
> 
> Two issues here:
> 
> * We have not been clear about the approach.
> * I am not convinced that this approach is more efficient than simply implementing a flexible `AppUtilities` class from the outset.

@seanpearsonuk 

We are using existing pattern if it's Fluent version older than 25R2 and app utilities otherwise.

It's not clear to me what exactly means flexible. We will have to evaluate the Fluent version everywhere to switch between scheme eval and app utilities so it's not clear how can we make it more efficient.

seanpearsonuk:
> > > > > @hpohekar Is the AppUtilities interface implemented differently for older (pre 252) versions of Fluent?
> > > > 
> > > > 
> > > > Good point, eventually we'll need another implementation of AppUtilities using scheme-eval to get it working with Fluent version < 25.2. We'll keep the usage code of AppUtilities unchanged, abstracting the version-dependenent implementation inside the services module.
> > > > @hpohekar
> > > 
> > > 
> > > @seanpearsonuk @mkundu1
> > > Yes, got it now. `app_utilities.proto` will be available in the 25R2 only not in the prior versions. Yes, I agree with both of you. I considered ansys-api-fluent only, for the availability of the `app_utilities.proto` but it is present in Fluent as well.
> > 
> > 
> > Finally, to answer my own question, the following (a commit from today) is the kind of change we are putting in all the calling code. This was my assumption, but I don't think that this question is being addressed anywhere.
> > image
> > Two issues here:
> > 
> > * We have not been clear about the approach.
> > * I am not convinced that this approach is more efficient than simply implementing a flexible `AppUtilities` class from the outset.
> 
> @seanpearsonuk
> 
> We are using existing pattern if it's Fluent version older than 25R2 and app utilities otherwise.
> 
> Do we want to use it as follows ? In the following case we are evaluating the Fluent version only once.
> 
> ```
> class AppUtilitiesOld:
>     """AppUtilitiesOld."""
> 
>     def __init__(self, service: SchemeEvalService):
>         """__init__ method of AppUtilitiesOld class."""
>         self.service = service
> ```
> 
> ```
> if FluentVersion(self.scheme_eval.version) < FluentVersion.252:
>             self._app_utilities = service_creator("app_utilities_old").create(
>                 self._app_utilities_service
>             )
> else:
>     self._app_utilities = service_creator("app_utilities").create(
>         self._app_utilities_service
>     )
> ```

`AppUtilities` would have two different implementations wit the same interface.

src/ansys/fluent/core/services/settings.py
seanpearsonuk: @hpohekar we want to hide the fact that there are different implementations depending on the Fluent version.

src/ansys/fluent/core/services/app_utilities.py
```diff
@@ -0,0 +1,247 @@
...
+
+from enum import Enum
+from typing import List, Tuple
+
+from google.protobuf.json_format import MessageToDict
+import grpc
+
+from ansys.api.fluent.v0 import app_utilities_pb2 as AppUtilitiesProtoModule
+from ansys.api.fluent.v0 import app_utilities_pb2_grpc as AppUtilitiesGrpcModule
+import ansys.fluent.core as pyfluent
```
seanpearsonuk, line 11: @hpohekar Please focus your imports to the things thath your code needs. We should not see this anywhere inside the code.

src/ansys/fluent/core/services/app_utilities.py
seanpearsonuk: `FluentVersion` should be usable without any such issues. There is no intrinsic problem there. Are you provoking this through your imports in this module? I commented earlier on the unnecessary import of `ansys.fluent.core`.

If you reject a superior design for such issues then it's better to comment it directly because reviewers miss these things otherwise.

hpohekar: @seanpearsonuk I'm not rejecting this design I had stated the issue that I was facing prior to localize imports. Now I have imported the ansys.fluent.core as per the requirement only instead of importing at top module level import this issue has been resolved and I have updated the code accordingly.

mkundu1: @seanpearsonuk @hpohekar `FluentVersion` doesn't support the patch version field. Also, `FluentVersion` is restricted to only few version numbers and won't work generally if this module is reused outside PyFluent.

I think it is better of simply return a tuple from this method.

hpohekar: @seanpearsonuk @mkundu1 

Actually I'm trying to make the current scheme eval usage and app utilities changes consistent first. I will address all suggestions in subsequent PRs after merging this PR.

hpohekar: https://github.com/ansys/pyfluent/issues/3579

hpohekar: > @seanpearsonuk @hpohekar `FluentVersion` doesn't support the patch version field. Also, `FluentVersion` is restricted to only few version numbers and won't work generally if this module is reused outside PyFluent.
> 
> I think it is better of simply return a tuple from this method.
> 
> `f"{response.major}.{response.minor}.{response.patch}"` is also fine and as per the [semvar syntax](https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions).

@mkundu1 Ok. Thanks a lot.

hpohekar: @seanpearsonuk We will revert this to the original state.

src/ansys/fluent/core/services/app_utilities.py
```diff
@@ -0,0 +1,402 @@
...
+    def get_controller_process_info(self) -> dict:
+        """Get controller process info."""
+        cortex_host = self.scheme_eval.scheme_eval("(cx-cortex-host)")
+        cortex_pid = self.scheme_eval.scheme_eval("(cx-cortex-id)")
+        cortex_pwd = self.scheme_eval.scheme_eval("(cortex-pwd)")
+        return {
+            "hostname": cortex_host,
+            "process_id": cortex_pid,
+            "working_directory": cortex_pwd,
+        }
```
hpohekar, line 159: @seanpearsonuk @mkundu1 

AppUtilities and AppUtilitiesOld are consistent with interface,  signature of methods. AppUtilities uses AppUtilitiesService and AppUtilitiesOld uses scheme_eval service, both are consistent in the output.

seanpearsonuk: Awesome, thanks, @hpohekar. Is there a test in place to show that we are consistent before and since 25R2?

seanpearsonuk:
Having difficulty finding GitHub comment after email notifications. In brief, I agree with this, thanks:
"
@seanpearsonuk @hpohekar FluentVersion doesn't support the patch version field. Also, FluentVersion is restricted to only few version numbers and won't work generally if this module is reused outside PyFluent.

I think it is better of simply return a tuple from this method.
"

src/ansys/fluent/core/services/app_utilities.py
hpohekar: @seanpearsonuk Actually the existing tests are already verifying this as we are using Fluent version markers and codegen step for each version.

hpohekar:
> Having difficulty finding GitHub comment after email notifications. In brief, I agree with this, thanks: " @seanpearsonuk @hpohekar FluentVersion doesn't support the patch version field. Also, FluentVersion is restricted to only few version numbers and won't work generally if this module is reused outside PyFluent.
> 
> I think it is better of simply return a tuple from this method. "

@seanpearsonuk We have reverted it to the original state. i.e.  `return f"{response.major}.{response.minor}.{response.patch}"`

image

.pre-commit-config.yaml
```diff
@@ -13,7 +13,7 @@ repos:
   hooks:
   - id: black
     args: [
-      --target-version=py39
+      --target-version=py310
```
mkundu1, line 16: 👍

src/ansys/fluent/core/session.py
```diff
@@ -133,10 +136,23 @@ def _build_from_fluent_connection(
         if self._start_transcript:
             self.transcript.start()
 
+        if FluentVersion(self.scheme_eval.version) < FluentVersion.v252:
+            self._app_utilities = AppUtilitiesOld(self.scheme_eval)
+        else:
+            self._app_utilities_service = self._fluent_connection.create_grpc_service(
+                AppUtilitiesService, self._error_state
+            )
+            self._app_utilities = service_creator("app_utilities").create(
```
mkundu1, line 145: I think this is already initialized within FluentConnection?

src/ansys/fluent/core/session.py
```diff
@@ -175,13 +191,17 @@ class Fields:
 
             def __init__(self, _session):
                 """Initialize Fields."""
+                self.is_data_valid = (
```
mkundu1, line 194: Is this getting evaluated here? We need to pass a callable to the service instances.

src/ansys/fluent/core/session.py
hpohekar: @mkundu1 Right. Thank you.

src/ansys/fluent/core/session.py
hpohekar: @mkundu1 Yes, right. Done. Thank you.

```
>>> solver = launch_fluent(ui_mode="gui")
>>> solver._app_utilities.is_solution_data_available()
False
>>> solver.fields._is_solution_data_valid()
False
>>> Fast-loading "D:\ANSYSDev\vNNN\fluent\fluent25.2\\addons\afd\lib\hdfio.bin"
Done.

Reading from AAPQ13pUvWxQtxN:"D:/Repos/example-data/pyfluent/mixing_elbow/mixing_elbow.cas.h5" in NODE0 mode ...
  Reading mesh ...
       66417 nodes,     1 node zone  ...
Warning: reading 2 partition grid onto 1 compute node machine.
         Combining every 2 partitions.
  Done.


Building...
     mesh
     materials,
     interface,
     domains,
        mixture
     zones,
        wall-elbow
        wall-inlet
        outlet
        cold-inlet
        hot-inlet
        symmetry-xyplane
        interior--elbow-fluid
        elbow-fluid
     parallel,
Done.

Reading from AAPQ13pUvWxQtxN:"D:/Repos/example-data/pyfluent/mixing_elbow/mixing_elbow.dat.h5" in NODE0 mode ...


  Reading results.
Parallel variables...
Done.

>>> solver._app_utilities.is_solution_data_available()
True
>>> solver.fields._is_solution_data_valid()
True
>>>
```

src/ansys/fluent/core/services/app_utilities.py
hpohekar: @seanpearsonuk We have added tests for app utilities.

src/ansys/fluent/core/session.py
```diff
@@ -133,10 +133,20 @@ def _build_from_fluent_connection(
         if self._start_transcript:
             self.transcript.start()
 
+        if FluentVersion(self.scheme_eval.version) < FluentVersion.v252:
+            self._app_utilities = AppUtilitiesOld(self.scheme_eval)
+        else:
+            self._app_utilities = (
+                self._fluent_connection._connection_interface._app_utilities
+            )
```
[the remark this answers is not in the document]
hpohekar: @seanpearsonuk We will address this in next PR.

seanpearsonuk, line 141:
```suggestion
        # TODO app_utilities module to be the APPUtilities factory
        # so that these complications will be hidden from calling code.
        if FluentVersion(self.scheme_eval.version) < FluentVersion.v252:
            self._app_utilities = AppUtilitiesOld(self.scheme_eval)
        else:
            self._app_utilities = (
                self._fluent_connection._connection_interface._app_utilities
            )
```

hpohekar: @seanpearsonuk We have created an issue to track this.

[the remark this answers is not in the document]
hpohekar: https://github.com/ansys/pyfluent/issues/3593

seanpearsonuk (approved):

PR #3567, opened 2024-12-11
test: Add tests for new datamodel API

mkundu1:
Adding StateEngine unittests corresponding to the new datamodel API. 

Added test_datamodel_api.py for testDataModelApi.cpp and testMappedApi.cpp. I've skipped testMFWMappedApi.cpp as this seems to be at workflow level - we can add it after updating the workflow classes based on the new API.

Some issues found are mentoned in TODO comments in the test code - will address them in separate PRs.

### DIFF
--- a/src/ansys/fluent/core/streaming_services/datamodel_event_streaming.py
+++ b/src/ansys/fluent/core/streaming_services/datamodel_event_streaming.py
@@ -66,12 +66,12 @@ class DatamodelEvents(StreamingService):
                         elif response.HasField("commandAttributeChangedEventResponse"):
                             value = response.commandAttributeChangedEventResponse.value
                             cb[1](_convert_variant_to_value(value))
-                        elif (
-                            response.HasField("modifiedEventResponse")
-                            or response.HasField("deletedEventResponse")
-                            or response.HasField("affectedEventResponse")
-                        ):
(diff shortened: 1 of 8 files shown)


src/ansys/fluent/core/streaming_services/datamodel_event_streaming.py
```diff
@@ -66,12 +66,12 @@ def _process_streaming(self, id, stream_begin_method, started_evt, *args, **kwar
...
-                            response.HasField("modifiedEventResponse")
-                            or response.HasField("deletedEventResponse")
-                            or response.HasField("affectedEventResponse")
-                        ):
+                        elif response.HasField(
+                            "modifiedEventResponse"
+                        ) or response.HasField("affectedEventResponse"):
                             cb[1](cb[0])
+                        elif response.HasField("deletedEventResponse"):
+                            cb[1]()
```
mkundu1, line 74: We should not capture the object which is going to be deleted, will update the PyFluent API.

mkundu1: https://github.com/ansys/pyfluent/issues/3570

hpohekar (approved):

PR #3568, opened 2024-12-12
feat: Add units and units_quantity in FieldInfo message

hpohekar:
closes https://github.com/ansys/pyfluent/issues/3569

As title says


src/ansys/fluent/core/post_objects/post_helper.py
```diff
@@ -131,11 +134,21 @@ def get_vector_fields(self):
 
     def get_field_unit(self, field):
         """Returns the unit of the field."""
-        quantity = self._field_unit_quantity(field)
-        if quantity == "*null*":
-            return ""
-        scheme_eval_str = f"(units/get-pretty-wb-units-from-dimension (units/inquire-dimension '{quantity}))"
-        return " ".join(self._scheme_str_to_py_list(scheme_eval_str))
+        session = self.obj.get_root().session
+        if hasattr(session, "scheme_eval"):
```
mkundu1, line 138: not sure why this check is required, we are already checking the fluent version to choose between scheme_eval and field_data based implementation

src/ansys/fluent/core/post_objects/post_helper.py
```diff
@@ -131,11 +134,21 @@ def get_vector_fields(self):
...
+            if FluentVersion(session.scheme_eval.version) < FluentVersion.v252:
```
mkundu1, line 139: Actually, it may be possible to encapsulate this inside FieldData service class.

class FieldData:
    def __init__(..., version, scheme_eval=None):
    ...
    def get_field_info(...):
    # implement either through scheme_eval or field_data service based on version

src/ansys/fluent/core/post_objects/post_helper.py
hpohekar: This check is present is `_scheme_str_to_py_list` therefore we have followed it. I think we can remove it from there as well.

src/ansys/fluent/core/post_objects/post_helper.py
hpohekar: @mkundu1 We are not using scheme eval to get field info. We are using scheme eval to get units only. We have encapsulated get_field_info in the FieldInfo class already.

prmukherj (approved):

PR #3576, opened 2024-12-16
feat: Return launch and connection info from standalone launcher dry-run

mkundu1:
Use case of the `dry_run` argument of the `launch_fluent()` to use with external launcher for launching Fluent in a different machine.

1. Assume a shared directory between the server machine (where Fluent is launched) and client machine (where PyFluent is run) is available at path `/mnt/shared` in the server machine and `\\server\shared` in the client machine.
2. Set the environment variable `SERVER_INFO_DIR` to `/mnt/shared` in the server machine and `\\server\shared` in the client machine.
3. In client, `launch_fluent(fluent_path=<path to fluent exe as in the server machine>, dry_run=True)` will return a tuple containing Fluent launch string and the absolute path of the server info file in client.
4. Fluent is launched by the external launcher using the Fluent launch string returned above.
5. In client, PyFluent connects using `connect_to_fluent()` with the server info file path returned above.

### DIFF
--- a/src/ansys/fluent/core/launcher/process_launch_string.py
+++ b/src/ansys/fluent/core/launcher/process_launch_string.py
@@ -137,7 +137,14 @@ def get_fluent_exe_path(**launch_argvals) -> Path:
             return fluent_root / "bin" / "fluent"
 
     # Look for Fluent exe path in the following order:
-    # 1. product_version parameter passed with launch_fluent
+    # 1. Custom Path provided by the user in launch_fluent
+    fluent_path = launch_argvals.get("fluent_path")
+    if fluent_path:
+        # Return the fluent_path string verbatim. The path may not even exist
+        # in the current machine if user wants to launch fluent externally (dry_run use case).
+        return fluent_path
+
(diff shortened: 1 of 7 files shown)


src/ansys/fluent/core/launcher/process_launch_string.py
```diff
@@ -137,7 +137,14 @@ def get_exe_path(fluent_root: Path) -> Path:
             return fluent_root / "bin" / "fluent"
 
     # Look for Fluent exe path in the following order:
-    # 1. product_version parameter passed with launch_fluent
+    # 1. Custom Path provided by the user in launch_fluent
```
mkundu1, line 140: The custom Fluent path will have the highest priority (this will also be useful while running codegen during the Fluent build)

hpohekar (approved):

prmukherj (approved):

src/ansys/fluent/core/launcher/process_launch_string.py
prmukherj: yes, this makes sense, Thank you @mkundu1.

PR #3577, opened 2024-12-17
feat: Walk API classes

seanpearsonuk:
Add a tool to walk static API classes, already being used for analysing the API through separate code.

### DIFF
--- a/src/ansys/fluent/core/codegen/walk_api.py
+++ b/src/ansys/fluent/core/codegen/walk_api.py
@@ -30,11 +30,15 @@ def walk_api(api_root_cls, on_each_path, current_path: str | List[str] = ""):
         on_each_path(current_path)
 
     # Get child names and their respective classes
-    child_names = getattr(api_root_cls, "child_names", [])
+    all_names = [
+        name
+        for attr in ("child_names", "argument_names", "command_names", "query_names")
+        for name in getattr(api_root_cls, attr, [])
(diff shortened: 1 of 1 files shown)


mkundu1:
@seanpearsonuk For NamedObject and ListObject, we have to get the `child_object_type` property which is a Group and then traverse inside that Group class.

src/ansys/fluent/core/codegen/walk_api.py
```diff
@@ -0,0 +1,48 @@
...
+    # Skip the root path
+    if current_path:
+        on_each_path(current_path)
+
+    # Get child names and their respective classes
+    child_names = getattr(api_root_cls, "child_names", [])
+    child_classes = getattr(api_root_cls, "_child_classes", {})
+
+    # Traverse each child
+    for child_name in child_names:
```
mkundu1, line 37: We also need to loop over `command_names` and `query_names`

mkundu1: and `argument_names` for Command and Query classes

seanpearsonuk: In a later version, I also want to pass each item's base type info to the callback for type-specific checks.

mkundu1 (approved):

prmukherj (approved):

PR #3606, opened 2025-01-02
feat: allow for pyansys-units dev version (0.4.dev0)

RobPasMue:
This will allow to use the dev version of PyAnsys units, otherwise it would not be possible


hpohekar (approved):

RobPasMue:
Looks like there are some issues with some tests in v25.2?

prmukherj:
> Looks like there are some issues with some tests in v25.2?

Yes, already looking into it.

PR #3624, opened 2025-01-08
feat: Datamodel cb signature for various events

mkundu1:
Details in comment.

### DIFF
--- a/src/ansys/fluent/core/services/datamodel_se.py
+++ b/src/ansys/fluent/core/services/datamodel_se.py
@@ -725,7 +725,7 @@ class DatamodelService(StreamingService):
         self.subscriptions.unsubscribe_all()
 
     def add_on_child_created(
-        self, rules: str, path: str, child_type: str, obj, cb: Callable[[str], None]
+        self, rules: str, path: str, child_type: str, cb: Callable[[str], None]
     ) -> EventSubscription:
         """Add on child created."""
         request_dict = {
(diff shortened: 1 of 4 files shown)


src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1253,23 +1285,32 @@ def create_command_arguments(self, command: str) -> str:
             self.rules, convert_path_to_se_path(self.path), command
         )
 
-    def add_on_child_created(self, child_type: str, cb: Callable) -> EventSubscription:
+    def add_on_child_created(
```
mkundu1, line 1288: This is the user-facing function and has identical implementation in PyFluent and PyConsole. In the user-facing function the callback takes an object of PyFluent's PyMenu class in both PyFluent and PyConsole. The user-facing workflow classes in PyFluent wraps further this function.

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1253,23 +1285,32 @@ def create_command_arguments(self, command: str) -> str:
...
+        cb : Callable[[PyMenuT], None]
             Callback function
 
         Returns
         -------
         EventSubscription
             EventSubscription instance which can be used to unregister the callback
         """
+
+        def cb_service(child_path: str):
```
mkundu1, line 1306: In the `add_on_child_created` method in the service class, the callback takes the child path.

I'm not sure if it is possible to pass either object or child path to the service level callback as we have only one implementation of the `add_on_child_created` method of the PyMenu class.

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -722,7 +723,7 @@ def unsubscribe_all_events(self) -> None:
         self.subscriptions.unsubscribe_all()
 
     def add_on_child_created(
```
mkundu1, line 725: This is the `add_on_child_created` method of the service class which has a different implmentation between PyFluent, old PyConsole and new PyConsole. In this function the callback takes the child path.

For the new PyConsole implementation, the callback function doesn't require any transformation, it can be passed directly to PyStateEngine/C++.

We need to update the old PyConsole implmentation to transform the callback function to take a PyStateEngine's ModelState object.

seanpearsonuk (approved):

hpohekar (approved):

mkundu1:
This will require a Fluent side change.

PR #3628, opened 2025-01-09
fix: Don't import error_details proto in Fluent's python environment

mkundu1:


src/ansys/fluent/core/services/interceptors.py
```diff
@@ -116,6 +114,9 @@ def _intercept_call(
             ex = response.exception()
             new_ex_cls = RuntimeError
             try:
+                from google.rpc import error_details_pb2
+                from grpc_status import rpc_status
```
mkundu1, line 118: There is an issue in Fluent's Python environment while importing these module. Fluent's PyConsole won't use this client-side code.

seanpearsonuk (approved):

hpohekar (approved):

prmukherj (approved):

PR #3630, opened 2025-01-09
fix: Map multiple ports to container

mkundu1:


src/ansys/fluent/core/launcher/fluent_container.py
```diff
@@ -239,16 +239,23 @@ def configure_container_dict(
...
         port = get_free_port()
+        port_mapping = {port: port}
 
-    container_dict.update(ports={str(port): port})  # container port : host port
+    container_dict.update(
+        ports={str(x): y for x, y in port_mapping.items()}
+    )  # container port : host port
+    container_grpc_port = next(
+        iter(port_mapping.values())
+    )  # the first port in the mapping is chosen as the gRPC port
```
mkundu1, line 258: @lagacep-ans If multiple ports are sent, the first port will be chosen as the grpc port.

hpohekar (approved):

prmukherj (approved):

seanpearsonuk (approved):

PR #3645, opened 2025-01-15
fix: Fix node indexing for multi-zone case

mkundu1:
Fluent node ids are not contagious within a zone for a multi-zone case.


seanpearsonuk:
> Fluent node ids are not generally contagious within a zone for a multi-zone case.

contiguous?

seanpearsonuk (approved):

hpohekar (approved):

prmukherj (approved):

Issue #3634, opened 2025-01-10
Higher-level exceptions


seanpearsonuk:
Investigate PyFluent behaviour to find scenarios where we could catch and rethrow to provide improved error information to users. See this post: https://softwareengineering.stackexchange.com/a/231059.

Note some usage examples where the first example could reraise an exception closer to the user's context:

```
>>> pf.connect_to_fluent(ip="1.2.3.4",  port=5555)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\ANSYSDev\pyfluent\src\ansys\fluent\core\launcher\launcher.py", line 349, in connect_to_fluent
    fluent_connection = FluentConnection(
  File "C:\ANSYSDev\pyfluent\src\ansys\fluent\core\fluent_connection.py", line 417, in __init__
    self.health_check.check_health()
  File "C:\ANSYSDev\pyfluent\src\ansys\fluent\core\services\health_check.py", line 62, in check_health
    response = self._stub.Check(request, metadata=self._metadata)
  File "C:\ANSYSDev\pyfluent\.env-local-2\lib\site-packages\grpc\_interceptor.py", line 277, in __call__
    response, ignored_call = self._with_call(
  File "C:\ANSYSDev\pyfluent\.env-local-2\lib\site-packages\grpc\_interceptor.py", line 329, in _with_call
    call = self._interceptor.intercept_unary_unary(
  File "C:\ANSYSDev\pyfluent\src\ansys\fluent\core\services\interceptors.py", line 124, in intercept_unary_unary
    return self._intercept_call(continuation, client_call_details, request)
  File "C:\ANSYSDev\pyfluent\src\ansys\fluent\core\services\interceptors.py", line 114, in _intercept_call
    raise new_ex from None
RuntimeError: failed to connect to all addresses; last error: UNKNOWN: ipv4:1.2.3.4:5555: tcp handshaker shutdown
>>>
>>> pf.connect_to_fluent(ip="1.2.3.4")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\ANSYSDev\pyfluent\src\ansys\fluent\core\launcher\launcher.py", line 348, in connect_to_fluent
    ip, port, password = _get_server_info(server_info_file_name, ip, port, password)
  File "C:\ANSYSDev\pyfluent\src\ansys\fluent\core\launcher\server_info.py", line 56, in _get_server_info
    raise IpPortNotProvided()
[...]

PR #3646, opened 2025-01-16
feat: Improved error handling

hpohekar:
closes https://github.com/ansys/pyfluent/issues/3634

Check whether provided `ip` and `port` are valid.

```
# Incorrect ip and port
>>> s = pf.connect_to_fluent(ip="1.2.3.4", port=5555)
  File "D:\pyfluent\src\ansys\fluent\core\launcher\launcher.py", line 361, in connect_to_fluent
    _check_ip_port_2(ip=ip, port=port)
  File "D:\pyfluent\src\ansys\fluent\core\launcher\launcher.py", line 92, in _check_ip_port_2
    raise exceptions.IncorrectIpPortProvided()
ansys.fluent.core.launcher.error_handler.IncorrectIpPortProvided: Provide correct 'ip' and 'port'.
```


src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -345,6 +361,7 @@ def connect_to_fluent(
     :class:`~ansys.fluent.core.session_solver_icing.SolverIcing`]
         Session object.
     """
+    _check_ip_port(ip=ip, port=port)
```
seanpearsonuk, line 364: @hpohekar I belive that if we are to do this (which seems like a reasonable idea) then we should move it one line down to catch the scenario where an out-of-date server file is specified.

hpohekar: @seanpearsonuk Yes, I agree. Actually this should be inside `_get_server_info`. We have moved it there. Thank you.

src/ansys/fluent/core/launcher/error_handler.py
```diff
@@ -29,6 +29,14 @@ def __init__(self):
         super().__init__("Provide either 'ip' and 'port' or 'server_info_file_name'.")
 
 
+class IncorrectIpPortProvided(ValueError):
+    """Raised when IP address and port are incorrect."""
+
+    def __init__(self):
+        """Initialize IncorrectIpPortProvided."""
+        super().__init__("Provide correct 'ip' and 'port'.")
```
hpohekar, line 37:
```suggestion
        super().__init__("Provide correct 'ip' and 'port' or 'server_info_file_name'.")
```

src/ansys/fluent/core/launcher/error_handler.py
```diff
@@ -29,6 +29,14 @@ def __init__(self):
         super().__init__("Provide either 'ip' and 'port' or 'server_info_file_name'.")
 
 
+class IncorrectIpPortProvided(ValueError):
+    """Raised when IP address and port are incorrect."""
+
+    def __init__(self):
+        """Initialize IncorrectIpPortProvided."""
+        super().__init__("Provide correct 'ip' and 'port' or 'server_info_file_name'.")
```
seanpearsonuk, line 37:
```suggestion
        super().__init__("Provide a valid 'ip' and 'port' or a 'server_info_file_name'.")
```
I'm also not sure if 'server_info_file_name' should be mentioned as a specific argument name like this because it could appear out of context (for instance if PyFluent is used in a higher-level workflow)?

src/ansys/fluent/core/launcher/server_info.py
```diff
@@ -45,6 +48,21 @@ def _get_server_info_file_names(use_tmpdir=True) -> tuple[str, str]:
...
+    """Check if a port is open on a given IP address."""
+    import socket
+
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    sock.settimeout(2)
+    try:
+        default_port = port if port else 0000
+        result = sock.connect_ex((ip, default_port))
+        if result != 0:
+            raise IncorrectIpPortProvided()
```
seanpearsonuk, line 61:
```suggestion
            raise InvalidIpPort()
```

src/ansys/fluent/core/launcher/server_info.py
```diff
@@ -68,4 +86,6 @@ def _get_server_info(
     if not port:
         raise PortNotProvided()
 
+    _check_ip_port(ip=ip, port=port)
```
seanpearsonuk, line 89: The above check can be rolled into the new function - it is anyway a pre-requisite of testing the connection.

src/ansys/fluent/core/launcher/error_handler.py
hpohekar: Ok. Removed.

src/ansys/fluent/core/launcher/server_info.py
```diff
@@ -45,6 +48,21 @@ def _get_server_info_file_names(use_tmpdir=True) -> tuple[str, str]:
...
 
+def _check_ip_port(ip: str, port: int):
+    """Check if a port is open on a given IP address."""
+    import socket
+
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    sock.settimeout(2)
+    try:
+        default_port = port if port else 0000
+        result = sock.connect_ex((ip, default_port))
```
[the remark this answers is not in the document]
hpohekar: Port should not be of `None` type. It always should be an integer. Considering this scenario we are setting a default port so that `_check_ip_port` is self contained and does not depend on any other external validation as you mentioned above.

src/ansys/fluent/core/launcher/server_info.py
hpohekar: Done. Thank you.

seanpearsonuk (approved):

Issue #3623, opened 2025-01-08
Fluent container freezing - possibly due to launch arguments?


raph-luc:
Fluent 24.1 was freezing in the middle of iterations: 
https://github.com/ansys/pyansys-workflows/actions/runs/12675239102/job/35325797089
https://github.com/ansys/pyansys-workflows/actions/runs/12584809146/job/35075221738

After commit which only changed launch arguments: https://github.com/ansys/pyansys-workflows/pull/44/commits/a72cd6b2db1bcbd5b7cd45d6d53f4183dd709323

After this commit, for some reason the run passed: https://github.com/ansys/pyansys-workflows/actions/runs/12676140368/job/35328690975?pr=44

I am guessing there is some inappropriate parsing of the launch arguments

Note: I recall seeing in the command that was sent to the container that `-gu` and/or `-g` were duplicated, this is probably related to the issue

hpohekar:
@seanpearsonuk @mkundu1  @raph-luc 

I'm not able to reproduce this issue.

Refer the following PR commits and workflow runs.

https://github.com/ansys/pyfluent/pull/3647

1. `solver = launch_fluent(cwd="/mnt/pyfluent")`

https://github.com/ansys/pyfluent/actions/runs/12805935881/job/35703331220

2. `solver = launch_fluent(cwd="/mnt/pyfluent", ui_mode="no_gui_or_graphics")`

https://github.com/ansys/pyfluent/actions/runs/12806313514/job/35704493227

3.  `solver = launch_fluent(cwd="/mnt/pyfluent", start_container=True, ui_mode="no_gui_or_graphics")`

https://github.com/ansys/pyfluent/actions/runs/12806570729/job/35705313173

4.  `solver = launch_fluent(cwd="/mnt/pyfluent", start_container=True, ui_mode="no_gui_or_graphics",  cleanup_on_exit=False)`

https://github.com/ansys/pyfluent/actions/runs/12806808585/job/35706053404

raph-luc:
Hi @hpohekar can you instead verify whether the launch arguments are being processed and sent to the container properly?

I think trying to reproduce the freeze is going to be difficult unless we run that exact same case they were using there

hpohekar:
@seanpearsonuk @mkundu1 @raph-luc 

I have verified that Fluent container is not hanging even if `-gu` is duplicated.

[...]

PR #3649, opened 2025-01-16
fix: Remove `-gu` from command

hpohekar:
closes https://github.com/ansys/pyfluent/issues/3623

### DIFF
--- a/src/ansys/fluent/core/launcher/fluent_container.py
+++ b/src/ansys/fluent/core/launcher/fluent_container.py
@@ -343,7 +343,7 @@ def configure_container_dict(
             container_dict["environment"] = {}
         container_dict["environment"]["FLUENT_LAUNCHED_FROM_PYFLUENT"] = "1"
 
-    fluent_commands = ["-gu", f"-sifile={container_server_info_file}"] + args
+    fluent_commands = [f"-sifile={container_server_info_file}"] + args
 
     container_dict_default = {}
     container_dict_default.update(
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -12,7 +12,10 @@ from typing import Any, Dict
 import ansys.fluent.core as pyfluent
 from ansys.fluent.core.fluent_connection import FluentConnection
 from ansys.fluent.core.launcher.container_launcher import DockerLauncher
-from ansys.fluent.core.launcher.launcher_utils import _confirm_watchdog_start
+from ansys.fluent.core.launcher.launcher_utils import (
+    _confirm_watchdog_start,
+    is_windows,
+)
 from ansys.fluent.core.launcher.pim_launcher import PIMLauncher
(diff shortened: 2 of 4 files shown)


mkundu1 (approved):

hpohekar:
@seanpearsonuk @mkundu1 

I think we don't need this PR now.

https://github.com/ansys/pyfluent/issues/3623#issuecomment-2597440219

seanpearsonuk:
> @seanpearsonuk @mkundu1
> 
> I think we don't need this PR now.
> 
> [#3623 (comment)](https://github.com/ansys/pyfluent/issues/3623#issuecomment-2597440219)

If the code is always adding the argument redundantly, then irrespective of the other issue, then this correction should still stand.

raph-luc:
I've noticed that when launching containers on Windows hosts, it is starting them with `-gu` and then `-hidden` as well in the argument list, which is always going to lead into issues. This is causing the launch to fail on my end. The default for containers should be `-gu` regardless of host OS, and it should respect whatever the user specifies for UI mode otherwise.

I think we also need tests covering this, could start with something like:
```
 def test_container_launcher_args():
    container_dict = pyfluent.launch_fluent(start_container=True, dry_run=True)
    commands = container_dict["command"]
    graphics_args = ['-gu', '-hidden', '-g']
    graphics_arg_count = 0
    for arg in graphics_args:
        graphics_arg_count += commands.count(arg)
    assert graphics_arg_count == 1, "Too many graphics arguments"
```

hpohekar:
> I've noticed that when launching containers on Windows hosts, it is starting them with `-gu` and then `-hidden` as well in the argument list. The default for containers should be `-gu` regardless of host OS, and it should respect whatever the user specifies for UI mode otherwise.
> 
> I think we also need tests covering this, could start with something like:
> 
> ```
>  def test_container_launcher_args():
>     container_dict = pyfluent.launch_fluent(start_container=True, dry_run=True)
>     commands = container_dict["command"]
>     graphics_args = ['-gu', '-hidden', '-g']
>     graphics_arg_count = 0
>     for arg in graphics_args:
>         graphics_arg_count += commands.count(arg)
>     assert graphics_arg_count == 1, "Too many graphics arguments"
> ```

That's right.

Current default for windows is "hidden_gui" and linux is "no_gui".

image

prmukherj (approved):

src/ansys/fluent/core/launcher/fluent_container.py
raph-luc, line 1: > Current default for windows is "hidden" and linux is "gu".

@hpohekar Yes that is correct. The Fluent containers always run Linux, therefore as I said in my previous comment, they should always run with `-gu`, unless the user specifies something else. This is not what is happening right now

(Moving to a conversation as this is more organized than isolated PR comments)

hpohekar: Yes, I agree.

hpohekar: @raph-luc Done. Thank you.

Issue #3664, opened 2025-01-21
Get float data for single precision


hpohekar:
Get float data for single precision.

PR #3665, opened 2025-01-21
fix: Get float mesh data for single precision

hpohekar:
closes https://github.com/ansys/pyfluent/issues/3664

Get float data for single precision.


mkundu1 (approved):
Thanks a lot, @hpohekar

Issue #3672, opened 2025-01-27
Pyfluent not starting 2ddp with pim launcher


chrisg564:
### 🔍 Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### 🐞 Description of the bug

From Ansys Notebook we are using pyfluent with the pyansys 25R1 meta package (0.28.2) and 25R1 product pods for solver instances. They are launched with PIM. However users are only able to start `3ddp` and `fluent-meshing` products. I can confirm that `2ddp` is avaiable, but even when we specify dimension=2 it starts `3ddp`. 

```python
# Both start 3ddp
solver = pyfluent.launch_fluent(dimension=Dimension.TWO)

solver = pyfluent.launch_fluent(dimension=2)
```

Running directly from PIM:

<img width="1546" alt="Image" src="https://github.com/user-attachments/assets/b655b2a4-168a-4c8a-8f76-3c810f0cb5f6" />

I suspect that the issue is we are hitting the default `3ddp` product because dimensionality is returning `2` instead of of `2d`, which is what is checked [here](https://github.com/ansys/pyfluent/blob/main/src/ansys/fluent/core/launcher/pim_launcher.py#L232)

### 📝 Steps to reproduce

Go to any PIM enabled environment and try to launch fluent in 2d. 

### 💻 Which operating system are you using?

Linux

### 📀 Which ANSYS version are you using?

25R1 but also experience with 24R2.

### 🐍 Which Python version are you using?

3.11

### 📦 Installed packages

```shell
Ansys packages are sourced from pyansys metapackage 2025.1.1


I0000 00:00:1738012939.033204     114 fork_posix.cc:77] Other threads are currently calling into gRPC, skipping fork() handlers
aiohappyeyeballs==2.4.0
aiohttp==3.10.5
aiosignal==1.3.1
alembic @ file:///home/conda/feedstock_root/build_artifacts/alembic_1694690212886/work
annotated-types==0.7.0
ans.protocol.amk==0.1.0b9
ansys-acp-core==0.1.0
ansys-additive-core==0.19.0
ansys-additive-widgets==0.2.1
ansys-api-acp==0.2.0
ansys-api-additive==2.2.1
ansys-api-dbu==0.3.6
ansys-api-dyna==0.4.2
ansys-api-edb==1.0.10
ansys-api-fluent==0.3.32
ansys-api-geometry==0.4.16
ansys-api-mapdl==0.5.2
[...]

PR #3673, opened 2025-01-28
fix: Dimensionality correction in PIM launcher

hpohekar:
closes https://github.com/ansys/pyfluent/issues/3672

Earlier - 

image

Now - 

image

### DIFF
--- a/src/ansys/fluent/core/launcher/pim_launcher.py
+++ b/src/ansys/fluent/core/launcher/pim_launcher.py
@@ -229,7 +229,7 @@ def launch_remote_fluent(
         product_name=(
             "fluent-meshing"
             if FluentMode.is_meshing(mode)
-            else "fluent-2ddp" if dimensionality == "2d" else "fluent-3ddp"
+            else "fluent-2ddp" if dimensionality == Dimension.TWO else "fluent-3ddp"
         ),
         product_version=product_version,
     )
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/launcher/pim_launcher.py
```diff
@@ -229,7 +229,7 @@ def launch_remote_fluent(
         product_name=(
             "fluent-meshing"
             if FluentMode.is_meshing(mode)
-            else "fluent-2ddp" if dimensionality == "2d" else "fluent-3ddp"
+            else "fluent-2ddp" if dimensionality == Dimension.TWO else "fluent-3ddp"
```
seanpearsonuk, line 232: I think we can also create an issue for future work to factor out code duplication from these launchers.

seanpearsonuk (approved):

src/ansys/fluent/core/launcher/pim_launcher.py
hpohekar: @seanpearsonuk Can you please point out to code duplication in these launchers.

prmukherj (approved):

src/ansys/fluent/core/launcher/pim_launcher.py
seanpearsonuk: Some logic for handling arguments (setting to a chosen default where the argument value is None) appears in more than one place. There is opportunity for consolidation there. I also note that some arguments are unused and I'm not sure if that's intentional in some way.

hpohekar: @seanpearsonuk Ok. Thanks a lot.

Issue #3405, opened 2024-10-18
New module for pyfluent module constants


mkundu1:
Should we move the pyfluent module constants to a separate module `ansys.fluent.core.constants`? This could fix the circular dependency issue we often see while importing constants into modules directly or indirectly imported in `ansys.fluent.core.__init__`. E.g., try to import one of the constants in `src\ansys\fluent\core\launcher\standalone_launcher.py`.

We still want to preserve the usage: `ansys.fluent.core.<CONSTANT_NAME>`.

PR #3677, opened 2025-01-28
feat: New module for PyFluent module constants

hpohekar:
closes https://github.com/ansys/pyfluent/issues/3405


src/ansys/fluent/core/solver/flunits.py
```diff
@@ -104,119 +104,7 @@
...
-    "velocity": "m s^-1",
-    "viscosity": "kg m^-1 s^-1",
-    "voltage": "V",
-    "volume": "m^3",
-    "volume-flow-rate": "m^3 s^-1",
-    "volume-flow-rate-per-depth": "m^3 s^-1 m^-1",
-    "volume-inverse": "m^-3",
-    "youngs-modulus": "N m^-2",
-}
+from ansys.fluent.core.constants import QuantityT, _fl_unit_table
```
seanpearsonuk, line 107: @hpohekar I'm not sure why moving these tables to a common `constants` module is needed(?)  cc: @prmukherj

src/ansys/fluent/core/codegen/datamodelgen.py
```diff
@@ -12,36 +12,14 @@
...
 )
+from ansys.fluent.core.constants import _PY_TYPE_BY_DM_TYPE
 from ansys.fluent.core.utils.fluent_version import (
     FluentVersion,
     get_version_for_file_name,
 )
 
 _ROOT_DIR = Path(__file__) / ".." / ".." / ".." / ".." / ".." / ".."
 
-_PY_TYPE_BY_DM_TYPE = {
```
seanpearsonuk, line 22: @hpohekar Similar to my other comment in `flunits`, I cannot see the benefit of this. cc: @prmukherj

prmukherj: @seanpearsonuk, if you see the task item: https://github.com/ansys/pyfluent/issues/3405, created by Mainak, this will solve a lot of the circular dependency issues. I think that's the only benefit.
cc. @hpohekar

src/ansys/fluent/core/constants.py
```diff
@@ -0,0 +1,259 @@
...
+
+_INDENT_STEP = 4
+
+_logging_file_enabled = False
+
+# Store the top level class names and their data hash.
+# This is used to avoid name collisions and data duplication.
+_NAME_BY_HASH = {}
+
+_PY_TYPE_BY_DM_TYPE = {
```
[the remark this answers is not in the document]
prmukherj: yes, there are some things that needs proper planning to move. Those we can discuss.

src/ansys/fluent/core/codegen/datamodelgen.py
seanpearsonuk: @prmukherj The benefit associated with the proposal in #3405 is not provided by this change. These are private module constants. Would you extract all private module constants from and put them all in one module? Absolutely never. Lets not code like robots.

src/ansys/fluent/core/constants.py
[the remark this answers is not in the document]
seanpearsonuk: What needs to be discussed about these? Take `_NAME_BY_HASH` - it's clearly a non-constant `dict` use and owned by its current module. Nobody is considering/discussing moving it.

hpohekar:
Hi @seanpearsonuk @prmukherj 

There are lot of places where we are importing module or sub-modules inside a function just to avoid the circular import error. I thought this PR will try to improve this handling gracefully but this is not a great idea. 

I agree with your views.

Thanks for sharing your views.

Issue #2307, opened 2023-12-11
Should all `.msh` files in `example-data/pyfluent` be compressed?


ansjsia:
These are the `.msh` files in `example-data/pyfluent`:

| File | Compressed File | Uploader | PR |
|--------|--------|--------|--------|
| `box/poly.msh`						| `box/poly.msh.h5`				 | @akulshre-qa			| https://github.com/ansys/example-data/pull/61 |
| `cavitation/cav.msh`					| `cavitation/cav.msh.gz`			 | @ansaboutin			| https://github.com/ansys/example-data/pull/200 |
| `examples/Brake-Thermal-PyVista-Matplotlib/brake.msh` | `examples/Brake-Thermal-PyVista-Matplotlib/brake.msh.h5` | @hpohekar | https://github.com/ansys/example-data/pull/79
| `exhaust_manifold/manifold.msh`		| `exhaust_manifold/manifold.msh.h5` | @akulshre-qa			| https://github.com/ansys/example-data/pull/61 |
| `fan/fan_bc_vent.msh`					| `fan/fan_bc_vent.msh.h5`			 | @akulshre-qa			| https://github.com/ansys/example-data/pull/61 |
| `lunar_lander_thermal/lander_mesh.msh`| `lunar_lander_thermal/lander_mesh.msh.h5` | @ansjsia | https://github.com/ansys/example-data/pull/207 |
| `nozzle/nozzle_3d.msh`				| `nozzle/nozzle_3d.msh.h5`		 | @akulshre-qa			| https://github.com/ansys/example-data/pull/61 |
| `rotating_disk/disk.msh`				| `rotating_disk/disk.msh.gz`		 | @akulshre-qa			| https://github.com/ansys/example-data/pull/55 |
| `static_mixer/StaticMixer.msh`		| `static_mixer/StaticMixer.msh.h5` | @akulshre-qa			| https://github.com/ansys/example-data/pull/61	|
| `wigley_hull/wigley.msh`				| `wigley_hull/wigley.msh.h5`		 | @akulshre-qa			| https://github.com/ansys/example-data/pull/60 |

**To reduce storage requirements and to avoid slowing down text searches in the `example-data` repo, should all `.msh` files there be removed? Compressed versions of these files have already been uploaded.**

The following PRs have modified/will modify all (as much as I could find) references to `.msh` files in PyFluent repositories to `.msh.gz` or `.msh.h5`:

[...]

PR #3686, opened 2025-01-30
refactor: Update mesh file format

hpohekar:
closes https://github.com/ansys/pyfluent/issues/2307

example-data PR - https://github.com/ansys/example-data/pull/328

### DIFF
--- a/src/ansys/fluent/core/filereader/case_file.py
+++ b/src/ansys/fluent/core/filereader/case_file.py
@@ -618,6 +618,7 @@ class CaseFile(RPVarProcessor):
                     rp_vars_str = rpvars.decode()
             elif (
                 Path(case_file_name).match("*.cas")
+                or Path(case_file_name).match("*.msh")
                 or Path(case_file_name).match("*.msh.h5")
                 or Path(case_file_name).match("*.msh.gz")
             ):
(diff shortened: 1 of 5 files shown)


src/ansys/fluent/core/filereader/case_file.py
```diff
@@ -616,8 +616,10 @@ def __init__(
                     settings = _file["settings"]
                     rpvars = settings["Rampant Variables"][0]
                     rp_vars_str = rpvars.decode()
-            elif Path(case_file_name).match("*.cas") or Path(case_file_name).match(
-                "*.msh"
+            elif (
```
seanpearsonuk, line 619: @hpohekar I'm surprised to see chnges in the file read logic. Please do not end support for .msh files. The goal of this task is only to change the way we are storing the example data such that it's more efficient. cc: @prmukherj

hpohekar: Done. Thank you.

hpohekar: @seanpearsonuk We want to support all 3 *.msh, *.msh.h5 and *.msh.gz right ?

seanpearsonuk: @hpohekar (cc: @prmukherj)

Note that .h5 is already handled above. Ignore the .gz pattern here. In other words, no code changes - just focus on the data file removal. 

If we do happen to find a bug in this code, we should only address it properly with creation of a defect issue and a separate PR with code change plus new unit test(s). 

That approach should apply generally -- we should avoid bundling additional changes into PRs, which is a hazardous strategy. Note also that this code is very sensitive for optiSLang functionality.

hpohekar: @seanpearsonuk Yes, right. Thanks a lot.

seanpearsonuk (approved):

prmukherj (approved):

Issue #2945, opened 2024-06-12
Command to change Fluent working directory


raph-luc:
Suggestion by Andreas (Jun 12): do we already have any "Python native" non-Scheme command to change the Fluent working directory from PyFluent?

Currently I believe we can do that using `session.scheme_eval.scheme_eval("(chdir <directory>)")` or `session.execute_tui(...)`, but a nicer interface would be great.

The PyFluent working directory can be changed using `os.chdir()`.

Note that Scheme's `(chdir)` only changes the Cortex working directory, so there is some additional things to keep in mind here, but this one is enough in most situations.

To make things more intuitive, a `pyfluent.chdir()` command could be added that changes both the PyFluent working directory and Fluent's simultaneously, keeping both synchronized.

Edit: Ryan also requested this (Sep 18)

mkundu1:
I think having a `session.chdir()` will give better control to users. `pyfluent.chdir()` will indicate chdir for all current and future connected sessions.

Related: `session.execute_tui()` can be modified to execute Python code in Fluent. Or we can plan to remove it - we should have less options to execute arbitrary code through grpc.

hpohekar:
Yes, I agree with both this idea and Mainak's suggestion.

mkundu1:
Actually, we need to continue to support `session.execute_tui()` to playback Fluent journals.

dnwillia-work:
Can't you just support `session.execute_tui()` in Fluent's Python console?  I would tend to agree that the remote execution of Scheme calls should now be removed entirely.  Even `session.scheme_eval` should probably go away from the remote interface entirely.

raph-luc:
What do you mean by "Fluent's Python console" in this case? Fluent's Python console is becoming basically just a way to use PyFluent to interact with Fluent, right?

And yeah, not only you can support that (and we already do), but as Mainak said above, I think we won't be able to drop support for `session.execute_tui()` even though it allows execution of some level of arbitrary code. Still, I think an actual Python API command to chdir would be better.

PR #3691, opened 2025-01-31
feat: Change working directory

hpohekar:
closes https://github.com/ansys/pyfluent/issues/2945

```
D:\pyfluent>python
Python 3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)] on win32
>>> solver = launch_fluent()
>>> solver._app_utilities.get_controller_process_info()
{'hostname': 'AAPQ13pUvWxQtxN', 'process_id': 339960, 'working_directory': 'D:\\pyfluent'}
>>> solver._app_utilities.get_solver_process_info()     
{'hostname': 'AAPQ13pUvWxQtxN.win.ansys.com', 'process_id': 350440, 'working_directory': 'D:/pyfluent'}
>>> solver.chdir("D:\pyfluent\codegen")
>>> solver._app_utilities.get_controller_process_info()
{'hostname': 'AAPQ13pUvWxQtxN', 'process_id': 339960, 'working_directory': 'D:\\pyfluent\\codegen'}
>>> solver._app_utilities.get_solver_process_info()     
{'hostname': 'AAPQ13pUvWxQtxN.win.ansys.com', 'process_id': 350440, 'working_directory': 'D:/pyfluent/codegen'}
>>> solver.exit()
```

### DIFF
--- a/src/ansys/fluent/core/session.py
+++ b/src/ansys/fluent/core/session.py
@@ -435,7 +435,7 @@ class BaseSession:
         path : str
             Path of the directory to change.
         """
-        self._app_utilities.change_client_cortex_dir(path)
+        self._app_utilities.set_cortex_working_directory(path)
 
     def __enter__(self):
         return self
--- a/src/ansys/fluent/core/session.py
+++ b/src/ansys/fluent/core/session.py
@@ -435,7 +435,7 @@ class BaseSession:
         path : str
             Path of the directory to change.
         """
-        self._app_utilities.set_cortex_working_directory(path)
+        self._app_utilities.set_working_directory(path)
 
     def __enter__(self):
         return self
--- a/src/ansys/fluent/core/session.py
+++ b/src/ansys/fluent/core/session.py
@@ -427,6 +427,16 @@ class BaseSession:
         if self._file_transfer_service:
             return self._file_transfer_service.download(file_name, local_directory)
 
+    def chdir(self, path: str) -> None:
+        """Change Fluent working directory.
+
+        Parameters
+        ----------
+        path : str
+            Path of the directory to change.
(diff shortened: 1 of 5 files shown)


src/ansys/fluent/core/session.py
```diff
@@ -435,7 +435,7 @@ def chdir(self, path: str) -> None:
         path : str
             Path of the directory to change.
         """
-        self._app_utilities.change_client_cortex_dir(path)
+        self._app_utilities.set_cortex_working_directory(path)
```
seanpearsonuk, line 438: @hpohekar The issue discusses simultaneously changing the client working directory (`os.chdir(...)`) but that would not make the change session-specific so our options appear limited there and I think we should take no additional action in that respect. I think that the user can just call `os.chdir(...)` independently. Perhaps we should document that in the docstring of this method.

hpohekar: @seanpearsonuk `os.chdir()` does not change client working directory. I have already verified it. Only changing Cortex directory is not sufficient. The changed directory should be synchronized with the client as well.

seanpearsonuk: @hpohekar I don't follow what you mean when you say that `os.chdir(...)` does not change the client working directory. I think you mean something different to what you're saying but I don't know what that is. Maybe you mean that even though the working directory has changed, some other side effect that you had assumed would happen has not happened. Please could you help by clarifying.

hpohekar: discussed offline.

seanpearsonuk (approved):

PR #3728, opened 2025-02-10
feat: Use new data-model api.

prmukherj:
Use new data-model api.


src/ansys/fluent/core/launcher/launcher_utils.py
```diff
@@ -64,6 +64,7 @@ def _get_subprocess_kwargs_for_fluent(env: Dict[str, Any], argvals) -> Dict[str,
     if pyfluent.CLEAR_FLUENT_PARA_ENVS:
         del fluent_env["PARA_NPROCS"]
         del fluent_env["PARA_MESH_NPROCS"]
+    fluent_env["REMOTING_NEW_DM_API"] = "1"
```
seanpearsonuk, line 67: :+1:

src/ansys/fluent/core/__init__.py
```diff
@@ -106,10 +106,10 @@ def version_info() -> str:
 pydoc.text.docother = fldoc.docother.__get__(pydoc.text, pydoc.TextDoc)
 
 # Whether to use datamodel state caching
-DATAMODEL_USE_STATE_CACHE = True
+DATAMODEL_USE_STATE_CACHE = False
```
seanpearsonuk, line 109: @prmukherj Is there an issue that necessitates this change? If so, then we need to investigate why.

prmukherj: @seanpearsonuk, no. Just making sure that caching is not playing a role here as I was facing some issues while local testing. Will revert it soon and merge.

seanpearsonuk (approved):

mkundu1 (approved):

PR #3751, opened 2025-02-19
fix: Better way to copy the current function argument values

mkundu1:


src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -185,7 +186,11 @@ def __init__(
         """
         import ansys.fluent.core as pyfluent
 
-        self.argvals, self.new_session = _get_argvals_and_session(locals().copy())
```
mkundu1, line 188: `pyfluent` was getting copied here which causes `launch_fluent` error in the background session.

tests/test_launcher.py
```diff
@@ -171,7 +171,6 @@ def test_case_load():
     session.exit()
 
 
-@pytest.mark.standalone
```
mkundu1, line 174: These tests originally had some issue in running them in github, checking if the issue still exists ..

mkundu1:
I shall create a patch release with this change.

hpohekar (approved):

prmukherj (approved):

seanpearsonuk (approved):

seanpearsonuk (approved):

Issue #2714, opened 2024-04-18
Use hyphens in documentation file names


raph-luc:
### Description of the modifications

Apparently better for search engines:
![image](https://github.com/ansys/pyfluent/assets/132297401/82f09d68-bd65-4614-9a1f-7530e0880566)

### Useful links and references

https://developers.google.com/style/filenames
This tip came from a PyAnsys presentation: [Search Engine Optimization.pptx](https://ansys-my.sharepoint.com/:p:/p/revathy_venugopal/Ea9K3TWZezJKs10ilKMDpLYBUJdITmzoEMwnL1GdjMJ68A?e=n7jGPX)

hpohekar:
We can not import Python files with hyphens in their name.

seanpearsonuk:
Shall we close this or debate it further?

@raph-luc @mkundu1 @hpohekar @prmukherj

mkundu1:
As a large part of our documentation is generated from Python source, I think it is OK to have Python names as documentation file names. 

I have checked other Python libraries also have underscores in doc file names. We shall look into if there is any specific recommendation from pyansys in future.

raph-luc:
Sorry I only saw your comment now @hpohekar, this is only about documentation files (i.e. rst, md, etc), not Python files themselves. Did you mean something else?

@mkundu1 this came from a PyAnsys meeting at the time when I quickly wrote this down so we didn't forget. We can check if this is still the up-to-date recommendation, I don't know if guidance has changed (I would think it hasn't).

If this is considered worthwhile, when automatically generating the documentation files, it shouldn't be much work to convert them from snake_case to kebab-case, right?

hpohekar:
> Sorry I only saw your comment now [@hpohekar](https://github.com/hpohekar), this is only about documentation files (i.e. rst, md, etc), not Python files themselves. Did you mean something else?
> 
> [@mkundu1](https://github.com/mkundu1) this came from a PyAnsys meeting at the time when I quickly wrote this down so we didn't forget. We can check if this is still the up-to-date recommendation, I don't know if guidance has changed (I would think it hasn't).
> 
> If this is considered worthwhile, when automatically generating the documentation files, it shouldn't be much work to convert them from snake_case to kebab-case, right?

@raph-luc No issues at all. I agree with @mkundu1.

raph-luc:
A few examples where we are using camelCase or putting multiple words together, and could have been using hyphens or underscores instead:
[...]

Issue #2713, opened 2024-04-18
Avoid having index.rst pages


raph-luc:
### Description of the modifications

Seems that this is less than ideal for search engines, page names should be meaningful. See also #2714.

### Useful links and references

This tip came from a PyAnsys presentation: [Search Engine Optimization.pptx](https://ansys-my.sharepoint.com/:p:/p/revathy_venugopal/Ea9K3TWZezJKs10ilKMDpLYBUJdITmzoEMwnL1GdjMJ68A?e=n7jGPX)

seanpearsonuk:
Interesting.

raph-luc:
FYI this was based on recommendations by the PyAnsys team in a meeting. In hindsight I should've added more details and perhaps a link to their slides/presentation

hpohekar:
I have checked PyACP, PyMAPDL and PyAEDT repos, they are still using index.rst.

`index.rst` pages are useful to give introduction of various sections of docs. 

We have also added detailed information in some index.rst pages. 

We have added hyperlinks to key APIs  in index.rst for API reference recently.

In summary, we are using index.rst pages constructively, we are not using them for listing toctrees only.

hpohekar:
https://github.com/ansys/pyfluent/pull/3761

PR #3761, opened 2025-02-21
docs: search engine optimization

hpohekar:
closes https://github.com/ansys/pyfluent/issues/2714, https://github.com/ansys/pyfluent/issues/2713


doc/api_rstgen.py
```diff
@@ -144,7 +144,7 @@ def _write_common_rst_members(rst_file):
 
 def _generate_api_source_rst_files(folder: str, files: list):
     for file in files:
-        if "index" in file:
+        if "contents" in file:
```
seanpearsonuk, line 147: Is the condition that `file` ends with "_contents"?

seanpearsonuk (approved):

doc/api_rstgen.py
hpohekar: Yes, we can use that condition as well, updated. Thank you.

mkundu1 (approved):

PR #3764, opened 2025-02-21
chore: Reduce min grpcio-status version

mkundu1:


hpohekar (approved):

pyproject.toml
```diff
@@ -27,7 +27,7 @@ dependencies = [
     "docker>=7.1.0",
     "grpcio>=1.30.0",
     "grpcio-health-checking>=1.30.0",
-    "grpcio-status>=1.30.0",
+    "grpcio-status>=1.26.0",
```
mkundu1, line 30: This is the version in FBU deps. There will not be any runtime issue even there is a pip dependency incompatibility in FBU deps with grpcio-status>=1.30.0, so we don't need to patch any previous releases.

seanpearsonuk (approved):

prmukherj (approved):

PR #3767, opened 2025-02-21
chore: Specify PyFluent package version at a single location

mkundu1:


src/ansys/fluent/core/__init__.py
```diff
@@ -63,8 +62,13 @@
 from ansys.fluent.core.utils.fluent_version import FluentVersion  # noqa: F401
 from ansys.fluent.core.utils.setup_for_fluent import setup_for_fluent  # noqa: F401
 
+__version__ = "0.30.dev2"
```
mkundu1, line 65: Only place where version is specified

prmukherj (approved):

src/ansys/fluent/core/__init__.py
prmukherj: Great change.

seanpearsonuk (approved):

PR #3770, opened 2025-02-25
fix: Minor issues in PyConsole.

prmukherj:
1197165, 1187804, and getting appropriate error message when user tries to access a wrong settings object.

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -1107,6 +1107,9 @@ class Group(SettingsBase[DictStateType]):
             error_msg = allowed_name_error_message(
                 trial_name=name,
                 message=ex.args[0],
+                allowed_values=sorted(
+                    list(set(self.get_active_child_names() + self.command_names))
+                ),
             )
             ex.args = (error_msg,)
             raise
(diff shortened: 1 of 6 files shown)


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1107,6 +1107,9 @@ def __getattribute__(self, name):
             error_msg = allowed_name_error_message(
                 trial_name=name,
                 message=ex.args[0],
+                allowed_values=sorted(
+                    list(set(self.get_active_child_names() + self.command_names))
```
seanpearsonuk, line 1111: `set` is already iterable so creating a `list` is redundant (and innefficient).

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1105,18 +1104,12 @@ def __getattribute__(self, name):
...
-                match_whole_word=False,
-            )
-            pyfluent.PRINT_SEARCH_RESULTS = True
-            results = search_results if search_results else []
             error_msg = allowed_name_error_message(
                 trial_name=name,
                 message=ex.args[0],
-                search_results=results,
+                allowed_values=sorted(
+                    list(set(self.get_active_child_names() + self.command_names))
```
prmukherj, line 1111:
```suggestion
                    set(self.get_active_child_names() + self.command_names)
```

seanpearsonuk (approved):

PR #3773, opened 2025-02-25
fix: Invalid surface_ids check for field_data.

prmukherj:
Fix error handling for invalid surface_ids check for field_data.

### DIFF
--- a/src/ansys/fluent/core/services/field_data.py
+++ b/src/ansys/fluent/core/services/field_data.py
@@ -338,7 +338,7 @@ class _AllowedSurfaceNames(_AllowedNames):
             If surface name is invalid.
         """
         if validate_inputs and not self.is_valid(surface_name):
-            raise DisallowedValuesError("surface", surface_name, self())
+            raise DisallowedValuesError("surface", surface_name, list(self()))
         return surface_name
 
 
@@ -818,7 +818,11 @@ def _get_surface_ids(
                 ]
             )
         else:
-            surface_ids.append(surf)
+            allowed_surf_ids = _AllowedSurfaceIDs(field_info)()
+            if surf in allowed_surf_ids:
+                surface_ids.append(surf)
+            else:
+                raise DisallowedValuesError("surface", surf, allowed_surf_ids)
     return surface_ids
 
 


src/ansys/fluent/core/workflow.py
```diff
@@ -1365,6 +1365,7 @@ def __init__(
                     "service",
                     "task_object",
                     "workflow",
+                    "rename",
```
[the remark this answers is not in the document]
prmukherj: sounds good @seanpearsonuk. Will separate it out.
Thank you.

src/ansys/fluent/core/services/field_data.py
```diff
@@ -338,7 +338,7 @@ def valid_name(self, surface_name: str) -> str:
             If surface name is invalid.
         """
         if validate_inputs and not self.is_valid(surface_name):
```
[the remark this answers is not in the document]
seanpearsonuk: The parts of the implementation that do the work do not depend on the allowed values being a list. The limitation is actually only due to an easily corrected line of code. Annotations in the implementation alsoo need to be corrected.

mkundu1 (approved):

seanpearsonuk (approved):

hpohekar (approved):

src/ansys/fluent/core/solver/error_message.py
```diff
@@ -45,14 +45,15 @@ def allowed_name_error_message(
     if not message:
         message = f"'{context}' has no attribute '{trial_name}'"
     message += ".\n"
-    matches = None
     if allowed_values:
-        if isinstance(allowed_values, list) and isinstance(allowed_values[0], str):
+        matches = None
+        if all(isinstance(item, str) for item in allowed_values) and allowed_values:
```
prmukherj, line 50: @seanpearsonuk, we have to put in the 'str' check as of now because the field_data is also using this with the surface ids as allowed values which are int. We need to update these methods in a separate PR to correct the error message as well.

prmukherj: I have updated the check to test for a homogenous 'str' list.

seanpearsonuk: That's nice, thanks, @prmukherj.

seanpearsonuk: Just wondering:
1. Why is the caller passing in a `str` `name` with an `Iterable[int]` to constrain that name. How is `Iterable[int]` used if at all? A bit confused by this.
2. The annotation shows `Iterable[str | int]`. Does the code require that specifically? It seems that it's either `Iterable[str]` or `Iterable[Any]` depending on the outcome of pt. 1 above.

prmukherj: 1. Actually the closest match is required if the allowed values and the entered name is str, however there is one particular case where user passes surface id list (int) and in this case only allowed values is displayed.
2. Right, actually here it should be Iterable[str] | Iterable[int]

prmukherj: Updated the type-hints.

PR #3774, opened 2025-02-25
fix: resolve mro for _InputFile

hpohekar:


mkundu1 (approved):
@hpohekar Should we fix this in the codegen-side in a later version, so that file-like arguments are derived from classes having the file upload/download behaviour and avoid any runtime manipulation of the instances? We do something similar in flobject.py.

I'm approving the current PR as this looks to be an urgent fix that must go in for pyconsole.

hpohekar:
> @hpohekar Should we fix this in the codegen-side in a later version, so that file-like arguments are derived from classes having the file upload/download behaviour and avoid any runtime manipulation of the instances? We do something similar in flobject.py.
> 
> I'm approving the current PR as this looks to be an urgent fix that must go in for pyconsole.

@mkundu1 Yes, I agree because this is not happening in settings side. i.e. flobject.py

seanpearsonuk (approved):

PR #3802, opened 2025-03-04
feat: Use 'deprecated-version' flag for settings-api classes

prmukherj:
Use 'deprecated-version' flag for settings-api classes. The flag is specific to Fluent version.


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1058,26 +1058,50 @@ def get_active_child_names(self):
...
     def get_active_query_names(self):
         """Names of queries that are currently active."""
         ret = []
         for query in self.query_names:
-            if getattr(self, query).is_active():
+            if getattr(self, query).is_active() and not self._is_deprecated(query):
                 ret.append(query)
         return ret
 
+    def _is_deprecated(self, child_name: str) -> bool:
```
prmukherj, line 1081: Any scope to optimize this code. Please suggest. Thank you.
@seanpearsonuk

tests/test_settings_api.py
```diff
@@ -639,3 +639,68 @@ def test_return_types_of_operations_on_named_objects(mixing_elbow_settings_sessi
...
+        == "25.1"
+    )
+    assert (
+        graphics.contour["contour-velocity"].coloring.get_attr("deprecated-version")
+        == "25.1"
+    )
+
+    # in 'get_state'
+    if solver.get_fluent_version() >= FluentVersion.v252:
+        # From v252 'get_state' behaviour is to be corrected in Fluent.
```
prmukherj, line 670: This 25R2 test will pass after the settings-api side updates.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1058,26 +1058,50 @@ def get_active_child_names(self):
...
+        """If the 'child_name' is deprecated in a specific Fluent version.'"""
+        deprecated_version = getattr(self, child_name).get_attr("deprecated-version")
+        if deprecated_version and FluentVersion(self.version) >= FluentVersion(
+            deprecated_version
+        ):
+            return True
+        return False
```
seanpearsonuk, line 1088: Can just return the value of the predicate directly.

prmukherj: yes, Thanks.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1058,26 +1058,48 @@ def get_active_child_names(self):
         """Names of children that are currently active."""
         ret = []
         for child in self.child_names:
-            if getattr(self, child).is_active():
+            if getattr(self, child).is_active() and not self._is_deprecated(child):
```
seanpearsonuk, line 1061: @prmukherj Deprecated items should already be inactive, right? So, it's `is_active` itself that would be updated.

seanpearsonuk, line 1061: @prmukherj This is seems blatantly illogical. We're saying these are the active children. The existing code asks each child if it's active. That's already logical and doesn't need to be changed; i.e. why add an additional predicate at this level? By doing it this way, we also get inconsistent behaviour between asking "which are the active children?" and "is this child active?".

seanpearsonuk: @prmukherj 

We have to give an accurate, consistent picture to the Python user, and that includes `obj.is_active()` and `obj.get_active_child_names()`. 

Is there a use case where we need a different `is_active()` behaviour?

prmukherj: I don't think so. Yes, I agree, will update accordingly.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1058,26 +1058,48 @@ def get_active_child_names(self):
...
         return ret
 
+    def _is_deprecated(self, child_name: str) -> bool:
+        """If the 'child_name' is deprecated in a specific Fluent version.'"""
+        deprecated_version = getattr(self, child_name).get_attr("deprecated-version")
+        return deprecated_version and FluentVersion(self.version) >= FluentVersion(
+            deprecated_version
+        )
+
+    def __dir__(self):
```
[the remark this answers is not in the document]
prmukherj: Done, Thank you.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1078,6 +1088,16 @@ def get_active_query_names(self):
                 ret.append(query)
         return ret
 
+    def __dir__(self):
+        dir_list = set(list(self.__dict__.keys()) + dir(type(self)))
+        return dir_list - set(
+            [
+                child
+                for child in self.child_names + self.command_names + self.query_names
```
mkundu1, line 1096: Please also include `self.argument_names`

prmukherj: Done, Thank you.

mkundu1 (approved):

mkundu1 (approved):

PR #3826, opened 2025-03-12
build: Update ansys-units version

RobPasMue:
The current dev version is at 0.5.dev0 and this condition is not allowing it. Can we make sure we upgrade this limit whenever a new version is released on ansys-units?


prmukherj (approved):

hpohekar (approved):

hpohekar:
> The current dev version is at 0.5.dev0 and this condition is not allowing it. Can we make sure we upgrade this limit whenever a new version is released on ansys-units?

@RobPasMue Yes, sure. Thanks a lot!

pyproject.toml
```diff
@@ -22,7 +22,7 @@ dependencies = [
     "ansys-api-fluent>=0.3.35",
     "ansys-platform-instancemanagement~=1.1",
     "ansys-tools-filetransfer>=0.1,<0.3",
-    "ansys-units>=0.3.3,<0.5",
+    "ansys-units>=0.3.3,<0.6",
```
seanpearsonuk, line 25: @hpohekar What us the purpose of the upper limit?

hpohekar: @seanpearsonuk To allow installation of dev version.

seanpearsonuk: @hpohekar I mean what is the purpose of having any upper limit?

RobPasMue: I would remove it if possible - I just did the less intrusive modification

hpohekar: @RobPasMue Actually we can keep it as `"ansys-units>=0.4.0"` . Earlier it was in same format without an upper limit.

seanpearsonuk: Yes, please, @hpohekar.

seanpearsonuk: > I would remove it if possible - I just did the less intrusive modification

@RobPasMue That was my assumption, too, and thank you.

hpohekar: As per the @mkundu1's suggestion we are keeping major version as an upper limit. Thank you.

PR #3827, opened 2025-03-13
feat: Object based field data access (unifying the interface).

prmukherj:
Object based field data access. 
Unifying the interface for transaction based query and direct query.

### DIFF
--- a/src/ansys/fluent/core/services/field_data.py
+++ b/src/ansys/fluent/core/services/field_data.py
@@ -593,7 +593,7 @@ class _ReturnFieldData:
         return path_lines_dict
 
 
-class SurfaceFieldData:
+class SurfaceFieldDataRequest:
     """Container for storing parameters for surface data."""
 
     def __init__(
@@ -602,13 +602,13 @@ class SurfaceFieldData:
         surfaces: List[int | str],
         overset_mesh: bool | None = False,
     ):
-        """__init__ method of SurfaceFieldData class."""
+        """__init__ method of SurfaceFieldDataRequest class."""
         self.data_types = data_types
         self.surfaces = surfaces
         self.overset_mesh = overset_mesh
 
 
-class ScalarFieldData:
+class ScalarFieldDataRequest:
     """Container for storing parameters for scalar field data."""
 
     def __init__(
@@ -618,14 +618,14 @@ class ScalarFieldData:
         node_value: bool | None = True,
         boundary_value: bool | None = True,
     ):
-        """__init__ method of ScalarFieldData class."""
+        """__init__ method of ScalarFieldDataRequest class."""
         self.field_name = field_name
         self.surfaces = surfaces
         self.node_value = node_value
         self.boundary_value = boundary_value
 
 
-class VectorFieldData:
+class VectorFieldDataRequest:
     """Container for storing parameters for vector field data."""
 
     def __init__(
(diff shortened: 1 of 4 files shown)


tests/test_field_data.py
```diff
@@ -375,6 +529,120 @@ def test_field_data_objects_3d(new_solver_session) -> None:
...
+        data_types=[SurfaceDataType.Vertices], surfaces=["cold-inlet"]
+    )
+    vertices_data = field_data.get_field_data(su1)
+    assert vertices_data["cold-inlet"][SurfaceDataType.Vertices].shape == (241, 3)
+    assert (
+        round(float(vertices_data["cold-inlet"][SurfaceDataType.Vertices][5][0]), 2)
+        == -0.2
+    )
+
+    su2 = SurfaceFieldData(
```
seanpearsonuk, line 571: @prmukherj If this is a request type then please append the name with `Request`.

prmukherj: Actually it can again be re-used for fetching the data as well. That's why this naming.

src/ansys/fluent/core/services/field_data.py
```diff
@@ -593,6 +593,88 @@ def _pathlines_data(
...
+        self.step_size = step_size
+        self.skip = skip
+        self.reverse = reverse
+        self.accuracy_control_on = accuracy_control_on
+        self.tolerance = tolerance
+        self.coarsen = coarsen
+        self.velocity_domain = velocity_domain
+        self.zones = zones
+
+
```
seanpearsonuk, line 677: @prmukherj What was the rationale for deciding against use of `NamedTuple`?

tests/test_field_data.py
seanpearsonuk: @prmukherj It's always a request.

src/ansys/fluent/core/services/field_data.py
prmukherj: Actually the other methods will cease to exist if we have this and there we can just pass **kwargs. So for documenting the args I have kept it like it is. We can decide on this.

tests/test_field_data.py
prmukherj: got it. Makes sense. Will update.

src/ansys/fluent/core/services/field_data.py
prmukherj: Thank you for the context. Will update it.

src/ansys/fluent/core/services/field_data.py
```diff
@@ -799,16 +808,59 @@ def get_pathlines_field_data(
...
+                node_value=obj.node_value,
+                steps=obj.steps,
+                step_size=obj.step_size,
+                skip=obj.skip,
+                reverse=obj.reverse,
+                accuracy_control_on=obj.accuracy_control_on,
+                tolerance=obj.tolerance,
+                coarsen=obj.coarsen,
+                velocity_domain=obj.velocity_domain,
+                zones=obj.zones,
```
mkundu1, line 861: Dataclasses have a method to return the internal state as a dictionary. Maybe by using frozen dataclass instead of named tuple, we can simplify the above argument assignment.

mkundu1: OK, namedtuple also has a method to return the internal state as dictionary.

tests/test_field_data.py
```diff
@@ -176,35 +180,133 @@ def test_field_data_transactions(new_solver_session) -> None:
...
+        provide_particle_time_field=True,
     )
-    vector_trn = transaction3.add_vector_fields_request(
-        surfaces=[3, "hot-inlet"],
-        field_name="velocity",
+
+    transaction = transaction.add_request(su1)  # adding single request.
+    data = transaction.add_request(
+        su2, sc1, sc2, vc1, pt1
+    ).get_fields()  # adding multiple requests.
```
prmukherj, line 250: get_fields -> get_response

PR #3829, opened 2025-03-13
chore: Replace FLUENT_PRECISION_MODE global with runtime check

mkundu1:


src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -310,8 +310,6 @@ def launch_fluent(
     if env is None:
         env = {}
 
-    pyfluent.FLUENT_PRECISION_MODE = precision
```
mkundu1, line 313: This won't work for connected sessions.

src/ansys/fluent/core/services/field_data.py
```diff
@@ -1522,7 +1523,12 @@ def get_mesh(self, zone: str | int) -> Mesh:
         nodes_request = FieldDataProtoModule.GetSolverMeshNodesRequest(
             domain_id=ROOT_DOMAIN_ID, thread_id=zone_info._id
         )
-        nested_nodes = self._service.get_solver_mesh_nodes(nodes_request)
+        # TODO: Add precision query in AppUtilities service
```
seanpearsonuk, line 1526: :+1:

seanpearsonuk: @mkundu1 I propose that the `FieldData` service itself should also include a precision() method. This would provide a more complete and self-contained interface for `FieldData`, enabling its reuse in different contexts as well as simplifying client-side logic.

seanpearsonuk (approved):

prmukherj (approved):

hpohekar (approved):

PR #3848, opened 2025-03-18
ci: Run unittests conditionally

mkundu1:
Remove an unintended condition at the unittest job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -487,7 +487,6 @@ jobs:
 
   test:
     name: Unit Testing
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -487,7 +487,6 @@ jobs:
 
   test:
     name: Unit Testing
-    if: ${{ always() }}
```
mkundu1, line 490: I think due to this line, unittests tries to get executed even if the build is failed, resulting occational hangs.

prmukherj (approved):

PR #3851, opened 2025-03-18
ci: Reduce timeout for unittests

mkundu1:
Reduce the unittests timeout to 60 min to reduce hung time. Sometimes hung jobs cannot be cancelled from the UI and we have to wait for 2 hours currently. Generally, unittests jobs are run within 30 min, so a double time for timeout is sufficient.


hpohekar:
> Reduce the unittests timeout to 60 min to reduce hung time. Sometimes hung jobs cannot be cancelled from the UI and we have to wait for 2 hours currently. Generally, unittests jobs are run within 30 min and a double time for timeout is sufficient.
> 
> Merged the changes from #3848 also to avoid rerunning CI.

This change was really needed. I had also raised this point of long waiting time in the past. Thanks a lot.

hpohekar (approved):

prmukherj (approved):

PR #3859, opened 2025-03-20
fix: field-level-help at runtime for datamodel_se objects

mkundu1:
Field-level help is coming through static-info which can be shown for command arguments at datamodel_se level at runtime.

I'll look into next the codegen for command arguments to have those help-strings as docstrings.

### DIFF
--- a/doc/api_rstgen.py
+++ b/doc/api_rstgen.py
@@ -115,7 +115,7 @@ hierarchy = {
         "file_session",
         "fluent_connection",
         "journaling",
-        "logging",
+        "logger",
         "parametric",
         "rpvars",
         "search",
--- a/doc/changelog.d/3861.fixed.md
+++ b/doc/changelog.d/3861.fixed.md
@@ -1,0 +1,1 @@
+Avoid file name as native Python package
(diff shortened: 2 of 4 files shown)


tests/test_datamodel_service.py
```diff
@@ -833,3 +834,21 @@ def test_dynamic_dependency(new_meshing_session):
     d = ic.Refaceting.Deviation.get_state()
     cd = ic.Refaceting.CustomDeviation.get_state()
     assert d == cd
+
+
+def test_field_level_help(new_meshing_session, capsys):
+    meshing = new_meshing_session
+    import_geometry = meshing.meshing.ImportGeometry.create_instance()
+    assert isinstance(import_geometry.FileFormat, PyCommandArgumentsSubItem)
+    import_geometry.FileFormat.help()
```
mkundu1, line 843: @prmukherj Can you please check if this function is mapped to the workflow level?

hpohekar (approved):

seanpearsonuk (approved):

tests/test_datamodel_service.py
prmukherj: It is there at the task object level but not individual commands.

prmukherj: I think "helpString" is not present at this level?

mkundu1: Is it because `watertight.import_geometry.file_format.help()` doesn't call `PyCommandArgumentsSubItem.help`?

Issue #3620, opened 2025-01-07
Placeholder: podman support


mkundu1:
See Linux Container ADR

mkundu1:
The primary issue is Docker Python SDK doesn't work with podman server (raised by Patrick)

mkundu1:
Note the for linux container support, ideally we should support all container runtime mentioned there.

PR #3872, opened 2025-03-24
feat: Docker compose support

hpohekar:
closes https://github.com/ansys/pyfluent/issues/3620

### Introduction

Project Name: Fluent Docker and Podman compose 

### Overview
Docker and Podman compose support same compose.yaml file format to launch Fluent in container mode. This document outlines the technical design for the Fluent compose application.

### Problem Statement

Current Fluent container mode supports Docker only. Compose project aims to support both Docker and Podman using same compose file. 
### Goals and Objectives

1. Support both Docker and Podman for Fluent container launch mode.

2. Provide more robust control on container handling.

3. Ensure scalability to support multiple simultaneous instances of Fluent containers.

### Compose Overview

It is a tool for defining and running multi-container applications. We use a compose.yaml file to describe the services and volumes required for an application. Using a single command  `docker compose up` or `podman compose up`, you can start all services defined in the YAML file.

### Design Considerations

Assumptions: Users have local installation of Docker and Podman.

Dependencies: Docker, Podman.

### Architecture Design: PyFluent Container Launch Workflow

This architecture enables PyFluent to launch an Ansys Fluent session inside a container (Docker or Podman) based on environment configuration. The workflow dynamically generates a Compose file as text, launches the container using the available engine, and connects to the running Fluent session via a server info file shared between the host and container.

#### Architecture Diagram (Textual)

```
User
  |
  | (sets PYFLUENT_LAUNCH_CONTAINER=1 and PYFLUENT_USE_DOCKER_COMPOSE=1 or PYFLUENT_USE_PODMAN_COMPOSE=1)
  v
PyFluent
  |
  |--[Detect Docker/Podman]---> Container Engine (Docker/Podman)
  |                                 ^
  |                                 |
  |--[subprocess.communicate(input=compose_file)]--|
  |
  | (mounts host dir to container)
  v
Fluent Container
  |
  |--[creates server info file in mount target]---> Host Filesystem (mount source)
  |
  v
PyFluent Session Connector
  |
  |--[reads server info file, connects to Fluent]---> Returns Session Object
```

### Key Design Features

*Dynamic Compose Input*: No physical Compose file is written; YAML is passed as text directly to the container engine.

*Engine Agnostic*: PyFluent detects and uses either Docker or Podman, depending on availability.

*Mount Sharing*: Host and container share a directory for server info file exchange, enabling seamless connection setup.

*Session Management*: PyFluent connects to Fluent using details from the server info file, abstracting away containerization from the user.

### API Design docker_compose.py module

`launcher = ComposeLauncher(container_dict)`

`launcher.check_image_exists()` – Check if a Docker image exists locally

`launcher.pull_image()` - Pull a Docker image if it does not exist locally.

`launcher.start()` - Start the services.

`launcher.stop()` - Stop the services.

`launcher.ports` - Return the ports of the launched services.

### End-to-end tests using Pytest

1. Verified on both Windows and Linux, Docker and Podman works fine.

2. CI/CD pipeline already integrated for Docker in the GitHub.

3. Added timeout in `subprocess.communicate()` and `subprocess.wait()`.

4. Used a mechanism to remove the following docker services:
        - all stopped containers
        - all networks not used by at least one container
        - all volumes not used by at least one container

### Summary
This architecture allows PyFluent to flexibly launch Fluent in a containerized environment using either Docker or Podman, with all orchestration handled programmatically and connection to the running session managed via a shared server info file. The process is fully automated and requires minimal user intervention beyond setting the appropriate environment variable and, optionally, mount paths


hpohekar:
> @hpohekar Shall we roll this out as a full feature replacing the usage of the Python docker SDK for container launch mode?
 
@mkundu1 Yes, sure.

prmukherj (approved):

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -127,35 +127,6 @@ def run(self) -> None:
...
-        docker_client = _docker().from_env()
-        container = docker_client.containers.get(container_id_or_name)
-    except _docker().errors.NotFound:  # NotFound is a child from DockerException
-        return False
-    except _docker().errors.DockerException as exc:
-        logger.info(f"{type(exc).__name__}: {exc}")
-        return None
-    return container
-
-
```
hpohekar, line 158: This was based on Docker SDK hence removed.

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -573,21 +533,8 @@ def force_exit(self):
...
-        if get_container(container_id):
-            try:
-                container.exec_run(["bash", cleanup_file_name], detach=True)
-            except _docker().errors.APIError as e:
-                logger.info(f"{type(e).__name__}: {e}")
-                logger.debug(
-                    "Caught Docker APIError, Docker container probably not running anymore."
-                )
-        else:
-            logger.debug("Container not found, cancelling cleanup script execution.")
```
hpohekar, line 590: Provided a separate `exit` method on container object which is compatible for both Docker and Podman.

src/ansys/fluent/core/launcher/container_launcher.py
```diff
@@ -217,12 +236,7 @@ def __call__(self):
             file_transfer_service=self.file_transfer_service,
             start_transcript=self.argvals["start_transcript"],
         )
-        session._container = container
 
-        if self.argvals["start_watchdog"] is None and self.argvals["cleanup_on_exit"]:
-            self.argvals["start_watchdog"] = True
-        if self.argvals["start_watchdog"]:
-            logger.debug("Launching Watchdog for Fluent container...")
-            watchdog.launch(os.getpid(), port, password)
```
hpohekar, line 226: watchdog was based on Docker SDK and not compatible with Podman, hence removed.

src/ansys/fluent/core/launcher/watchdog_exec
```diff
@@ -143,25 +137,6 @@ if __name__ == "__main__":
...
-                args=(cortex_host,),
-                expected="falsy",
-            )
-            if response:
-                logger.info(
-                    "Fluent container still alive somehow, directly terminating it..."
-                )
-                subprocess.run(["docker", "kill", cortex_host])
-            else:
-                logger.info("Fluent container successfully shut down.")
```
hpohekar, line 164: `exit` method of container object is already being called inside force_exit() of both `FluentConnection` and `Session` hence removed from here.

tests/test_datamodel_api.py
```diff
@@ -66,6 +66,7 @@
 )
 
 
+@pytest.mark.skip(reason="https://github.com/ansys/pyfluent/issues/3910")
```
mkundu1, line 69: @hpohekar I haven't seen this failing in other branches, I merged one yesterday. I'll first check the running status in the dev nightly.

hpohekar: @mkundu1 Ok. right. I haven't checked nightly.

src/ansys/fluent/core/docker/docker-compose.yaml
```diff
@@ -0,0 +1,20 @@
+services:
+  fluent:
+    image: ${IMAGE_NAME}
+    environment:
```
mkundu1, line 4: These environment variables are interpreted by Fluent. I think only `ANSYSLMD_LICENSE_FILE`, `REMOTING_PORTS` and `FLUENT_NO_AUTOMATIC_TRANSCRIPT` are relevant here.

src/ansys/fluent/core/docker/docker-compose.yaml
```diff
@@ -0,0 +1,20 @@
...
+      - FLUENT_NO_AUTOMATIC_TRANSCRIPT=${NO_TRANSCRIPT}
+      - COMMAND=${COMMAND}
+      - PORT_1=${PORT_1}
+      - PORT_2=${PORT_2}
+      - MOUNT_SOURCE=${MOUNT_SOURCE}
+      - MOUNT_TARGET=${MOUNT_TARGET}
+    command: ${COMMAND}
+    ports:
+      - ${PORT_1}:${PORT_1}
+      - ${PORT_2}:${PORT_2}
```
mkundu1, line 17: This is odd, but I see there is an open issue in docker compose to support settings array of values via env var (https://github.com/docker/compose/issues/4249).

src/ansys/fluent/core/launcher/fluent_container.py
```diff
@@ -451,34 +457,23 @@ def start_fluent_container(
...
+        config_dict["fluent_port"] = port
 
-        container = docker_client.containers.run(
-            config_dict.pop("fluent_image"), **config_dict
+        docker_compose_container = DockerComposeLauncher(
+            container_dict=config_dict, config=DockerComposeLaunchConfig()
         )
 
-        success = timeout_loop(
-            lambda: host_server_info_file.stat().st_mtime > last_mtime, timeout
```
mkundu1, line 471: So, this logic no loger works?

src/ansys/fluent/core/docker/docker-compose.yaml
mkundu1: https://stackoverflow.com/a/72498747 - this might be the current workaround to set a variable number of ports programmatically.

src/ansys/fluent/core/launcher/fluent_container.py
hpohekar: No.

src/ansys/fluent/core/docker/docker-compose.yaml
hpohekar: If we don't specify other remaining env vars under environment then it gives error that env var is not defined therefore we have added them here and it is working fine.

tests/test_datamodel_api.py
hpohekar: @mkundu1 Yes, it's strange. Error logs shows that all these tests are not working properly hence created the issue.

src/ansys/fluent/core/docker/docker-compose.yaml
hpohekar: There is no other way to map multiple ports to service through docker compose, we have to specify them explicitly.

hpohekar: > https://stackoverflow.com/a/72498747 - this might be the current workaround to set a variable number of ports programmatically.

@mkundu1 It writes ports to docker-compose.yaml file directly which is not suitable for multiple instances using same docker-compose.yaml as we need the same compose file for both start and stop the container.

src/ansys/fluent/core/docker/docker-compose.yaml
hpohekar: @mkundu1 Done. Thank you. It's working fine now.

src/ansys/fluent/core/docker/docker-compose.yaml
hpohekar: Ok. Let me try this.

src/ansys/fluent/core/docker/docker-compose.yaml
mkundu1: This will prevent us from looking into logs etc. after starting the compose container for troubleshooting purpose.
```
mkundu@AAPAomeLhXO1rdt:~/devel/pyfluent/src/ansys/fluent/core/docker$ docker compose logs
WARN[0000] The "LICENSE_FILE" variable is not set. Defaulting to a blank string. 
WARN[0000] The "REMOTE_PORTS" variable is not set. Defaulting to a blank string. 
WARN[0000] The "NO_TRANSCRIPT" variable is not set. Defaulting to a blank string. 
WARN[0000] The "COMMAND" variable is not set. Defaulting to a blank string. 
WARN[0000] The "PORT_1" variable is not set. Defaulting to a blank string. 
WARN[0000] The "PORT_1" variable is not set. Defaulting to a blank string. 
WARN[0000] The "PORT_2" variable is not set. Defaulting to a blank string. 
WARN[0000] The "PORT_2" variable is not set. Defaulting to a blank string. 
WARN[0000] The "MOUNT_TARGET" variable is not set. Defaulting to a blank string. 
WARN[0000] The "MOUNT_SOURCE" variable is not set. Defaulting to a blank string. 
WARN[0000] The "MOUNT_TARGET" variable is not set. Defaulting to a blank string. 
WARN[0000] The "IMAGE_NAME" variable is not set. Defaulting to a blank string. 
1 error(s) decoding:

* 'services[fluent].ports[0]' expected a map, got 'string'
```

src/ansys/fluent/core/docker/docker_compose.py
```diff
@@ -0,0 +1,304 @@
...
+        "COMMAND": " ".join(container_dict["command"]),
+        "MOUNT_TARGET": container_dict.get("mount_target"),
+        "MOUNT_SOURCE": container_dict.get("mount_source"),
+    }
+
+    env_vars["PORT_1"] = str(ports[0]) if ports else port
+    env_vars["PORT_2"] = str(ports[1]) if len(ports) > 1 else str(get_free_port())
+
+    for key, value in env_vars.items():
+        os.environ[key] = value
```
mkundu1, line 144: Instead of setting these env vars in the host process, please send them through `subprocess.run`.

hpohekar: @mkundu1 I have already tried this. 

It gives the following error.

image

src/ansys/fluent/core/docker/docker-compose.yaml
mkundu1: I see that I can generally inspect the container using docker desktop. So the above is not a blocking issue.

hpohekar: @mkundu1 I use `docker logs pyfluent_compose_a6204b778fe44e19abed4ff6dbf5c005-fluent-1` to see logs as follows.

image

mkundu1: Closing this as we have ways to inspect inside container.

src/ansys/fluent/core/docker/docker_compose.py
mkundu1: @hpohekar Please check the `env` parameter in https://docs.python.org/3/library/subprocess.html#subprocess.run

hpohekar: @mkundu1 We are setting them for this particular instance only. These env vars are erased once this instance gets closed so it does not affect other simultaneous instances, as we are already running multiple instances simultaneously in the GitHub and all of them are working fine without polluting original env vars of the system.

hpohekar: > @hpohekar Please check the `env` parameter in https://docs.python.org/3/library/subprocess.html#subprocess.run

@mkundu1 I used the same `env` for both subprocess.run and subprocess.check_call, same error is there.

mkundu1: Here is the exceprt from the standalone launcher code:
```
fluent_env["var1"] = val1
fluent_env["var2"] = val2
...
subprocess.run(..., env=fluent_env)
```
What is the error you're seeing?

hpohekar: @mkundu1 Done. Thank you. It's resolved after using `os.environ.copy()` then updating it with Fluent env vars. I was passing Fluent env vars dict directly.

src/ansys/fluent/core/docker/docker-compose.yaml
hpohekar: @mkundu1 Done. Thank you.

src/ansys/fluent/core/docker/docker_compose.py
```diff
@@ -0,0 +1,342 @@
...
+            ]
+
+            process = subprocess.Popen(
+                self._set_compose_cmds() + cmd,
+                stdin=subprocess.PIPE,
+                text=True,
+                env=self._env,
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+            )
```
hpohekar, line 283: @mkundu1 

If we pass env vars using `env` argument then podman gives the following error. Therefore I had preferred to write docker-compose.yml separately for each session as it works for both docker and podman. 

```
pyfluent.launcher WARNING: Starting Fluent container mounted to /home/hpohekar/pyfluent, with this path available as /home/container/workdir for the Fluent session running inside the container.
>>>> Executing external compose provider "/bin/docker-compose". Please refer to the documentation for details. <<<<

WARN[0000] The "IMAGE_NAME" variable is not set. Defaulting to a blank string. 
WARN[0000] The "LICENSE_FILE" variable is not set. Defaulting to a blank string. 
WARN[0000] The "REMOTE_PORTS" variable is not set. Defaulting to a blank string. 
WARN[0000] The "NO_TRANSCRIPT" variable is not set. Defaulting to a blank string. 
WARN[0000] The "COMMAND" variable is not set. Defaulting to a blank string. 
WARN[0000] The "MOUNT_TARGET" variable is not set. Defaulting to a blank string. 
WARN[0000] The "MOUNT_SOURCE" variable is not set. Defaulting to a blank string. 
WARN[0000] The "MOUNT_TARGET" variable is not set. Defaulting to a blank string. 
[+] Running 1/2
 ✔ Network pyfluent_compose_2804e29fe17a45099f7377acc19ef254_default     Created                                                                                0.0s 
 ⠋ Container pyfluent_compose_2804e29fe17a45099f7377acc19ef254-fluent-1  Creating                                                                               0.1s 
Error response from daemon: fill out specgen: invalid container path ":", must be an absolute path
Error: executing /bin/docker-compose -f /home/hpohekar/pyfluent/src/ansys/fluent/core/docker/docker-compose.yaml -f - --project-name pyfluent_compose_2804e29fe17a45099f7377acc19ef254 up --detach: exit status 1
  File "/home/hpohekar/pyfluent/src/ansys/fluent/core/docker/debug.py", line 9, in <module>
    solver = pyfluent.launch_fluent(product_version=pyfluent.FluentVersion.v251, 
  File "/home/hpohekar/pyfluent/src/ansys/fluent/core/utils/deprecate.py", line 72, in wrapper
    return func(*args, **kwargs)
  File "/home/hpohekar/pyfluent/src/ansys/fluent/core/utils/deprecate.py", line 72, in wrapper
    return func(*args, **kwargs)
  File "/home/hpohekar/pyfluent/src/ansys/fluent/core/launcher/launcher.py", line 340, in launch_fluent
    return launcher()
  File "/home/hpohekar/pyfluent/src/ansys/fluent/core/launcher/container_launcher.py", line 217, in __call__
    port, config_dict, container = start_fluent_container(
  File "/home/hpohekar/pyfluent/src/ansys/fluent/core/launcher/fluent_container.py", line 482, in start_fluent_container
    docker_compose_container.start()
  File "/home/hpohekar/pyfluent/src/ansys/fluent/core/docker/docker_compose.py", line 292, in start
    raise subprocess.CalledProcessError(
subprocess.CalledProcessError: Command '['sudo', 'podman', 'compose', '-f', '/home/hpohekar/pyfluent/src/ansys/fluent/core/docker/docker-compose.yaml', '-f', '-', '--project-name', 'pyfluent_compose_2804e29fe17a45099f7377acc19ef254', 'up', '--detach']' returned non-zero exit status 1.
(venv310) {hpohekar@punfdevblade9}{/home/hpohekar/pyfluent}$ 
```

mkundu1: @hpohekar Have we checked why this error is coming?

hpohekar: @mkundu1 I'm looking into this but not found any proper cause and solution as this works fine for Docker.

hpohekar: @mkundu1 If we hard code volumes it resolves above error.

```
>>>> Executing external compose provider "/bin/docker-compose". Please refer to the documentation for details. <<<<

WARN[0000] The "COMMAND" variable is not set. Defaulting to a blank string. 
WARN[0000] The "MOUNT_TARGET" variable is not set. Defaulting to a blank string. 
WARN[0000] The "IMAGE_NAME" variable is not set. Defaulting to a blank string. 
WARN[0000] The "LICENSE_FILE" variable is not set. Defaulting to a blank string. 
WARN[0000] The "REMOTE_PORTS" variable is not set. Defaulting to a blank string. 
WARN[0000] The "NO_TRANSCRIPT" variable is not set. Defaulting to a blank string. 
[+] Running 2/2
 ✔ Network pyfluent_compose_3f843411ab5e48ecab2a9b6cccc24c98_default     Created                                                                                                 0.0s 
 ✔ Container pyfluent_compose_3f843411ab5e48ecab2a9b6cccc24c98-fluent-1  Started                                                                                                 0.4s 
  File "/home/hpohekar/pyfluent/src/ansys/fluent/core/docker/debug.py", line 10, in <module>
    solver = pyfluent.launch_fluent(product_version=pyfluent.FluentVersion.v251, 
  File "/home/hpohekar/pyfluent/src/ansys/fluent/core/utils/deprecate.py", line 72, in wrapper
    return func(*args, **kwargs)
  File "/home/hpohekar/pyfluent/src/ansys/fluent/core/utils/deprecate.py", line 72, in wrapper
    return func(*args, **kwargs)
  File "/home/hpohekar/pyfluent/src/ansys/fluent/core/launcher/launcher.py", line 340, in launch_fluent
    return launcher()
  File "/home/hpohekar/pyfluent/src/ansys/fluent/core/launcher/container_launcher.py", line 221, in __call__
    _, _, password = _get_server_info_from_container(config_dict=config_dict)
  File "/home/hpohekar/pyfluent/src/ansys/fluent/core/launcher/container_launcher.py", line 79, in _get_server_info_from_container
    raise FileNotFoundError(f"{host_server_info_file} not found.")
FileNotFoundError: /home/hpohekar/serverinfo-nnezlp34.txt not found.
(venv310) {hpohekar@punfdevblade9}{/home/hpohekar}$ 
```

mkundu1: Whe we setting the paths via environment variables can we check how that is resolved at runtime? Maybe there is some issue in the path syntx when it is provided via env var. [docker inspect](https://docs.docker.com/reference/cli/docker/inspect/) should return the volume information of a running container.

hpohekar: @mkundu1 Same syntax is working fine for Docker.

hpohekar: This is a broader issue not just limited to paths as we can see that all env vars are not picked up by Podman. Once this is resolved then path syntax will resolve on it's own.

hpohekar:
> @hpohekar We can have a quick CI step (with only few tests) to test the podman-only configuration in a GitHub runner. This can done separately from the current PR.

@mkundu1 Yes, sure. We need to install Podman on all 3 runners and use an env var to select Podman as a compose source for selective tests.

mkundu1:
> > @hpohekar We can have a quick CI step (with only few tests) to test the podman-only configuration in a GitHub runner. This can done separately from the current PR.
> 
> @mkundu1 Yes, sure. We need to install Podman on all 3 runners and use an env var to select Podman as a compose source for selective tests.

Should test it only in a github-runner, better to avoid any change in the self-hosted runners.

src/ansys/fluent/core/docker/docker_compose.py
seanpearsonuk, line 1: @hpohekar Please eliminate code duplication in this module.

hpohekar: @seanpearsonuk removed duplication from start, stop, check_image_exists and pull_image methods. Thanks a lot for your suggestions.

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -453,18 +454,20 @@ def __init__(
             self._connection_interface.get_cortex_connection_properties()
         )
         self._cleanup_on_exit = cleanup_on_exit
-
+        self._container = container
+        self._compose = os.getenv("PYFLUENT_USE_DOCKER_COMPOSE") or os.getenv(
```
mkundu1, line 458: Instead of checking these env vars at vaious places, can we move these checks to method within launcher_utils.py? Note that, we generally compare boolean-type env-var values against "1" in pyfluent code.

mkundu1:
@hpohekar As you said in the new plan we are not making any change in the default behaviour. But, it seems there are some changes in the exiting code, like usage of `inside_container` is replaced with `get_container()`, which don't seem to be related to compose support.

Ideally, to minimize the changes in existing code, can we maintain some structure like the following?
```
self._compose = <value of the new config>
...
if self._compose:
  <new code>
else:
  <exisiting code>
```
If there are some useful changes in the existing code, please move them to a separate PR. It will be easier to review just the new changes in the current PR.

hpohekar:
> @hpohekar As you said, in the new plan we are not making any change in the default behaviour. But, it seems there are some changes in the exiting code, like usage of `inside_container` is replaced with `get_container()`, which don't seem to be related to compose support.
> 
> Ideally, to minimize the changes in existing code, can we maintain some structure like the following?
> 
> ```
> self._compose = <value of the new config>
> ...
> if self._compose:
>   <new code>
> else:
>   <exisiting code>
> ```
> 
> If there are some useful changes in the existing code, please move them to a separate PR. It will be easier to review just the new changes in the current PR.

@mkundu1 

Yes, that's true. We are not making any changes in default **behavior** as passing of all unit tests already proved this. 

Default behavior means Docker SDK should work fine for unit testing in CI and at the same time `test_docker_compose` should pass using Docker compose and both of them are working fine. 

The usage of inside_container is replaced with get_container() because there is an issue in the existing code and we have resolved this.

src/ansys/fluent/core/fluent_connection.py
hpohekar: Yes, we can try this.

hpohekar: Done. Thank you.

hpohekar: @mkundu1 

When we have moved this to a single file and used the following cases, it is failing consistently.

1. compose = os.getenv("PYFLUENT_USE_DOCKER_COMPOSE") == "1" or os.getenv("PYFLUENT_USE_DOCKER_COMPOSE") == "1"

2. compose = os.getenv("PYFLUENT_USE_DOCKER_COMPOSE") or os.getenv("PYFLUENT_USE_DOCKER_COMPOSE")

hpohekar: I have investigated it further, we are not importing the env var like this anywhere in the source code. We are evaluating it as per the requirement at each place. As per our initial discussion we were agreed that it will be OK to have evaluation of this env vars at various places. Reverting back to that change as test is failing consistently.

hpohekar: Done. Thank you.

hpohekar:
@seanpearsonuk @mkundu1 This PR is ready. Please share if you have any suggestions otherwise we can merge this now. Thank you.

mkundu1 (approved):

Issue #3897, opened 2025-04-04
PyFluent hang waiting for health_check to return


seanpearsonuk:
This is in `FluentConnection.__init__` which waits indefinitely for health check to return. If cortex is stuck failing to start properly, it does not return.

PR #3917, opened 2025-04-10
refactor: Add timeout_loop for health check

hpohekar:
closes https://github.com/ansys/pyfluent/issues/3897

Add timeout for health check request.


src/ansys/fluent/core/fluent_connection.py
```diff
@@ -436,7 +436,13 @@ def __init__(
         # throws, we should not proceed.
         # TODO: Show user-friendly error message.
         if pyfluent.CHECK_HEALTH:
-            self.health_check.check_health()
+            connected = timeout_loop(
```
mkundu1, line 439: Instead of adding a loop here, we can add a timeout directly in the grpc method call: https://github.com/grpc/grpc/blob/master/examples/python/timeout/greeter_client.py

Also, please make the timeout variable configurable via module-level global var.

mkundu1: In the defect case, does the `self.health_check.check_health()` return `NOT_SERVING`? In that case, we should simply throw from here.

hpohekar: There are 4 possibilities UNKNOWN, SERVING, NOT_SERVING and SERVICE_UNKNOWN therefore we are checking explicitly for SERVING only.

hpohekar: > Instead of repeatedly calling the method within a loop here, we can add the timeout directly in the grpc method call: https://github.com/grpc/grpc/blob/master/examples/python/timeout/greeter_client.py
> 
> Also, please make the timeout variable configurable via module-level global var.

@mkundu1 We have `wait_for_server` method on HealthCheckService where we are already checking whether grpc.StatusCode.DEADLINE_EXCEEDED or not. We are not using `wait_for_server` anywhere. We can use it here right ?

src/ansys/fluent/core/services/health_check.py
```diff
@@ -39,6 +39,8 @@
 
 logger: logging.Logger = logging.getLogger("pyfluent.general")
 
+TIMEOUT = 60
```
mkundu1, line 42: Please move this to src\ansys\fluent\core\__init__.py as a new configuarion `CHECK_HEALTH_TIMEOUT`. We already have `CHECK_HEALTH` there.

mkundu1:
> @seanpearsonuk Please could you share the details of how to reproduce this issue locally so that we can confirm the behavior.
> 
> cc: @mkundu1

I'd like to understand the scenario when the grpc call can hang. If server is not avaialble, we generally expect `UNAVAILABLE` or `DEADLINE_EXCEEDED` errors.

src/ansys/fluent/core/fluent_connection.py
seanpearsonuk: > Instead of repeatedly calling the method within a loop here, we can add the timeout directly in the grpc method call: https://github.com/grpc/grpc/blob/master/examples/python/timeout/greeter_client.py
> 
> Also, please make the timeout variable configurable via module-level global var.

Did the original solution repeatedly poll the server? That would allow the code to keep the user informed about what's cooking between polls (particularly useful for durations close to the timeout).
@mkundu1 @hpohekar

mkundu1: When we reach at this code, the Fluent server should already be running (for `launch_fluent` we have already parsed the generated server-info file and for `connect_to_fluent` the server-info file from the server has been provided). We don't need any polling here. This check_health is a single call to check the sanity of the already running server. If there are any error from the running server, this check_health will fail instead of a later gRPC call.
@seanpearsonuk @hpohekar

src/ansys/fluent/core/services/health_check.py
hpohekar: Done. Thank you.

seanpearsonuk:
> > @seanpearsonuk Please could you share the details of how to reproduce this issue locally so that we can confirm the behavior.
> > cc: @mkundu1
> 
> I'd like to understand the scenario when the grpc call can hang. If server is not avaialble, we generally expect `UNAVAILABLE` or `DEADLINE_EXCEEDED` errors.

It mentions in the linked that, "This is in FluentConnection.__init__ which waits indefinitely for health check to return. If **cortex is stuck failing to start properly**, it does not return." We saw this recently. Was this the issue where the parametric API code was checking for the optiSLang installation? You could simulate the issue by putting an infinte loop in the cortex loop.

src/ansys/fluent/core/fluent_connection.py
seanpearsonuk: The server can be already running but cortex can still execute more code in the meantime. I think that this is the issue. If cortex gets into a bad state then the remote health check call can hang.

prmukherj (approved):

Issue #3877, opened 2025-03-26
More flexible version of register_callback for events


seanpearsonuk:
#### **Current Function Signature:**  
```python
def register_callback(
    self,
    event_name: TEvent | str,
    callback: Callable,
    *args,
    **kwargs,
)
```
Currently, `register_callback()` allows registering a callback for **only one event type** at a time.  

#### **Proposed Enhancement:**  
- Enable **multiple event types** to be registered in a **single call**.  
- Rename `event_name` → `event_types` to reflect the new capability.  
- Determine the best syntax for passing multiple events.  

#### **Suggested Design Change:**  
Modify `register_callback()` to accept **one or more event types** using a tuple:  
```python
def register_callback(
    self,
    event_types: SolverEvent | tuple[SolverEvent, ...],  # Accepts one or multiple events
    callback: Callable,
    *args,
    **kwargs,
):
    if not isinstance(event_types, tuple):
        event_types = (event_types,)  # Convert single event to tuple

    for event in event_types:
        # Existing logic for registering the callback
        self._register_single_callback(event, callback, *args, **kwargs)
```
#### **Usage Example:**  
```python
events.register_callback(
    (SolverEvent.CALCULATIONS_STARTED, SolverEvent.CALCULATIONS_ENDED),
    my_callback_function
)
```

PR #3924, opened 2025-04-11
feat: Register multiple event types in register_callback()

hpohekar:
closes https://github.com/ansys/pyfluent/issues/3877

Register multiple event types in `register_callback()`.


src/ansys/fluent/core/streaming_services/events_streaming.py
```diff
@@ -545,6 +545,51 @@ def register_callback(
                 }
             return callback_id
 
+    def register_callback(
+        self,
+        event_types: (
+            SolverEvent
+            | MeshingEvent
+            | tuple[SolverEvent, ...]
```
mkundu1, line 553: Allow both list and tuple, use Sequence as typehint

hpohekar: Done. Thank you.

seanpearsonuk (approved):

PR #3936, opened 2025-04-16
chore: Remove the temporary fix for Fluent image 25.2

mkundu1:


hpohekar:
I think we need to run the Force update stable Fluent image version workflow first then these CI jobs will pass.

prmukherj (approved):

PR #3937, opened 2025-04-16
fix: Create directory within the current user's home

mkundu1:
@hpohekar This fixes the permission issue locally for me (running docker from within wsl)

### DIFF
--- a/tests/test_file_transfer_service.py
+++ b/tests/test_file_transfer_service.py
@@ -45,11 +45,11 @@ def test_remote_grpc_fts_container():
         "mixing_elbow.cas.h5", "pyfluent/mixing_elbow", return_without_path=False
     )
 
-    source_path = "/home/ansys/ansys_fluent_core_examples"
-    if not Path(source_path).exists():
-        Path(source_path).mkdir(parents=True, exist_ok=True)
+    source_path = Path.home() / "Downloads" / "ansys_fluent_core_examples"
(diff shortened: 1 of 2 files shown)


seanpearsonuk (approved):

tests/test_file_transfer_service.py
```diff
@@ -45,11 +45,11 @@ def test_remote_grpc_fts_container():
         "mixing_elbow.cas.h5", "pyfluent/mixing_elbow", return_without_path=False
     )
 
-    source_path = "/home/ansys/ansys_fluent_core_examples"
-    if not Path(source_path).exists():
-        Path(source_path).mkdir(parents=True, exist_ok=True)
+    source_path = Path.home() / "Downloads" / "ansys_fluent_core_examples"
```
hpohekar, line 48: @mkundu1 We don't want to use default location of `ansys_fluent_core_examples` as `mount_source` because Remote FTS will try to upload and download the file from same directory which corrupts the file being uploaded and downloaded. Therefore I was trying to use another location which gives me the permission error. This is a general permission issue on Linux.

hpohekar: Ok. I see. Test is passing without any errors, last time we had faced the file corruption issue may be at that time we were using the same folder as `mount_source` for all tests but we are using for single test now.

hpohekar (approved):

PR #3944, opened 2025-04-21
fix: Manual stage for add-license-header [skip tests]

hpohekar:
Added manual stage for add-license-header in CI and added it in the required checks.


.pre-commit-config.yaml
```diff
@@ -43,12 +43,12 @@ repos:
...
-  hooks:
-  - id: docformatter
-    stages: [manual]
-    args: ["--config", "pyproject.toml"]
+# - repo: https://github.com/PyCQA/docformatter
+#   rev: v1.7.5
+#   hooks:
+#   - id: docformatter
+#     stages: [manual]
+#     args: ["--config", "pyproject.toml"]
```
hpohekar, line 51: This hook was giving the following error consistently and we were not using it in CI and locally therefore we have commented it.

image

prmukherj: That means it is not getting used either locally or in CI right?

hpohekar: Yes, that was the existing behavior therefore we have commented it now.

prmukherj: Shall we remove it then?

hpohekar: We have kept the job below it therefore applied the same for this.

mkundu1: Yeah, let's it as commented. Maybe we'll try to fix it in a future version of docformatter.

mkundu1 (approved):

src/ansys/fluent/core/filereader/pre_processor.py
mkundu1, line 1: 👍

prmukherj (approved):

seanpearsonuk (approved):

PR #3945, opened 2025-04-22
fix: Nightly dev doc CI run.

prmukherj:
Fix the Nightly dev doc CI run.

PR #3948, opened 2025-04-22
fix: Fix deprecated behaviour

mkundu1:
For "deprecated" settings, we should hide the exposure, but not block the access. With the current PR changes, we get the following behaviour in both pyfluent and pyconsole:
```
>>> v1 = solver.settings.results.graphics.vector.create()
>>> v1.scale.scale_f = 4
>>> "scale" in dir(v1)
False
```
Nightly dev doc run - https://github.com/ansys/pyfluent/actions/runs/14606954687

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -410,6 +410,8 @@ class Base:
     def _is_deprecated(self) -> bool:
         """Whether the object is deprecated in a specific Fluent version.'"""
         deprecated_version = self.get_attrs(["deprecated-version"])
+        if deprecated_version:
+            deprecated_version = deprecated_version.get("attrs", deprecated_version)
         deprecated_version = (
             deprecated_version.get("deprecated-version") if deprecated_version else None
         )
@@ -419,7 +421,7 @@ class Base:
 
     def is_active(self) -> bool:
         """Whether the object is active."""
-        attr = self.get_attr(_InlineConstants.is_active) and not self._is_deprecated()
+        attr = self.get_attr(_InlineConstants.is_active)
         return False if attr is False else True
 
     def _check_stable(self) -> None:
@@ -955,7 +957,7 @@ def _command_query_name_filter(
     for name in names:
         if name not in excluded and name.startswith(prefix):
             child = getattr(parent, name)
-            if child.is_active():
+            if child.is_active() and not child._is_deprecated():
                 ret.append([name, child.__class__.__bases__[0].__name__, child.__doc__])
     return ret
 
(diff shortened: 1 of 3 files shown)


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -419,7 +421,7 @@ def _is_deprecated(self) -> bool:
 
     def is_active(self) -> bool:
         """Whether the object is active."""
-        attr = self.get_attr(_InlineConstants.is_active) and not self._is_deprecated()
+        attr = self.get_attr(_InlineConstants.is_active)
```
mkundu1, line 424: User won't normally find deprecated objects in the settings API, so it is OK to leave them active.

Only the active children can be accessed from parent, so the deprecated objects should be kept active for backward compatibility of journals.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -410,6 +410,8 @@ def get_attr(
     def _is_deprecated(self) -> bool:
         """Whether the object is deprecated in a specific Fluent version.'"""
         deprecated_version = self.get_attrs(["deprecated-version"])
+        if deprecated_version:
+            deprecated_version = deprecated_version.get("attrs", deprecated_version)
```
mkundu1, line 414: Extra unwrapping required in PyConsole. Without this `_is_deprecated` was returning `None` in PyConsole and that's why the old journal was passing in PyConsole (#3945).

tests/test_settings_api.py
```diff
@@ -678,7 +678,6 @@ def test_return_types_of_operations_on_named_objects(mixing_elbow_settings_sessi
     assert var3.obj_name == "air-copied"
 
 
-@pytest.mark.skip("https://github.com/ansys/pyfluent/issues/3813")
```
mkundu1, line 681: Enabled the test, kept only the failing part disabled.

hpohekar (approved):

src/ansys/fluent/core/solver/flobject.py
prmukherj: sounds right, Thanks.

seanpearsonuk (approved):

tests/test_settings_api.py
```diff
@@ -703,23 +702,23 @@ def test_settings_with_deprecated_flag(mixing_elbow_settings_session):
         == "25.1"
     )
 
-    # Deprecated objects should not be active
-    assert not graphics.contour["contour-velocity"].range_option.is_active()
+    # User won't normally find deprecated objects in the settings API, so it is OK to leave them active.
     assert graphics.contour["contour-velocity"].range_options.is_active()
 
+    # https://github.com/ansys/pyfluent/issues/3813
```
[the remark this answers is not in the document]
mkundu1: Yes, this still doesn't work from server-side.

PR #3949, opened 2025-04-23
fix: Raise AttributeError for non-existing meshing objects after switch_to solver.

prmukherj:
Raise AttributeError for non-existing meshing objects after switch_to solver.

Earlier instead of raising an Attribute error (general python behaviour), it was returning None. This has been corrected now.

There is an "is_active" method associated with session objects to check it's state.


prmukherj:
> Hi @prmukherj,
> 
> Thanks for the improvements here — I agree that raising `AttributeError` on access to a switched meshing session is more idiomatic and aligns better with general Python expectations around invalid states.
> 
> That said, I wanted to highlight a few related areas that may warrant attention to ensure consistent and intuitive behavior:
> 
> 1. **`dir(<mesh_session>)` still raises a low-level exception**, both before and after this change, due to access to internal attributes during `__dir__` resolution. This could be surprising and uninformative to users. We may need to explicitly handle the switched state in `__dir__()` and return a helpful message or an empty list.
> 2. **`help(<mesh_session>)` now presents stale info** after the switch (which is a regression). It gives the impression the interface is still usable. This could mislead users who turn to `help()` for post-switch diagnostics. If there’s a way to influence `help()` output (e.g., overriding `__doc__` or `__getattribute__` with a custom response), we might consider doing so for clarity.
> 3. **No public API to check if the session is active.** At the moment, the only way to determine whether the meshing session is still valid is to access the private `_switched` attribute. It would be helpful to expose a method like `is_active()` or `has_been_switched()` to allow users to make this check in a clean and forward-compatible way.
> 
> Happy to discuss any of these further — let me know what you think.

Yes @seanpearsonuk, I agree with this proposal. These changes are necessary. Others please put in your views as well.
Thank you.

mkundu1 (commented):
Can you please explain the *_wo_exit changes in tests?

src/ansys/fluent/core/session.py
```diff
@@ -449,6 +454,8 @@ def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any):
         self.exit()
 
     def __dir__(self):
+        if self._fluent_connection is None:
+            return ["is_active"]
```
seanpearsonuk, line 458: :+1:

src/ansys/fluent/core/session.py
```diff
@@ -370,6 +374,7 @@ def exit(self, **kwargs) -> None:
         logger.debug("session.exit() called")
         if self._fluent_connection:
             self._fluent_connection.exit(**kwargs)
+            self._fluent_connection = None
```
seanpearsonuk, line 377: :+1:

mkundu1 (approved):
The private `_switched` attribute is probably no longer required now.

seanpearsonuk:
Is there a test for `help(<meshing_session>)`?

prmukherj:
> @prmukherj Is there a test for `help(<meshing_session>)`?

No, the help behavior was not updated yet. Will update it as part of this PR only then. Converting it to draft for further testing.

seanpearsonuk (approved):

PR #3951, opened 2025-04-23
fix: Preprocess xml content before sending it to ElementTree parser

mkundu1:
XMLParser gives a ParseError if there are colons in the tag name. This PR pro-processes the XML content from FLuent's .flprj file to convert double colons to double underscores. E.g. the XML fragment
```
<ProjectStoragePolicy::FolderEnabled class="string" value="true"/>
```
will be converted to
```
<ProjectStoragePolicy__FolderEnabled class="string" value="true"/>
```

### DIFF
--- a/doc/changelog.d/3951.fixed.md
+++ b/doc/changelog.d/3951.fixed.md
@@ -1,0 +1,1 @@
+Preprocess xml content before sending it to ElementTree parser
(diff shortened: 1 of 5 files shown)


seanpearsonuk (approved):

hpohekar (approved):

mkundu1:
Force merging this PR as the GitHub Actions CI has been passed here - https://github.com/ansys/pyfluent/actions/runs/14622751386

The status is not reflected even after multiple runs.

PR #3953, opened 2025-04-24
fix: deprecated flag for flobject.py

prmukherj:
In deprecated version check we did not consider the version can be less than 22.2 (minimum version form which we support PyFluent). However recently, a deprecated version was introduced as 21.1. That's why this fix.


seanpearsonuk (approved):

hpohekar (approved):

hpohekar:
Yes, It was introduced just 2 days ago in report definitions.

mkundu1 (approved):

PR #3965, opened 2025-04-28
chore: Update 2d meshing test and doc.

prmukherj:
In the 2d meshing test and doc, the attributes under advanced options are accessed before setting the advanced options as True. This was showing up as an error message. This is corrected now.


hpohekar (approved):

seanpearsonuk:
@prmukherj 

1. Please avoid prefixing the title with _refactor_ unless it is appropiate. Please check the meaning of refactor to see what I mean.
2. Please add a comment like this one before each of the added lines in both doc and test files:
```python
# NOTE: Setting `show_advanced_options = True` is required to configure advanced preferences.
# This dependency may be removed in a future release as the API evolves.
```

prmukherj:
> @prmukherj
> 
> 1. Please avoid prefixing the title with _refactor_ unless it is appropiate. Please check the meaning of refactor to see what I mean.
> 2. Please add a comment like this one before each of the added lines in both doc and test files:
> 
> ```python
> # NOTE: Setting `show_advanced_options = True` is required to configure advanced preferences.
> # This dependency may be removed in a future release as the API evolves.
> ```

Done, Thank you.

mkundu1 (approved):

PR #3971, opened 2025-04-28
chore: Update codecov.yml

mkundu1:
Updating codecov.yml from the latest PyAnsys developer's guide: https://dev.docs.pyansys.com/how-to/testing.html#enforce-code-coverage


codecov.yml
```diff
@@ -1,21 +1,22 @@
...
 coverage:
-  range: 70..100
-  round: down
-  precision: 2
   status:
     project:
       default:
-        target: 80%
+        # basic
+        # target: 50%
```
hpohekar, line 10: Here target is disabled does it mean we want to increase code coverage to 100% first then for each PR we will check that code coverage has been maintained otherwise the PR will not get merged ?

mkundu1: The PR requirement is determined by the patch target which is 90%. That means 90% of new changes should have test coverage. The project codebase (default target) doesn't have any requirement, however we shall work on to generally improve the coverage.

I'm also not sure how it is enforced during PR merge. E.g. in the main branch already has a similar target requirement, but I've never seen any status check corresponding to that while merging PRs.

We also need to fix why the coverage report in the main github page is showing 36%. I think the report is based on old codebase, although it should be updated during each nightly test run. I'll first update the .yaml as per the latest guide (this PR) and then look into if the CI requires some additional fixes.

hpohekar: I see. Last time I was working on increasing test coverage of `file_transfer_service.py`, however even after adding appropriate tests it was showing same code coverage around 25%.

 https://app.codecov.io/gh/ansys/pyfluent/blob/main/src%2Fansys%2Ffluent%2Fcore%2Futils%2Ffile_transfer_service.py
 
 I had mentioned this issue to @seanpearsonuk but I understood now that it was not updating.
 
 I have create a PR to check the coverage manually. 
 
 https://github.com/ansys/pyfluent/pull/4006

Let's see.

hpohekar: We get code coverage around 64%. https://github.com/ansys/pyfluent/actions/runs/14836953788/job/41652990983?pr=4006

Not sure why there is a difference with code coverage mentioned in the PR description.

PR #3982, opened 2025-04-29
chore: Update error message for attribute errors from Solver.

prmukherj:
Update error message for attribute errors from Solver. If the attribute is not there in the settings-root, then we should raise the Attribute error from the same class only.

### DIFF
--- a/src/ansys/fluent/core/session_solver.py
+++ b/src/ansys/fluent/core/session_solver.py
@@ -363,7 +363,10 @@ class Solver(BaseSession):
                     f"'{attr}' is deprecated. Use 'settings.{attr}' instead.",
                     DeprecatedSettingWarning,
                 )
-        return getattr(self._settings_api_root, attr)
+        try:
+            return getattr(self._settings_api_root, attr)
+        except AttributeError:
+            return object.__getattribute__(self, attr)
 
     def __dir__(self):
         if self._fluent_connection is None:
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/session_solver.py
```diff
@@ -363,7 +363,10 @@ def __getattr__(self, attr):
                     f"'{attr}' is deprecated. Use 'settings.{attr}' instead.",
                     DeprecatedSettingWarning,
                 )
-        return getattr(self._settings_api_root, attr)
+        try:
+            return getattr(self._settings_api_root, attr)
+        except AttributeError:
+            return object.__getattribute__(self, attr)
```
seanpearsonuk, line 369:
```suggestion
        # Try forwarding attribute access to the settings API root.
        # If that fails with AttributeError, fall back to normal attribute lookup.
        # Using object.__getattribute__ triggers the standard AttributeError with default messaging.
        try:
            return getattr(self._settings_api_root, attr)
        except AttributeError:
            # Let standard attribute access raise the appropriate error
            return object.__getattribute__(self, attr)
```

seanpearsonuk: @prmukherj 
Better version of the if statement: `if not attr.startswith("_") and attr in dir(self._settings_api_root):`

prmukherj: Done, Thank you.

seanpearsonuk: @prmukherj Just checking that you saw the final comment, above.

prmukherj: @seanpearsonuk, yes, done.

seanpearsonuk (approved):

PR #3984, opened 2025-04-29
chore: Search for undocumented fields in codebase

mkundu1:
Based on #3980's findings.

The script is mostly generated by AI. Please see the module doc on how to run the script. There are around 400-450 undocumented attributes. We'll plan how to divide the workload to add the missing attribute docs.


seanpearsonuk (approved):

hpohekar (approved):

devel/undocumented_fields/undocumented_fields.txt
```diff
@@ -0,0 +1,616 @@
...
+    - service
+  Class: AppUtilitiesOld
+    - scheme_eval
+
+  Module: ansys.fluent.core.services.batch_ops
+  Class: BatchOps
+    - batching
+
+  Module: ansys.fluent.core.services.datamodel_se
+  Class: Attribute
```
prmukherj, line 270: It seems it is picking up attributes from Enums as well. Anyways, you're right. We can decide and work on this. We can ignore Enums as well I think.

prmukherj (approved):

devel/undocumented_fields/undocumented_fields.txt
mkundu1: Thanks for pointing it out, I'll skip the enums.

mkundu1: Enums are skipped now.

PR #3988, opened 2025-04-30
feat: physical quantities baseline code

seanpearsonuk:
This is just for demonstration purposes.

Provides a structured, extensible way to represent physical quantities (e.g., pressure, velocity) independently of any product-specific naming conventions.

### DIFF
--- a/src/ansys/fluent/core/__init__.py
+++ b/src/ansys/fluent/core/__init__.py
@@ -25,6 +25,9 @@
 import os
 from pathlib import Path
 import pydoc
+import sys
+
+from . import physicalquantities
 
 # isort: off
 # Logging has to be imported before importing other PyFluent modules
@@ -176,3 +179,9 @@ LAUNCH_FLUENT_PORT = None
 
 # Skip password check during rpc execution when Fluent is launched from PyFluent
 LAUNCH_FLUENT_SKIP_PASSWORD_CHECK = False
+
+# `ansys.physicalquantities` will eventually be a separate package. Until then:
+# Inject `physicalquantities` into the `ansys` namespace if it doesn't already exist
+if not hasattr(sys.modules["ansys"], "physicalquantities"):
+    sys.modules["ansys.physicalquantities"] = physicalquantities
+    sys.modules["ansys"].physicalquantities = physicalquantities
--- a/src/ansys/fluent/core/physicalquantities/__init__.py
+++ b/src/ansys/fluent/core/physicalquantities/__init__.py
@@ -1,0 +1,53 @@
+# Copyright (C) 2021 - 2025 ANSYS, Inc. and/or its affiliates.
+# SPDX-License-Identifier: MIT
+#
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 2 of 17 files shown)


src/ansys/fluent/core/__init__.py
```diff
@@ -176,3 +179,9 @@ def version_info() -> str:
 
 # Skip password check during rpc execution when Fluent is launched from PyFluent
 LAUNCH_FLUENT_SKIP_PASSWORD_CHECK = False
+
+# `ansys.physicalquantities` will eventually be a separate package. Until then:
+# Inject `physicalquantities` into the `ansys` namespace if it doesn't already exist
```
seanpearsonuk, line 184: In order to use it this way requires importing PyFluent up front.

src/ansys/fluent/core/__init__.py
```diff
@@ -176,3 +179,9 @@ def version_info() -> str:
...
+if not hasattr(sys.modules["ansys"], "physicalquantities"):
+    sys.modules["ansys.physicalquantities"] = physicalquantities
+    sys.modules["ansys"].physicalquantities = physicalquantities
```
hpohekar, line 187: 1. Hardcoding module names as strings can lead to typos and makes refactoring harder. We can use constants or variables so that it will improve the error handling as follows::

```
ANSYS_MODULE = "ansys"
QUANTITIES_ATTR = "physicalquantities"
ANSYS_QUANTITIES_MODULE = f"{ANSYS_MODULE }.{QUANTITIES_ATTR }" 

ansys_module = sys.modules.get(ANSYS_MODULE)

if ansys_module is not None and not hasattr(ansys_module , QUANTITIES_ATTR):
    sys.modules[ANSYS_QUANTITIES_MODULE] = QUANTITIES_ATTR 
    setattr(ansys_module, QUANTITIES_ATTR, QUANTITIES_ATTR )
else:
    raise ImportError(f"Module '{ANSYS_MODULE}' not found in sys.modules")
```

src/ansys/fluent/core/physicalquantities/__init__.py
```diff
@@ -0,0 +1,53 @@
+# Copyright (C) 2021 - 2025 ANSYS, Inc. and/or its affiliates.
```
hpohekar, line 1: Search engine optimization techniques suggest that we should avoid a long name to a file. We can use `physical_quantities.py` here as it also helps to improve both search performance and the readability. We had done the similar change to `system_coupling.py`.

src/ansys/fluent/core/physicalquantities/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+
+.. code-block:: python
+
+    >>> from ansys.fluent.core.physicalquantities import PhysicalQuantities
+    >>> from ansys.fluent.core.physicalquantities.strategies.fluent import FluentSVarStrategy
+
+    >>> quantity = PhysicalQuantities.PRESSURE
+    >>> strategy = FluentSVarStrategy()
+
+    >>> print(strategy.to_string(quantity))
```
hpohekar, line 40: 👍

This will help to improve both user guide and API reference section.

PR #4018, opened 2025-05-13
chore: Minor improvements.

prmukherj:
1. Expose `register_finalizer_cb` from 'fluent_connection' to 'session'. This is used in the visualization repo, and now can directly be accessed from session level.
2. Fixes a 'meshing_utilities' test which was causing a Fluent bug. Actually the test was passing in PyFluent without any exception while it was raising an exception in the PyConsole.


tests/test_meshing_utilities.py
```diff
@@ -697,9 +697,9 @@ def test_meshing_utilities(new_meshing_session):
 
     assert (
         meshing_session.meshing_utilities.get_maxsize_cell_zone_by_count(
-            zone_name_list=["outlet", "inlet", "wall", "internal"]
+            zone_name_list=["elbow-fluid"]
```
hpohekar, line 700: 👍 It is passing in PyFluent but raising an error in PyConsole. Not sure why this is failing in PyConsole. Thanks for updating this.

hpohekar (approved):

tests/test_meshing_utilities.py
prmukherj: Yes, it is strange. There is no exception in PyFluent, even in the Fluent console.

seanpearsonuk: What was the previous test doing instead of returning `None`?

prmukherj: Nothing, just returning None.

src/ansys/fluent/core/session.py
```diff
@@ -152,6 +152,7 @@ def __init__(
             event_type,
             get_zones_info,
         )
+        self.register_finalizer_cb = fluent_connection.register_finalizer_cb
```
seanpearsonuk, line 155: @prmukherj If exposing this, we have to give it a better named. The problem is that "cb" is a meaningless abbreviation. What naming format are we using for similar methods elsewhere?

tests/test_meshing_utilities.py
prmukherj: That's why the test was passing in CI.

src/ansys/fluent/core/session.py
prmukherj: "register_finalizer_callback" should be fine I think, as other methods are named "register_callback"

seanpearsonuk: @prmukherj Yes, that's meaningful.

tests/test_meshing_utilities.py
prmukherj: yes, actually the assertions were corrected in test as well.

src/ansys/fluent/core/session.py
prmukherj, line 155:
```suggestion
        self.register_finalizer_callback = fluent_connection.register_finalizer_cb
```

Issue #4023, opened 2025-05-14
Fluent session will not launch if AWP_ROOTXXX environment variable is for later release than installed Fluent release.


EDCarman:
### 🔍 Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### 🐞 Description of the bug

If the AWP_ROOTXXX environment variable exists for a later release than the installed Fluent release, ``launch_fluent`` will fail with default arguments. e.g. AWP_ROOT252 will break launching for if the latest version of Fluent is 251.
This can break automated workflows.

See https://github.com/ansys/ansys-tools-path/issues/296 and https://github.com/ansys/pysherlock/issues/523 for similar issues.

### 📝 Steps to reproduce

Install Ansys Fluent version 2025 R1, with no later version of any Ansys product installed.

Run the following commands:
```
>>> import os
>>> from ansys.fluent.core import launch_fluent
>>> print(os.getenv('AWP_ROOT252'))  # Confirm no 2025 R2 installed
None
>>> solver = launch_fluent()  # Correctly launches
>>> solver.is_active()
True
>>> solver.exit()
>>> os.environ['AWP_ROOT252'] = 'C:\\Program Files\\ANSYS Inc\\v252'  # Mimic 2025 R2 installation
>>> solver = launch_fluent()  # Launch fails due to missing .exe
The system cannot find the file C:\Program Files\ANSYS Inc\v252\fluent\ntbin\win64\fluent.exe.
```

### 💻 Which operating system are you using?

Windows

### 🐍 Which Python version are you using?

3.12

### 📦 Installed packages

```shell
ansys-api-fluent==0.3.35
ansys-api-platform-instancemanagement==1.1.3
ansys-api-tools-filetransfer==0.1.1
ansys-fluent-core @ git+https://github.com/ansys/pyfluent.git@26d2edd26a885a5514d02f186c6505c1420f7889
ansys-platform-instancemanagement==1.1.2
ansys-tools-filetransfer==0.1.1
ansys-units==0.4.0
certifi==2025.4.26
charset-normalizer==3.4.2
click==8.2.0
colorama==0.4.6
defusedxml==0.7.1
Deprecated==1.2.18
docker==7.1.0
googleapis-common-protos==1.70.0
grpcio==1.71.0
grpcio-health-checking==1.62.3
grpcio-status==1.62.3
idna==3.10
importlib_metadata==8.7.0
joblib==1.5.0
nltk==3.9.1
[...]

PR #4024, opened 2025-05-14
feat: Check if fluent exe exists in AWP_ROOT path while determining the Fluent version to launch

mkundu1:
This is to handle the scenario when a later version of Ansys is installed without Fluent (#4023)

### DIFF
--- a/src/ansys/fluent/core/utils/fluent_version.py
+++ b/src/ansys/fluent/core/utils/fluent_version.py
@@ -25,6 +25,8 @@
 from enum import Enum
 from functools import total_ordering
 import os
+from pathlib import Path
+import platform
 
 import ansys.fluent.core as pyfluent
 
@@ -114,8 +116,22 @@ class FluentVersion(Enum):
         AnsysVersionNotFound
             If an Ansys version cannot be found.
         """
+
+        def fluent_exe_exists(awp_var):
+            awp_root = os.getenv(awp_var)
+            if awp_root is None:
+                return False
+
+            fluent_root = Path(awp_root) / "fluent"
+            fluent_exe = (
+                fluent_root / "ntbin" / "win64" / "fluent.exe"
+                if platform.system() == "Windows"
+                else fluent_root / "bin" / "fluent"
+            )
+            return fluent_exe.exists()
+
         for member in cls:
-            if member.awp_var in os.environ:
+            if fluent_exe_exists(member.awp_var):
                 return member
 
         raise AnsysVersionNotFound(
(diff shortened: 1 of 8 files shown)


src/ansys/fluent/core/utils/fluent_version.py
```diff
@@ -113,8 +115,22 @@ def get_latest_installed(cls):
         AnsysVersionNotFound
             If an Ansys version cannot be found.
         """
+
```
seanpearsonuk, line 118: @mkundu1

In this class, we can add ~ `get_latest_installed_fluent_exe_path()` implemented in terms of `get_latest_installed()` and adding the new code that maps Ansys path to Fluent path. Then we can return the full path that we are checking below. The calling module for `get_latest_installed()` also does the same Ansys to Fluent path mapping so by adding this method we can eliminate the duplication.

src/ansys/fluent/core/utils/fluent_version.py
```diff
@@ -113,8 +115,22 @@ def get_latest_installed(cls):
...
+            fluent_exe = (
+                fluent_root / "ntbin" / "win64" / "fluent.exe"
+                if platform.system() == "Windows"
+                else fluent_root / "bin" / "fluent"
+            )
+            return fluent_exe.exists()
+
         for member in cls:
-            if member.awp_var in os.environ:
+            if fluent_exe_exists(member.awp_var):
```
seanpearsonuk, line 133: @mkundu1 
The following code,
```python
raise AnsysVersionNotFound(
            "Verify the value of the 'AWP_ROOT' environment variable."
        )
```
can become
```python
raise AnsysVersionNotFound()
```
The message, fixed in the exception class, can be something like "Unable to locate a compatible Ansys Fluent installation. Ensure that an environment variable like 'AWP_ROOT242' or 'AWP_ROOT251' points to a supported Ansys version, and that Fluent is included in the installation.".

src/ansys/fluent/core/utils/fluent_version.py
```diff
@@ -115,11 +117,29 @@ def get_latest_installed(cls):
             If an Ansys version cannot be found.
         """
         for member in cls:
-            if member.awp_var in os.environ:
+            if member.awp_var in os.environ and member.get_fluent_exe_path().exists():
                 return member
 
         raise AnsysVersionNotFound(
-            "Verify the value of the 'AWP_ROOT' environment variable."
+            "Unable to locate a compatible Ansys Fluent installation. "
```
seanpearsonuk, line 124: @mkundu1 Noting that we have three places where theis exception type is raised. One is version-specific and two are not. We can consolidate the message creation in the class itself.

seanpearsonuk (approved):

hpohekar (approved):

PR #4028, opened 2025-05-16
chore: Add a way to extract Fluent launch command and subprocess arguments from StandaloneLauncher

mkundu1:
```
import subprocess
from ansys.fluent.core.launcher.standalone_launcher import StandaloneLauncher
launcher = StandaloneLauncher(ui_mode="gui", dimension=2)
subprocess.run(launcher.command, **launcher.subprocess_keywords)
```

### DIFF
--- a/src/ansys/fluent/core/launcher/standalone_launcher.py
+++ b/src/ansys/fluent/core/launcher/standalone_launcher.py
@@ -184,7 +184,7 @@ class StandaloneLauncher:
         In job scheduler environments (e.g., SLURM, LSF, PBS), resources and compute nodes are allocated,
         and core counts are queried from these environments before being passed to Fluent.
         """
-        import ansys.fluent.core as pyfluent
+        env = env or {}
 
         locals_ = locals().copy()
         argvals = {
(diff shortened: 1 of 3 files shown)


src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -246,6 +248,16 @@ def __init__(
...
+    @property
+    def command(self) -> str:
+        """Return the command for launching Fluent."""
+        return self._launch_string
+
+    @property
+    def subprocess_keywords(self) -> dict[str, Any]:
+        """Return the keyword arguments for launching Fluent via Python subprocess."""
+        return self._kwargs
+
```
hpohekar, line 260: Where are we going to use this information ? Does it require for the standalone mode only ?

seanpearsonuk:
@mkundu1 I would prefer to create a `DetachedLauncher` class. **The absolutely simplest implementation would just wrap the code you showed in the description.** A more elaborate design would use a new shared base class, which would need to expose `launcher.command` and `launcher.subprocess_keywords` to subclasses.

PR #4031, opened 2025-05-19
test: Update test w.r.t. recent exposure flag changes in data-model.

prmukherj:
Update couple of tests with respect to changes in the exposure level flags to developer.
This has to be merged into the release v0.30 with a patch as well.


prmukherj:
We need to update the tests here as the testing is using these tests as reference. It should be fine even if we don't create a patch release for it. The docs should be up to date there.

hpohekar (approved):

PR #4056, opened 2025-05-22
chore: skip tests for nightly tests to pass [skip tests]

prmukherj:
FAILED tests/test_launcher_remote.py::test_launch_remote_instance - ValueError: The request failed to complete because the connection has not been validated. You can validate the connection by connecting to a connection service with a password or by sending a password as a key, value pair in the context of each request. The key should be the password.

This test is skipped for now. Investigate to fix it later.


tests/test_reduction.py
```diff
@@ -408,6 +408,7 @@ def static_mixer_case_session2(static_mixer_case_session: Any):
     return static_mixer_case_session
 
 
+@pytest.mark.skip("https://github.com/ansys/pyfluent/issues/4055")
```
prmukherj, line 411:
```suggestion
```

hpohekar (approved):

hpohekar:
@prmukherj Updated title with '[skip tests]' so that it will save our time.

mkundu1 (approved):

PR #4119, opened 2025-06-09
docs: Address inconsistencies in naming svars.

prmukherj:


src/ansys/fluent/core/services/solution_variables.py
```diff
@@ -293,13 +302,12 @@ def get_zones_info(self) -> ZonesInfo:
 class SvarError(ValueError):
     """Exception class for errors in solution variable name."""
 
-    def __init__(self, solution_variable_name: str, allowed_values: List[str]):
+    def __init__(self, svar_name: str, allowed_values: List[str]):
```
seanpearsonuk: @prmukherj I believe it's better to rename it variable. We have deliberately not exposed svar so much. The exception class itself could have been renamed SolutionVariableNameError.

prmukherj: @seanpearsonuk, Done, Thank you. I have named the exception class: "InvalidSolutionVariableNameError" just to be consistent with other exception classes in PyFluent.

mkundu1 (approved):

hpohekar (approved):

PR #4135, opened 2025-06-10
ci: Remove docker images after any job that consumes Fluent image

mkundu1:
Cleanup docker images after use to avoid the disk-space issue in self-hosted runners.

I'll push similar changes to other repos that consumes Fluent images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,9 @@ version-info:
 docker-pull:
 	@bash .ci/pull_fluent_image.sh
 
+docker-clean-images:
+	@docker image prune -a -f
+
 test-import:
 	@python -c "import ansys.fluent.core as pyfluent"
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -240,6 +240,10 @@ jobs:
           path: HTML-Documentation-tag-${{ env.DOC_DEPLOYMENT_IMAGE_TAG }}.zip
           retention-days: 7
 
+      - name: Remove all docker images
+        if: always()
+        run: make docker-clean-images
+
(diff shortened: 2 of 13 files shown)


prmukherj (approved):

prmukherj:
> > Cleanup docker images after usage to avoid the disk-space issue in self-hosted runners.
> > I'll push similar changes to other repos that consumes Fluent images.
> 
> Will do it for Viz.

https://github.com/ansys/pyfluent-visualization/pull/533

Makefile
```diff
@@ -21,6 +21,9 @@ version-info:
 docker-pull:
 	@bash .ci/pull_fluent_image.sh
 
+docker-clean-images:
+	@docker image prune -a -f
```
hpohekar, line 25:
```suggestion
	@docker system prune --volumes -a -f 
```

We shall use `docker system prune --volumes -a -f` because our main objective here is to free maximum space. 
`docker system prune --volumes -a -f` removes all unused images, stopped containers, unused networks, build cache and dangling volumes. 

Removal of unused network, dangling volumes and build cache will give more space and it will help to avoid hang issues which are not covered in `docker image prune -a -f`.

We also do not need the separate job `cleanup-previous-docker-containers` in other workflows. 

`docker system prune --volumes -a -f ` will address all issues at once and we can avoid adding separate job for each specific case.

hpohekar: I have verified that `docker system prune --volumes -a -f` takes similar time to complete the Build job.

https://github.com/ansys/pyfluent/actions/runs/15575645692/job/43860002228?pr=4139

PR #4140, opened 2025-06-11
chore: Add warning while using Fluent develop branch.

prmukherj:
The below error messages will be updated as follows:
image

1. 
image

2. 
image

3. 
image

4. 
image

5. 
image

6. 
image

7. 
image

8. Connect to fluent also gives correct behaviour:
image

### DIFF
--- a/src/ansys/fluent/core/exceptions.py
+++ b/src/ansys/fluent/core/exceptions.py
@@ -24,7 +24,6 @@
 from typing import Any, Iterable
 import warnings
 
-import ansys.fluent.core as pyfluent
 from ansys.fluent.core.solver.error_message import allowed_name_error_message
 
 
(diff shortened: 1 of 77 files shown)


prmukherj:
The warning message provided here is almost a replica of the Fluent warning message. Please suggest if a better more informative or self explaining message can be provided.

prmukherj:
One query here is that when the user has a release version of Fluent and from it's 'commonfiles' he tries using 'ansys-fluent-core', he will get this message, as we have built and packaged it during develop. Will this not be confusing for users? Shall we document something that the commonfiles pyfluent should not be used or something? or we can improve out warning message to inform the users better.
@seanpearsonuk, @mkundu1, @hpohekar

src/ansys/fluent/core/exceptions.py
```diff
@@ -51,3 +52,21 @@ class InvalidArgument(ValueError):
...
+
+    if (
+        pyfluent.utils.load_module(
+            f"fluent_version_{version}", CODEGEN_OUTDIR / f"fluent_version_{version}.py"
+        ).FLUENT_BRANCH
+        == "develop"
+    ):
+        warnings.warn(
+            "WARNING:\n"
+            "This is using a prototype version of FLUENT that has not yet been tested and validated.\n"
```
prmukherj: do version checking.

"The APIs are still under development for this version of Fluent."

mkundu1 (approved):

src/ansys/fluent/core/session_shared.py
```diff
@@ -67,9 +69,10 @@ def _make_datamodel_module(session, module_name):
 
         file_name = datamodel_file_name_map[module_name]
         module = pyfluent.utils.load_module(
-            f"{module_name}_{session._version}",
+            f"{module_name if module_name != 'MeshingUtilities' else file_name}_{session._version}",
```
seanpearsonuk: @prmukherj 
- This kind of mysterious edge case handling deserves a comment
- The special handling is for one value of one of the arguments. Therefore the caller should handle the problem. `_make_datamodel_module()` can accept an additional argument `module_file_name=None`

src/ansys/fluent/core/pyfluent_warnings.py
```diff
@@ -38,6 +38,18 @@ class PyFluentUserWarning(UserWarning):
...
+def warning_for_fluent_dev_version(version):
+    """Provides warning if Fluent develop branch is used."""
+    from ansys.fluent.core import FLUENT_RELEASE_VERSION, FluentVersion
+
+    if FluentVersion(version) > FluentVersion(FLUENT_RELEASE_VERSION):
+        warnings.warn(
+            "⚠️ Warning: You are using PyFluent with an unreleased or development version of Fluent.\n"
+            "Compatibility is not guaranteed, and unexpected behavior may occur. Please use a released "
+            "version of Fluent that is officially supported by this version of PyFluent."
+        )
```
hpohekar: Can we use `PyFluentUserWarning` here ? (e.g. `warnings.warn(<message>, PyFluentUserWarning)` as we have 2 distinct categories for warnings as `PyFluentDeprecationWarning` and `PyFluentUserWarning`.

prmukherj: Done.

hpohekar (approved):

Issue #4143, opened 2025-06-11
Make graphics driver `null` except in `gui` / `hidden_gui` modes


seanpearsonuk:
### Problem

When launching Fluent in batch mode (e.g., `ui_mode=None` or `"no_gui"`), plots may still try to open unless `-driver null` is explicitly passed. This can cause Fluent to hang in headless environments (especially on Linux). Most users are not aware of this subtlety.

### Solution

We should handle this automatically in the launcher by appending `-driver null` unless the UI mode explicitly supports/needs GUI rendering.

---

### ✅ Requirement 1 — Default to `-driver null`

Add logic to automatically append `-driver null` unless the UI mode is `"gui"` or `"hidden_gui"`.  
This includes handling the case where `ui_mode` is `None` (defaulting to `"NO_GUI"` on Linux).

#### Pseudocode

```python
def should_add_driver_null(ui_mode: str | None) -> bool:
    if ui_mode is None:
        ui_mode = get_default_ui_mode()  # e.g., "NO_GUI" on Linux
    return ui_mode.lower() not in {"gui", "hidden_gui"}

args = existing_user_args.copy()
if should_add_driver_null(ui_mode):
    args.extend(["-driver", "null"])
```

### ✅ Requirement 2 — Detect invalid driver setting
We already have an argument to specify the driver. If the user specifies any non-null driver (including auto) and we are in a UI mode that deactivates the UI then we should fdo all the follwowing:
- do not respect the driver selection
- show a warning
If the UI mode is GUI or hidden, then we need to start fluent with driver=auto which should be automatic (no action needed on PyFluent side).
~Add a new boolean argument to the launcher, `disable_driver_null=False`.~
~If this is True, then the automatic addition of -driver null is skipped, even if the UI mode is "no_gui" or similar.~

~This enables advanced users to deliberately bypass the default behavior for special use cases (e.g., using a virtual framebuffer like Xvfb).~

seanpearsonuk:
@tejalprabhu01 @mkundu1

PR #4149, opened 2025-06-12
feat: Make graphics driver null except in gui / hidden_gui modes

hpohekar:
closes https://github.com/ansys/pyfluent/issues/4143

Make graphics driver null except in gui / hidden_gui modes.

### DIFF
--- a/src/ansys/fluent/core/launcher/process_launch_string.py
+++ b/src/ansys/fluent/core/launcher/process_launch_string.py
@@ -25,6 +25,8 @@
 import json
 import os
 from pathlib import Path
+import platform
+import warnings
 
 import ansys.fluent.core as pyfluent
 from ansys.fluent.core.launcher import launcher_utils
@@ -34,11 +36,24 @@ from ansys.fluent.core.launcher.launch_options import (
     Precision,
     UIMode,
 )
+from ansys.fluent.core.pyfluent_warnings import PyFluentUserWarning
 from ansys.fluent.core.scheduler import build_parallel_options, load_machines
 from ansys.fluent.core.utils.fluent_version import FluentVersion
 
 _THIS_DIR = os.path.dirname(__file__)
 _OPTIONS_FILE = os.path.join(_THIS_DIR, "fluent_launcher_options.json")
+
+
+def _get_default_ui_mode() -> str:
+    """Get the default UI mode based on the operating system."""
+    if platform.system() == "Linux":
+        return UIMode.NO_GUI
+
+
+def _should_add_driver_null(ui_mode: UIMode | None = None) -> bool:
+    if ui_mode is None:
+        ui_mode = _get_default_ui_mode()
+    return ui_mode not in {UIMode.GUI, UIMode.HIDDEN_GUI}
 
 
 def _build_fluent_launch_args_string(**kwargs) -> str:
@@ -100,11 +115,25 @@ def _build_fluent_launch_args_string(**kwargs) -> str:
     elif isinstance(gpu, list):
         launch_args_string += f" -gpu={','.join(map(str, gpu))}"
     ui_mode = UIMode(kwargs.get("ui_mode"))
-    if ui_mode and ui_mode.get_fluent_value()[0]:
-        launch_args_string += f" -{ui_mode.get_fluent_value()[0]}"
+    ui_mode_value = ui_mode.get_fluent_value()[0]
+    if ui_mode_value:
+        launch_args_string += f" -{ui_mode_value}"
     graphics_driver = kwargs.get("graphics_driver")
-    if graphics_driver and graphics_driver.get_fluent_value()[0]:
-        launch_args_string += f" -driver {graphics_driver.get_fluent_value()[0]}"
+    graphics_driver_value = graphics_driver.get_fluent_value()[0]
+    if _should_add_driver_null(ui_mode):
+        launch_args_string += " -driver null"
+    if graphics_driver_value != "null" and ui_mode not in {
+        UIMode.GUI,
+        UIMode.HIDDEN_GUI,
+    }:
+        warnings.warn(
+            """\nIf the UI mode is ``UIMode.GUI`` or ``UIMode.HIDDEN_GUI``,
+                      then we need to start fluent with ``FluentWindowsGraphicsDriver.AUTO`` or ``FluentLinuxGraphicsDriver.AUTO``
+                      which should be automatic (no action needed on PyFluent side).\n""",
+            PyFluentUserWarning,
+        )
+    if graphics_driver_value != "null" and ui_mode in {UIMode.GUI, UIMode.HIDDEN_GUI}:
+        launch_args_string += f" -driver {graphics_driver_value}"
     return launch_args_string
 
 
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/launcher/process_launch_string.py
```diff
@@ -34,13 +36,26 @@
...
 _THIS_DIR = os.path.dirname(__file__)
 _OPTIONS_FILE = os.path.join(_THIS_DIR, "fluent_launcher_options.json")
 
 
+def _get_default_ui_mode(ui_mode: UIMode | None = None) -> UIMode:
+    """Get the default UI mode based on the operating system."""
+    if platform.system() == "Linux":
+        return UIMode.NO_GUI
+    else:
+        return ui_mode
```
seanpearsonuk: @prmukherj Please could you explain the logic here? Is this correct? This seems to be overriding the mode that has been input by the user, thereby completely ignoring the user's selection. What am I missing?

prmukherj: It is for getting the default mode for a specific system, right? If the system is linux the default mode is no_gui. Is it not true?

prmukherj: Although we already have a default behaviour that is platform dependent already:
image

src/ansys/fluent/core/launcher/process_launch_string.py
```diff
@@ -99,12 +114,31 @@ def _build_fluent_launch_args_string(**kwargs) -> str:
         launch_args_string += " -gpu"
     elif isinstance(gpu, list):
         launch_args_string += f" -gpu={','.join(map(str, gpu))}"
-    ui_mode = UIMode(kwargs.get("ui_mode"))
-    if ui_mode and ui_mode.get_fluent_value()[0]:
-        launch_args_string += f" -{ui_mode.get_fluent_value()[0]}"
+    ui_mode_input = kwargs.get("ui_mode")
+    ui_mode = _get_default_ui_mode(UIMode(ui_mode_input))
```
prmukherj: Thank you @seanpearsonuk, had missed this part where we were over-writing the user's input as well.

src/ansys/fluent/core/launcher/process_launch_string.py
hpohekar: We don't need `_get_default_ui_mode` it is alreday handled by `_default`.

src/ansys/fluent/core/launcher/process_launch_string.py
```diff
@@ -99,12 +105,30 @@ def _build_fluent_launch_args_string(**kwargs) -> str:
...
-        launch_args_string += f" -{ui_mode.get_fluent_value()[0]}"
+    ui_mode = kwargs.get("ui_mode")
+    ui_mode_value = ui_mode.get_fluent_value()[0] if ui_mode else None
+
+    if ui_mode_value:
+        launch_args_string += f" -{ui_mode_value}"
+
     graphics_driver = kwargs.get("graphics_driver")
-    if graphics_driver and graphics_driver.get_fluent_value()[0]:
-        launch_args_string += f" -driver {graphics_driver.get_fluent_value()[0]}"
```
hpohekar: @seanpearsonuk @mkundu1 @prmukherj 

If graphics driver is `AUTO` its value is `''` and it needs to be treated separately.

I just realized that we were not appending `-driver` if it is `AUTO` (i.e. `''`)

We had not considered this case in the issue description, as codegen started failing then I investigated it.

Following was the difference between current PR and main branch behavior -

PR - 
`Fluent launch string: D:\ANSYSDev\vNNN\fluent\ntbin\win64\fluent.exe 3ddp -hidden -driver  -sifile=D:\TEMP\serverinfo-t1wbaxs1.txt -nm`

Main branch
`Fluent launch string: D:\ANSYSDev\vNNN\fluent\ntbin\win64\fluent.exe 3ddp -hidden -sifile=D:\TEMP\serverinfo-qa8m1vg7.txt -nm`

It's corrected now and we have also added a test for it.

mkundu1: @hpohekar We already have a function `_get_graphics_driver` in `launch_options.py`. Can we modify that to return the correct graphics driver based on ui_mode? The warning can also be shown there. The function `_build_fluent_launch_args_string` doesn't need to be changed.

tests/test_launcher.py
```diff
@@ -583,3 +558,95 @@ def test_docker_compose(monkeypatch):
...
     solver.exit()
+
+
+def test_respect_driver_is_not_null():
+    assert (
+        _build_fluent_launch_args_string(
+            ui_mode=UIMode.GUI,
+            graphics_driver=(
+                FluentWindowsGraphicsDriver.DX11
+                if is_windows()
```
seanpearsonuk: I'm confused here. We should be controlling the behaviour of `is_windows()` within the test so that we deliberately test each platform-specific behaviour. The outcome of the test shouldn't depend on the actual, real platform that it happens to be running on. Perhaps I'm misunderstanding?

@hpohekar

tests/test_launcher.py
```diff
@@ -583,3 +558,95 @@ def test_docker_compose(monkeypatch):
...
+                else FluentLinuxGraphicsDriver.X11
+            ),
+            additional_arguments="",
+            processor_count=None,
+        ).strip()
+        == "3ddp -driver dx11"
+        if is_windows()
```
seanpearsonuk: @hpohekar Similar kind of question runs throughout.

src/ansys/fluent/core/launcher/process_launch_string.py
hpohekar: @mkundu1 Sure. Thank you.

tests/test_launcher.py
hpohekar: Actually I referred existing tests and added tests based on this, otherwise I had added separate tests for both Windows and Linux. We will update these tests. Thank you.

tests/test_launcher.py
hpohekar: We will update this. Thank you.

src/ansys/fluent/core/launcher/process_launch_string.py
hpohekar: Done. Thank you.

tests/test_launcher.py
hpohekar: We have updated the tests now.

src/ansys/fluent/core/launcher/launch_options.py
```diff
@@ -294,6 +302,22 @@ def _get_graphics_driver(
         if is_windows()
         else FluentLinuxGraphicsDriver(graphics_driver)
     )
+    if graphics_driver.value != "null" and ui_mode_ not in {
+        UIMode.GUI,
+        UIMode.HIDDEN_GUI,
+    }:
+        warnings.warn(
+            """\nIf the UI mode is ``UIMode.GUI`` or ``UIMode.HIDDEN_GUI``,
```
mkundu1: I don't think there is any restriction for setting graphics driver in `GUI` and `HIDDEN_GUI` mode.

We want to warn user if they specify graphics driver in other modes.

hpohekar: @seanpearsonuk We have updated the warning message.

hpohekar: @mkundu1 Thanks for the suggestion.

Issue #4151, opened 2025-06-12
Resolve unpinned tag security warnings


hpohekar:
As title says

PR #4152, opened 2025-06-12
ci: Resolve unpinned tag security warnings [skip tests]

hpohekar:
closes https://github.com/ansys/pyfluent/issues/4151

Resolve unpinned tag security warnings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -199,7 +199,7 @@ jobs:
 
       - name: Login to GitHub Container Registry
         if: steps.cache-api-code.outputs.cache-hit != 'true'
-        uses: docker/login-action@v3.4.0
+        uses: docker/login-action@c65cd883420fd2eb864698a825fc4162dd94482c # v44
         with:
           registry: ghcr.io
           username: ansys-bot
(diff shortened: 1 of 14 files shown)


.github/workflows/ci.yml
```diff
@@ -199,7 +199,7 @@ jobs:
 
       - name: Login to GitHub Container Registry
         if: steps.cache-api-code.outputs.cache-hit != 'true'
-        uses: docker/login-action@v3
+        uses: docker/login-action@74a5d142397b4f367a81961eba4e8cd7edddf772 # v3.4.0
```
mkundu1, line 202: Does 74a5d142397b4f367a81961eba4e8cd7edddf772 point to v3.4.0? Pinning the version would be preferrred over pinning a SHA id.

hpohekar: @mkundu1 

Does 74a5d142397b4f367a81961eba4e8cd7edddf772 point to v3.4.0? - Yes

Pinning the version would be preferred over pinning a SHA id. - That's why it is giving security warning and we should avoid pinning version. Roberto has explained this in the channel.

mkundu1 (approved):

.github/workflows/ci.yml
mkundu1: Thanks for explaining

PR #4153, opened 2025-06-12
feat: beta features access

prmukherj:
Allow enabling of beta features both in meshing and solver mode.

For now only 2 features are there:
1. In meshing mode: topology based meshing.
2. In solver mode, "switch_to_meshing" mode.

Both of these are accessible only after enabling beta features. Please refer tests for details.


src/ansys/fluent/core/session_pure_meshing.py
```diff
@@ -238,3 +237,15 @@ def transfer_mesh_to_solvers(
...
+
+    def enable_beta_features(self):
+        """Enable access to Fluent beta-features"""
+        if not self._app_utilities.is_beta_enabled():
+            self.tui.beta_feature_access("yes", "yes")
+
+    def __dir__(self):
+        if self._fluent_connection is None:
+            return ["is_active"]
+        if self._app_utilities.is_beta_enabled():
```
seanpearsonuk: @prmukherj I believe that we should implement `_is_beta_enabled()` at base-class level.

prmukherj: yes, sounds right.

src/ansys/fluent/core/session_pure_meshing.py
```diff
@@ -238,3 +237,15 @@ def transfer_mesh_to_solvers(
             clean_up_mesh_file,
             overwrite_previous,
         )
+
```
seanpearsonuk, line 240: @prmukherj I believe that this should be a base class method that looks like:
```python
    def enable_beta_features(self):
        """Enable access to Fluent beta-features"""
        # - app utilities encapsulates checking of current status
        # - server figures out how to concretely enable beta (given that the action might be different per Fluent run mode)
        self._app_utilities.enable_beta(self)
```

src/ansys/fluent/core/session_pure_meshing.py
```diff
@@ -238,3 +237,15 @@ def transfer_mesh_to_solvers(
...
+    def enable_beta_features(self):
+        """Enable access to Fluent beta-features"""
+        if not self._app_utilities.is_beta_enabled():
+            self.tui.beta_feature_access("yes", "yes")
+
+    def __dir__(self):
+        if self._fluent_connection is None:
+            return ["is_active"]
```
seanpearsonuk: @prmukherj The strategy of what to do when detached is a core decision, not for concrete leaf classes.

src/ansys/fluent/core/session_pure_meshing.py
prmukherj: AppUtilities should contain a enable_beta?

seanpearsonuk: I'm convinced of it.

Anything like a concrete TUI/Scheme dependency needs to be pushed as far into the back-end as possible &mdash; otherwise we'll slide back to where we were. For older servers, we can implement it in the app utilities client wrapper.

prmukherj: Right, will do it. It's actually agrees with something suggested by Mainak as well. @mkundu1

prmukherj: Done, Thank you.

src/ansys/fluent/core/session_pure_meshing.py
prmukherj: Done, Thank you.

prmukherj:
This PR will be merged after v26R1 Fluent changes are in place.

src/ansys/fluent/core/services/app_utilities.py
```diff
@@ -255,6 +261,18 @@ def is_beta_enabled(self) -> bool:
...
+    def enable_beta(self):
+        """Enable beta features.
+
+        Raises
+        ------
+        RuntimeError
+            Not supported before Fluent 2025 R2.
+        """
+        raise RuntimeError(
+            "Enabling beta is not supported py PyFluent for Fluent versions before 2025 R2."
```
seanpearsonuk: by

prmukherj:
```suggestion
            "Enabling beta is not supported by PyFluent for Fluent versions before 2025 R2."
```

PR #4158, opened 2025-06-12
chore: Fix github caching for codegen

mkundu1:
GitHub caching for codegen is not working due to the following warning while saving the cache:
Warning: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.

https://github.com/ansys/pyfluent/actions/runs/15606958362/job/43970126130#step:88:2

This PR fixes the paths included in cache.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -190,12 +190,7 @@ jobs:
         with:
           path: |
             src/ansys/fluent/core/generated
-            doc/source/api/meshing/tui
-            doc/source/api/meshing/datamodel
-            doc/source/api/solver/tui
-            doc/source/api/solver/datamodel
           key: API-Code-v${{ env.API_CODE_CACHE }}-${{ steps.version.outputs.PYFLUENT_VERSION }}-${{ env.DOC_DEPLOYMENT_IMAGE_TAG }}-${{ hashFiles('src/ansys/fluent/core/codegen/**') }}
-          restore-keys: API-Code-v${{ env.API_CODE_CACHE }}-${{ steps.version.outputs.PYFLUENT_VERSION }}-${{ env.DOC_DEPLOYMENT_IMAGE_TAG }}
 
       - name: Login to GitHub Container Registry
         if: steps.cache-api-code.outputs.cache-hit != 'true'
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -294,12 +294,7 @@ jobs:
         id: cache-231-api-code
         with:
           path: src/ansys/fluent/core/generated
-            doc/source/api/core/meshing/tui
-            doc/source/api/core/meshing/datamodel
-            doc/source/api/core/solver/tui
-            doc/source/api/core/solver/datamodel
           key: API-Code-v${{ env.API_CODE_CACHE }}-${{ steps.version.outputs.PYFLUENT_VERSION }}-v23.1.0-${{ hashFiles('src/ansys/fluent/core/codegen/**') }}
-          restore-keys: API-Code-v${{ env.API_CODE_CACHE }}-${{ steps.version.outputs.PYFLUENT_VERSION }}-v23.1.0
```
mkundu1, line 302: We don't need the `restore-keys` as the full codegen output is rebuilt when the cache is missed. `restore-keys` are useful for incremental builds.

.github/workflows/ci.yml
```diff
@@ -289,26 +284,23 @@ jobs:
...
+        id: cache-api-code
         with:
           path: src/ansys/fluent/core/generated
-            doc/source/api/core/meshing/tui
-            doc/source/api/core/meshing/datamodel
-            doc/source/api/core/solver/tui
-            doc/source/api/core/solver/datamodel
-          key: API-Code-v${{ env.API_CODE_CACHE }}-${{ steps.version.outputs.PYFLUENT_VERSION }}-v23.1.0-${{ hashFiles('src/ansys/fluent/core/codegen/**') }}
-          restore-keys: API-Code-v${{ env.API_CODE_CACHE }}-${{ steps.version.outputs.PYFLUENT_VERSION }}-v23.1.0
+          # Combined cache key for all versions:
```
mkundu1, line 292: As restored path is identical for all Fluent versions, it makes sense to use a combined cache.

prmukherj (approved):

hpohekar (approved):

Issue #4161, opened 2025-06-13
Looking up static class attributes within the client-side generated settings classes should not require is_active() = True precondition.


mkundu1:
Update flobject.py code to allow client-side lookup of such attributes without requiring is_active() checks.

prmukherj:
Also do as part of this:
https://github.com/ansys/pyfluent/pull/4162#discussion_r2149342489

PR #4162, opened 2025-06-13
chore: Perform deprecation call from generated data.

prmukherj:
Perform deprecation call from generated data.

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -97,7 +97,6 @@ class InactiveObjectError(RuntimeError):
 
 class _InlineConstants:
     is_active = "active?"
-    is_stable = "webui-release-active?"
     is_read_only = "read-only?"
     default_value = "default"
     min = "min"
@@ -195,7 +194,7 @@ _to_field_name_str = naming_strategy().to_string
 def _get_python_path_comps(obj):
     """Get python path components for traversing class hierarchy."""
     comps = []
-    while obj:
+    while obj is not None:
         python_name = obj.python_name
         obj = obj._parent
         if isinstance(obj, (NamedObject, ListObject)):
@@ -304,12 +303,10 @@ class Base:
 
         Supports file upload and download.
         """
-        with warnings.catch_warnings():
-            warnings.filterwarnings(action="ignore", category=UnstableSettingWarning)
-            if self._file_transfer_service:
-                return self._file_transfer_service
-            elif self._parent:
-                return self._parent._file_transfer_handler
+        if self._file_transfer_service:
+            return self._file_transfer_service
+        elif self._parent:
+            return self._parent._file_transfer_handler
 
     _name = None
     fluent_name = None
@@ -425,20 +422,6 @@ class Base:
         """Whether the object is active."""
         attr = self.get_attr(_InlineConstants.is_active)
         return False if attr is False else True
-
-    def _check_stable(self) -> None:
-        """Whether the object is stable."""
-        if not self.is_active():
-            return
-        attr = self.get_attr(_InlineConstants.is_stable)
-        attr = True if attr is None else attr
-        if not attr:
-            warnings.warn(
-                f"The API feature at '{self.path}' is not stable. "
-                f"It is not guaranteed that it is fully validated and "
-                f"there is no commitment to its backwards compatibility.",
-                UnstableSettingWarning,
-            )
 
     def is_read_only(self) -> bool:
         """Whether the object is read-only."""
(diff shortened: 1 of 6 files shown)


doc/settings_rstgen.py
```diff
@@ -139,8 +139,9 @@ def _populate_rst_from_settings(rst_dir, cls, version):
         r.write(f'{"="*(len(cls_orig_name))}\n\n')
         deprecated = getattr(cls, "_deprecated_version", None)
         if deprecated:
-            r.write(f".. deprecated:: Ansys {cls._deprecated_version}\n\n")
-            deprecated_class_version.update({cls_name: deprecated})
+            release_version = "20" + cls._deprecated_version.replace(".", "R")
```
prmukherj, line 142: This is only required for documentation, so moved it here.
@hpohekar, could you please confirm that there will be no change in behaviour for docs?

src/ansys/fluent/core/codegen/settingsgen.py
```diff
@@ -230,8 +230,7 @@ def _write_data(cls_name: str, python_name: str, data: dict, f: IO, f_stub: IO |
     s.write(f"    _version = {data['version']!r}\n")
     deprecated = data["deprecated_version"]
     if deprecated:
-        release_version = "20" + data["deprecated_version"].replace(".", "R")
-        s.write(f"    _deprecated_version = {release_version!r}\n")
+        s.write(f"    _deprecated_version = {deprecated!r}\n")
```
prmukherj, line 233: Leave it as it is coming from server.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -224,6 +224,15 @@ def _get_class_from_paths(root_cls, some_path: list[str], other_path: list[str])
     return cls, full_path
 
 
+def _is_deprecated(obj) -> bool | None:
```
prmukherj, line 226: A module level method to perform the localized deprecation checks from generated data.

mkundu1 (approved):

doc/settings_rstgen.py
hpohekar: This should work fine.

doc/settings_rstgen.py
```diff
@@ -139,8 +139,9 @@ def _populate_rst_from_settings(rst_dir, cls, version):
...
+            r.write(f".. deprecated:: Ansys {release_version}\n\n")
+            deprecated_class_version.update({cls_name: release_version})
```
seanpearsonuk, line 144: @prmukherj 
This function is getting increasingly complex and hard to follow, especially with the addition of edge-case handling like the `if deprecated:` block. A few concerns:

- It’s now doing too much. We should consider decomposing it into smaller, well-named helper functions for clarity and testability.

- The magic string logic for converting versions (e.g., `replace(".", "R")`) is brittle and obscure. This logic belongs in the `FluentVersion` class, where it can be centrally maintained and tested.

- Some parts (e.g., `cls_orig_name == "root"` and hardcoded RST section generation) would benefit from explanatory comments.

- As the doc generation grows, we need to ensure we’re not making the code harder for others (or future us) to safely maintain.

Could we take a moment to refactor this function as part of this PR, or otherwise track it for near-term cleanup?

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -224,6 +224,15 @@ def _get_class_from_paths(root_cls, some_path: list[str], other_path: list[str])
...
+    """Whether the object is deprecated in a specific Fluent version."""
+    deprecated_version = getattr(obj, "_deprecated_version", None)
+    return deprecated_version and (
+        float(deprecated_version) <= 22.2
```
seanpearsonuk: @prmukherj Should this comparison go via the `FluentVersion` class? Is it possible? (It might require `FluentVersion` to be updated methinks.)

doc/settings_rstgen.py
prmukherj: I strongly agree with all your points and we had a discussion regarding this as well with @mkundu1. Actually it was an already existing code and as part of this PR we just moved it. There will be another part to this PR that will contain some refactoring changes, for which @mkundu1 has already created a task. I'll address this as well then.

Thank you for your suggestions.

src/ansys/fluent/core/solver/flobject.py
prmukherj: yes, possible. Will do it.

doc/settings_rstgen.py
prmukherj: https://github.com/ansys/pyfluent/issues/4161

PR #4164, opened 2025-06-13
feat: Add builtin commands

mkundu1:
Add built-in commands at solver level:
```
import ansys.fluent.core as pf
pf.solver.ReadCase(...)
pf.solver.Initialize()
```
TODO: update doc examples

### DIFF
--- a/src/ansys/fluent/core/solver/settings_builtin_data.py
+++ b/src/ansys/fluent/core/solver/settings_builtin_data.py
@@ -1168,4 +1168,11 @@ DATA = {
     "DesignPoints": ("Singleton", "parametric_studies.design_points"),
     "DesignPoint": ("NamedObject", "parametric_studies.design_points"),
     "ReadCase": ("Command", "file.read_case"),
+    "ReadData": ("Command", "file.read_data"),
+    "ReadCaseData": ("Command", "file.read_case_data"),
+    "WriteCase": ("Command", "file.write_case"),
+    "WriteData": ("Command", "file.write_data"),
+    "WriteCaseData": ("Command", "file.write_case_data"),
+    "Initialize": ("Command", "solution.initialization.initialize"),
+    "Calculate": ("Command", "solution.run_calculation.calculate"),
 }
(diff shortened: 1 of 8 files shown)


src/ansys/fluent/core/solver/settings_builtin_data.py
```diff
@@ -1167,4 +1167,5 @@
     "ParametricStudy": ("NamedObject", "parametric_studies"),
     "DesignPoints": ("Singleton", "parametric_studies.design_points"),
     "DesignPoint": ("NamedObject", "parametric_studies.design_points"),
+    "ReadCase": ("Command", "file.read_case"),
```
mkundu1, line 1170: More to be added

src/ansys/fluent/core/solver/settings_builtin_data.py
```diff
@@ -1170,9 +1170,36 @@
     "ReadCase": ("Command", "file.read_case"),
     "ReadData": ("Command", "file.read_data"),
     "ReadCaseData": ("Command", "file.read_case_data"),
-    "WriteCase": ("Command", "file.write_case"),
-    "WriteData": ("Command", "file.write_data"),
-    "WriteCaseData": ("Command", "file.write_case_data"),
+    "WriteCase": (
+        "Singleton",
+        {
+            FluentVersion.v261: "file.write_case",
```
seanpearsonuk, line 1176: @mkundu1 

How would you feel about introducing a small DSL here, with helpers like `since()`, `until()`, `only_at()`, to declare version-specific paths more clearly?

I'm thinking this might be worth building as reusable infrastructure around `FluentVersion`. Version ranges could be helpful in other areas of the codebase too.

What do you think?

mkundu1: Good idea. I've added an implementation [here](https://github.com/ansys/pyfluent/pull/4172), will discuss further.

seanpearsonuk: Independent of that side discussion, we can still aprrove this for merge. @hpohekar @prmukherj

prmukherj (approved):

src/ansys/fluent/core/solver/settings_builtin_data.py
hpohekar: LGTM! Thank you.

mkundu1: Thanks, I'll update the current file to use those new functions in the other PR.

Issue #4128, opened 2025-06-10
Require ability to filter out settings API beta and below


seanpearsonuk:
Settings API items can have an optional exposure attribute. Either just the API node or the underlying feature can be beta but we can represent these the same way. 

If `exposure <= beta` the API node is unexposed by default. It can be made unexposed via the `is-active` attribute provided by Fluent so that the client does not have to do any extra work.

We will use alpha/beta labels in our documentation so that users understand why settings are inactive.

When the user sends a `file.beta_settings.enable()` call to Fluent it needs to enable beta API items, not just Fluent beta features.

PR #4166, opened 2025-06-13
fix: Fix performance issues while settings boundary condition

mkundu1:
There were 111 gRPC calls for each `set_state` at a boundary condition object. Those huge number of calls are due to the following 2 issues:
1. `_check_stable()` makes a number of `active?` and `webui-release-active?` attribute queries.
2. There is an unintentional call to `boundary_conditions.__len__` (->`_ChildNamedObjectAccessorMixin.__len__`) which further calls `active?` and `get_object_names()` for each boundary types.

I have cleaned up the code related to `_check_stable()`/`webui-release-active?` as we are not planning to follow that approach for PyFluent (current plan - https://github.com/ansys/pyfluent/issues/4128). Issue 2 is also fixed.

With this PR, there are now 4 gRPC calls for each `set_state` which is the expected number. The runtime for the defect case has been reduced from 12 sec. to 0.7 sec.

### DIFF
--- a/tests/test_flobject.py
+++ b/tests/test_flobject.py
@@ -1251,6 +1251,7 @@ def test_default_argument_names_for_commands(static_mixer_settings_session):
     assert solver.results.graphics.contour.list.argument_names == []
 
 
+@pytest.mark.fluent_version(">=25.1")
 def test_bc_set_state_performance(static_mixer_settings_session, monkeypatch):
     solver = static_mixer_settings_session
 


mkundu1:
TODO: Add test

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -195,7 +194,7 @@ def to_python_name(fluent_name: str) -> str:
 def _get_python_path_comps(obj):
     """Get python path components for traversing class hierarchy."""
     comps = []
-    while obj:
```
mkundu1, line 198: This makes a call to `obj.__len__` due to booleanness check

hpohekar (approved):

PR #4171, opened 2025-06-16
fix: container timeout issue and new environment variables

raph-luc:
Changes:
- Fixed issue related to the container launcher not recognizing timeout settings coming through `launch_fluent`. Now the container launcher respects the `start_timeout` argument, and the `container_dict["timeout"]` argument is marked as deprecated
- Now when a container launch times out, it will attempt to stop the container. I noticed during the tests that timeout or failure during launch were causing zombie containers to be left behind. Now that should not happen anymore
- `test_unsuccessful_fluent_connection` now works for containers as well and tests that their timeout config is working properly
- New `test_container_timeout_deprecation` and `test_container_timeout_deprecation_override` for testing the deprecation behavior
- Support for new `PYFLUENT_FLUENT_LAUNCH_TIMEOUT` and `PYFLUENT_FLUENT_AUTOMATIC_TRANSCRIPT` env vars - were useful previously for testing/development
- Few other minor env var fixes to follow PyFluent's convention that boolean env vars are truthy when set to `1`, false otherwise


src/ansys/fluent/core/launcher/fluent_container.py
```diff
@@ -180,7 +185,9 @@ def configure_container_dict(
         Path inside the container where ``mount_source`` will be mounted. This will be the working directory path
         visible to the Fluent process running inside the container.
     timeout : int, optional
-        Time limit  for the Fluent container to start, in seconds. By default, 30 seconds.
+        Time limit for the Fluent container to start, in seconds.
+        .. deprecated:: v0.33.dev0
+                Use the ``start_timeout`` argument of ``launch_fluent`` instead.
```
hpohekar: We don't use this format to document deprecations.

Please remove it from here and add it in `doc\deprecated_pyfluent_apis.py` with the existing format.

https://github.com/ansys/pyfluent/blob/main/doc/deprecated_pyfluent_apis.py#L92

src/ansys/fluent/core/utils/__init__.py
```diff
@@ -55,3 +56,12 @@ def get_user_data_dir():
...
         return Path.home() / ".local" / "share" / "Ansys" / "ansys_fluent_core"
+
+
+def env_var_to_bool(var_name: str, default: str = "false") -> bool:
+    """Converts environment variable to boolean. By default, returns 'False' if the variable is not set."""
+    value = os.getenv(var_name, default).strip().lower()
+    if value in ("0", "off", "false", "no"):
+        return False
+    else:
+        return True
```
hpohekar: I think to avoid the check for such a number of possibilities, the existing code was specifically verifying the value "1". It is mentioned in the docs that `PYFLUENT_LAUNCH_CONTAINER`, `PYFLUENT_USE_DOCKER_COMPOSE` and `PYFLUENT_USE_PODMAN_COMPOSE` should be set to "1" explicitly. We are also verifying all tests using the value "1".

prmukherj: yes, I agree, I think it's better to stick to 0 and 1 for env vars in PyFluent as a convention.

seanpearsonuk: Or specifically, the convention is that a value of 1 means true such that anything else means false; 0 has no special meaning.

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -746,13 +746,16 @@ def exit(
...
+                    f"Found PYFLUENT_TIMEOUT_FORCE_EXIT env var: '{env_timeout}'"
+                )
                 try:
                     timeout = float(env_timeout)
                     logger.debug(f"Setting TIMEOUT_FORCE_EXIT to {timeout}")
                 except ValueError:
                     logger.debug(
-                        "Off or unrecognized PYFLUENT_TIMEOUT_FORCE_EXIT value, not enabling timeout force exit"
+                        "Off or unrecognized PYFLUENT_TIMEOUT_FORCE_EXIT value (floats and integers are also supported), "
+                        "disabling timeout forced exit."
```
seanpearsonuk: The new message tries to say a lot in one sentence.
```
logger.debug(
    "Invalid or unset PYFLUENT_TIMEOUT_FORCE_EXIT. Must be a float or int. "
    "Timeout forced exit is disabled."
)
```

seanpearsonuk: ...should it be elevated to `warning` level?

src/ansys/fluent/core/utils/__init__.py
raph-luc: I wasn't aware of that PyFluent convention, is that documented somewhere? 

`PYFLUENT_WATCHDOG_DEBUG` was previously accepting `1`, `ON` or `on`

`PYFLUENT_LOGGING` is not a boolean, so not the best example, but accepts many different levels including `0`, `OFF` or `off` to disable it

`PYFLUENT_WATCHDOG_EXCEPTION_ON_ERROR` also accepts any value, not only `1`

raph-luc: I am not aware of a general convention for boolean env vars. Other applications often accept values such as "true" and "false" instead of "1" and "0"

I personally think that users would appreciate the flexibility of being able to specify more "truthy" values to activate PyFluent features through env vars, I did not think there were any downsides to doing this

Regardless, it's fine if you prefer that PyFluent only accept "1" exclusively, with anything else being interpreted as "falsy", I will adjust my changes for that

src/ansys/fluent/core/launcher/fluent_container.py
raph-luc: Thanks for the heads-up, I wasn't aware of that, will do that

src/ansys/fluent/core/fluent_connection.py
raph-luc: I agree, thanks for both suggestions, will do that

src/ansys/fluent/core/__init__.py
```diff
@@ -155,7 +155,7 @@ def version_info() -> str:
 FLUENT_SHOW_MESH_AFTER_CASE_READ = False
 
 # Whether to write the automatic transcript in Fluent
-FLUENT_AUTOMATIC_TRANSCRIPT = False
+FLUENT_AUTOMATIC_TRANSCRIPT = env_var_to_bool("PYFLUENT_FLUENT_AUTOMATIC_TRANSCRIPT")
```
hpohekar, line 158: `FLUENT_AUTOMATIC_TRANSCRIPT` is not an environment variable. It's a PyFluent attribute.

src/ansys/fluent/core/utils/__init__.py
hpohekar: I agree with @seanpearsonuk. "1" is true and anything else is false.

src/ansys/fluent/core/fluent_connection.py
raph-luc: On second thought, going to keep it at the `debug` level: the way to make sure this env var is disabled, is to set it to non-number values such as "off" or "false". I don't think we want to raise a warning in that situation. 

A timeout of zero currently means "return immediately", so zero can't be used to disable this one, right now. This is something that we could change though.

src/ansys/fluent/core/launcher/fluent_container.py
raph-luc: Addressed by https://github.com/ansys/pyfluent/pull/4171/commits/441dd6adf98a733f1911dad1996279514d62fb09, please have a look to see if it is appropriate now

src/ansys/fluent/core/utils/__init__.py
raph-luc: Addressed by https://github.com/ansys/pyfluent/pull/4171/commits/dab26202646aca4c29d234365fcc75b4744bd4c7, also fixed a few of the other env vars that were not following PyFluent's convention, mentioned in my earlier reply

src/ansys/fluent/core/fluent_connection.py
raph-luc: Addressed by [changes to fluent_connection.py in commit: `dab26202`](https://github.com/ansys/pyfluent/pull/4171/commits/dab26202646aca4c29d234365fcc75b4744bd4c7#diff-00c898f4e09231f8a736f0d6a5d8903afe4b6e4ae8a8a12b30b9234f8d70bddb)

src/ansys/fluent/core/launcher/fluent_container.py
hpohekar: Yes. Looks good. Thank you.

src/ansys/fluent/core/utils/__init__.py
hpohekar: LGTM! Thank you.

hpohekar (approved):

prmukherj (approved):

PR #4172, opened 2025-06-16
feat: Add helper methods to specify a set of Fluent versions

mkundu1:
Please see the unittest for usage.

settings_builtin_data.py is updated to use the new methods.


tests/test_fluent_version.py
```diff
@@ -89,3 +97,21 @@ def test_eq():
 def test_str_output():
     assert str(FluentVersion.v232) == "Fluent version 2023 R2"
     assert str(FluentVersion.v251) == "Fluent version 2025 R1"
+
+
+def test_fluent_version_set():
```
seanpearsonuk, line 103: @mkundu1 I was thinking of a smaller set of functions that could be applied as compound functions but I prefer what you have done: when one sees this set of available functions, it is immediately clear how they should be used.

devel/update_settings_builtin_data.py
mkundu1, line 1: the updater script is no longer needed

prmukherj (approved):

hpohekar (approved):

hpohekar:
@mkundu1 Merging to create a release.

PR #4174, opened 2025-06-17
chore: static class attributes

prmukherj:
Refactor static class attributes access


doc/settings_rstgen.py
```diff
@@ -131,6 +140,7 @@ def _populate_rst_from_settings(rst_dir, cls, version):
     with open(rstpath, "w") as r:
         # Populate initial rst
         r.write(":orphan:\n\n")
+        # ``root`` used to create a hyperlink for settings API
```
hpohekar, line 144: I have pushed a change in this PR to add this comment as discussed with @prmukherj.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -69,6 +69,7 @@
 import warnings
 import weakref
 
+from ansys.fluent.core.codegen.settingsgen import _static_class_attributes
```
```suggestion
```

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1132,6 +1131,9 @@ def get_completer_info(self, prefix="", excluded=None) -> List[List[str]]:
         return ret
 
     def __getattribute__(self, name):
+        # Static class attributes should not do server query
```
```suggestion
        # Static class attributes should not do server query
        from ansys.fluent.core.codegen.settingsgen import _static_class_attributes
```

doc/settings_rstgen.py
prmukherj: Thank you @hpohekar

doc/settings_rstgen.py
```diff
@@ -119,6 +119,15 @@ def _populate_parents_list(cls):
         _populate_parents_list(getattr(cls, "child_object_type"))
 
 
+def _write_common(initial_param, r, cls, attr):
```
seanpearsonuk, line 122: Should I be concerned about the absence of comments here?

doc/settings_rstgen.py
seanpearsonuk: Comments like this save time later.

@prmukherj @hpohekar

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1132,6 +1139,9 @@ def get_completer_info(self, prefix="", excluded=None) -> List[List[str]]:
...
```
```suggestion
        # Avoiding server queries for static attributes
```

doc/settings_rstgen.py
hpohekar: Yes, I agree.

seanpearsonuk (approved):

doc/settings_rstgen.py
seanpearsonuk, line 122:
```suggestion
def _write_common(initial_param, r, cls, attr):
    # TODO Add clarifying comments here
```

hpohekar:
@prmukherj Have you verified that settings API docs are rendering correctly by running Nightly Dev Doc Build workflow using this branch ?

doc/settings_rstgen.py
```diff
@@ -139,8 +150,12 @@ def _populate_rst_from_settings(rst_dir, cls, version):
         r.write(f'{"="*(len(cls_orig_name))}\n\n')
         deprecated = getattr(cls, "_deprecated_version", None)
         if deprecated:
-            release_version = "20" + cls._deprecated_version.replace(".", "R")
-            r.write(f".. deprecated:: Ansys {release_version}\n\n")
+            if float(cls._deprecated_version) < 22.2:
```
prmukherj: Here if the deprecated version is less that 22.2, we provide the least version supported by pyfluent, i.e. 2022R2, as the versions before that are not supported by our FluentVersion enum,

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -236,8 +244,10 @@ def _is_deprecated(obj) -> bool | None:
         deprecated_version = (
             deprecated_version.get("deprecated-version") if deprecated_version else None
         )
+    # The comparison is made in terms of float because some deprecated versions,
+    # for instance 21.1 might not be part of FluentVersion
```
seanpearsonuk: `FluentVersion` is intended to be used for all such comparisons rather than relying on brittle float comparisons. We should extend the functionality as necessary.

prmukherj: ok @seanpearsonuk, sounds good. Will implement it in FluentVersion.

prmukherj: Actually this case will not be encountered in Fluent > 25R2 anyways as we are dependent on static query, and static query is restricted to FluentVersion.222. So do we need to change FluentVersion for something temporary?

@mkundu1, @seanpearsonuk

doc/settings_rstgen.py
seanpearsonuk: `float(cls._deprecated_version) < 22.2` has no reliable meaning. The attribute value is invalid. Ignore it and create a Fluent defect.

src/ansys/fluent/core/solver/flobject.py
seanpearsonuk: I don't understand the above comment

doc/settings_rstgen.py
prmukherj: Fixed now. We are ignoring the entry like discussed.

src/ansys/fluent/core/solver/flobject.py
prmukherj: This has been completely removed. We are not populating it as discussed.

mkundu1:
Merging this as CI is passed except the known cheatsheet generation issue.

PR #4176, opened 2025-06-17
ci: Fix codacy warnings and quarto-cli [skip tests]

hpohekar:


.codacy.yml
```diff
@@ -0,0 +1,4 @@
+---
+languages:
+  python:
+    enabled: false
```
raph-luc, line 4: So the fix was to disable Codacy for Python? Is there even any point to have it running in this repo, then?

hpohekar: We have already integrated vulnerabilities check therefore we can disable this.

hpohekar: Codacy is not running for PyAEDT https://github.com/ansys/pyaedt/pull/6261.

They also have a separate vulnerabilities job as an alternative to this.

raph-luc: Ok I see, thanks. Can we simply remove Codacy from this repo, then? Are there any benefits to keeping it if it is not doing anything for Python?

hpohekar: I'm searching for a way to remove it completely but could not find it yet.

raph-luc: Got it, thanks. Then, what do you think about changing the PR title/description, this is not "fixing Codacy warnings", it is disabling Codacy instead, right? Or "removing Codacy" instead if we can figure out how to do that

hpohekar: Sure.

tests/journals/test_DatamodelService_Journal.py
```diff
@@ -386,3 +386,7 @@ def test_datamodel_api_static_info(solver):
 test_datamodel_api_static_info(solver)  # noqa: F821
 
 print("\n Testing finished. All tests passed.\n")
+
+
+def test_codacy():
+    assert True
```
raph-luc: Reminder to remove this when you finish testing. This test is not useful, right?

.github/workflows/ci.yml
```diff
@@ -158,7 +158,7 @@ jobs:
             Python-${{ runner.os }}-${{ env.PYTHON_VERSION }}
 
       - name: Install Quarto
-        uses: quarto-dev/quarto-actions/setup@v2
+        uses: quarto-dev/quarto-actions/setup@9e48da27e184aa238fcb49f5db75469626d43adb # v2.1.9
```
raph-luc, line 161: Just curious, what are the guidelines, do you have a link to them? Is it actually a good idea to do this? So we will be updating these actions ourselves, manually?

I am not sure if pinning it like this isn't a greater security risk than just using the latest v2 version, for example

raph-luc (approved):

tests/journals/test_DatamodelService_Journal.py
hpohekar: Yes, we can remove this now.

.github/workflows/ci.yml
hpohekar: @raph-luc Yes, https://github.com/ansys/pyfluent/security/code-scanning/150 Just click on `Show more` to see guidelines. 

If you see ansys/actions repo we are already using commit SHAs for all ansys actions therefore we don't get warnings for them. We get warnings for third party actions only, which don't use commit SHAs, therefore we are updating third party actions.

PR #4177, opened 2025-06-17
fix: Fix an issue in returning parameter units

mkundu1:
We retrieve the Fluent unit strings via scheme calls from PyFluent at the end of the execution of `parameters.output_parameters.list()` command in PyFluent so we can populate Fluent unit strings in that command's return value. This PR prevents a Fluent-side error which was coming from those scheme calls.

`parameters.output_parameters.list()` will be fixed in 26.1 to contain the Fluent unit strings in the return value. The above PyFluent workaround will still be needed in Fluent versions <= 25.2.

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -1780,10 +1780,16 @@ def _fix_parameter_list_return(val):
                 # Symbols are not stripped in the command return in PyConsole.
                 # Following code will work in both PyConsole and PyFluent.
                 unit = units[0].lstrip("'")
-                unit_labels = _fix_parameter_list_return.scheme_eval(
-                    f"(units/inquire-available-label-strings-for-quantity '{unit})"
-                )
-                unit_label = unit_labels[0] if len(unit_labels) > 0 else ""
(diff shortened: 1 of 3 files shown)


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1780,10 +1780,16 @@ def _fix_parameter_list_return(val):
                 # Symbols are not stripped in the command return in PyConsole.
                 # Following code will work in both PyConsole and PyFluent.
                 unit = units[0].lstrip("'")
-                unit_labels = _fix_parameter_list_return.scheme_eval(
-                    f"(units/inquire-available-label-strings-for-quantity '{unit})"
-                )
-                unit_label = unit_labels[0] if len(unit_labels) > 0 else ""
+                if unit != "*null*":
```
mkundu1, line 1783: This condition safeguards against the reported error in the defect. The try-catch around the scheme-eval has been added an additional measure for safeguarding. I haven't put the try-catch at the outer level as that will still producde the Fluent-side error which will be shown in the transcript.

seanpearsonuk (approved):

PR #4180, opened 2025-06-17
fix: Relax error message comparison in test

mkundu1:


tests/test_error_handling.py
```diff
@@ -52,5 +53,10 @@ def test_custom_python_error_via_grpc(datamodel_api_version_new, new_solver_sess
     # This may need to be updated if the error type changes in the server
     with pytest.raises(RuntimeError, match="prefereces not found!"):
         solver._se_service.get_state("prefereces", "General")
-    with pytest.raises(ValueError, match="Datamodel rules for prefereces not found!"):
+    translator = str.maketrans("", "", string.punctuation)
+    with pytest.raises(ValueError) as ex:
         solver._se_service.get_specs("prefereces", "General")
+    assert (
+        ex.value.args[0].translate(translator)
+        == "Datamodel rules for prefereces not found"
```
mkundu1, line 61: The failure is due to the slight formatting which has been added to the error message in Fluent (quotes around the rules name).

hpohekar (approved):

tests/test_error_handling.py
hpohekar: Right, this was tricky. Thank you.

prmukherj: If we just check for the ValueError, that should be enough right? @mkundu1, @seanpearsonuk

prmukherj (approved):

tests/test_error_handling.py
hpohekar: Started nightly dev test - https://github.com/ansys/pyfluent/actions/runs/15726554216

hpohekar: Nightly Dev Test passed successfully.

hpohekar:
@mkundu1 Merging to create a release.

tests/test_error_handling.py
seanpearsonuk: > If we just check for the ValueError, that should be enough right? @mkundu1, @seanpearsonuk

Agreed. For maintainability, we generally avoid asserting error message strings. If we need stronger guarantees, we should raise and check for a specific custom exception rather than rely on the message text.

PR #4190, opened 2025-06-18
feat: Set string settings with allowed values via string constants

mkundu1:
The enumerated values are exposed as constant attributes of the instance. Please see the unittest for usage.

There is already an attribute `allowed-values-with-display-text` in settings API to fetch all possible allowed values which has been added to the static-info (Fluent PR 599447).

Below is the fragment of generated class:
```
class model_2(String, AllowedValuesMixin):
    """
    Viscous Model.
    """
    _version = '261'
    fluent_name = 'model'
    _python_name = 'model'
    LAMINAR = _FlConstant('laminar')
    MIXING_LENGTH = _FlConstant('mixing-length')
    KE1E = _FlConstant('ke1e')
    INVISCID = _FlConstant('inviscid')
...
```

TODO: doc update

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -187,7 +187,7 @@ _ttable = str.maketrans(string.punctuation, "_" * len(string.punctuation), "?'")
 def to_python_name(fluent_name: str) -> str:
     """Convert a scheme string to a Python variable name.
 
-    This function replaces symbols with _. Any ``?`` symbols are
+    This function replaces symbols with _. ``'`` and ``?`` symbols are
     ignored.
     """
     if not fluent_name:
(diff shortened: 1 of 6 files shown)


prmukherj:
@mkundu1, existing tests suggest that something has gone wrong.

mkundu1:
Validation against all possible allowed values during set_state is now possible in the client side, will add it.

hpohekar:
> Validation against all possible allowed values during set_state is now possible in the client side, will add it.

That's great! Thank you.

seanpearsonuk (approved):

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -2025,6 +2039,19 @@ def allowed_values(self):
...
+    """A descriptor class to hold a constant value."""
+
+    def __init__(self, value):
+        self._value = value
+
+    def __get__(self, instance, owner):
+        return self._value
+
+    def __set__(self, instance, value):
+        raise AttributeError("Cannot set a constant value.")
```
seanpearsonuk: @mkundu1 Should we add an `is_active()` method here so that the user can check if the constant is currently in the dynamic allowed values list?

prmukherj (approved):

src/ansys/fluent/core/solver/flobject.py
mkundu1: yes, added, thanks

PR #4214, opened 2025-06-25
refactor: get_completer_info implementation based on Python's inspect module

mkundu1:
Simpler implementation of `get_completer_info` based on `inspect.getmembers`. This removes some duplicated logic in PyFluent code and will avoid some hardcoded logic in PyConsole code.

PR #4217, opened 2025-06-26
fix: interior zones are inactive in the latest Fluent image

mkundu1:
Updating tests based on Fluent PR 599405 changes.

Nightly test CI passed - https://github.com/ansys/pyfluent/actions/runs/15906604497

### DIFF
--- a/tests/test_builtin_settings.py
+++ b/tests/test_builtin_settings.py
@@ -175,10 +175,6 @@ def test_builtin_settings(mixing_elbow_case_data_session):
         == solver.setup.boundary_conditions.interior
     )
     assert (
-        InteriorBoundary(settings_source=solver, name="interior--elbow-fluid")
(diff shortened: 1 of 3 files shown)


tests/test_builtin_settings.py
```diff
@@ -547,9 +543,9 @@ def test_builtin_settings(mixing_elbow_case_data_session):
     else:
         with pytest.raises(RuntimeError):
             CustomVectors(settings_source=solver)
-    tmp_save_path = Path(tempfile.mkdtemp(dir=pyfluent.EXAMPLES_PATH))
```
mkundu1, line 550: This line is failing locally (probably due to some limiation of pytest's `tmp_path` fixture). Ideally, relative paths should work accross host and container working directories even though they have different absolute paths.

mkundu1:
Covered in #4214

PR #4244, opened 2025-07-05
feat: Get raw value for faces connectivity data.

prmukherj:
This feature updates existing method to add a flag "flatten_connectivity" which changes its default return types to retrieve raw face connectivity data from surface field queries. It enables advanced users to directly access the vertex indices that define each face, which can be useful for custom post-processing or geometry interpretation.

Key Changes
- Update get_faces_connectivity_data to provide raw-outputs.

Example usage:

```
python 

faces_connectivity_request = SurfaceFieldDataRequest(
    surfaces=[VelocityInlet(settings_source=solver_session, name="inlet")],
    data_types=[SurfaceDataType.FacesConnectivity],
    flatten_connectivity=True,
)
faces_connectivity_data = field_data.get_field_data(faces_connectivity_request)

# Accessing face connectivity
faces_connectivity_data["inlet"].connectivity
# Output: array([  4,   3,   2, ..., 379, 382, 388], shape=(1518,), dtype=int32)

# Accessing structured data:
transform_faces_connectivity_data(faces_connectivity_data["inlet"].connectivity)
```

Notes
- The raw format directly returns vertex indices for each face without additional formatting or interpretation.
- This enhancement aligns with use cases requiring low-level mesh topology access.

Please see tests and documentation for detailed usage.

### DIFF
--- a/tests/test_file_session.py
+++ b/tests/test_file_session.py
@@ -708,8 +708,12 @@ def test_error_handling_multi_phase_deprecated():
 
 
 def test_faces_connectivity_behaviour():
-    case_file_name = examples.download_file("elbow1.cas.h5", "pyfluent/file_session")
-    data_file_name = examples.download_file("elbow1.dat.h5", "pyfluent/file_session")
+    case_file_name = examples.download_file(
+        "elbow1.cas.h5", "pyfluent/file_session", return_without_path=False
+    )
+    data_file_name = examples.download_file(
+        "elbow1.dat.h5", "pyfluent/file_session", return_without_path=False
+    )
     file_session = FileSession(case_file_name, data_file_name)
     transaction = file_session.fields.field_data.new_transaction()
     vertices_and_faces_connectivity_request = SurfaceFieldDataRequest(


prmukherj:
@abhishekchitwar, this will suffice the requirement you had for getting raw data from faces_connectivity?

doc/source/user_guide/fields/field_data.rst
```diff
@@ -95,6 +95,23 @@ To obtain face connectivity data, specify ``FacesConnectivity`` as the ``data_ty
...
+
+  >>> faces_connectivity_request = SurfaceFieldDataRequest(
+  >>>     surfaces=[VelocityInlet(settings_source=solver_session, name="inlet")],
+  >>>     data_types=[SurfaceDataType.FacesConnectivity]
+  >>>     raw_data=True,
+  >>> )
+  >>> faces_connectivity_data = field_data.get_field_data(faces_connectivity_request)
+
+  # FacesConnectivity provides indices of vertices for each face. For example:
+  # Face 6 is connected to vertices 4, 5, 12, and 11.
```
seanpearsonuk: @prmukherj Is the connection between the comment and the code/output obvious?

prmukherj: Thank you @seanpearsonuk. This was  an error in the doc. Fixed it now.

doc/source/user_guide/fields/field_data.rst
```diff
@@ -95,6 +95,21 @@ To obtain face connectivity data, specify ``FacesConnectivity`` as the ``data_ty
...
+.. code-block:: python
+
+  >>> faces_connectivity_request = SurfaceFieldDataRequest(
+  >>>     surfaces=[VelocityInlet(settings_source=solver_session, name="inlet")],
+  >>>     data_types=[SurfaceDataType.FacesConnectivity]
+  >>>     raw_data=True,
+  >>> )
+  >>> faces_connectivity_data = field_data.get_field_data(faces_connectivity_request)
+
+  >>> faces_connectivity_data["inlet"].connectivity
```
seanpearsonuk, line 112: @prmukherj What means do we provide to help users interpret individual faces in the flat raw data array? For example, is there a way to query face-level metadata like face types or offsets?

prmukherj: Updated as per discussions.
Thank you.

doc/source/user_guide/fields/field_data.rst
```diff
@@ -80,7 +80,14 @@ in the ``data_types`` list.
   # Example: The centroid of the 16th face has coordinates [-0.3463, 0.0, -0.0328].
   array([-0.34634298,  0.        , -0.03276413], dtype=float32)
 
-To obtain face connectivity data, specify ``FacesConnectivity`` as the ``data_types`` parameter.
+To obtain face connectivity data, specify ``FacesConnectivity`` as the ``data_types`` parameter. The
+returned data provides face connectivity in a flat numpy array of vertex indices.
```
seanpearsonuk: @prmukherj 

NumPy.

doc/source/user_guide/fields/field_data.rst
seanpearsonuk: @prmukherj I still feel that it lacks clarity.

doc/source/user_guide/fields/field_data.rst
prmukherj: Was giving vale error. Will add it to accept list then. Actually I referred other docs in our repo and saw lower case only.
Will fix it.

doc/source/user_guide/fields/field_data.rst
prmukherj: In-line comments in the code will help improve clarity?

seanpearsonuk: A full explanation of the format.

prmukherj: Done. @seanpearsonuk, could you please have a look if clarity is better?
Thank you,

doc/source/user_guide/fields/field_data.rst
```diff
@@ -80,7 +80,25 @@ in the ``data_types`` list.
...
+when constructing a ``SurfaceFieldDataRequest``. The returned data provides a flat, NumPy array
+of vertex indices that describe how each face is connected to the mesh's vertices.
+
+The data is stored in a compact "flattened" format. Each face is represented by a sequence of
+vertex indices, preceded by an integer specifying how many vertices the face contains. This means
+the array is structured as:
+
+::
+
+   [N₀, V₀₁, V₀₂, ..., V₀ₙ₀, N₁, V₁₁, V₁₂, ..., V₁ₙ₁, ...]
```
seanpearsonuk, line 93: @prmukherj Great, now the user can see what it is.

abhishekchitwar:
> > > > @abhishekchitwar, this will suffice the requirement you had for getting raw data from faces_connectivity?
> > > 
> > > 
> > > Thanks, @prmukherj! This avoids the need to explicitly flatten the face connectivity data when converting it to VTK or any other format.
> > 
> > 
> > @abhishekchitwar, we have decided to update the behavior so as to return the flat structure by default. So no changes required from your end to flatten it out or to add the flag. Thank you.
> 
> @prmukherj We did propose deprecating the previous format, which would usually imply initially providing a separate, temporary means for accessing the new format (possibly a different `data_type`). Should we not be concerned if instead we're simply making a breaking change without any prior notification?

I believe it would be good to introduce a `flatten = True` flag, while keeping the default behavior set too `False` to ensure that existing workflows remain unaffected.

prmukherj:
> > > > > @abhishekchitwar, this will suffice the requirement you had for getting raw data from faces_connectivity?
> > > > 
> > > > 
> > > > Thanks, @prmukherj! This avoids the need to explicitly flatten the face connectivity data when converting it to VTK or any other format.
> > > 
> > > 
> > > @abhishekchitwar, we have decided to update the behavior so as to return the flat structure by default. So no changes required from your end to flatten it out or to add the flag. Thank you.
> > 
> > 
> > @prmukherj We did propose deprecating the previous format, which would usually imply initially providing a separate, temporary means for accessing the new format (possibly a different `data_type`). Should we not be concerned if instead we're simply making a breaking change without any prior notification?
> 
> I believe it would be good to introduce a `flatten = True` flag, while keeping the default behavior set too `False` to ensure that existing workflows remain unaffected.

yes, doing that only. Just that the flatten behavior will be the default one later, going forward. So we need a deprecation message.

mkundu1 (approved):

hpohekar (approved):

PR #4256, opened 2025-07-09
feat: raise the attribute error with a correct message

hpohekar:
Earlier

```
solver.setup.named_expressions.child_names
  File "<string>", line 1, in <module>
  File "D:\Repos\pyfluent\src\ansys\fluent\core\solver\flobject.py", line 1458, in __getattr__
    return getattr(super(), name)
AttributeError: 'super' object has no attribute 'child_names'
```

Now

```
solver.setup.named_expressions.child_names
  File "D:\Repos\pyfluent\src\ansys\fluent\core\solver\flobject.py", line 1459, in __getattr__
    return getattr(super(), name)
AttributeError: 'super' object has no attribute 'child_names'

The above exception was the direct cause of the following exception:

  File "d:\Repos\pyfluent\debug.py", line 323, in <module>
    solver.setup.named_expressions.child_names
  File "D:\Repos\pyfluent\src\ansys\fluent\core\solver\flobject.py", line 1461, in __getattr__
    raise AttributeError(self.__class__.__name__ + " has no attribute " + name) from ex
AttributeError: named_expressions has no attribute child_names

```

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -1474,7 +1474,7 @@ class NamedObject(SettingsBase[DictStateType], Generic[ChildTypeT]):
                 return getattr(super(), name)
             except AttributeError as ex:
                 raise AttributeError(
-                    self.__class__.__name__ + " has no attribute " + name
+                    f"'{self.__class__.__name__}' has no attribute '{name}'"
                 ) from ex
 
     def __add__(self, other):


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1470,7 +1470,12 @@ def __getattr__(self, name: str):
                 )
             return alias_obj
         else:
-            return getattr(super(), name)
+            try:
+                return getattr(super(), name)
+            except AttributeError as ex:
+                raise AttributeError(
+                    self.__class__.__name__ + " has no attribute " + name
+                ) from ex
```
prmukherj, line 1478: Do we need the traceback or shall we raise it from None?

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1470,7 +1470,12 @@ def __getattr__(self, name: str):
                 )
             return alias_obj
         else:
-            return getattr(super(), name)
+            try:
+                return getattr(super(), name)
+            except AttributeError as ex:
+                raise AttributeError(
+                    self.__class__.__name__ + " has no attribute " + name
```
```suggestion
                    f"'{self.__class__.__name__}' has no attribute '{name}'"
```

src/ansys/fluent/core/solver/flobject.py
hpohekar: Yes, we need it, it helps to identify the exact cause.  This change has been done to give more informative message only.

src/ansys/fluent/core/solver/flobject.py

src/ansys/fluent/core/solver/flobject.py
seanpearsonuk: > Yes, we need it, it helps to identify the exact cause. This change has been done to give more informative message only.
Yes, @hpohekar makes a good point. Transparency is generally important in error messages.

mkundu1: We could rethrow the original `ex` after updating its message, no need to chain exceptions.

prmukherj: yes, that's good as well.

hpohekar: @mkundu1 Do you mean like this ? I think this a hacky approach.

```
try:
    getattr(super(), name)
except AttributeError as ex:
    ex.args = (f"'{self.__class__.__name__}' has no attribute '{name}'",)
    raise
```

Output - 

```
solver.setup.named_expressions.child_names
  File "d:\Repos\pyfluent\debug.py", line 323, in <module>
    solver.setup.named_expressions.child_names
  File "D:\Repos\pyfluent\src\ansys\fluent\core\solver\flobject.py", line 1474, in __getattr__
    return getattr(super(), name)
AttributeError: 'named_expressions' has no attribute 'child_names'
```

@seanpearsonuk What do you think ?

mkundu1: Yeah, right, changing the argument is bit hacky. My concern was that 2 exceptions of the same type would be shown in the traceback. But transparency and immutability are more important requirements here, so please keep the existing code.

mkundu1 (approved):

src/ansys/fluent/core/solver/flobject.py
hpohekar: @mkundu1 Thank you.

Issue #4233, opened 2025-07-01
Bug with Fluent R251


XingchunCat:
### 🔍 Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### 🐞 Description of the bug

When PyFluent tries launching Fluent R251 it received an error says "pyfluent.launcher ERROR: Exception caught - RuntimeError: Deadline Exceeded". 

It seems even execute solver.exit() some resources are not released properly, this could also happen if open a Fluent R251 session but then closed. 

While you cannot launch Fluent R251 through PyFluent, you can still launch R242 just by changing the version number,

### 📝 Steps to reproduce

It seems random to me; I sometimes won't get this error. One way I can think of to reproduce is launch R251 once and force quit. then launch it again.

### 💻 Which operating system are you using?

Windows

### 📀 Which ANSYS version are you using?

Fluent R251

### 🐍 Which Python version are you using?

3.9

### 📦 Installed packages

```shell
ansys-api-fluent==0.3.35
ansys-api-platform-instancemanagement==1.1.3
ansys-api-tools-filetransfer==0.1.1
ansys-fluent-core==0.32.0
ansys-platform-instancemanagement==1.1.2
ansys-tools-filetransfer==0.1.1
ansys-units==0.7.dev0
anyio @ file:///home/conda/feedstock_root/build_artifacts/bld/rattler-build_anyio_1742243108/work
argon2-cffi @ file:///home/conda/feedstock_root/build_artifacts/argon2-cffi_1733311059102/work
argon2-cffi-bindings @ file:///D:/bld/argon2-cffi-bindings_1725356700772/work
arrow @ file:///home/conda/feedstock_root/build_artifacts/arrow_1733584251875/work
asttokens @ file:///home/conda/feedstock_root/build_artifacts/asttokens_1733250440834/work
async-lru @ file:///home/conda/feedstock_root/build_artifacts/bld/rattler-build_async-lru_1742153708/work
attrs @ file:///home/conda/feedstock_root/build_artifacts/attrs_1741918516150/work
babel @ file:///home/conda/feedstock_root/build_artifacts/babel_1738490167835/work
beautifulsoup4 @ file:///home/conda/feedstock_root/build_artifacts/beautifulsoup4_1744783198182/work
bleach @ file:///home/conda/feedstock_root/build_artifacts/bld/rattler-build_bleach_1737382993/work
Brotli @ file:///D:/bld/brotli-split_1725267609074/work
[...]

PR #4274, opened 2025-07-11
fix: Check localhost for grpc connection before other ips

mkundu1:
This is a potential fix for #4233. We have seen PyFluent can connect to Fluent using `localhost` as the ip in abhishekchitwar's system.

### DIFF
--- a/doc/changelog.d/4274.fixed.md
+++ b/doc/changelog.d/4274.fixed.md
@@ -1,0 +1,1 @@
+Check localhost for grpc connection before other ips


src/ansys/fluent/core/utils/networking.py
```diff
@@ -81,14 +81,24 @@ def find_remoting_ip() -> str:
...
+            family=socket.AF_INET,
+            type=socket.SOCK_STREAM,
+            flags=socket.AI_PASSIVE,
+        )
+    ]
+    # Check if we can establish a gRPC connection using localhost first
+    # before trying other IPs. It has been observed that in some systems,
+    # although we can establish a test gRPC connection using one of the
+    # resolved IP addresses in addrinfo, PyFluent fails to connect to Fluent
+    # using that IP address. Using localhost usually helps in such cases.
```
hpohekar, line 98: I see. Thanks for the information.

hpohekar (approved):

src/ansys/fluent/core/utils/networking.py
hpohekar: I think it was a platform specific issue (Windows only). Both the users reported for Windows only.

seanpearsonuk (approved):

src/ansys/fluent/core/utils/networking.py
mkundu1: Yes, it could be specific to network configuration of the system. From the observation, if we pass `localhost` directly to the grpc library and let the grpc library resolves it, that works. But passing the numeric ips from the ipconfig output doesn't work. The workaround is based on this particular observation.

hpohekar: I see. Thanks for the explanation.

PR #4284, opened 2025-07-15
feat: Replace compose env vars with parameters

hpohekar:
This pull request replaces the use of environment variables for selecting Docker Compose or Podman Compose with explicit parameters in the codebase and updates associated documentation accordingly. The changes improve clarity and maintainability by introducing a `ComposeConfig` class to manage these configurations.

#### Environment Variable Replacement
* Removed the `PYFLUENT_USE_DOCKER_COMPOSE` and `PYFLUENT_USE_PODMAN_COMPOSE` environment variables, replacing them with parameters `use_docker_compose` and `use_podman_compose` in relevant methods and constructors [[1]](diffhunk://#diff-992f70f7dd23aabc64cacb1b4e6b08af2dbcb824b18e3225e36e78eeb7b7523fL22) [[2]](diffhunk://#diff-03e8f95919668d5ef0bc3cbded8e1f391e91d9fb50998ca46898488f8582a9d1L135-R140) [[3]](diffhunk://#diff-6388218efa10000a687f0828690e9072e2752a00c66a5cccb0fa2e737344a112R111-R112).

#### Codebase Refactoring
* Introduced the `ComposeConfig` class to encapsulate Docker Compose and Podman Compose configurations, replacing direct environment variable checks throughout the codebase [[1]](diffhunk://#diff-00c898f4e09231f8a736f0d6a5d8903afe4b6e4ae8a8a12b30b9234f8d70bddbL45-R45) [[2]](diffhunk://#diff-6388218efa10000a687f0828690e9072e2752a00c66a5cccb0fa2e737344a112R207-R215).
* Updated the `ComposeBasedLauncher` and other related classes to use the new `ComposeConfig` for determining the compose commands and configurations [[1]](diffhunk://#diff-03e8f95919668d5ef0bc3cbded8e1f391e91d9fb50998ca46898488f8582a9d1L48-L50) [[2]](diffhunk://#diff-00c898f4e09231f8a736f0d6a5d8903afe4b6e4ae8a8a12b30b9234f8d70bddbL552-R559).

#### Documentation Updates
* Updated user guides and examples to reflect the removal of environment variables and the introduction of new parameters for specifying compose options [[1]](diffhunk://#diff-bf02dfc62a4ee617b477db473be809d4230216dc4181515d90a6c2e1736e1311L89-R91) [[2]](diffhunk://#diff-d7fcc4b62ce469fc8d0c903395bc6b3e6e9575e11c1d0bacdc1a60843c68fbe3R45) [[3]](diffhunk://#diff-d7fcc4b62ce469fc8d0c903395bc6b3e6e9575e11c1d0bacdc1a60843c68fbe3R436).
* Removed references to deprecated environment variables in the documentation and added instructions for using the new parameters [[1]](diffhunk://#diff-fa34d41524bd759dfa9579a94f61bdf4e013434cb0596f6202d4c000f29f221fL77-L80) [[2]](diffhunk://#diff-d7fcc4b62ce469fc8d0c903395bc6b3e6e9575e11c1d0bacdc1a60843c68fbe3L461).

#### Deprecation Notices
* Added deprecation warnings for the removed environment variables in the `doc/deprecated_pyfluent_apis.py` file, specifying their replacements.

#### Changelog
* Added a changelog entry summarizing the replacement of environment variables with parameters.


hpohekar:
> @hpohekar
> 
> **General comments on the current approach:**
> 
> 1. Accessing the environment at various points to infer program state and make logical decisions in the code is dangerous. The environment can change underneath the logic, which makes it unreliable. Instead, we should read the relevant environment variables once at the highest-level construction point and ignore them after that.
> 2. Environment variables should always play second fiddle to arguments explicitly set by the user. This is a good general principle: argument values are in a narrower scope (more localized) and should therefore take precedence. This is especially important here since these environment variables are now deprecated.
> 
> Let’s make sure the logic reflects this hierarchy and avoids reading the environment more than once.

@seanpearsonuk Yes, I totally agree. Done. Thank you.

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -381,6 +381,8 @@ def __init__(
         slurm_job_id: str | None = None,
         inside_container: bool | None = None,
         container: ContainerT | None = None,
+        use_docker_compose: bool = False,
+        use_podman_compose: bool = False,
```
seanpearsonuk: @hpohekar Would there be any downside to defaulting these two arguments to `None`? It would allow us to properly detect the user's selection and thereby override any environment settings.

seanpearsonuk:
> > @hpohekar
> > **General comments on the current approach:**
> > 
> > 1. Accessing the environment at various points to infer program state and make logical decisions in the code is dangerous. The environment can change underneath the logic, which makes it unreliable. Instead, we should read the relevant environment variables once at the highest-level construction point and ignore them after that.
> > 2. Environment variables should always play second fiddle to arguments explicitly set by the user. This is a good general principle: argument values are in a narrower scope (more localized) and should therefore take precedence. This is especially important here since these environment variables are now deprecated.
> > 
> > Let’s make sure the logic reflects this hierarchy and avoids reading the environment more than once.
> 
> @seanpearsonuk Yes, I totally agree. Done. Thank you.

@hpohekar It's still the same in the version I'm seeing. I might be missing a commit.

src/ansys/fluent/core/fluent_connection.py
hpohekar: @seanpearsonuk Yes, we will need to explicitly handle the None case instead of just checking if `use_docker_compose`.

hpohekar:
> > > @hpohekar
> > > **General comments on the current approach:**
> > > 
> > > 1. Accessing the environment at various points to infer program state and make logical decisions in the code is dangerous. The environment can change underneath the logic, which makes it unreliable. Instead, we should read the relevant environment variables once at the highest-level construction point and ignore them after that.
> > > 2. Environment variables should always play second fiddle to arguments explicitly set by the user. This is a good general principle: argument values are in a narrower scope (more localized) and should therefore take precedence. This is especially important here since these environment variables are now deprecated.
> > > 
> > > Let’s make sure the logic reflects this hierarchy and avoids reading the environment more than once.
> > 
> > 
> > @seanpearsonuk Yes, I totally agree. Done. Thank you.
> 
> @hpohekar It's still the same in the version I'm seeing. I might be missing a commit.

@seanpearsonuk Please have a look at this commit - https://github.com/ansys/pyfluent/pull/4284/files#diff-03e8f95919668d5ef0bc3cbded8e1f391e91d9fb50998ca46898488f8582a9d1

src/ansys/fluent/core/fluent_connection.py
hpohekar: Done. We have used None as default values. Thank you.

seanpearsonuk (approved):

PR #4291, opened 2025-07-16
ci: Use the new image name for version >=26.1

mkundu1:
Use the image from `ghcr.io/ansys/fluent` for 26.1 development and CI runs.


.ci/pull_fluent_image.py
mkundu1, line 1: Changed the bash script to python to keep the code simple

hpohekar: 👍

seanpearsonuk (approved):

PR #4309, opened 2025-07-24
fix: Issue with 'wait_process_finished'

prmukherj:
Expose 'wait_process_finished' from session level.

### DIFF
--- a/src/ansys/fluent/core/session.py
+++ b/src/ansys/fluent/core/session.py
@@ -164,7 +164,7 @@ class BaseSession:
     ):
         """Build a BaseSession object from fluent_connection object."""
         self._fluent_connection = fluent_connection
-        # Stores the backup once fluent connection is nullified.
+        # Stores the backup of the fluent connection for later reference.
         self._fluent_connection_backup = self._fluent_connection
         self._file_transfer_service = file_transfer_service
         self._error_state = fluent_connection._error_state


src/ansys/fluent/core/session.py
```diff
@@ -357,12 +359,35 @@ def _exit_compose_service(self):
...
+        Raises
+        ------
+        UnsupportedRemoteFluentInstance
+            If current Fluent instance is running remotely.
+        WaitTypeError
+            If ``wait`` is specified improperly.
+        """
+        if self._fluent_connection is None:
+            return self._fluent_connection_backup.wait_process_finished()
+        return self._fluent_connection.wait_process_finished()
```
seanpearsonuk: @prmukherj What is the reason for the boolean logic rather than a single call?

src/ansys/fluent/core/session.py
```diff
@@ -164,6 +164,8 @@ def _build_from_fluent_connection(
     ):
         """Build a BaseSession object from fluent_connection object."""
         self._fluent_connection = fluent_connection
+        # Stores the backup once fluent connection is nullified.
```
seanpearsonuk: @prmukherj Why isn't the backup always valid? It would be less complicated that way.

prmukherj: ok, sounds right. This will answer your other point as well.

prmukherj:
```suggestion
        # Stores the backup of the fluent connection for later reference.
```

mkundu1 (approved):
Looking at the current code, ideally `wait_process_finished` should not have been at `fluent_connection` level as Fluent process and Fluent connection are different concepts and they have different lifetimes. They are quite tightly coupled in the current implementation, it may worth investigating in future if it is possible to separate them.

prmukherj:
> Thinking about the existing code, ideally `wait_process_finished` should not have been at `fluent_connection` level as Fluent process and Fluent connection are different concepts and they have different lifetimes. They are quite tightly coupled in the current implementation, it may worth investigating in future whether it is possible to separate them.

yes, agreed.

hpohekar (approved):

seanpearsonuk (approved):

PR #4340, opened 2025-08-01
feat: Expose both object and method classes for builtin commands

mkundu1:


src/ansys/fluent/core/solver/settings_builtin_data.py
```diff
@@ -225,26 +225,14 @@
...
-            since(FluentVersion.v241): "setup.boundary_conditions.non_reflecting_bc",
-        },
-    ),
     "NonReflectingBoundary": (
-        "NamedObject",
-        {
-            since(FluentVersion.v241): "setup.boundary_conditions.non_reflecting_bc",
-        },
-    ),
-    "PerforatedWallBoundaries": (
```
mkundu1, line 240: These 2 were mistakenly marked as named-object

PR #4342, opened 2025-08-04
chore: Extract allowed values

hpohekar:
image

### DIFF
--- a/src/ansys/fluent/core/codegen/settingsgen.py
+++ b/src/ansys/fluent/core/codegen/settingsgen.py
@@ -316,6 +316,8 @@ def _write_data(cls_name: str, python_name: str, data: dict, f: IO, f_stub: IO |
         s_stub.write(
             f"    {to_constant_name(allowed_value)}: Final[str] = {allowed_value!r}\n"
         )
+    s.write(f"    _allowed_values = {data['allowed_values']!r}\n")
+    s_stub.write("    _allowed_values: list\n")
     s.write("\n")
     s_stub.write("\n")
     for name, (python_name, data, hash_, should_write_stub) in classes_to_write.items():


seanpearsonuk:
@hpohekar Please can we have description text: images won't play well in the logs.

seanpearsonuk (approved):

hpohekar:
> @hpohekar Please can we have description text: images won't play well in the logs.

@seanpearsonuk Done. Thank you.

prmukherj:
@hpohekar, for these things we can add a small test in test_codegen?

hpohekar:
We need to wait for some more time as Fluent side changes are not available in the latest Fluent image.

src/ansys/fluent/core/codegen/settingsgen.py
```diff
@@ -316,6 +316,8 @@ def _write_data(cls_name: str, python_name: str, data: dict, f: IO, f_stub: IO |
         s_stub.write(
             f"    {to_constant_name(allowed_value)}: Final[str] = {allowed_value!r}\n"
         )
+    s.write(f"    _allowed_values = {data['allowed_values']!r}\n")
```
mkundu1: @hpohekar Can we skip writing _allowed_values when it is empty to minimize the size of the generated file? You can add a private `_allowed_values = []` in the base class in flobject.py if that is required.

src/ansys/fluent/core/codegen/settingsgen.py
```diff
@@ -316,6 +316,8 @@ def _write_data(cls_name: str, python_name: str, data: dict, f: IO, f_stub: IO |
...
+    s_stub.write("    _allowed_values: list\n")
```
mkundu1: list -> list[str]

src/ansys/fluent/core/codegen/settingsgen.py
hpohekar: @mkundu1 Sure. We will add only if they are present.

seanpearsonuk (approved):

src/ansys/fluent/core/codegen/settingsgen.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/codegen/settingsgen.py
hpohekar: Done. Thank you.

PR #4345, opened 2025-08-04
fix: Unavaialble field data for some surfaces.

prmukherj:
Sometimes some surfaces might be created but field data on them might not be available. For example please refer the issue: https://github.com/ansys/pyfluent-visualization/issues/552

Now we are getting and returning the field data as empty array for these surfaces, just like Fluent does. We are additionally providing an warning that theses surfaces does not have field data.

image

We need to patch this with v0.34 to address an user bug.

### DIFF
--- a/doc/changelog.d/4345.fixed.md
+++ b/doc/changelog.d/4345.fixed.md
@@ -1,0 +1,1 @@
+Unavaialble field data for some surfaces.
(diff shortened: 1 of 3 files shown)


seanpearsonuk:
This is a tricky one where I'm not sre what the best behaviour would be. Please make sure that we are also documenting the API behaviour clearly. IMHO that's more important than runtime warnings.

seanpearsonuk (approved):

doc/changelog.d/4345.fixed.md
```diff
@@ -0,0 +1 @@
+Unavaialble field data for some surfaces.
```
prmukherj, line 1:
```suggestion
Unavailable field data for some surfaces.
```

PR #4349, opened 2025-08-05
feat: File handling for Ansys Notebook

hpohekar:
This pull request updates the `FileTransferService` logic to introduce an intermediate local directory for file uploads and downloads, optimizing file handling by checking for and using a per-instance directory before interacting with the remote file service.

**File transfer logic enhancements:**

* Added `cwd`, `instance_name`, and `instance_dir` attributes to the class, setting up a per-instance directory based on the current working directory and the instance name (`__init__` in `file_transfer_service.py`).
* Modified the `upload` method to copy files to the local `instance_dir` if it exists, bypassing remote upload if possible (`upload` in `file_transfer_service.py`).
* Updated the `download` method to first check for the file in the local `instance_dir` and copy it to the current working directory if found, reducing unnecessary remote downloads (`download` in `file_transfer_service.py`).

### DIFF
--- a/doc/changelog.d/4345.fixed.md
+++ b/doc/changelog.d/4345.fixed.md
@@ -1,1 +1,1 @@
-Unavaialble field data for some surfaces.
+Unavailable field data for some surfaces.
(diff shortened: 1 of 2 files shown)


seanpearsonuk (approved):
:+1:

PR #4361, opened 2025-08-08
docs: Added example on the battery cell simulation.

MohammedAnsys:
Added End to End PyFluent example on Single Cell Battery Simulation based on [Ansys Fluent documentation]( https://ansyshelp.ansys.com/public/account/secured?returnurl=/Views/Secured/corp/v242/en/flu_tg/flu_bat_tutorial_cell.html).

Please review the same!

Thanks!

### DIFF
--- a/examples/00-fluent/single_battery_cell_workflow.py
+++ b/examples/00-fluent/single_battery_cell_workflow.py
@@ -121,7 +121,11 @@ battery.zone_assignment.positive_tab = ["tab_p"]
 #
 # .. note::
 #   Chemical formula values are arbitrary identifiers for demonstration.
-
+#
+# Material definition for battery cell, positive tab and negative tab. User define
+# scalars are defined for e-material and positive material to specify the
+# electric conductivity with ``defined-per-uds`` and ``constant`` option respectively.
+#
 materials = [
     {
         "name": "e_material",


doc/source/conf.py
```diff
@@ -151,7 +151,7 @@ def _stop_fluent_container(gallery_conf, fname):
     # path where to save gallery generated examples
     "gallery_dirs": ["examples"],
     # Pattern to search for example files
-    "filename_pattern": r"exhaust_system_settings_api\.py|external_compressible_flow\.py|mixing_elbow_settings_api\.py|modeling_cavitation\.py|species_transport\.py|ahmed_body_workflow\.py|brake\.py|DOE_ML\.py|radiation_headlamp\.py|parametric_static_mixer_1\.py|conjugate_heat_transfer\.py|tyler_sofrin_modes\.py|lunar_lander_thermal\.py|modeling_ablation\.py|frozen_rotor_workflow\.py|mixing_tank_workflow\.py",
+    "filename_pattern": r"exhaust_system_settings_api\.py|external_compressible_flow\.py|mixing_elbow_settings_api\.py|modeling_cavitation\.py|species_transport\.py|ahmed_body_workflow\.py|brake\.py|DOE_ML\.py|radiation_headlamp\.py|parametric_static_mixer_1\.py|conjugate_heat_transfer\.py|tyler_sofrin_modes\.py|lunar_lander_thermal\.py|modeling_ablation\.py|frozen_rotor_workflow\.py|mixing_tank_workflow\.py|single_battery_cell_workflow\.py|",
```
hpohekar, line 154: 👍

hpohekar:
@MohammedAnsys 

Please refer [workflow file](https://github.com/ansys/pyfluent/blob/main/.github/workflows/execute-examples-weekly.yml) and add similar section on the following line: 

https://github.com/ansys/pyfluent/blob/main/.github/workflows/execute-examples-weekly.yml#L153

```
- name: Execute single_battery_cell_workflow.py
   run: |
```

This will help to run this example in weekly in the GitHub.

examples/00-fluent/single_battery_cell_workflow.py
```diff
@@ -0,0 +1,727 @@
...
+)
+
+#######################################################################################
+# Load Mesh file and display it with surfaces
+# =====================================================================================
+
+
+solver.settings.file.read_case(file_name=unit_battery_mesh)  # Read the mesh file
+# List of surfaces to display
+surfaces = [
```
prmukherj: Can we use built-in settings objects here, like wall boundaries?

MohammedAnsys:
> Looks, good. You can try using ansys-fluent-visualization for rendering graphics as well from your side and provide your views on how it works: https://github.com/ansys/pyfluent-visualization

Hi @prmukherj, Thanks for reviewing the PR.
Surely will add visualization example on the repo.

.github/workflows/execute-examples-weekly.yml
```diff
@@ -150,6 +150,10 @@ jobs:
         run: |
           python examples/00-fluent/tyler_sofrin_modes.py
 
+      - name: Execute single_battery_cell_workflow.py
+        run: |
+          python examples/00-fluent/single_battery_cell_workflow.py
+
```
hpohekar, line 156: Looks good. Thank you. 👍

hpohekar (approved):

src/ansys/fluent/core/utils/file_transfer_service.py
```diff
@@ -647,7 +647,7 @@ def download_file(self, file_name: str, local_directory: str | None = None):
             else:
                 raise FileNotFoundError("Remote file does not exist.")
 
-    def download(self, file_name: list[str] | str, local_directory: str | None = None):
+    def download(self, file_name: list[str] | str, local_directory: str | None = "."):
```
hpohekar: @MohammedAnsys The changes in this file are not relevant to this PR. These are old changes. This should be resolved before merging the PR. You can pull latest changes from main branch to resolve this.

abhishekchitwar (approved):

hpohekar (changes requested):
@abhishekchitwar `src/ansys/fluent/core/utils/file_transfer_service.py` has been deleted. This should be resolved properly.

cc: @seanpearsonuk

src/ansys/fluent/core/utils/file_transfer_service.py
hpohekar: @MohammedAnsys This is not resolved properly. Unresolving again.

examples/00-fluent/single_battery_cell_workflow.py
```diff
@@ -0,0 +1,719 @@
...
+# =====================================================================================
+# The battery mesh file is downloaded from the Fluent examples repository and
+# saved to the current working directory:
+#
+unit_battery_mesh = examples.download_file(
+    "unit_battery.msh.h5",
+    "pyfluent/battery_thermal_simulation",
+    save_path=os.getcwd(),
+)
+
```
seanpearsonuk, line 70: I’ve noticed that comments in this PR are written in the passive voice, which differs from the active style used in other PyFluent examples. The active style is preferred for consistency and readability.

We should also aim for economy in comments—only add them when they communicate something genuinely useful to the reader. Redundant or overly literal comments can slow readers down and may indicate that the script or API itself could be clearer. If many comments feel necessary, that might be a sign the underlying flow or naming could be improved.

examples/00-fluent/single_battery_cell_workflow.py
```diff
@@ -0,0 +1,719 @@
...
+solver.settings.file.read_case(file_name=unit_battery_mesh)  # Read the mesh file
+# List of surfaces to display
+surface_list = solver.settings.setup.boundary_conditions.wall.get_object_names()
+surfaces = surface_list
+mesh = solver.settings.results.graphics.mesh
+mesh.create("mesh-1")
+mesh["mesh-1"].surfaces_list = surfaces
+mesh["mesh-1"].options.edges = True
+mesh["mesh-1"].display()
```
seanpearsonuk: - `create()` already returns the new object, so you can store it directly and avoid repeatedly indexing `mesh["mesh-1"]`. This makes the code cleaner and more object-oriented.

- Setting each property in separate calls is relatively inefficient, especially when each call triggers back-end processing and interprocess communication (which may be remote). Where possible, group property assignments into a single update.

- In cases where repeated access is unavoidable, aliasing (`mesh_1 = mesh["mesh-1"]`) helps reduce duplication and improves readability.

examples/00-fluent/single_battery_cell_workflow.py
```diff
@@ -0,0 +1,719 @@
...
+graphics = solver.settings.results.graphics
+
+graphics.picture.x_resolution = 650
+graphics.picture.y_resolution = 450
+
+graphics.views.restore_view(view_name="isometric")
+
+graphics.picture.save_picture(
+    file_name="Single_Battery_Cell_Mesh.png",
+)
```
seanpearsonuk: Does this work for all Fluent run modes?

**Aside:** I have concerns about our examples being **dominated** by Fluent graphics usage which is **not a first class part of the API** and can easily be rendered unusable in a clean headless mode. We need to think about structuring the examples so that users can learn about the full range of alternatives.

examples/00-fluent/single_battery_cell_workflow.py
```diff
@@ -0,0 +1,719 @@
...
+# This checks if Fluent understood your setup and shows how everything is connected.
+#
+battery = solver.settings.setup.models.battery
+battery.enabled = True
+battery.echem_model = "ntgk/dcir"
+battery.zone_assignment.active_zone = ["e_zone"]
+battery.zone_assignment.passive_zone = ["tab_nzone", "tab_pzone"]
+battery.zone_assignment.negative_tab = ["tab_n"]
+battery.zone_assignment.positive_tab = ["tab_p"]
+battery.zone_assignment.print_battery_connection()
```
seanpearsonuk: This appears to print directly to the console, which makes it more of a debugging or inspection utility than a proper API method. Such functions generally shouldn’t be part of the public API, and we should avoid using them in examples. Examples should demonstrate supported, programmatic workflows that return data rather than only printing it.

examples/00-fluent/single_battery_cell_workflow.py
```diff
@@ -0,0 +1,719 @@
...
+# You're telling Fluent which parts are the active battery
+# region (electrochemical zone) and which are conductive parts like tabs or connectors.
+#
+# Assign External Connectors:
+# Defines where the current enters and leaves the battery. These are the positive and
+# negative terminals.
+#
+# Print Battery System Connection Info:
+# This checks if Fluent understood your setup and shows how everything is connected.
+#
```
seanpearsonuk: Is a lot of this just telling the reader what the defaults are or is it all actioned somewhere in the script? The comments should make this clearer.

examples/00-fluent/single_battery_cell_workflow.py
```diff
@@ -0,0 +1,719 @@
...
+#  Defining New Materials for Cell and Tabs
+# =====================================================================================
+# Material properties for the battery cell and tabs are defined to accurately
+# model electrical and thermal behavior:
+#
+# what materials the battery is made of
+#
+# How well each material conducts electricity
+#
+# This helps fluent calculate voltage, current, and heat correctly during the simulation
```
seanpearsonuk: Fluent should always be capitalised, `"Fluent"`.

examples/00-fluent/single_battery_cell_workflow.py
```diff
@@ -0,0 +1,719 @@
...
+        "thermal_conductivity": 387.6,
+        "uds_diffusivity": {"option": "constant", "value": 10000000},
+    },
+    {
+        "name": "n_material",
+        "chemical_formula": "nmat",
+        "density": 8978,
+        "specific_heat": 381,
+        "thermal_conductivity": 387.6,
+    },
```
seanpearsonuk, line 156: The chemical_formula values here (`"e"`, `"pmat"`, `"nmat"`) are placeholders, not actual chemical formulas. We should either use realistic values to avoid confusion, or make it clear in a comment that these are arbitrary identifiers for demonstration purposes only.

examples/00-fluent/single_battery_cell_workflow.py
seanpearsonuk: Also, the list is inconsistently formatted: **w**hat, **H**ow, **T**his.

examples/00-fluent/single_battery_cell_workflow.py
```diff
@@ -0,0 +1,719 @@
...
+]
+
+solids = solver.settings.setup.materials.solid
```
seanpearsonuk, line 159: We should aim for direct assignment. The existence of the following code in an example script would only imply that the API is difficult to use.

examples/00-fluent/single_battery_cell_workflow.py
```diff
@@ -0,0 +1,719 @@
...
+# This helps fluent calculate voltage, current, and heat correctly during the simulation
+
+materials = [
+    {
+        "name": "e_material",
+        "chemical_formula": "e",
+        "density": 2092,
+        "specific_heat": 678,
+        "thermal_conductivity": 18.2,
+        "uds_diffusivity": {
```
seanpearsonuk, line 136: Are we missing a definition for UDS Diffusivity?

examples/00-fluent/single_battery_cell_workflow.py
```diff
@@ -0,0 +1,719 @@
...
+# =====================================================================================
+# Convective heat transfer is applied to the battery’s external surfaces to model heat
+# loss to the environment, with a heat transfer coefficient of 5 W/m²·K:
+
+wall = solver.settings.setup.boundary_conditions.wall
+wall["wall_active"].thermal.thermal_condition = "Convection"
+wall["wall_active"].thermal.heat_transfer_coeff.value = 5
+
+solver.settings.setup.boundary_conditions.copy(
+    from_="wall_active", to=["wall_n", "wall_p"], verbosity=True
```
seanpearsonuk: Since it's unobvious what the implication of `verbosity=True` is, that should be commented here, or **preferably** the argument should be omitted.

**Reason:** We need to report the API design in this instance, as once again, **this does not belong in the API:** `verbosity:` `"Copy boundary conditions: Print more information."` in other words, the `verbosity=True` argument makes this method print directly to the console, which is undesirable for API calls shown in examples (or in general). 

**Aside:** `verbose=True` and `verbosity=1|2|3` would make more sense.

examples/00-fluent/single_battery_cell_workflow.py
```diff
@@ -0,0 +1,719 @@
...
+)
+
+#######################################################################################
+# Specifying solution settings
+# =====================================================================================
+# Since the simulation models a solid battery without fluid flow,
+# flow and turbulence equations are disabled, and residual criteria are set to none:
```
seanpearsonuk: For `none` write `"none"` to clearly disambiguate from Python `None`.

examples/00-fluent/single_battery_cell_workflow.py
```diff
@@ -0,0 +1,719 @@
...
+#
+# No flow = No need to solve equations for velocity or pressure.
+#
+# No turbulence = No air or liquid movement that could become chaotic.
```
seanpearsonuk: It isn't made sufficiently clear how this correlates with the following code. The issue is partly due to API design.

examples/00-fluent/single_battery_cell_workflow.py
```diff
@@ -0,0 +1,719 @@
...
+#######################################################################################
+# Creating Report Definitions
+# =====================================================================================
+# Report definitions are created to monitor the average voltage on the positive tab and
+# the maximum temperature across the battery zones. Output files and plots are
+# generated to track these quantities:
+
+# Surface report for voltage
+surface_reports = solver.settings.solution.report_definitions.surface
+surface_reports.create("voltage_vp")
```
seanpearsonuk: Avoid obscure abbreviations.

examples/00-fluent/single_battery_cell_workflow.py
```diff
@@ -0,0 +1,719 @@
...
+vector_plot.options.vector_style = "arrow"
+vector_plot.range_options.compute()
+
+# Set view, display, and save the vector plot image
+graphics.views.restore_view(view_name="front")
+vector_plot.display()
+graphics.picture.save_picture(file_name="Single_Battery_Cell_5.png")
+
+# Save case file
+solver.settings.file.write(file_name="unit_battery.cas.h5", file_type="case")
```
seanpearsonuk: I realise that the API contains such general methods as `read()` and `write()` as well as the specific versions. I have a preference for the latter, especially while `file_type` is a string argument.

examples/00-fluent/single_battery_cell_workflow.py
MohammedAnsys: Updated all the comments and instruction with the active style as recommended.
Thanks

examples/00-fluent/single_battery_cell_workflow.py
MohammedAnsys: Updated as per review comment.

examples/00-fluent/single_battery_cell_workflow.py
MohammedAnsys: Removed dubugging API as recommended.

examples/00-fluent/single_battery_cell_workflow.py
MohammedAnsys: Added the existing documentation as reference so one can refer in detail, removed all the duplicate information

examples/00-fluent/single_battery_cell_workflow.py
MohammedAnsys: Updated as per review comment.

examples/00-fluent/single_battery_cell_workflow.py
MohammedAnsys: Updated as per review comment.

examples/00-fluent/single_battery_cell_workflow.py
MohammedAnsys: I have tested the API with available mode and added as a NOTE

examples/00-fluent/single_battery_cell_workflow.py
MohammedAnsys: Updated as per review comment.

examples/00-fluent/single_battery_cell_workflow.py
MohammedAnsys: Added NOTE for chemical formulas.

examples/00-fluent/single_battery_cell_workflow.py
MohammedAnsys: This information is well captured in references as well as Fluent documentation specific example appendix.

examples/00-fluent/single_battery_cell_workflow.py
MohammedAnsys: Added comment to avoid confusion.

examples/00-fluent/single_battery_cell_workflow.py
MohammedAnsys: Yes, I removed this comment to avoid confusion.

examples/00-fluent/single_battery_cell_workflow.py
MohammedAnsys: I have updated proper naming convention which relates to the report defination.

examples/00-fluent/single_battery_cell_workflow.py
MohammedAnsys: Updated as per comment.

examples/00-fluent/single_battery_cell_workflow.py
seanpearsonuk: Please refer to @abhishekchitwar to clarify the requirements.

abhishekchitwar:
> @abhishekchitwar, what was the reason for deleting an utility file from pyfluent's core repo? [4853b00](https://github.com/ansys/pyfluent/commit/4853b00311520158de9338d4e31fda6abded1a9e)
> 
> cc. @seanpearsonuk, @MohammedAnsys

Hi @hpohekar, It seems this happened by mistake at first instance. I thought it's not related to @MohammedAnsys working branch.

src/ansys/fluent/core/utils/file_transfer_service.py
hpohekar: It's resolved now.

hpohekar (approved):

PR #4370, opened 2025-08-11
fix: grpc connection

mkundu1:


mkundu1:
Abandoning as covered in another PR

PR #4371, opened 2025-08-12
test: Add test to check PyFluent launched PyConsole by default.

prmukherj:
Add test to check PyFluent launched PyConsole by default.

### DIFF
--- a/tests/test_launcher.py
+++ b/tests/test_launcher.py
@@ -759,4 +759,3 @@ def test_warning_for_deprecated_compose_env_vars(monkeypatch):
 def test_default_launch_mode_is_py():
     fluent_launch_string, _ = pyfluent.launch_fluent(dry_run=True)
     assert "-py" in fluent_launch_string
-


hpohekar (approved):

tests/test_launcher.py
```diff
@@ -753,3 +753,9 @@ def test_warning_for_deprecated_compose_env_vars(monkeypatch):
     monkeypatch.setattr(pyfluent.config, "use_podman_compose", True)
     with pytest.warns(PyFluentDeprecationWarning):
         ComposeConfig()
+
+
```
prmukherj, line 756:
```suggestion
@pytest.mark.standalone
```

tests/test_launcher.py
```diff
@@ -753,3 +753,9 @@ def test_warning_for_deprecated_compose_env_vars(monkeypatch):
     monkeypatch.setattr(pyfluent.config, "use_podman_compose", True)
     with pytest.warns(PyFluentDeprecationWarning):
         ComposeConfig()
+
```
prmukherj, line 755:
```suggestion


```

tests/test_launcher.py
```diff
@@ -753,3 +753,10 @@ def test_warning_for_deprecated_compose_env_vars(monkeypatch):
...
+
+@pytest.mark.standalone
+@pytest.mark.fluent_version(">=25.1")
+def test_default_launch_mode_is_py():
+    fluent_launch_string, _ = pyfluent.launch_fluent(dry_run=True)
+    assert "-py" in fluent_launch_string
```
prmukherj, line 761:
```suggestion
    assert "-py" in fluent_launch_string

```

hpohekar:
@prmukherj 

This is incorrect behaviour. We need to resolve this. The issue description already states that we must add unit test with fix.

We just need to update the following condition and I can confirm that it resolves this issue.

```
from ansys.fluent.core import config

if (
            (fluent_version and fluent_version >= FluentVersion.v251) or config.fluent_root
            and self.argvals["py"] is None
        ):
            self.argvals["py"] = True
```

https://github.com/ansys/pyfluent/blob/main/src/ansys/fluent/core/launcher/standalone_launcher.py#L207

cc: @seanpearsonuk @mkundu1

prmukherj:
> @prmukherj
> 
> This is incorrect behaviour. We need to resolve this. The issue description already states that we must add unit test with fix.
> 
> We just need to update the following condition and I can confirm that it resolves this issue.
> 
> ```
> from ansys.fluent.core import config
> 
> if (
>             (fluent_version and fluent_version >= FluentVersion.v251) or config.fluent_root
>             and self.argvals["py"] is None
>         ):
>             self.argvals["py"] = True
> ```
> 
> https://github.com/ansys/pyfluent/blob/main/src/ansys/fluent/core/launcher/standalone_launcher.py#L207
> 
> cc: @seanpearsonuk @mkundu1

I don't think it makes sense. I think the earlier behaviour was more correct. What if user sets fluent_root to 22R2?

prmukherj:
> > @prmukherj
> > This is incorrect behaviour. We need to resolve this. The issue description already states that we must add unit test with fix.
> > We just need to update the following condition and I can confirm that it resolves this issue.
> > ```
> > from ansys.fluent.core import config
> > 
> > if (
> >             (fluent_version and fluent_version >= FluentVersion.v251) or config.fluent_root
> >             and self.argvals["py"] is None
> >         ):
> >             self.argvals["py"] = True
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > https://github.com/ansys/pyfluent/blob/main/src/ansys/fluent/core/launcher/standalone_launcher.py#L207
> > cc: @seanpearsonuk @mkundu1
> 
> I don't think it makes sense. I think the earlier behaviour was more correct. What if user sets fluent_root to 22R2?

Moreover, if we want to make this change, '_get_standalone_launch_fluent_version' is the correct place to do so I believe.

seanpearsonuk (approved):

PR #4374, opened 2025-08-12
feat: Interactive settings UI.

prmukherj:
The interactive **Settings UI** provides an intuitive way to explore and modify
settings objects exposed by PyFluent. It lets you navigate through the hierarchy,
expand sections, and directly interact with individual parameters or commands.

### DIFF
--- a/doc/source/user_guide/ui.rst
+++ b/doc/source/user_guide/ui.rst
@@ -44,14 +44,14 @@ You can also launch the UI for a specific subset of settings:
 Features
 --------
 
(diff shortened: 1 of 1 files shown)


seanpearsonuk:
`interactive` is obscure and `_version` is redundant.

src/ansys/fluent/core/ui/standalone_web_ui.py
seanpearsonuk, line 1: Can rename this `web_ui.py`.

PR #4380, opened 2025-08-12
fix: Some config and env var cleanup

mkundu1:
1. Remove the config equivalent to `PYFLUENT_FLUENT_ROOT` environment variable. That environment variable is only for development use.
2. Remove the mentions of the environment `PYFLUENT_LAUNCH_CONTAINER` from user documentation. That environment variable is only for CI use.
3. Fix a typo in slurm launcher.

### DIFF
--- a/src/ansys/fluent/core/module_config.py
+++ b/src/ansys/fluent/core/module_config.py
@@ -300,11 +300,6 @@ class Config:
         lambda instance: instance._env.get("PYFLUENT_HIDE_LOG_SECRETS") == "1"
     )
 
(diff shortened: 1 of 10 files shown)


src/ansys/fluent/core/module_config.py
```diff
@@ -300,11 +300,6 @@ class Config:
         lambda instance: instance._env.get("PYFLUENT_HIDE_LOG_SECRETS") == "1"
     )
 
-    #: The Fluent root directory to be used for PyFluent, defaults to the value of ``PYFLUENT_FLUENT_ROOT`` environment variable.
-    fluent_root = _ConfigDescriptor["Config"](
```
prmukherj, line 304: 👍

hpohekar (approved):

PR #4382, opened 2025-08-12
fix: Textual type in flobject.

prmukherj:
Fix Textual type in flobject.


hpohekar (approved):

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -676,7 +677,14 @@ def set_state(self, state: StateT | None = None, **kwargs):
...
         kwargs : Any
             Keyword arguments.
+
+        Raises
+        ------
+        TypeError
+            If state is not a string.
         """
+        if not isinstance(state, (str, VariableDescriptor)):
+            raise TypeError(f"Expected state to be str, got {type(state).__name__}.")
```
```suggestion
        allowed_types = (str, VariableDescriptor)

        if not isinstance(state, allowed_types):
            expected = " or ".join(t.__name__ for t in allowed_types)
            raise TypeError(
                f"Expected state to be {expected}, got {type(state).__name__}."
            )
```

The prior message will lead to confusion.

PR #4394, opened 2025-08-19
feat: Get launch mode

hpohekar:
We can retrieve the actual mode a Fluent session is launched in by using `session.launch_mode`

```
>>> solver = launch_fluent()
>>> solver.launch_mode
<LaunchMode.STANDALONE: 'standalone'>
>>>
```

```
>>> meshing = pyfluent.Meshing.from_install()
>>> meshing.launch_mode
<LaunchMode.STANDALONE: 'standalone'>
>>>
```

### DIFF
--- a/src/ansys/fluent/core/session.py
+++ b/src/ansys/fluent/core/session.py
@@ -143,7 +143,6 @@ class BaseSession:
         """
         self._start_transcript = start_transcript
         self._launcher_args = launcher_args
-        self.launch_mode = self._launcher_args.get("launch_mode")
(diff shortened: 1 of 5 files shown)


src/ansys/fluent/core/session.py
```diff
@@ -143,6 +143,7 @@ def __init__(
         """
         self._start_transcript = start_transcript
         self._launcher_args = launcher_args
+        self.launch_mode = self._launcher_args.get("launch_mode")
```
prmukherj: self._launcher_args = {} if launcher_args is None else launcher_args

PR #4407, opened 2025-08-26
docs: Added example on one way fsi [skip tests]

MohammedAnsys:


.github/workflows/execute-examples-weekly.yml
```diff
@@ -154,6 +154,10 @@ jobs:
         run: |
           python examples/00-fluent/single_battery_cell_workflow.py
 
+      - name: Execute fsi_1way_workflow.py
+        run: |
+          python examples/00-fluent/fsi_1way_workflow.py
+
```
hpohekar, line 175: 👍

doc/source/conf.py
```diff
@@ -151,7 +151,7 @@ def _stop_fluent_container(gallery_conf, fname):
     # path where to save gallery generated examples
     "gallery_dirs": ["examples"],
     # Pattern to search for example files
-    "filename_pattern": r"exhaust_system_settings_api\.py|external_compressible_flow\.py|mixing_elbow_settings_api\.py|modeling_cavitation\.py|species_transport\.py|ahmed_body_workflow\.py|brake\.py|DOE_ML\.py|radiation_headlamp\.py|parametric_static_mixer_1\.py|conjugate_heat_transfer\.py|tyler_sofrin_modes\.py|lunar_lander_thermal\.py|modeling_ablation\.py|frozen_rotor_workflow\.py|mixing_tank_workflow\.py|single_battery_cell_workflow\.py|",
+    "filename_pattern": r"exhaust_system_settings_api\.py|external_compressible_flow\.py|mixing_elbow_settings_api\.py|modeling_cavitation\.py|species_transport\.py|ahmed_body_workflow\.py|brake\.py|DOE_ML\.py|radiation_headlamp\.py|parametric_static_mixer_1\.py|conjugate_heat_transfer\.py|tyler_sofrin_modes\.py|lunar_lander_thermal\.py|modeling_ablation\.py|frozen_rotor_workflow\.py|mixing_tank_workflow\.py|single_battery_cell_workflow\fsi_1way_workflow\.py|",
```
hpohekar: 👍

examples/00-fluent/fsi_1way_workflow.py
```diff
@@ -0,0 +1,220 @@
...
+
+Modeling One-Way Fluid-Structure Interaction
+-------------------------------------------------------------
+"""
+
+#######################################################################################
+# Objective
+# =====================================================================================
+#
+# The Simulation focuses on simulating turbulent airflow through a cylindrical test chamber
```
seanpearsonuk:
```suggestion
# The simulation focuses on simulating turbulent airflow through a cylindrical test chamber
```

examples/00-fluent/fsi_1way_workflow.py
```diff
@@ -0,0 +1,220 @@
...
+# containing a steel probe, and analyzing the deformation of the probe due to
+# aerodynamic forces. The deformation is assumed to be small enough that it
+# does not influence the fluid flow, allowing for a one-way coupling approach.
+
```
seanpearsonuk, line 48: I had two concerns about the current **Objective** paragraph:
- The intro feels a little clunky.
- It assumes too much background knowledge.

Even for expert readers, examples are more engaging when they begin with a fuller explanation of the context. I’d like us to adopt this style across all examples so they are more accessible to a broader range of users, including non-experts and students.

Here’s a suggested replacement text (from ChatGPT):

### Objective

This example models turbulent airflow through a cylindrical test chamber that contains a steel probe. The airflow generates aerodynamic forces on the probe, causing it to deform. In this case, the deformation is expected to be small compared with the overall flow field. Because the probe’s motion does not significantly alter the airflow, we can treat the problem using a **one-way fluid–structure interaction (FSI)** approach.

In a one-way FSI analysis, the fluid flow is solved first and the resulting forces are transferred to the structural model. The structural response is then computed independently, without feeding back into the fluid solution. This contrasts with a two-way FSI analysis, where structural deformation and fluid flow are solved in a fully coupled manner. The one-way approach is computationally more efficient and appropriate when structural feedback on the flow can be neglected.

examples/00-fluent/fsi_1way_workflow.py
```diff
@@ -0,0 +1,220 @@
...
+    save_path=os.getcwd(),
+)
+
+examples.download_file(
+    "fsi_1way.msh.h5",
+    "pyfluent/fsi_1way",
+    save_path=os.getcwd(),
+)
+
+solver.tui.file.read_journal(journal_file)  # Read the journal file
```
seanpearsonuk: The comment only repeats the code.

examples/00-fluent/fsi_1way_workflow.py
```diff
@@ -0,0 +1,220 @@
...
+
+#######################################################################################
+# Download and Read the journal file
+# =====================================================================================
+#
+# .. note::
+#   Mesh file is required as input for the journal file.
+#
+# Journal file serves as a script that instructs Ansys Fluent on sequential operations
+
```
seanpearsonuk, line 97: Some of the above grammar is wrong due to missing articles.

examples/00-fluent/fsi_1way_workflow.py
```diff
@@ -0,0 +1,220 @@
...
+# Launch Fluent session in solver mode
+# =====================================================================================
+
+solver = pyfluent.launch_fluent(
+    precision=Precision.DOUBLE,
+    mode=FluentMode.SOLVER,
+)
+
+#######################################################################################
+# Download and Read the journal file
```
seanpearsonuk: **R**ead: capital mid-sentence again.

examples/00-fluent/fsi_1way_workflow.py
seanpearsonuk: `read_journal(<legacy-file>)` is not suitable for a PyFluent example. We should do the scripting using the _modern_ Python interface _and_ from PyFluent, not by uploading some arbitrary code to execute on the server.

examples/00-fluent/fsi_1way_workflow.py
```diff
@@ -0,0 +1,220 @@
...
+)
+
+examples.download_file(
+    "fsi_1way.msh.h5",
+    "pyfluent/fsi_1way",
+    save_path=os.getcwd(),
+)
+
+solver.tui.file.read_journal(journal_file)  # Read the journal file
+
```
seanpearsonuk, line 138: Fluent postprocessing is not the future for PyFluent. We want to use Fluent as a solver, extract its data, and do the rest _Pythonically_.

examples/00-fluent/fsi_1way_workflow.py
```diff
@@ -0,0 +1,220 @@
...
+# =====================================================================================
+# configure Fluent to define the steel probe's support and movement using
+# structural boundary conditions
+
+wall = solver.settings.setup.boundary_conditions.wall
+
+# Configure solid-symmetry boundary
+wall["solid-symmetry"] = {
+    "structure": {
+        "z_disp_boundary_value": 0,
```
seanpearsonuk, line 175: _disp_ (`z_disp_boundary_value` etc) is a disappointing API naming choice, and goes against the guidelines. I will report this as appropriate.

examples/00-fluent/fsi_1way_workflow.py
MohammedAnsys: updated as per comments

seanpearsonuk: @MohammedAnsys 

I appreciate the update to the introduction. A couple of points still stand out to me:
- **Sentence length and grammar**: The first sentence is quite long, and the phrasing (“simulates … and to analyze”) doesn’t read correctly. Shorter, cleaner sentences will improve readability.
- **Paragraph structure**: The whole text is written as a single block. In the earlier suggestion, splitting into two paragraphs gave each idea its own space:
  - Paragraph 1: Problem setup.
  - Paragraph 2: Why one-way FSI is appropriate.

This separation helps both readability and comprehension, especially for newer users or students.

MohammedAnsys: Updated as per review comments

.github/workflows/ci.yml
```diff
@@ -647,7 +647,7 @@ jobs:
 
       - name: Upload to Private PyPi
         run: |
-          pip install 'twine<6.2.0'
```
hpohekar, line 687: This is unrelated change. latest change is `pip install 'twine<6.2.0'`. Please update your branch with respect to main branch.

PR #4433, opened 2025-09-02
fix: Resolve examples workflow issues [skip tests]

hpohekar:


examples/00-fluent/single_battery_cell_workflow.py
```diff
@@ -55,7 +55,6 @@
     precision=Precision.DOUBLE,
     processor_count=4,
     mode=FluentMode.SOLVER,
-    ui_mode=UIMode.GUI,
```
hpohekar: @MohammedAnsys As we run example scripts in container mode in GitHub workflow, we do not use `UIMODE.GUI` otherwise it fails to execute the example as follows:

https://github.com/ansys/pyfluent/actions/runs/17396669497/job/49380020707

MohammedAnsys: @hpohekar, UI mode has been disabled in the GitHub workflow. #4436 this PR includes the fix for that issue.

hpohekar: @MohammedAnsys I had already pushed the fix in this PR but I will revert it now as you have created a separate PR for this.

seanpearsonuk (approved):

PR #4452, opened 2025-09-08
docs: example on the steady vortex dynamics

abhishekchitwar:
This PR provides a comprehensive example of simulating steady vortex dynamics, specifically focusing on vortex behavior within continuously stirred tank reactors. 

The example tries to follow the provided guidelines with #4378 and uses the new API implementation.

Thanks!

### DIFF
--- a/examples/00-fluent/steady_vortex.py
+++ b/examples/00-fluent/steady_vortex.py
@@ -295,7 +295,7 @@ solution_initialization.initialize()
 
 # Patch the water volume fraction in the defined cell register to set the initial
 # liquid region in the tank.
-# "mp" refers to the volume fraction of the primary phase (here water). 
+# "mp" refers to the volume fraction of the primary phase (here water).
 # Setting value=1 fills the patch region entirely with water.
 solution_initialization.patch.calculate_patch(
     domain="water",


.github/workflows/execute-examples-weekly.yml
```diff
@@ -157,6 +157,11 @@ jobs:
         run: |
           python examples/00-fluent/single_battery_cell_workflow.py
 
+      - name: Execute steady_vortex.py
+        run: |
+          python examples/00-fluent/steady_vortex.py
```
hpohekar, line 162: 👍

doc/source/conf.py
```diff
@@ -151,7 +151,7 @@ def _stop_fluent_container(gallery_conf, fname):
     # path where to save gallery generated examples
     "gallery_dirs": ["examples"],
     # Pattern to search for example files
-    "filename_pattern": r"exhaust_system_settings_api\.py|external_compressible_flow\.py|mixing_elbow_settings_api\.py|modeling_cavitation\.py|species_transport\.py|ahmed_body_workflow\.py|brake\.py|DOE_ML\.py|radiation_headlamp\.py|parametric_static_mixer_1\.py|conjugate_heat_transfer\.py|tyler_sofrin_modes\.py|lunar_lander_thermal\.py|modeling_ablation\.py|frozen_rotor_workflow\.py|mixing_tank_workflow\.py|single_battery_cell_workflow\.py|",
+    "filename_pattern": r"exhaust_system_settings_api\.py|external_compressible_flow\.py|mixing_elbow_settings_api\.py|modeling_cavitation\.py|species_transport\.py|ahmed_body_workflow\.py|brake\.py|DOE_ML\.py|radiation_headlamp\.py|parametric_static_mixer_1\.py|conjugate_heat_transfer\.py|tyler_sofrin_modes\.py|lunar_lander_thermal\.py|modeling_ablation\.py|frozen_rotor_workflow\.py|mixing_tank_workflow\.py|single_battery_cell_workflow\.py|steady_vortex\.py",
```
hpohekar, line 154: 👍

examples/00-fluent/steady_vortex.py
```diff
@@ -0,0 +1,568 @@
...
+    Materials,
+    Mesh,
+    Methods,
+    Models,
+    NamedExpression,
+    PlaneSurface,
+    RunCalculation,
+    Scene,
+    WallBoundary,
+)
```
hpohekar, line 101: 👍

hpohekar (approved):

examples/00-fluent/steady_vortex.py
```diff
@@ -0,0 +1,568 @@
...
+general_settings.operating_conditions.gravity.components = [0.0, 0.0, -g]
+
+# %%
+# Copy Materials from Fluent Database
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# Copy water liquid materials from the Fluent database.
+
+copy_material_from_database = Materials(solver_session)
+copy_material_from_database.database.copy_by_name(type="fluid", name="water-liquid")
```
seanpearsonuk: @abhishekchitwar 
Shouldn't this just read something like the following?
```suggestion
materials = Materials(solver_session)
materials.database.copy_by_name(type="fluid", name="water-liquid")
```

examples/00-fluent/steady_vortex.py
```diff
@@ -0,0 +1,568 @@
...
+# %%
+# Define Named Expression for Agitation Speed
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# Create a named expression for the agitation speed and specify as an input parameter.
+stirring_speed = NamedExpression(solver_session, new_instance_name="stirring_speed")
+
+stirring_speed.definition = "240 [rev min^-1]"
+stirring_speed.input_parameter = True
+
```
seanpearsonuk, line 211: :+1:

examples/00-fluent/steady_vortex.py
```diff
@@ -0,0 +1,568 @@
...
+# MRF zone parameters
+# ^^^^^^^^^^^^^^^^^^^
+#
+# Define MRF zone parameters for the mrf cell zone.
+
+# fluid_cell_zone = solver_session.settings.setup.cell_zone_conditions.fluid["mrf"]
+fluid_cell_zone = CellZoneCondition(solver_session, name="mrf")
+fluid_cell_zone.reference_frame.frame_motion = True
+fluid_cell_zone.reference_frame.reference_frame_axis_origin = [0, 0, 0]
+fluid_cell_zone.reference_frame.reference_frame_axis_direction = [0, 0, 1]
```
seanpearsonuk, line 221: Noting that `reference_frame` is repeated along the path. Creating an _issue_ for the API design...  #4461

examples/00-fluent/steady_vortex.py
```diff
@@ -0,0 +1,568 @@
...
+stirring_speed.input_parameter = True
+
+# %%
+# MRF zone parameters
+# ^^^^^^^^^^^^^^^^^^^
+#
+# Define MRF zone parameters for the mrf cell zone.
+
+# fluid_cell_zone = solver_session.settings.setup.cell_zone_conditions.fluid["mrf"]
+fluid_cell_zone = CellZoneCondition(solver_session, name="mrf")
```
seanpearsonuk: @abhishekchitwar Noting that you commented the previous line out, may we delete it now?

examples/00-fluent/steady_vortex.py
```diff
@@ -0,0 +1,568 @@
...
+fluid_cell_zone.reference_frame.frame_motion = True
+fluid_cell_zone.reference_frame.reference_frame_axis_origin = [0, 0, 0]
+fluid_cell_zone.reference_frame.reference_frame_axis_direction = [0, 0, 1]
+fluid_cell_zone.reference_frame.mrf_omega.value = "stirring_speed"
```
seanpearsonuk, line 222: @abhishekchitwar One guideline is to favour `dict`-style assignment, primarily for efficiency. Do you prefer this style for readability?

examples/00-fluent/steady_vortex.py
```diff
@@ -0,0 +1,568 @@
...
+
+# %%
+# Rotating Wall BC parameters
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# Define the rotating wall boundary condition for the shaft.
+
+# wall_boundary = solver_session.settings.setup.boundary_conditions.wall["shaft_tank"]
```
seanpearsonuk: I think we could definitely remove this one.

examples/00-fluent/steady_vortex.py
abhishekchitwar: @seanpearsonuk That's right, It will be helpful for user to understand, but I'll agree on the code performance with dict, will update to dict. Please let me know your views!

examples/00-fluent/steady_vortex.py
```diff
@@ -0,0 +1,568 @@
...
+        "max_point": [100.0, 100.0, 0.19],
+        "min_point": [-100.0, -100.0, -100.0],
+    },
+}
+
+solution_initialization.initialize()
+
+# Patch the water volume fraction in the defined cell register to set the initial
+# liquid region in the tank.
+
```
seanpearsonuk:
```suggestion
# "mp" refers to the volume fraction of the primary phase (here water). 
# Setting value=1 fills the patch region entirely with water.
```
@abhishekchitwar I'm proposing this comment because of the cryptic nature of this API. Please correct as appropriate if I have it wrong.  "mp" seems not to be included currently in the Variable Catalog currently which could be rectified...  

Created PyAnsys Units Issue https://github.com/ansys/pyansys-units/issues/403

examples/00-fluent/steady_vortex.py
abhishekchitwar: @seanpearsonuk, This wall is adjacent to the stationary domain, in that case it is required to define the wall explicitly as rotating. I am not sure if you're pointing to it or something else. Please explain if that is not the case. Thanks!

examples/00-fluent/steady_vortex.py
seanpearsonuk: In fact, I don't have a strong opinion on this topic.

examples/00-fluent/steady_vortex.py
abhishekchitwar: Yeah! That would be great! Thanks!

examples/00-fluent/steady_vortex.py
abhishekchitwar: Yeah! My bad! Thanks!

examples/00-fluent/steady_vortex.py
abhishekchitwar: So, for now will keep as is. Thanks!

examples/00-fluent/steady_vortex.py
seanpearsonuk: @abhishekchitwar I was referring to any lines of commented-out code like the one above which appear to have been replaced by an alternative format, such as the following line in this case. Once we have replaced this line of code, we should remove it entirely rather than leaving it in as a commented-out line. It would be untidy and the purpose of such lines is ambiguous.

abhishekchitwar: @seanpearsonuk, yeah, got it! Thanks!

PR #4456, opened 2025-09-09
chore: Added backwards compatibility for solution_varialbe_name.

prmukherj:
Added backwards compatibility for solution_varialbe_name.

### DIFF
--- a/src/ansys/fluent/core/services/solution_variables.py
+++ b/src/ansys/fluent/core/services/solution_variables.py
@@ -42,6 +42,7 @@ from ansys.fluent.core.services.interceptors import (
     TracingInterceptor,
 )
 from ansys.fluent.core.solver.error_message import allowed_name_error_message
+from ansys.fluent.core.utils.deprecate import all_deprecators
 from ansys.fluent.core.variable_strategies import (
     FluentSVarNamingStrategy as naming_strategy,
 )
@@ -343,6 +344,20 @@ class _AllowedSvarNames:
             zone_names=zone_names, domain_name=domain_name
         ).solution_variables
 
+    #   pylint: disable=unused-argument
+    @all_deprecators(
+        deprecate_arg_mappings=[
+            {
+                "old_arg": "solution_variable_name",
+                "new_arg": "variable_name",
+                "converter": lambda old_arg_val: old_arg_val,
+            },
+        ],
+        data_type_converter=None,
+        deprecated_version="v0.35.1",
+        deprecated_reason="'solution_variable_name' is deprecated. Use 'variable_name' instead.",
+        warn_message="",
+    )
     def is_valid(
         self,
         variable_name,
@@ -352,6 +367,20 @@ class _AllowedSvarNames:
         """Check whether solution variable name is valid or not."""
         return variable_name in self(zone_names=zone_names, domain_name=domain_name)
 
+    #   pylint: disable=unused-argument
+    @all_deprecators(
+        deprecate_arg_mappings=[
+            {
+                "old_arg": "solution_variable_name",
+                "new_arg": "variable_name",
+                "converter": lambda old_arg_val: old_arg_val,
+            },
+        ],
+        data_type_converter=None,
+        deprecated_version="v0.35.1",
+        deprecated_reason="'solution_variable_name' is deprecated. Use 'variable_name' instead.",
+        warn_message="",
+    )
     def valid_name(
         self,
         variable_name,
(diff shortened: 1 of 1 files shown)


src/ansys/fluent/core/services/solution_variables.py
```diff
@@ -343,6 +344,20 @@ def __call__(
             zone_names=zone_names, domain_name=domain_name
         ).solution_variables
 
+    #   pylint: disable=unused-argument
+    @all_deprecators(
+        deprecate_arg_mappings=[
+            {
+                "old_arg": "solution_variable_name",
+                "new_arg": "variable_name",
+                "converter": lambda old_arg_val: old_arg_val,
```
prmukherj:
```suggestion
```

src/ansys/fluent/core/services/solution_variables.py
```diff
@@ -352,6 +366,20 @@ def is_valid(
         """Check whether solution variable name is valid or not."""
         return variable_name in self(zone_names=zone_names, domain_name=domain_name)
 
+    #   pylint: disable=unused-argument
+    @all_deprecators(
+        deprecate_arg_mappings=[
+            {
+                "old_arg": "solution_variable_name",
+                "new_arg": "variable_name",
+                "converter": lambda old_arg_val: old_arg_val,
```
prmukherj:
```suggestion
```

src/ansys/fluent/core/services/solution_variables.py
```diff
@@ -581,6 +608,20 @@ def _update_solution_variable_info(self):
             self._solution_variable_info
         )
 
+    #   pylint: disable=unused-argument
+    @all_deprecators(
+        deprecate_arg_mappings=[
+            {
+                "old_arg": "solution_variable_name",
+                "new_arg": "variable_name",
+                "converter": lambda old_arg_val: old_arg_val,
```
prmukherj:
```suggestion
```

src/ansys/fluent/core/services/solution_variables.py
```diff
@@ -605,6 +645,20 @@ def create_empty_array(
                     dtype=solution_variables_info[variable_name].field_type,
                 )
 
+    #   pylint: disable=unused-argument
+    @all_deprecators(
+        deprecate_arg_mappings=[
+            {
+                "old_arg": "solution_variable_name",
+                "new_arg": "variable_name",
+                "converter": lambda old_arg_val: old_arg_val,
```
prmukherj:
```suggestion
```

src/ansys/fluent/core/services/solution_variables.py
```diff
@@ -650,6 +703,20 @@ def get_data(
             extract_svars(self._service.get_data(svars_request)),
         )
 
+    #   pylint: disable=unused-argument
+    @all_deprecators(
+        deprecate_arg_mappings=[
+            {
+                "old_arg": "solution_variable_name",
+                "new_arg": "variable_name",
+                "converter": lambda old_arg_val: old_arg_val,
```
prmukherj:
```suggestion
```

src/ansys/fluent/core/services/solution_variables.py
```diff
@@ -667,6 +733,20 @@ def get_svar_data(
             domain_name=domain_name,
         )
 
+    #   pylint: disable=unused-argument
+    @all_deprecators(
+        deprecate_arg_mappings=[
+            {
+                "old_arg": "solution_variable_name",
+                "new_arg": "variable_name",
+                "converter": lambda old_arg_val: old_arg_val,
```
prmukherj:
```suggestion
```

src/ansys/fluent/core/services/solution_variables.py
```diff
@@ -767,6 +846,20 @@ def generate_set_data_requests():
 
         self._service.set_data(generate_set_data_requests())
 
+    #   pylint: disable=unused-argument
+    @all_deprecators(
+        deprecate_arg_mappings=[
+            {
+                "old_arg": "solution_variable_name",
+                "new_arg": "variable_name",
+                "converter": lambda old_arg_val: old_arg_val,
```
prmukherj:
```suggestion
```

src/ansys/fluent/core/services/solution_variables.py
```diff
@@ -343,6 +344,19 @@ def __call__(
             zone_names=zone_names, domain_name=domain_name
         ).solution_variables
 
+    #   pylint: disable=unused-argument
```
seanpearsonuk, line 347: @prmukherj 
I'm slightly recoiling on reading this. Let me break it down:
- In the "internal" code (as reviewed today) we see use of `deprecate_argument` in the same context which reads very intuitively. This is a bit more obscure. Does it need to be?
- `deprecated_reason` could be generated. Not generating that message means that each deprecation contains boilerplate that is larger than the non-boilerplate.
- Field names here are generally prefixed with `deprecate_`/`deprecated_` making the usage verbose (and inconsistent). That contrasts with the terseness and clarity of `deprecate_argument`.
- The structure is deeply nested, making it hard to read and hard to use.

Is the issue that `deprecated_version` is missing from `deprecate_argument`? Can we add it there and then use `deprecate_argument` more widely?

prmukherj: yes, that can be done. This is actually the decorator wrapper for documentation as well as deprecation warning.

prmukherj: Task-item added in backlog: https://github.com/ansys/pyfluent/issues/4462

seanpearsonuk (approved):

PR #4476, opened 2025-09-17
feat: Create UI for meshing workflow.

prmukherj:


src/ansys/fluent/core/ui/standalone_web_ui.py
```diff
@@ -269,7 +280,10 @@ def _loader(name=child_name, parent=obj, lvl=indent + 1):
             try:
                 child_obj = getattr(parent, name)
             except AttributeError:
-                child_obj = parent[name]
+                try:
+                    child_obj = parent[name]
+                except TypeError:
+                    child_obj = getattr(parent, name.split(":")[0])[name.split(":")[-1]]
```
Gobot1234, line 286:
```suggestion
                    attr, *_, key = name.split(":")
                    child_obj = getattr(parent, attr)[key]
```
I think is a bit better and less nested

src/ansys/fluent/core/session_meshing.py
Gobot1234, line 134: This shouldn't be required and it will mess with the type hints for the class, I've removed all of these in a separate PR

src/ansys/fluent/core/session_pure_meshing.pyi
Gobot1234, line 52: Would it be possible to add the type hint to the session_pure_meshing file here for a similar reason to my previous comment

PR #4477, opened 2025-09-17
docs: Update ReadME

prmukherj:
Update ReadME

Remove reference to parametric repo.
Update code to not show tui usage.


prmukherj:
Please provide any suggestions as required.
Thank you,

README.rst
```diff
@@ -125,21 +125,21 @@ To launch Fluent from Python, use the ``launch_fluent`` function:
 
 Basic usage
 ~~~~~~~~~~~
-You can use the ``solver_session.tui`` interface to run all Fluent TUI commands:
+The ``solver_session`` interface provides a convenient way to interact with Fluent:
 
 .. code:: python
```
seanpearsonuk, line 132: @prmukherj

I see the intention here of giving users a quick taste of PyFluent in the README, but I have a couple of concerns:

1. **Best practices**: The snippet shown here uses the older settings path rather than the newer settings classes (VelocityInlet, PressureOutlet, etc.). If this appears in the README it risks confusing new users and sending them down a path that isn’t aligned with where we want them to go. The README should showcase the canonical way to use PyFluent.
2. **README vs. User Guide**: The README’s job is to orient and point users to the docs. A minimal “hello world” style snippet to prove you can launch Fluent is fine, but longer uncommented examples blur the line with the user guide. It may be better to trim this down and direct users to the guide for full usage patterns.

My suggestion would be either:
- **Preferred**: keep a very minimal snippet (just launch Fluent), then link prominently to the user guide.
- Or, if we want to keep a usage example here, update it to reflect current best practice with the new API and add brief comments.

prmukherj: @seanpearsonuk, Thank you for the suggestion. I think it makes sense to go with the preferred suggestion. Will update accordingly.

Thanks.

prmukherj: Done, Thank you.

hpohekar (approved):

PR #4483, opened 2025-09-19
fix: chown server-info file by client user in compose mode

mkundu1:
This PR removes the `FLUENT_SERVER_INFO_PERMISSION_SYSTEM` env var for compose launch. The server writes the server-info file with the default 600 permission. The file is then owned by client-user via a subsequent scheme command, so it can be read in the client side.

### DIFF
--- a/src/ansys/fluent/core/docker/docker_compose.py
+++ b/src/ansys/fluent/core/docker/docker_compose.py
@@ -46,7 +46,9 @@ class ComposeBasedLauncher:
         self._container_source = self._set_compose_cmds()
         self._container_source.remove("compose")
 
-        container_dict["command"].append(f"""-command="(system \\\"chown {os.getuid()}:{os.getgid()} {container_dict["container_server_info_file"]}\\\")" """)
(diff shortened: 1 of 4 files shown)


mkundu1:
@hpohekar Please verify this from your side. I got the test `test_docker_compose` passing locally with docker compose. I haven't tested podman compose.

src/ansys/fluent/core/docker/docker_compose.py
```diff
@@ -45,6 +46,10 @@ def __init__(self, compose_config, container_dict):
         self._container_source = self._set_compose_cmds()
         self._container_source.remove("compose")
 
+        container_dict["command"].append(
+            f"""-command="(system \\\"chown {os.getuid()}:{os.getgid()} {container_dict["container_server_info_file"]}\\\")" """
+        )
```
mkundu1, line 51: This will send the scheme command `(system "chown <uid>:<gid> <server-info-file>")` to Fluent.

mkundu1:
Superseded by #4484

PR #4488, opened 2025-09-22
Fix misspelling of postprocess in examples

Gobot1234:


prmukherj (approved):

examples/00-fluent/exhaust_system_settings_api.py
```diff
@@ -569,7 +569,7 @@
 # Solve and postprocess
 # ---------------------
 # Once you have completed the fault tolerate meshing workflow, you can solve and
-# postprcess the results.
+# postprocess the results.
```
seanpearsonuk, line 572: Thanks @Gobot1234.

We supposedly have a safety net to prevent this but based on recent PRs, it isn't being applied effectively.

N.b. previous line has "the fault **tolerate** meshing workflow".

@prmukherj @mkundu1 @hpohekar @mayankansys

prmukherj: The issue is that vale is only part of our documentation, so it checks only the `doc` directory. Overall the codebase is checked by code-spell but it seems to ignore comments and docstrings although it is configured correctly only. I will try to figure out the exact issue and let everyone know.

seanpearsonuk: @Gobot1234 The above correction would be **fault-tolerant**

PR #4496, opened 2025-09-23
build: Update codespell check in pre-commit.

prmukherj:
Update codespell check in pre-commit to check for spelling issues in docstings and comments..


prmukherj:
Closing this as this needs additional investigation.

PR #4503, opened 2025-09-26
fix: Ansys Lab authentication

hpohekar:
Resolved Ansys Lab authentication issue.


src/ansys/fluent/core/launcher/pim_launcher.py
```diff
@@ -245,6 +248,16 @@ def launch_remote_fluent(
 
     instance.wait_for_ready()
 
+    file_service = PimFileTransferService(pim_instance=instance)
+
+    local_sifile = f"sifile_{uuid.uuid4().hex}"
+
+    file_service.download_file("sifile.txt", os.path.join(os.getcwd(), local_sifile))
+
+    time.sleep(10)
```
mkundu1: Can use a retry loop based on the exception beahviour instead of the hardcoded sleep. Something like below:
```
max_retries = ...
wait_time_between_retires = 1

for i in range(max_retries):
  try:
    <code>
    break
  except <observed exception(s)> as ex:
    if i == max_retries - 1:
      raise TimeoutError from ex
    sleep(wait_time_between_retires)
  
```

src/ansys/fluent/core/launcher/pim_launcher.py
```diff
@@ -245,6 +248,16 @@ def launch_remote_fluent(
...
+
+    ip, port, password = _parse_server_info_file(f"{local_sifile}/sifile.txt")
```
jerome-blanche: Can use a local variable for the local directory:


```suggestion
    local_directory = os.path.join(os.getcwd(), f"sifile_{uuid.uuid4().hex}")

    file_service.download_file("sifile.txt", local_directory)

    time.sleep(10)

    ip, port, password = _parse_server_info_file(os.path.join(local_directory, "sifile.txt"))
```

hpohekar: @jerome-blanche Thank you.

src/ansys/fluent/core/launcher/pim_launcher.py

src/ansys/fluent/core/launcher/pim_launcher.py
```diff
@@ -245,6 +254,30 @@ def launch_remote_fluent(
...
+                raise TimeoutError(
+                    "Failed to download file after multiple retries."
+                ) from ex
+            time.sleep(wait_time_between_retries)
+
+    ip, port, password = _parse_server_info_file(
+        os.path.join(local_directory, "sifile.txt")
+    )
+
+    shutil.rmtree(local_directory)
```
hpohekar: @jerome-blanche @mkundu1 

We are removing `local_directory` after extracting ip, port and password successfully, otherwise it would keep creating many unnecessary folders as follows:

image

Therefore users don't need to remove them manually.

jerome-blanche: And, security-wise, it's better to delete these sensitive data.

hpohekar: Yes, I agree.

seanpearsonuk:
```python
import tempfile
# ...
with tempfile.TemporaryDirectory(prefix="fluent_sifile_") as tmpdir:
        target_path = os.path.join(tmpdir, filename)
```
You don't need to remove it explicitly and you do want to release resources in an exception-safe manner.

src/ansys/fluent/core/launcher/pim_launcher.py
```diff
@@ -245,6 +254,30 @@ def launch_remote_fluent(
...
     instance.wait_for_ready()
 
+    file_service = PimFileTransferService(pim_instance=instance)
+
+    local_directory = os.path.join(os.getcwd(), f"sifile_{uuid.uuid4().hex}")
+
+    max_retries = 20
+    wait_time_between_retries = 1  # seconds
+
+    for i in range(max_retries):
```
seanpearsonuk: If this `range()` starts at 1... (`for attempt in range(1, max_retries + 1)`) then the following use of `max_retries` reads more easily. N.b. `i` -> `attempt` for overall readability.

src/ansys/fluent/core/launcher/pim_launcher.py
```diff
@@ -245,6 +254,30 @@ def launch_remote_fluent(
 
     instance.wait_for_ready()
 
+    file_service = PimFileTransferService(pim_instance=instance)
```
seanpearsonuk: This whole new chunk of code belongs in a separate function/method where all the currently hard-coded constants become arguments, potentially overridden by the user's configuration.

src/ansys/fluent/core/launcher/pim_launcher.py
```diff
@@ -245,6 +254,30 @@ def launch_remote_fluent(
...
+    wait_time_between_retries = 1  # seconds
+
+    for i in range(max_retries):
+        try:
+            file_service.download_file("sifile.txt", local_directory)
+            break
+        except Exception as ex:
+            if i == max_retries - 1:
+                raise TimeoutError(
+                    "Failed to download file after multiple retries."
```
seanpearsonuk: Since we know how the retries were configured at this point, we can (easily) add that information in the message for full transparency.

src/ansys/fluent/core/launcher/pim_launcher.py
```diff
@@ -245,6 +254,30 @@ def launch_remote_fluent(
...
+    local_directory = os.path.join(os.getcwd(), f"sifile_{uuid.uuid4().hex}")
+
+    max_retries = 20
+    wait_time_between_retries = 1  # seconds
+
+    for i in range(max_retries):
+        try:
+            file_service.download_file("sifile.txt", local_directory)
+            break
+        except Exception as ex:
```
seanpearsonuk: We can log each failure.

src/ansys/fluent/core/launcher/pim_launcher.py
hpohekar: @seanpearsonuk Yes, I agree. Done. Thank you.

src/ansys/fluent/core/launcher/pim_launcher.py

src/ansys/fluent/core/launcher/pim_launcher.py

src/ansys/fluent/core/launcher/pim_launcher.py

src/ansys/fluent/core/launcher/pim_launcher.py

seanpearsonuk (approved):

prmukherj (approved):

PR #4518, opened 2025-10-07
fix: Update exception type for scheme call

hpohekar:
**Description**

This change fixes the exception type caught in the get_cortex_connection_properties method of fluent_connection.py. Previously the code only caught grpc._channel._InactiveRpcError — but with the introduction of the GrpcErrorInterceptor, failures may now propagate as RuntimeError.

With this PR:

- The except _InactiveRpcError: branch is replaced by except RuntimeError: to correctly intercept failures from the new interceptor path.

- Logging and fallback behavior remain unchanged.

A changelog fragment has been added (in doc/changelog.d/4518.fixed.md) to capture this change for release notes.

**Motivation & Context**

This change is necessary because errors previously not being caught would bubble up and potentially break upstream logic. Ensuring the correct exception is caught maintains expected behavior and stability, especially when using the newer gRPC error interceptor mechanism.

**Impact**

There should be no user‑facing API change; behavior when connections fail is preserved.


src/ansys/fluent/core/fluent_connection.py
```diff
@@ -315,7 +314,7 @@ def get_cortex_connection_properties(self):
             cortex_pid = cortex_info.process_id
             cortex_pwd = cortex_info.working_directory
             logger.debug("Cortex connection properties successfully obtained.")
-        except _InactiveRpcError:
+        except RuntimeError:
```
prmukherj: 👍

prmukherj (approved):

src/ansys/fluent/core/fluent_connection.py
seanpearsonuk: Could you please add a brief code comment explaining why `RuntimeError` is the most appropriate exception type here?

The same rationale should also be included in the PR description to aid future understanding.

Thanks.

hpohekar: Done.

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -315,7 +314,7 @@ def get_cortex_connection_properties(self):
             cortex_pid = cortex_info.process_id
             cortex_pwd = cortex_info.working_directory
             logger.debug("Cortex connection properties successfully obtained.")
-        except _InactiveRpcError:
+        except RuntimeError:  # GrpcErrorInterceptor raises RuntimeError on failure
```
seanpearsonuk, line 317: @hpohekar Now that you have done that it also provides an LLM with some context to construct a stronger PR description from the git diffs. I would be interested to see the output of that.

hpohekar: Done.

PR #4530, opened 2025-10-08
fix: Meshing tests.

prmukherj:
## Context
Recent updates to the server-side datamodel-api and renaming of some entries in meshing FDL file have caused certain PyFluent tests to fail. These updates affected how some settings are exposed and validated through the API.

## Change Summary
This PR updates the affected PyFluent tests to align with the latest server-side datamodel changes and FDL renaming.

## Rationale
The failures were due to changes in the server-side behavior rather than issues in the test logic itself.

- The introduction of the "exposed_state" in the datamodel altered the activation state of some objects, making certain earlier test scenarios invalid (for example, tests expecting inactive objects to be accessible). Those tests have been removed or adjusted accordingly.

- The fixes were verified through manual debugging to ensure the tests now correctly reflect the new datamodel behavior and server API expectations.

## Impact
Only the affected test cases have been modified. No functional changes to PyFluent code or behavior.

## Final notes:
All tests are fixed and Nightly dev test is running fine in this branch and hence the main branch.

Only "test_ansys_units_integration" needs fixing that @hpohekar is looking into. It has been skipped as of now.

### DIFF
--- a/doc/changelog.d/4530.fixed.md
+++ b/doc/changelog.d/4530.fixed.md
@@ -1,0 +1,1 @@
+Meshing tests.
--- a/tests/test_datamodel_service.py
+++ b/tests/test_datamodel_service.py
@@ -387,11 +387,11 @@ def test_get_and_set_state_for_command_arg_instance(new_meshing_session):
 
     assert x.LengthUnit.get_state() == "ft"
 
-    assert x.CadImportOptions.ExtractFeatures()
+    assert not x.ImportCadPreferences.ShowImportCadPreferences()
 
-    x.CadImportOptions.ExtractFeatures.set_state(False)
+    x.ImportCadPreferences.ShowImportCadPreferences.set_state(True)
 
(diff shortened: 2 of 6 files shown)


hpohekar (approved):

seanpearsonuk:
@prmukherj I’m finding the PR description a bit hard to follow, particularly the Rationale section:

_Manually debugged the tests to put the fix, for example 'is_active' check has rendered some tests unusable now, so removed them, and some renaming in fdl has caused failures, hence fixed them._

It’s not clear how the 'is_active' check has rendered some tests unusable, especially since `is_active` isn’t referenced elsewhere in the description.

Also, this section reads more like a summary of what you did rather than an explanation of why those actions were necessary. The rationale should focus on the reasoning behind the changes such as what underlying issue was being addressed and why the chosen approach was appropriate.

Could you please revise this section to make that clearer?

Issue #3886, opened 2025-04-02
Garbled UI mode doc


seanpearsonuk:
```Python
>>> import ansys.fluent.core as pf
>>> help(pf.launch_fluent)  
...
 ui_mode : UIMode or str, optional
        Fluent user interface mode. Options are either the values of the ``UIMode``
        enum or any of ``"no_gui_or_graphics"``, ``"no_gui"``, ``"hidden_gui"``,
        ``"no_graphics"`` or ``"gui"``. The default is ``UIMode.HIDDEN_GUI`` in
        Windows and ``UIMode.NO_GUI`` in Linux. ``"no_gui_or_graphics"`` and
        ``"no_gui"`` user interface modes are supported in Windows starting from Fluent
        version 2024 R1.
...
```
It mixes strings and enums in the description which is confusing.

PR #4533, opened 2025-10-08
docs: Update ui_mode doc [skip tests]

hpohekar:
closes https://github.com/ansys/pyfluent/issues/3886

**Context**

Prior to this change, the ui_mode parameter in several parts of the PyFluent codebase had an incomplete or slightly misleading docstring. Users could pass either members of the UIMode enum or specific string literals (e.g. "no_gui", "hidden_gui", etc.), but that flexibility was not well documented. This ambiguity could lead to confusion about which string names are valid and how they map to enum values.

**Change Summary**

Updated the docstrings in launcher.py, slurm_launcher.py, and session_utilities.py to better describe accepted ui_mode values (both enum and string forms).

Clarified which string names correspond to which UIMode values (e.g. "no_gui_or_graphics", "no_gui", "hidden_gui", "gui", etc.).

Removed or refined outdated or redundant lines in the existing doc comments.

Ensured consistency across the codebase about how ui_mode is explained.

Marked the PR to skip tests, since only docstrings were changed.

**Rationale**

Better user experience / usability: Clear documentation helps users (and maintainers) know exactly what ui_mode values are supported.

Reduce confusion / errors: Users may try unsupported strings or misuse the enum; this aims to guide them.

Maintain consistency: Before, docstrings in different modules had slight variations or omissions. This change harmonizes those descriptions.

Low risk: Since only documentation is modified (no logic changes), this is a safe improvement.

**Impact**

Documentation / developer experience: Anyone reading or using code involving ui_mode will get more precise information on acceptable values.

Minimal to no runtime impact: Because no functional logic was changed, existing behavior remains unchanged.

Future maintenance: When extending or deprecating a ui_mode, future contributors can rely on the docstrings being consistent across modules.

Testing: No tests should break (and the PR is marked [skip tests]), but doc validation tools (if present) may pick up formatting or lint changes.

### DIFF
--- a/doc/changelog.d/4533.documentation.md
+++ b/doc/changelog.d/4533.documentation.md
@@ -1,0 +1,1 @@
+Update ui_mode doc [skip tests]


src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -239,13 +239,8 @@ def launch_fluent(
...
         and ``transcript.stop()`` on the session object.
-    ui_mode : UIMode or str, optional
-        Fluent user interface mode. Options are either the values of the ``UIMode``
-        enum or any of ``"no_gui_or_graphics"``, ``"no_gui"``, ``"hidden_gui"``,
-        ``"no_graphics"`` or ``"gui"``. The default is ``UIMode.HIDDEN_GUI`` in
-        Windows and ``UIMode.NO_GUI`` in Linux. ``"no_gui_or_graphics"`` and
-        ``"no_gui"`` user interface modes are supported in Windows starting from Fluent
-        version 2024 R1.
+    ui_mode : UIMode
+        Defines the user interface mode for Fluent. Options correspond to values in the ``UIMode`` enum.
```
seanpearsonuk: @hpohekar

Should we still mention that `ui_mode` also accepts `str` values? The type hint allows it, but the new docstring implies only `UIMode` is valid. If deprecation of string values isn’t intended, maybe we could clarify with something like:

```python
    ui_mode : UIMode or str, optional  
        Defines the user interface mode for Fluent. Accepts either a ``UIMode`` value  
        or a corresponding string such as ``"no_gui"``, ``"hidden_gui"``, or ``"gui"``.
```

or

```python
    ui_mode : UIMode or str, optional  
        Defines the user interface mode for Fluent. Accepts either a ``UIMode`` value  
        or a corresponding string value.
```

hpohekar: @seanpearsonuk Yes, I agree.

The following one looks good to me.

 ```
   ui_mode : UIMode or str, optional  
        Defines the user interface mode for Fluent. Accepts either a ``UIMode`` value  
        or a corresponding string such as ``"no_gui"``, ``"hidden_gui"``, or ``"gui"``.
```

PR #4537, opened 2025-10-08
chore: Add security action in CI.

prmukherj:
## Context
Added security actions to PyFluent CI

## Rationale
Added: ansys/actions/check-actions-security

## Impact
Only an extra step runs in CI, no code is affected

@RobPasMue, could you please comment if this is fine?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,7 +117,7 @@ jobs:
 
   check-vulnerabilities:
     name: "Check library vulnerabilities"
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
       - uses: ansys/actions/check-vulnerabilities@main
         with:
@@ -128,6 +128,17 @@ jobs:
           upload-reports: True
           hide-log: false
           bandit-configfile: "pyproject.toml"
+
+  actions-security:
+    name: "Actions Security"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: ansys/actions/check-actions-security@c2fa7c93f6883114e0e643599431b33d29f0b13f # v10.1.4
+        with:
+          generate-summary: true
+          token: ${{ secrets.GITHUB_TOKEN }}
+          auditing-level: 'high'
+
 
   docs_build:
     name: Build Documentation


.github/workflows/ci.yml
```diff
@@ -531,6 +554,7 @@ jobs:
         if: ${{ !contains(github.event.pull_request.title, '[skip tests]') }}
         run: make docker-pull
 
+      # zizmor: ignore[template-injection] - matrix.version is defined in workflow and not user input
```
RobPasMue: Even if this is the case @prmukherj - you should map the matrix.version value to an environment variable and use it inside the section. Something like:

```yml
      - name: Unit Testing
        if: ${{ !contains(github.event.pull_request.title, '[skip tests]') }}
        env:
            MATRIX_VERSION: ${{ matrix.version }}
        run: |
          make install-test
          make unittest-dev-${MATRIX_VERSION}
```

This is the recommendation we are providing everybody. This also helps standarizing the rest of the workflow file

.github/workflows/ci.yml
```diff
@@ -226,6 +242,7 @@ jobs:
         env:
           FLUENT_IMAGE_TAG: ${{ env.DOC_DEPLOYMENT_IMAGE_TAG }}
 
+      # zizmor: ignore[template-injection] - env.DOC_DEPLOYMENT_IMAGE_TAG is set internally in workflow and not user-controlled
```
RobPasMue: Same comment as before. You should do the following in that case (in this case the environment variable already exists in your context):

```yml
      - name: Zip HTML Documentation before upload
        run: |
          sudo apt install zip -y
          pushd doc/_build/html
          zip -r ../../../HTML-Documentation-tag-${DOC_DEPLOYMENT_IMAGE_TAG}.zip .
          popd
```

prmukherj: Done, Thank you.

.github/workflows/ci.yml
prmukherj: Done, Thank you.

.github/workflows/codeql.yml
```diff
@@ -22,7 +27,7 @@ jobs:
     permissions:
       actions: read
       contents: read
-      security-events: write
+      security-events: read
```
moe-ad: I think it is better to remove this permission block since it is the same as the workflow level permission block (which the job should inherit automatically).

PR #4560, opened 2025-10-16
ci: Fix notifications [skip tests]

hpohekar:
## Context
Notifications job is skipped every time which is not expected.

## Change Summary
We are sending notifications for both success and failure.

## Rationale
The condition job.status != 'success' will always be false when evaluated. The job.status is not available during step execution and only becomes available in post-job hooks.

## Impact
`Send GitHub Actions Run Status Adaptive Card to Teams via Power Automate` step only.

### DIFF
--- a/doc/changelog.d/4371.test.md
+++ b/doc/changelog.d/4371.test.md
@@ -1,0 +1,1 @@
+Add test to check PyFluent launched PyConsole by default.
--- a/tests/test_launcher.py
+++ b/tests/test_launcher.py
@@ -752,3 +752,10 @@ def test_warning_for_deprecated_compose_env_vars(monkeypatch):
     monkeypatch.setattr(pyfluent.config, "use_podman_compose", True)
     with pytest.warns(PyFluentDeprecationWarning):
         ComposeConfig()
+
+
+@pytest.mark.standalone
+@pytest.mark.fluent_version(">=25.1")
+def test_default_launch_mode_is_py():
+    fluent_launch_string, _ = pyfluent.launch_fluent(dry_run=True)
+    assert "-py" in fluent_launch_string


.github/workflows/test-podman-compose.yml
```diff
@@ -121,7 +121,7 @@ jobs:
           sudo podman system prune -af
 
       - name: Send GitHub Actions Run Status Adaptive Card to Teams via Power Automate
-        if: github.event_name == 'schedule' && job.status != 'success'
+        if: github.event_name == 'schedule' && always()
```
hpohekar: @mkundu1 Previous condition was evaluated as false always.

mkundu1: @hpohekar Can we still send only the failure notification, possibly by modifying the run body?

hpohekar: I investigated other ways but we need to access job status for this purpose therefore I don't think we can send only failure notification in this step.

mkundu1: @hpohekar I've created a draft PR with changes which would that here - https://github.com/ansys/pyfluent/pull/4561/files?diff=unified&w=1 

Feel free to merge it if the changes look good to you.

hpohekar: @mkundu1 Sure. Merging in this PR.

hpohekar: @mkundu1 Thanks very much. Merged in this PR.

.github/workflows/test-run-dev-version-nightly.yml
```diff
@@ -113,62 +113,62 @@ jobs:
         run: make docker-clean-images
 
       - name: Send GitHub Actions Run Status Adaptive Card to Teams via Power Automate
-        if: github.event_name == 'schedule' && job.status != 'success'
+        if: github.event_name == 'schedule' && always()
         run: | # zizmor: ignore[template-injection]
-          STATUS="Success"
           if [ "${{ job.status }}" != "success" ]; then
             STATUS="Failure"
```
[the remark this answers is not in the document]
mkundu1: @hpohekar If this is true, the current code won't still work.

I've created https://github.com/ansys/pyfluent/pull/4562 which will define the condition based on the testing job status.

.github/workflows/test-podman-compose.yml
mkundu1: No problem

seanpearsonuk (approved):

mkundu1 (approved):

PR #4565, opened 2025-10-21
test: Fix tests which are failing in the latest nightly [skip tests]

mkundu1:
I'm not able to reproduce tests/test_field_data.py::test_field_data_errors's failure (yet).

### DIFF
--- a/doc/changelog.d/4565.test.md
+++ b/doc/changelog.d/4565.test.md
@@ -1,0 +1,1 @@
+Fix tests which are failing in the latest nightly [skip tests]


tests/test_flobject.py
```diff
@@ -1223,17 +1223,17 @@ def test_default_argument_names_for_commands(static_mixer_settings_session):
     solver = static_mixer_settings_session
 
     if solver.get_fluent_version() >= FluentVersion.v251:
-        assert set(solver.results.graphics.contour.command_names) == {
-            "create",
-            "delete",
-            "rename",
-            "list",
-            "list_properties",
```
mkundu1: In 26.1, above 2 commands are replaced with base class methods

PR #4572, opened 2025-10-28
fix: 'wait_process_finished' in sessions.

prmukherj:
[1347769]

## Context
1. 'wait_process_finished' was not available after session exit.
2. Inconsistent behavior between solver and meshing session post exit.

## Change Summary
1. 'wait_process_finished' was made available after session exit.
2. 'dir()' and attribute access amde consisten for both solver and meshing session post exit.

## Impact
'wait_process_finished' can be accessed after session exit and post exit only 'is_active' and 'wait_process_finished' can be accessed from a session, both in solver and meshing mode.

### DIFF
--- a/tests/test_events_manager.py
+++ b/tests/test_events_manager.py
@@ -167,7 +167,6 @@ def test_multiple_register_callback_event(static_mixer_case_session, caplog):
         solver.events.unregister_callback(cb_id)
     solver.settings.solution.run_calculation.iterate(iter_count=5)
     assert len(event_index) == len(iteration_index)
-    solver.exit()
 
 
 @pytest.mark.fluent_version(">=23.1")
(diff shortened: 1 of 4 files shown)


src/ansys/fluent/core/session_solver.py
```diff
@@ -347,18 +347,31 @@ def set_state(self, state: StateT | None = None, **kwargs):
     def __call__(self):
         return self.get_state()
 
-    def __getattr__(self, name):
+    def __getattribute__(self, item: str):
```
seanpearsonuk, line 350: @prmukherj these methods block access to dunder attributes.

```python
    # Always allow Python internals
    if item.startswith("__") and item.endswith("__"):
        return super().__getattribute__(item)
```

PR #4573, opened 2025-10-29
fix: Update get_si_unit_for_fluent_quantity and from_container

hpohekar:
---

## Context

Before this change, certain utility functions in PyFluent — specifically `get_si_unit_for_fluent_quantity` (in `flunits.py`) and `from_container` (in `session_utilities.py`) — had issues handling units or container-based settings correctly. For example, `get_si_unit_for_fluent_quantity` needed better handling when the input `quantity` was `None`, a list, or an empty string. Also, `from_container` had parameters or logic that required adjustment for container-mode launches and file-transfer workflows. The existing behavior could lead to unexpected results when working with quantity units or launching sessions with containers.

---

## Change Summary

The PR makes the following changes:

* In `flunits.py`, the `get_si_unit_for_fluent_quantity` function has been updated to improve handling of various cases: when `quantity` is `None`, when it's a list (especially an empty list), when it is not a string, or when it is a string but empty.
* In `session_utilities.py`, the `from_container` function signature and/or its default parameters were adjusted (added parameters like `gpu: bool | None = None`, `file_transfer_service: Any | None = None`, `use_podman_compose: bool | None = None`). This expands the function’s ability to launch Fluent in container mode with more flexibility.
* A changelog entry (`doc/changelog.d/4573.fixed.md`) was added summarizing “Update get_si_unit_for_fluent_quantity and from_container”.
* Some tests were modified (in `tests/test_flobject.py`) to reflect updated behavior or remove outdated assertions.

---

## Rationale

* The `get_si_unit_for_fluent_quantity` change ensures that unit handling is robust and predictable even for edge cases: lists, `None`, empty strings. Since PyFluent integrates with the `ansys-units` library and supports quantity objects, it is important that units are determined correctly.
* The `from_container` changes are motivated by the need to better support containerised workflows of Fluent (e.g., using Docker or Podman) and enabling file transfer services for these sessions. Having additional parameters gives users more control (GPU usage, whether to use Docker Compose or Podman Compose, file transfer service) in launching Fluent via Python.
* Overall, the changes aim to reduce brittleness in unit-handling and broaden the support of containerised execution scenarios, improving reliability and usability of PyFluent in advanced workflows.

---

## Impact

* The functions `get_si_unit_for_fluent_quantity` and `from_container` are internal utilities used by the PyFluent core library (`ansys.fluent.core`). Workflows that involve retrieving or interpreting units of Fluent quantities will benefit (or may require slight adjustments if they relied on previous behavior).
* Session launch workflows, especially those using container mode (Docker or Podman) or using custom file-transfer services, will see enhanced capabilities and possibly changed interfaces (because `from_container` now has more parameters/defaults).
* Test and code-quality improvements: since the tests were adjusted, developers extending or modifying PyFluent should see fewer false positives/failures around unit and container-session utilities.
* Downstream code that introspects or depends on unit strings (via `units()` or `as_quantity()` on settings objects) may need to consider the changed handling of lists/None/empty strings in `get_si_unit_for_fluent_quantity`.
* Because these are utility changes rather than front-end user APIs, the impact on everyday usage of PyFluent (e.g., reading cases, solving) may be small for many users. But for advanced users leveraging units or container workflows, the improvements should lead to more predictable behavior.

---

### DIFF
--- a/doc/changelog.d/4573.fixed.md
+++ b/doc/changelog.d/4573.fixed.md
@@ -1,0 +1,1 @@
+Update get_si_unit_for_fluent_quantity and from_container


prmukherj (approved):

src/ansys/fluent/core/solver/flunits.py
```diff
@@ -296,10 +296,12 @@ def get_si_unit_for_fluent_quantity(
...
     # The settings API should return None for the units-quantity
     # attribute only for dimensionless variables
-    if quantity is None:
-        return ""
-    if isinstance(quantity, list):  # real vector
+    if isinstance(quantity, list):
+        if not quantity:
+            return ""
         quantity = quantity[0]
+    if not quantity:
```
mkundu1, line 299: `if not quantity` will capture `None`, empty list and empty string. So, we can consolidate as:
```
if not quantity:
  return ""
if isinstance(quantity, list):
  quantity = quantity[0]
if not isinstance(quantity, str):
  ...
```

hpohekar: @mkundu1 Done. Thank you.

mkundu1 (approved):

PR #4574, opened 2025-10-30
chore: Remove 'get_specs' from datamodel help.

prmukherj:
## Context
'get_specs' was used for generating the dynamic help.

## Change Summary
Dynamic help does not depend on 'get_specs' any longer.

## Rationale
Dynamic help is constructed via attribute query as required.

## Impact
Mostly no impact on the end users as the help text is expected to remain same.

PyCommand:

```python
Command: Execute
  Supported keyword arguments:
  Bool: Force
    Currently active: True
    Default value: False
  Return type: Logical
  
Example usage: 
  Execute(Force=False)
```

PyCommandArguments:

```python
Bool: Force
  Currently active: True
  Default value: False
```


src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1115,15 +1115,75 @@ def is_read_only(self) -> bool:
...
+            param_list = [
+                item
+                for item in inspect.getmembers_static(self)
+                if isinstance(item[1], PyParameter)
+            ]
+            for _, param in param_list:
+                help_lines.extend(
+                    ["  " + line for line in param._get_help_string().splitlines()]
+                )
+        elif isinstance(self, PyParameter):
```
seanpearsonuk, line 1139: Why not just override it?

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1115,15 +1115,75 @@ def is_read_only(self) -> bool:
         """Checks whether the object is read only."""
         return false_if_none(self.get_attr(Attribute.IS_READ_ONLY.value))
 
+    def _get_help_string(self) -> str:
+        help_lines = []
+
+        cls_name = self.__class__.__name__
+        is_task_object = cls_name == "_TaskObject"
+
+        if is_task_object:
```
seanpearsonuk, line 1124: Related to the `PyParameter` comment below, this should also be handled polymorphically, or at least externally (a type-based strategy).

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1115,15 +1115,75 @@ def is_read_only(self) -> bool:
...
+            ]
+            for _, param in param_list:
+                help_lines.extend(
+                    ["  " + line for line in param._get_help_string().splitlines()]
+                )
+        elif isinstance(self, PyParameter):
+            name = str(self).split()[0].split(".")[-1]
+            default_val = getattr(self, "default_value", None)
+            default = default_val() if callable(default_val) else None
+            if isinstance(self, PyTextual):
```
seanpearsonuk, line 1143: Each of these conditions can be handled via an override per type.

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1115,15 +1115,75 @@ def is_read_only(self) -> bool:
...
+                if isinstance(item[1], PyParameter)
+            ]
+            for _, param in param_list:
+                help_lines.extend(
+                    ["  " + line for line in param._get_help_string().splitlines()]
+                )
+        elif isinstance(self, PyParameter):
+            name = str(self).split()[0].split(".")[-1]
+            default_val = getattr(self, "default_value", None)
+            default = default_val() if callable(default_val) else None
```
seanpearsonuk, line 1142: I don't get what's going on here. Can we please have some comments?

prmukherj: ok, sure. If the default value returns something then we add it. Will add some inline comments.

prmukherj: In this case, "default_val" can either be the attr or None. In the next line, it checks if it is not None then return the value of "default_val". Will simplify the logic as well.

src/ansys/fluent/core/services/datamodel_se.py
prmukherj: yes, will do.

PR #4579, opened 2025-11-03
fix: example deprecation warnings and improvements

Gobot1234:
## Context
When solver session settings access was deprecated directly on solver the code wasn't updated I've also sprinkled in a couple of small improvements to the examples 

## Change Summary
Updated all references direct access patterns (`solver.setup.models`) to the correct settings-based access pattern (`solver.settings.setup.models`) along with adding PEP 723 styles dependencies to the examples to make them easier to run as standalone scripts.

## Rationale
Fixes all the deprecation warnings

## Impact
Changes basically everywhere

### DIFF
--- a/examples/00-fluent/conjugate_heat_transfer.py
+++ b/examples/00-fluent/conjugate_heat_transfer.py
@@ -494,7 +494,9 @@ solver_session.settings.mesh.check()
 mesh1 = Mesh(solver=solver_session)
 
 wall_list = [
-    boundary_name for boundary_name in mesh1.surfaces.allowed_values if "wall" in boundary_name
+    boundary_name
+    for boundary_name in mesh1.surfaces.allowed_values
+    if "wall" in boundary_name
 ]
 
 #############################################################################


doc/source/cheatsheet/cheat_sheet.qmd
```diff
@@ -767,7 +767,7 @@ zone_info.zone_type
 
 ```{python}
 wall_fluid_info = solution_variable_info.get_variables_info(
-    zone_names=['wall' , "fluid"], 
+    zone_names=['wall' , "fluid"],
```
seanpearsonuk:
```suggestion
    zone_names=['wall', "fluid"],
```

doc/source/user_guide/session/launching_ansys_fluent.rst
```diff
@@ -3,12 +3,12 @@
...
+It covers multiple methods to start Fluent, including launching from a local installation, containerized environments (Docker or Podman), and connecting
 to existing Fluent sessions. Detailed examples illustrate how to initialize Fluent in various modes such as meshing, solution, and pre/post processing.
 
-Additionally, the guide explains advanced launching options such as setting precision, dimensions, and parallel execution configurations. 
-It also covers integration with job schedulers like Slurm, enabling efficient high-performance computing workflows. 
+Additionally, the guide explains advanced launching options such as setting precision, dimensions, and parallel execution configurations.
+It also covers integration with job schedulers like Slurm, enabling efficient high-performance computing workflows.
 Furthermore, instructions for launching Fluent within PIM environments and detailed procedures for cross-platform remote connections and file transfers between Windows, Linux, and WSL hosts are provided.
 
 This reference aims to equip users with the knowledge and tools required to flexibly and efficiently manage Fluent sessions in diverse computing environments.
```
seanpearsonuk: This section is bloated with redundant information and phrasing. Here's an alternative:
```text
This guide explains how to start and connect to Ansys Fluent sessions using PyFluent. It covers common workflows such as launching Fluent locally, running in containerized environments (Docker or Podman), or attaching to an existing session. Examples are provided for initializing Fluent in meshing, solution, and post-processing modes.

Advanced launch options, such as selecting precision, dimensionality, and parallel settings, are also described. The document includes guidance for running Fluent under job schedulers like Slurm, using PyFluent in PIM environments, and configuring remote connections across Windows, Linux, and WSL systems.

Overall, this reference aims to give users a clear, unified set of instructions for managing Fluent sessions in a wide range of computing environments.
```

doc/source/user_guide/session/launching_ansys_fluent.rst
```diff
@@ -429,9 +429,9 @@ using the ``additional_arguments`` parameter for distributed parallel processing
 
 Launching a `PIM <https://pypim.docs.pyansys.com/version/stable/>`_ session
 ---------------------------------------------------------------------------
-When PyFluent is used within a `PIM <https://pypim.docs.pyansys.com/version/stable/>`_ configured environment, 
-the :func:`launch_fluent() <ansys.fluent.core.launcher.launcher.launch_fluent>` function automatically launches 
-Fluent session in `PIM <https://pypim.docs.pyansys.com/version/stable/>`_ mode and in that same environment it 
+When PyFluent is used within a `PIM <https://pypim.docs.pyansys.com/version/stable/>`_ configured environment,
+the :func:`launch_fluent() <ansys.fluent.core.launcher.launcher.launch_fluent>` function automatically launches
+Fluent session in `PIM <https://pypim.docs.pyansys.com/version/stable/>`_ mode and in that same environment it
```
seanpearsonuk: - Too many grammatical errors have slipped through in the past. We can't keep doing this and we have to make sure that the English we produce is correct. Anyway, we can just say _launches Fluent_ instead because _the/a Fluent session_ is a bit verbose here.
- Continuing with _and in that same environment_ loads too much detail into a sigle sentence. It doesn't read well and we should break that sentence up.

doc/source/user_guide/session/launching_ansys_fluent.rst
```diff
@@ -523,7 +523,7 @@ Connecting to a Fluent container running inside Linux from a Windows host
 Connecting to Fluent on Windows from a Linux or WSL host
 --------------------------------------------------------
 
-This guide describes how to connect to an ANSYS Fluent instance running on a Windows machine from a Linux or WSL host. 
+This guide describes how to connect to an ANSYS Fluent instance running on a Windows machine from a Linux or WSL host.
```
seanpearsonuk:
```suggestion
This section outlines how to connect to Fluent running on Windows from a Linux or WSL host.
```

doc/source/user_guide/session/launching_ansys_fluent.rst
```diff
@@ -589,16 +589,16 @@ Run the following Python code to connect to Fluent and transfer files:
...
    # `elbow_remote.cas.h5` will be downloaded to local working directory
-   solver_session.file.write_case(file_name="elbow_remote.cas.h5")
+   solver_session.settings.file.write_case(file_name="elbow_remote.cas.h5")
 
 
 Connecting to Fluent on Linux or WSL from a Windows host
 --------------------------------------------------------
 
-This guide describes how to connect to an ANSYS Fluent instance running on a Linux or WSL machine from a Windows host. 
+This guide describes how to connect to an ANSYS Fluent instance running on a Linux or WSL machine from a Windows host.
```
seanpearsonuk:
```suggestion
This section outlines how to connect to an Fluent running on Linux or WSL from a Windows host.
```

doc/source/user_guide/solver_settings/solver_settings_contents.rst
```diff
@@ -32,15 +32,15 @@ Example usage
...
+This format provides a more natural way to create and interact with settings objects,
+making your code easier to read and maintain. By abstracting the underlying hierarchy,
+users can focus on the specific settings they need without dealing with potential changes
 in the Fluent API structure.
 
 Accessing solver settings
 -------------------------
 
-Following the introduction of the new format, the traditional method remains available for those 
+Following the introduction of the new format, the traditional method remains available for those
```
seanpearsonuk: This sentence is out of place and lacks context:

>“Following the introduction of the new format, the traditional method remains available for those who prefer the existing hierarchy.”

It doesn’t belong at the start of this section and should instead be integrated into the previous section—after introducing the new-style API—so that the relationship between the new and existing APIs is clear and non-redundant. The goal is simply to mention that the existing hierarchy-based API remains fully supported and is not deprecated.

Below are suggested rewrites for both sections.

### Suggested rewrite for the previous section (new-style API introduction)

To simplify access to solver settings and improve readability, you can now instantiate settings objects directly using a concise constructor-style syntax. This avoids navigating the full hierarchy of solver settings while preserving the same functional capabilities.

`<Example>`

The traditional hierarchy-based API remains fully supported. The new syntax is an additional, more convenient option and not a replacement.

### Suggested rewrite for the Accessing solver settings section

The hierarchy-based settings API continues to be available for users who prefer explicit navigation of the settings structure. For example, when working with string lists that support wildcard expansion:

`<Code>`

examples/00-fluent/DOE_ML.py
seanpearsonuk, line 112: The comments in this example are unclear and not especially helpful in their current form. It’s difficult to tell whether they were intended as internal notes or actual user‑facing guidance. As written, they read more like personal reminders than documentation.

If this example is meant to be instructive, the comments should be rewritten into a concise narrative explaining:

- what the DOE represents,
- why the arrays are defined this way,
- what the sweep or workflow will do with them,
- and what `resArr` is intended to store.

Alternatively, if the example is only meant to show variable initialization, it may be clearer to remove the commented headings entirely and introduce the example in the surrounding text instead.

Below is an example of how this could be presented more clearly.

---

## Suggested Rewrite of the DOE Example

```python
# Example: Manual Design of Experiments (DOE)

# Specify inlet velocities for the cold and hot streams.
# Each pair of (coldVel, hotVel) will be used to run one Fluent case.
coldVelArr = np.array([0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7])
hotVelArr = np.array([0.8, 1.0, 1.2, 1.4, 1.6, 1.8, 2.0])

# Allocate a results array. Entry (i, j) will hold the Mass‑Weighted
# Average Temperature at the outlet for coldVelArr[i] and hotVelArr[j].
resArr = np.zeros((coldVelArr.size, hotVelArr.size))
```

This keeps the example compact while giving readers enough information to understand the purpose of the arrays and how they relate to the DOE workflow.

seanpearsonuk: We shouldn't have any commented out imports.

doc/source/user_guide/session/launching_ansys_fluent.rst
Gobot1234: I've used this but also linked to the appropriate section

examples/00-fluent/DOE_ML.py
Gobot1234: I've added a note that these are optional to decide which model you want to use but the pprint one is gone

examples/00-fluent/conjugate_heat_transfer.py
```diff
@@ -489,13 +497,13 @@
...
-        periodic_list.append(item)
-    if len(item.split("symmetry")) > 1:
-        symmetry_list.append(item)
+for boundary_name in mesh1.surfaces.allowed_values:
+    if "wall" in boundary_name:
+        wall_list.append(boundary_name)
+    if "periodic" in boundary_name:
+        periodic_list.append(boundary_name)
+    if "symmetry" in boundary_name:
+        symmetry_list.append(boundary_name)
```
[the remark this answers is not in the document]
Gobot1234: This is just wrong

examples/00-fluent/conjugate_heat_transfer.py
```diff
@@ -477,7 +485,7 @@
 # Mesh Check; Review Fluent transcript for errors
 # ===============================================
 
-solver_session.mesh.check()
+solver_session.settings.mesh.check()
 
 #############################################################################
 # Create a few boundary list for display and post-processing
```
seanpearsonuk: This comment has incorrect grammar. We could update it to "Create boundary lists for display and post-processing"

examples/00-fluent/conjugate_heat_transfer.py
```diff
@@ -477,7 +485,7 @@
 # Mesh Check; Review Fluent transcript for errors
 # ===============================================
 
-solver_session.mesh.check()
+solver_session.settings.mesh.check()
```
seanpearsonuk, line 488: `mesh.check()` is not useful programmatically and there is no useful alternative. Raising issue. (We can mark this comment resolved.)

examples/00-fluent/conjugate_heat_transfer.py
seanpearsonuk: **Only** `wall_list` is actually used in the script. We could have a simple list comprehension for `wall_list` here. Making this code smaller is good because if there is a lot of code like this, the user could easily be distracted and fooled into thinking that the code is significant and highly generalisable.

PR #4581, opened 2025-11-04
chore: Update text for 'BetaFeaturesNotEnabled' exception.

prmukherj:
## Context
The exception message was not clear. The message was slightly inaccurate, as it suggested that beta features can only be enabled via a Python call.

## Change Summary
The exception message was updated to be more precise about the need of the python call.

## Impact
Users will see the updated error message when 'BetaFeaturesNotEnabled' is raised.

### DIFF
--- a/src/ansys/fluent/core/exceptions.py
+++ b/src/ansys/fluent/core/exceptions.py
@@ -57,7 +57,7 @@ class BetaFeaturesNotEnabled(RuntimeError):
 
     def __init__(self, feature_name: str | None = None) -> None:
         base_message = (
-            " is a Fluent beta feature. To enable them from Python "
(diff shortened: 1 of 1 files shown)


src/ansys/fluent/core/exceptions.py
```diff
@@ -56,9 +56,13 @@ class BetaFeaturesNotEnabled(RuntimeError):
     """Raised when a beta feature is accessed without enabling beta features."""
 
     def __init__(self, feature_name: str | None = None) -> None:
+        base_message = (
+            " is a Fluent beta feature. To enable them from Python "
+            "call '<session>.enable_beta_features()' "
```
prmukherj:
```suggestion
            "call '<session>.enable_beta_features()'."
```

PR #4588, opened 2025-11-06
chore: Update PyQuery.

prmukherj:
## Context
PyQuery in datamodel service was missing complete implementation.

## Change Summary
PyQuery is implemented in a similar manner to PyCommand.

## Rationale
Both the generated classes and back-end code was affected.

## Impact
Users will be able to use sub arguments in a query as objects. Please refer tests.

### DIFF
--- a/doc/changelog.d/4588.maintenance.md
+++ b/doc/changelog.d/4588.maintenance.md
@@ -1,0 +1,1 @@
+Update PyQuery.


src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -2081,26 +2112,26 @@ def __setattr__(self, key, value):
             getattr(self, key).set_state(value)
 
 
-class PyCommandArguments(PyStateContainer):
-    """Class representing command arguments in datamodel."""
+class _PyCommandQueryArguments(PyStateContainer):
```
prmukherj, line 2115: PyArguments

PR #4590, opened 2025-11-06
chore: Rename and clean up internal datamodel service classes.

prmukherj:
## Context
Some internal classes name were strictly bound to Commands where they could also be used for Query. We rename and make them usable by both.

## Change Summary
Renamed all internal datamodel classes so that they are generic.
'help' is removed which was earlier providing dynamic help for legacy datamodel objects.

## Impact
No user side impact is anticipated.

Key Changes:

1. Renamed PyCommandArguments to PyArguments and all related CommandArguments classes to just Arguments
2. Extended PyQuery to support argument instances similar to PyCommand
3. Removed duplicate help() methods from multiple classes

### DIFF
--- a/src/ansys/fluent/core/services/datamodel_se.py
+++ b/src/ansys/fluent/core/services/datamodel_se.py
@@ -1859,13 +1859,13 @@ class PyAction:
 
 
 class PyQuery(PyAction):
-    """Class representing query in datamodel."""
+    """Enables querying Fluent’s data model through a simple Python interface."""
 
     _operation = "query"
 
 
 class PyCommand(PyAction):
-    """Class representing command in datamodel."""
+    """Enables commanding Fluent’s data model through a simple Python interface."""
 
     _operation = "command"
 


src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1808,84 +1798,91 @@ def get_state(self):
     getState = __call__ = get_state
 
 
-class PyQuery:
-    """Query class using the StateEngine-based DatamodelService as the backend. Use this
-    class instead of directly calling the DatamodelService's method.
+class _PyAction:
```
seanpearsonuk: Since this is a base class of other public classes, it is also public class. The leading underscore should be omitted.

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1808,84 +1798,91 @@ def get_state(self):
...
+        """__init__ method of _PyAction class."""
         self.service = service
         self.rules = rules
-        self.query = query
-        if path is None:
-            self.path = []
-        else:
-            self.path = path
+        setattr(self, self._operation, name)
+        self.path = [] if path is None else path
```
seanpearsonuk: I realise that this is a moot point and the two options are not strictly equivalent but I prefer `self.path = path or []`.

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1808,84 +1798,91 @@ def get_state(self):
...
+                getattr(self, self._operation),
+                self.path.copy(),
+                id,
+            ]
+        except (RuntimeError, ValueError) as e:
+            logger.warning(
+                f"datamodels_se.{self.__class__.__name__} "
+                f"was unable to construct {self._operation} arguments. "
+                "This may be due to gRPC issues or unsupported Fluent version "
+                "(23.1+ required). Error details: %s",
```
seanpearsonuk: The error message is too speculative IMHO. More importantly, these classes are gRPC-independent and are already used in a non-gRPC setting.

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1808,84 +1798,91 @@ def get_state(self):
...
+        """Create an operation instance."""
+        args = self._get_create_instance_args()
+        if args is not None:
+            return PyArguments(*args)
+
+
+class PyQuery(_PyAction):
+    """Query class using the StateEngine-based DatamodelService as the backend. Use this
+    class instead of directly calling the DatamodelService's method."""
+
```
seanpearsonuk, line 1863: I notice that we have other similar docstrings elsewhere in this module. The docstring is too heavily into unimportant implementation details.

The style should be more user-centric; e.g., "Enables querying Fluent’s data model through a simple Python interface."

src/ansys/fluent/core/services/datamodel_se.py
prmukherj: I agree @seanpearsonuk, will update the error message. I was confused since this was an existing message.

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -2229,18 +2178,14 @@ def __init__(
...
         ["Real", "Int", "ListReal", "Real List", "Integer", "ListInt", "Integer List"],
-        PyNumericalCommandArgumentsSubItem,
-    ),
-    "Dict": PyDictionaryCommandArgumentsSubItem,
-    **dict.fromkeys(
-        ["Bool", "Logical", "Logical List"], PyParameterCommandArgumentsSubItem
+        PyNumericalArgumentsSubItem,
     ),
-    "ModelObject": PySingletonCommandArgumentsSubItem,
+    "Dict": PyDictionaryArgumentsSubItem,
```
seanpearsonuk: Since we're renaming things here, it seems timely to bring this up. All classes with names like `Py<Type>ArgumentsSubItem` are misnamed since `<Type>` describes `SubItem`rather than `Arguments`. Thus `PyArguments<Type>SubItem`; e.g., `PyArgumentsTextualSubItem` rather than `PyTextualArgumentsSubItem`.

src/ansys/fluent/core/services/datamodel_se.py
prmukherj: Done, Thank you.

src/ansys/fluent/core/services/datamodel_se.py
prmukherj: Done, Thank you.

src/ansys/fluent/core/services/datamodel_se.py
prmukherj: How does this look now? @seanpearsonuk

src/ansys/fluent/core/services/datamodel_se.py
prmukherj: I agree @seanpearsonuk. Will update it.

prmukherj: Done.

PR #4600, opened 2025-11-12
feat: Update client side 'enhanced' meshing workflow to use server side 'meshing_workflow' root.

prmukherj:
## Context
Update the client side enhanced meshing workflow with respect to the server side updates in Fluent 26R1.

## Change Summary
Simple wrappers to expose server side features with some additional improvement like the insertion and traversal mechanisms.

## Impact
The users to experience an enhanced workflow experience.

Please refer tests for updated usages.

TODO: The user facing docs needs updates after Fluent 26R1 is released.

### DIFF
--- a/src/ansys/fluent/core/workflow_new.py
+++ b/src/ansys/fluent/core/workflow_new.py
@@ -600,12 +600,10 @@ class TaskObject:
 
     def _get_next_possible_tasks(self):
         """Get display names of tasks that can be inserted after this task."""
-        task_obj = super().__getattribute__("_task_object")
+        task_obj = self._task_object
         ret_list = []
         for item in task_obj.get_next_possible_tasks():
-            snake_case_name = command_name_to_task_name(
-                super().__getattribute__("_meshing_root"), item
-            )
+            snake_case_name = command_name_to_task_name(self._meshing_root, item)
             if snake_case_name != item:
                 self._cache[snake_case_name] = item
             ret_list.append(snake_case_name)
@@ -618,10 +616,9 @@ class TaskObject:
         -----
         Internal method. Users should use `insertable_tasks.<task>.insert()` instead.
         """
-        task_obj = super().__getattribute__("_task_object")
-        self.get_next_possible_tasks()
+        self._get_next_possible_tasks()
         command_name = self._cache.get(task_name) or task_name
-        task_obj.insert_next_task(command_name=command_name)
+        self._task_object.insert_next_task(command_name=command_name)
 
     @property
     def insertable_tasks(self):
@@ -739,7 +736,7 @@ class TaskObject:
         -----
         Arguments take precedence over task object properties.
         """
-        task_obj = super().__getattribute__("_task_object")
+        task_obj = self._task_object
         args = task_obj.arguments
         if item in args():
             return getattr(args, item)
@@ -747,8 +744,7 @@ class TaskObject:
 
     def __setattr__(self, key, value):
         """Enable attribute assignment to task arguments."""
-        task_obj = super().__getattribute__("_task_object")
-        args = task_obj.arguments
+        args = self._task_object.arguments
         if hasattr(args, key):
             setattr(args, key, value)
         else:
@@ -756,15 +752,14 @@ class TaskObject:
 
     def __call__(self):
         """Execute the task when called as a function."""
-        task_obj = super().__getattribute__("_task_object")
-        return task_obj.execute()
+        return self._task_object.execute()
 
     def __getitem__(self, key):
-        task_obj = super().__getattribute__("_task_object")
-        name = super().__getattribute__("_name")
-        workflow = super().__getattribute__("_workflow")
-        parent = super().__getattribute__("_parent")
-        meshing_root = super().__getattribute__("_meshing_root")
+        task_obj = self._task_object
+        name = self._name
+        workflow = self._workflow
+        parent = self._parent
+        meshing_root = self._meshing_root
         name_1 = name
         name_2 = re.sub(r"\s+\d+$", "", task_obj.name().strip()) + f" {key}"
         try:
@@ -800,12 +795,9 @@ class TaskObject:
 
     def _task_names(self):
         """Gets the display names of the child tasks of a task item."""
-        task_obj = super().__getattribute__("_task_object")
-        task_list = task_obj.task_list()
+        task_list = self._task_object.task_list()
         if task_list:
-            return _convert_task_list_to_display_names(
-                super().__getattribute__("_workflow"), task_list
-            )
+            return _convert_task_list_to_display_names(self._workflow, task_list)
         else:
             return []
 
@@ -821,8 +813,7 @@ class TaskObject:
         if not child_names:
             return []
 
-        workflow = super().__getattribute__("_workflow")
-        meshing_root = super().__getattribute__("_meshing_root")
+        workflow = self._workflow
 
         # Create reverse lookup: display name -> task type
         name_to_type = {
@@ -842,7 +833,7 @@ class TaskObject:
                     task_type,
                     workflow,
                     self,
-                    meshing_root,
+                    self._meshing_root,
                 )
                 wrapped_children.append(wrapped)
 
@@ -878,7 +869,7 @@ class TaskObject:
             first_name = task_list[0]
         else:
             return None
-        workflow = super().__getattribute__("_workflow")
+        workflow = self._workflow
 
         type_to_name = {
             item.split(":")[0]: item.split(":")[-1] for item in workflow.task_object()
@@ -890,7 +881,7 @@ class TaskObject:
                     key,
                     workflow,
                     self,
-                    super().__getattribute__("_meshing_root"),
+                    self._meshing_root,
                 )
 
     def last_child(self):
@@ -913,7 +904,7 @@ class TaskObject:
             last_name = task_list[-1]
         else:
             return None
-        workflow = super().__getattribute__("_workflow")
+        workflow = self._workflow
 
         type_to_name = {
             item.split(":")[0]: item.split(":")[-1] for item in workflow.task_object()
@@ -925,7 +916,7 @@ class TaskObject:
                     key,
                     workflow,
                     self,
-                    super().__getattribute__("_meshing_root"),
+                    self._meshing_root,
                 )
 
     @staticmethod
@@ -1006,8 +997,7 @@ class TaskObject:
             - Workflow instance for top-level tasks
             - TaskObject instance for nested child tasks
         """
-        parent = super().__getattribute__("_parent")
-        return parent
+        return self._parent
 
     def has_next(self) -> bool:
         """Check if there is a next sibling task.
@@ -1020,8 +1010,7 @@ class TaskObject:
         bool
             True if a next sibling exists, False if this is the last task.
         """
-        parent = super().__getattribute__("_parent")
-        task_dict = parent._ordered_tasks()
+        task_dict = self._parent._ordered_tasks()
         try:
             self._get_next_key(task_dict, self.name())
             return True
@@ -1030,8 +1019,7 @@ class TaskObject:
 
     def next(self):
         """Returns the next sibling task item."""
-        parent = super().__getattribute__("_parent")
-        task_dict = parent._ordered_tasks()
+        task_dict = self._parent._ordered_tasks()
         next_key = self._get_next_key(task_dict, self.name())
         return task_dict[next_key]
 
@@ -1046,8 +1034,7 @@ class TaskObject:
         bool
             True if a previous sibling exists, False if this is the first task.
         """
-        parent = super().__getattribute__("_parent")
-        task_dict = parent._ordered_tasks()
+        task_dict = self._parent._ordered_tasks()
         try:
             self._get_previous_key(task_dict, self.name())
             return True
@@ -1056,8 +1043,7 @@ class TaskObject:
 
     def previous(self):
         """Returns the previous sibling task item."""
-        parent = super().__getattribute__("_parent")
-        task_dict = parent._ordered_tasks()
+        task_dict = self._parent._ordered_tasks()
         previous_key = self._get_previous_key(task_dict, self.name())
         return task_dict[previous_key]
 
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/workflow_new.py
```diff
@@ -0,0 +1,1024 @@
...
+            try:
+                return make_task_wrapper(
+                    getattr(workflow.task_object, name_1)[key],
+                    name_1,
+                    workflow,
+                    temp_parent,
+                )
+            except LookupError as ex2:
+                raise LookupError(
+                    f"Neither '{name_2}' nor '{key}' not found in task object '{name_1}'."
```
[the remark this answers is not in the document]
prmukherj: Will do.

src/ansys/fluent/core/session_base_meshing.py
```diff
@@ -138,8 +138,11 @@ def meshing_workflow(self):
 
     def watertight_workflow(self, initialize: bool = True):
         """Datamodel root of workflow."""
+        root_module = (
+            "meshing_workflow" if os.getenv("USE_SERVER_MW") == "1" else "workflow"
+        )
```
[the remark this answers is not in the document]
prmukherj: Will do.

src/ansys/fluent/core/workflow_new.py
```diff
@@ -0,0 +1,1024 @@
...
+        task_list = self._workflow.general.workflow.task_list()
+        if task_list:
+            last_name = _get_child_task_by_task_id(self._workflow, task_list[-1])
+        else:
+            return None
+        for task_obj in self.tasks():
+            if task_obj.name() == last_name:
+                return make_task_wrapper(
+                    task_obj,
+                    task_obj.__class__.__name__.lstrip("_"),
```
seanpearsonuk: This line needs a comment.

src/ansys/fluent/core/workflow_new.py
```diff
@@ -0,0 +1,1024 @@
...
+
+    def _ordered_tasks(self):
+        """Get ordered dictionary mapping task names to task objects."""
+        ordered_names = _convert_task_list_to_display_names(
+            self._workflow,
+            self._workflow.general.workflow.task_list(),
+        )
+        name_to_task = {
+            task_obj.name(): make_task_wrapper(
+                task_obj, task_obj.__class__.__name__.lstrip("_"), self._workflow, self
```
seanpearsonuk: `task_obj.__class__.__name__.lstrip("_")` needs a comment

src/ansys/fluent/core/workflow_new.py
```diff
@@ -0,0 +1,1024 @@
...
+        - Dynamically created tasks
+
+        The method rebuilds its internal task cache on each call to ensure
+        freshness, though this can be expensive for large workflows.
+        """
+        self._task_dict = {}
+        _state = self._workflow.task_object()
+        for task in sorted(_state):
+            name = task.split(":")[0]
+            display_name = task.split(":")[-1]
```
seanpearsonuk: This is a one-liner `x, y = split(...`.

src/ansys/fluent/core/workflow_new.py
```diff
@@ -0,0 +1,1024 @@
...
+            if task_obj.task_type() == "Compound Child":
+                if name not in self._compound_child_dict:
+                    self._compound_child_dict[name] = {
+                        name + "_child_1": task_obj,
+                    }
+                else:
+                    _name_list = []
+                    for key, value in self._compound_child_dict[name].items():
+                        _name_list.append(value._name_())
+                    if task_obj._name_() not in _name_list:
```
seanpearsonuk: This can be simplified by matching the name against an inline generator over `x.values()`.

src/ansys/fluent/core/workflow_new.py
```diff
@@ -0,0 +1,1024 @@
...
+                        child_key = (
+                            int(sorted(self._compound_child_dict[name])[-1][-1]) + 1
```
seanpearsonuk, line 266: This is hard to follow

src/ansys/fluent/core/workflow_new.py
```diff
@@ -0,0 +1,1024 @@
...
+        fluent_version: FluentVersion,
+    ) -> None:
+        """Initialize Workflow."""
+        self._workflow = workflow
+        self._command_source = command_source
+        self._fluent_version = fluent_version
+        self._task_dict = {}
+        self._compound_child_dict = {}
+
+    def tasks(self) -> list:
```
seanpearsonuk: Does this need to return a concrete list, or would any collection of Task be acceptable? If not, the copy `list(self._task_dict.values())` seems unnecessary. Also, the return type should specify the element type of the collection.

src/ansys/fluent/core/workflow_new.py
```diff
@@ -0,0 +1,1024 @@
...
+    Notes
+    -----
+    This is a workaround for Fluent 26R1. It attempts to load version-specific
+    datamodel files to perform the conversion, falling back to a generic approach if unavailable.
+    """
+    # TODO: This is a fix only for 26R1 as the server lacks the mechanism to return mapped values
+    #  for '<Task Object>.get_next_possible_tasks()'.
+    try:
+        module = load_module(
+            "meshing_261", config.codegen_outdir / "datamodel_261" / "meshing.py"
```
seanpearsonuk: I have two suggestions here.
1. Pass in all dependencies from above; those are the `meshing_workflow` and `meshing` data model instances.
2. Access the `meshing` data model instance in the usual way rather than breaking encapsulation by mentioning `codegen_outdir` etc.

src/ansys/fluent/core/workflow_new.py
```diff
@@ -0,0 +1,1024 @@
...
+        The method rebuilds its internal task cache on each call to ensure
+        freshness, though this can be expensive for large workflows.
+        """
+        self._task_dict = {}
+        _state = self._workflow.task_object()
+        for task in sorted(_state):
+            name = task.split(":")[0]
+            display_name = task.split(":")[-1]
+            task_obj = getattr(self._workflow.task_object, name)[display_name]
+            if task_obj.task_type() == "Compound Child":
```
seanpearsonuk: Just noting that in the current code, we only mention these strings once, like this:
```python
    kinds = {
        "Simple": SimpleTask,
        "Compound Child": CompoundChild,
        "Compound": CompoundTask,
        "Composite": CompositeTask,
        "Conditional": ConditionalTask,
    }
```

src/ansys/fluent/core/workflow_new.py
```diff
@@ -0,0 +1,1024 @@
...
+                            int(sorted(self._compound_child_dict[name])[-1][-1]) + 1
+                        )
+                        self._compound_child_dict[name][
+                            name + f"_child_{child_key}"
+                        ] = task_obj
+            else:
+                if name not in self._task_dict:
+                    self._task_dict[name] = task_obj
+                else:
+                    self._task_dict[name + f"_{task_obj.name().split()[-1]}"] = task_obj
```
seanpearsonuk, line 296: This is hard to follow

src/ansys/fluent/core/workflow_new.py
```diff
@@ -0,0 +1,1024 @@
...
+
+        for key, value in self._compound_child_dict.items():
+            for task_name, task_obj in value.items():
+                self._task_dict[task_name] = task_obj
+
+        return list(self._task_dict.values())
```
seanpearsonuk: This can be condensed into a single expression that wouldn't look too complicated (also dispensing with `list()` -- see above). N.b. `items()` not needed always.

src/ansys/fluent/core/workflow_new.py
```diff
@@ -0,0 +1,1024 @@
...
+    def task_names(self):
+        """Get Python-friendly names for all available tasks.
+
+        Returns the list of task names as they would be accessed via Python
+        attribute syntax (e.g., "import_geometry" for "Import Geometry").
+        """
+        names = []
+        for name in self._workflow.task_object():
+            names.append(name.split(":")[0])
+        return names
```
seanpearsonuk: Single expression would be OK here. Would we prefer to return a generator? Caveat: is `task_names()` a usseful method? We're returning the objects too, right?

src/ansys/fluent/core/workflow_new.py
```diff
@@ -0,0 +1,1024 @@
...
+        ordered_names = _convert_task_list_to_display_names(
+            self._workflow,
+            self._workflow.general.workflow.task_list(),
+        )
+        name_to_task = {
+            task_obj.name(): make_task_wrapper(
+                task_obj, task_obj.__class__.__name__.lstrip("_"), self._workflow, self
+            )
+            for task_obj in self.tasks()
+        }
```
seanpearsonuk: Here `name_to_task` is fully populated **including the values**...

src/ansys/fluent/core/workflow_new.py
prmukherj: shall we remove this method from our new interface then?

src/ansys/fluent/core/workflow_new.py
prmukherj: This is no longer required in our new interface, right?

src/ansys/fluent/core/workflow_new.py
prmukherj: I wanted to make the interface with minimal change, but yes the design was flawed. I'll update according to your suggestions.

src/ansys/fluent/core/workflow_new.py
```diff
@@ -0,0 +1,1024 @@
...
+
+        sorted_list = []
+        for name in ordered_names:
+            if name not in name_to_task:
+                continue
+            sorted_list.append(name_to_task[name])
```
seanpearsonuk: ...but only some keys are used, so some of the objects will be thrown away. 

In other words, we shouldn't have constructed the `dict` at all but should have done the name match more directly.

Also, the logic here is back-to-front, adding extra code (`not`, `continue`)

The `sorted_list` looks like it could be instantiated and return ed on the fly with less, but still uncomplicated.

src/ansys/fluent/core/workflow_new.py
prmukherj: Earlier it used to return a list, that's why. There is no concrete requirement though, we can update as required.

src/ansys/fluent/core/workflow_new.py
```diff
@@ -0,0 +1,1024 @@
...
+                )
+            except LookupError as ex2:
+                raise LookupError(
+                    f"Neither '{name_2}' nor '{key}' not found in task object '{name_1}'."
+                ) from ex2
+
+    def __delitem__(self, key):
+        self[key].delete()
+
+    def _task_list(self):
```
seanpearsonuk: For clarity, we could call this `task_names` or similar.

src/ansys/fluent/core/workflow_new.py
```diff
@@ -0,0 +1,1024 @@
...
+        """Gets the display names of the child tasks of a task item."""
+        task_obj = super().__getattribute__("_task_object")
+        # This is just a precaution in case this method is directly called from the task level.
```
seanpearsonuk: The comment mentions a precaution, but I don't know what it's meant to refer to.

src/ansys/fluent/core/workflow_new.py
```diff
@@ -0,0 +1,1024 @@
...
+                raise LookupError(
+                    f"Neither '{name_2}' nor '{key}' not found in task object '{name_1}'."
+                ) from ex2
+
+    def __delitem__(self, key):
+        self[key].delete()
+
+    def _task_list(self):
+        """Gets the display names of the child tasks of a task item."""
+        task_obj = super().__getattribute__("_task_object")
```
seanpearsonuk: Can you comment this, please. I don't understand it.

**EDIT: see my later comment on use of** ``__getattribute__``

src/ansys/fluent/core/workflow_new.py
```diff
@@ -0,0 +1,1024 @@
...
+
+        Returns
+        -------
+        List[TaskObject]
+            Ordered list of child task wrappers, or empty list if no children.
+        """
+        if not self._task_list():
+            return []
+
+        workflow = super().__getattribute__("_workflow")
```
seanpearsonuk: If we update `__getattr__()` 's implementation we should be able to dispense with all this tricky code. That way we deal with the dynamic attribute handling in one palce. There are at least two `__getattribute__()` calls here.

src/ansys/fluent/core/workflow_new.py
prmukherj: This has been handled into a function now. We only mention string once.

src/ansys/fluent/core/meshing/meshing_workflow_new.py
```diff
@@ -0,0 +1,321 @@
...
+        """
+        super().__init__(
+            workflow=workflow, command_source=meshing, fluent_version=fluent_version
+        )
+        self._meshing = meshing
+        self._unsubscribe_root_affected_callback()
+        if initialize:
+            self._create_workflow()
+        else:
+            self._activate_dynamic_interface(dynamic_interface=True)
```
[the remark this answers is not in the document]
prmukherj: Done.

src/ansys/fluent/core/meshing/meshing_workflow_new.py
```diff
@@ -0,0 +1,321 @@
...
+        fluent_version: FluentVersion
+            Version of Fluent in this session.
+        initialize: bool
+            Flag to initialize the workflow, defaults to True.
+        """
+        super().__init__(
+            workflow=workflow, command_source=meshing, fluent_version=fluent_version
+        )
+        self._meshing = meshing
+        self._unsubscribe_root_affected_callback()
```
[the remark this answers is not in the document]
prmukherj: Done.

src/ansys/fluent/core/workflow_new.py
prmukherj: Done.

src/ansys/fluent/core/workflow_new.py
prmukherj: Commented accordingly.

prmukherj: This can be factored out into smaller samples to make it more readable. Will do it.

PR #4629, opened 2025-11-21
chore: Update launcher tests for better coverage

prmukherj:
## Context
Extensive test coverage for the launcher APIs from session classes  is introduced. 

One issue is when a 'PureMeshing' session is launched from PyFluent and 'Meshing' session is connected to it using the launcher api, it might be confusing for users, as now no mechanism exists in server to inform client that it has been launched in 'PureMeshing' mode. **Hence, this limitation is to be documented.**

## Change Summary
Tests were added for extensive coverage of launcher APIs and the limitation is documented.

## Impact
No direct impact, only a limitation as mentioned above to be documented.

**There are four possible connection patterns:**

Meshing → Meshing | ✔ Safe | Full capabilities preserved

PureMeshing → PureMeshing | ✔ Safe | Capabilities match exactly

Meshing → PureMeshing | ✔ Allowed | API connects to a subset of available capabilities

PureMeshing → Meshing | ⚠ Not recommended | Meshing API expects solver features that a pure-meshing session cannot provide

The fourth scenario is problematic because Meshing may assume capabilities (e.g., solver mode switching) that are not enabled in a pure-meshing session, leading to inconsistent or undefined behavior.

In later Fluent release we might consider updating the app-utilities, as suggested by @mkundu1 

"An implementation can be done by adding a new method set_pure_meshing_app_mode in the app_utilities service. The new method will be called to set a flag in Fluent when the Fluent session is launched or connected from a PureMeshing PyFluent session. That flag will be checked in get_app_modemethod of app_utilities. We can also block switch-to-solver operation within Fluent when that flag is set."

### DIFF
--- a/doc/source/user_guide/session/launching_ansys_fluent.rst
+++ b/doc/source/user_guide/session/launching_ansys_fluent.rst
@@ -104,8 +104,8 @@ Use this method when:
 
     PyFluent offers two Python interfaces for meshing:
 
-    - ``Meshing`` – meshing interface with an additional method to switch to solver mode.
-    - ``PureMeshing`` – meshing interface without any solver switching features.
+    - ``Meshing``: meshing interface with an additional method to switch to solver mode.
+    - ``PureMeshing``: meshing interface without any solver switching features.
 
     The two interfaces expose the **same meshing functionality**. The only difference is that
     ``Meshing`` includes ``switch_to_solver()``.
--- a/src/ansys/fluent/core/session_pure_meshing.py
+++ b/src/ansys/fluent/core/session_pure_meshing.py
@@ -41,18 +41,16 @@ from ansys.fluent.core.utils.fluent_version import FluentVersion
 
 
 class PureMeshing(BaseSession):
-    """Encapsulates a Fluent meshing session.
+    """Encapsulates a Fluent meshing session with a meshing-only Python interface.
 
-    ``PureMeshing`` is useful in modular or containerized workflows
-    where meshing and solving are run in separate processes. It provides a
-    clear, restricted interface that matches the capabilities of a meshing-only
-    session, even if that session happens to run inside a Fluent build that
-    includes solver components.
+    ``PureMeshing`` is designed for workflows where meshing and solving are run as
+        separate stages or in different environments, such as modular or containerized
+        deployments. It provides a clean API that focuses solely on meshing tasks.
 
-    It has a ``tui`` object for meshing TUI commanding, and ``meshing``
-    and ``workflow`` objects for access to task-based meshing workflows.
+        This interface exposes:
 
-    No ``switch_to_solver`` method is available in this mode.
+        - ``workflow`` and ``meshing`` objects for task-based meshing operations.
+        - ``tui`` for scripting via the legacy Text User Interface (when needed).
     """
 
     _rules = [


seanpearsonuk:
It is ok to create a pure session from an "impure" session. But the reverse is not true because if it really is originally pure then switching to solver won't work. No such limitations exist the other way.

I believe that that is what I said in the meeting but feel free to share your views.

@prmukherj @mkundu1 @hpohekar @mayankansys @raph-luc @Gobot1234

Gobot1234:
Can I suggest the following in the body of the PureMeshing class if you are going to retain it but keep it as an error

```py
if TYPE_CHECKING:
    @typing_extensions.deprecated("This will always raise due a fluent limitation")
    def from_connection(cls, *args: Any, **kwargs: Any) -> typing_extensions.Never: ...

**Comment by seanpearsonuk** [comment]
Next steps?
- re-title PR
- update description
- update tests according to expected behaviour, as discussed

@prmukherj

**Comment by prmukherj** [comment]
> > Next steps?
> > 
> > * re-title PR
> > * update description
> > * update tests according to expected behaviour, as discussed
> > 
> > @prmukherj
> 
> Will do it after this release PR is merged. Thank you. Will convert the PR to draft for now.

Done.

File: doc/source/user_guide/session/launching_ansys_fluent.rst [hunk 1]
```diff
@@ -100,6 +100,45 @@ Use this method when:
    solver.exit()
    solver_connected.exit()
 
+.. note::
+
```
seanpearsonuk, line 104: If we doc this, we must be concrete, explicit and precise about the actual issue. Otherwise, most users would be completely confused. I believe that we MUST lay out all the exact information in logical order and in the most direct and plainest possible language.

seanpearsonuk: :bear: with me

seanpearsonuk:
```rst
.. note::

    PyFluent offers two Python interfaces for meshing:

    - ``Meshing`` – meshing interface with an additional method to switch to solver mode.
    - ``PureMeshing`` – meshing interface without any solver switching features.

    The two interfaces expose the **same meshing functionality**. The only difference is
    that ``Meshing`` includes ``switch_to_solver()``.

    When connecting to an existing Fluent session via ``from_connection()``:

    - Use ``PureMeshing.from_connection()`` if the session was launched for **meshing only**.
    - Use ``Meshing.from_connection()`` if the session supports **meshing and solving**.
    - You may also use ``PureMeshing.from_connection()`` with a session that supports solving,
      if you intentionally want access **only to meshing features**.

    A ``Meshing`` interface is not recommended for a **meshing-only** session,
    because ``switch_to_solver()`` would raise an error in that case.

    ``PureMeshing`` is useful in modular or containerized workflows where meshing and
    solving are run in separate processes. It provides a clear, restricted interface
    that matches the capabilities of a meshing-only session, even if that session
    happens to run inside a Fluent build that includes solver components.
```

seanpearsonuk: IMHO the original is too chatty:
- What the server tells us or can't tell us about capabilities is an implementation detail. Leave those out, always.
- "does not provide **certain features** expected by the Meshing interface (**such as** solver mode switching), which may lead to **inconsistent behavior**" is vague as well as verbose.
- The "patterns" complicate the message unnecessarily and most of it is redundant.
- "the client API does not distinguish automatically between these session variants" talks about _session variants_ which is a confusing and non-existent concept.
- "The PureMeshing interface remains available because...". This assumes that the reader believes that the interface either is or should be under some sort of threat.

prmukherj: Sounds good @seanpearsonuk, just one thing. Shall we get rid off the last paragraph entirely?

seanpearsonuk: `PureMeshing` docstring. I had mostly corrected "The PureMeshing interface remains available because..." which was problematic.

prmukherj: @seanpearsonuk, have updated as per suggestions, please confirm.

prmukherj:
Thank you @seanpearsonuk for the suggestions. Applied them.

src/ansys/fluent/core/session_pure_meshing.py
```diff
@@ -43,11 +43,16 @@
...
+    ``PureMeshing`` is useful in modular or containerized workflows
+    where meshing and solving are run in separate processes. It provides a
+    clear, restricted interface that matches the capabilities of a meshing-only
+    session, even if that session happens to run inside a Fluent build that
+    includes solver components.
+
+    It has a ``tui`` object for meshing TUI commanding, and ``meshing``
+    and ``workflow`` objects for access to task-based meshing workflows.
+
+    No ``switch_to_solver`` method is available in this mode.
```
seanpearsonuk:
```suggestion
class PureMeshing(BaseSession):
    """Encapsulates a Fluent meshing session with a meshing-only Python interface.

    ``PureMeshing`` is designed for workflows where meshing and solving are run as
    separate stages or in different environments, such as modular or containerized
    deployments. It provides a clean API that focuses solely on meshing tasks.

    This interface exposes:

    - ``workflow`` and ``meshing`` objects for task-based meshing operations.
    - ``tui`` for scripting via the legacy Text User Interface (when needed).
```

seanpearsonuk: Note that I'm de-emphasizing the TUI and omitting mention of absent methods. We should ensure that we de-emphasize the TUI elsewhere.

hpohekar:
> > > It is ok to create a pure session from an "impure" session. But the reverse is not true because if it really is originally pure then switching to solver won't work. No such limitations exist the other way.
> > > I believe that that is what I said in the meeting but feel free to share your views.
> > > @prmukherj @mkundu1 @hpohekar @mayankansys @raph-luc @Gobot1234
> > 
> > 
> > Ok, then we will remove it completely like discussed, because there is no way to check the session type for Pure Meshing session from the server side.
> > Thank you
> 
> Before we remove it completely, I’d like to restate why I think retaining the pure meshing interface is still valuable.
> 
> The motivation behind “Pure Meshing” wasn’t meant to be a PyFluent-only abstraction. It supports a modular workflow where users can deliberately choose to isolate meshing from solving. In particular:
> 
> * Users can build minimal container images dedicated only to meshing or only to solving.
> * In that workflow, transferring data between separate containers is preferred over switching modes inside one container.
> * For these users, offering a clean and clearly scoped API surface for each mode is helpful.
> 
> So while we can’t currently detect a “pure” session from the server, users can still intentionally choose to work in a segregated way. Removing this option takes away a valid pattern that is relevant in automated and cloud-centric deployments.

@seanpearsonuk I agree. The container specific requirements specified here should consider for 26R2 planning as well.

PR #4656, opened 2025-11-26
fix: Arguments in standalone launcher.

prmukherj:
## Context
'certificates_folder' and 'insecure_mode' was missing in 'StandaloneLauncher'

## Change Summary
'certificates_folder' and 'insecure_mode' were added as launcher args.

## Impact
'certificates_folder' and 'insecure_mode' will be exposed.

### DIFF
--- a/src/ansys/fluent/core/launcher/standalone_launcher.py
+++ b/src/ansys/fluent/core/launcher/standalone_launcher.py
@@ -195,6 +195,15 @@ class StandaloneLauncher:
         """
         import ansys.fluent.core as pyfluent
 
+        if certificates_folder is None and not insecure_mode:
+            raise ValueError(
+                "To launch Fluent in Slurm environment, set `certificates_folder`."
+            )
+        if certificates_folder is not None and insecure_mode:
+            raise ValueError(
+                "`certificates_folder` and `insecure_mode` cannot be set at the same time."
+            )
+
         self.certificates_folder = certificates_folder
         self.insecure_mode = insecure_mode
 
--- a/src/ansys/fluent/core/launcher/fluent_launcher_options.json
+++ b/src/ansys/fluent/core/launcher/fluent_launcher_options.json
@@ -32,5 +32,9 @@
     "scheduler_account": {
         "type": "str",
         "fluent_format": " -scheduler_account={}"
-    }
+    },
+    "insecure_mode": {
+        "type": "bool",
+        "fluent_format": " -grpc-allow-remote-host -grpc-insecure-mode"
+    },
 }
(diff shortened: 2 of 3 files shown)


src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -106,6 +106,8 @@ def __init__(
         topy: str | list | None = None,
         start_watchdog: bool | None = None,
         file_transfer_service: Any | None = None,
+        certificates_folder: str | None = None,
+        insecure_mode: bool = False,
     ):
```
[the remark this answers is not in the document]
seanpearsonuk: @prmukherj

mkundu1:
Although these modes are not required for standalone launch, it is OK to add them as they are supported. Some user may look for a common script for different launch modes.

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -105,7 +105,11 @@ def __init__(
...
+        self.certificates_folder = certificates_folder
+        self.insecure_mode = insecure_mode
         """
         Launch a Fluent session in standalone mode.
```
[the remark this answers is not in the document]
seanpearsonuk: @prmukherj - seems a valid point.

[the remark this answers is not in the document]
prmukherj: right, had missed this point. Thanks @seanpearsonuk .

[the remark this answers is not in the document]
prmukherj: Why this had happened was because I had committed it directly from the suggestion. This is something we should be careful.
Updated the file.

src/ansys/fluent/core/launcher/standalone_launcher.py
seanpearsonuk, line 196: Hadn't noticed this previously. 👎

prmukherj: This was done to prevent circular import I guess.. Will try to fix this.

seanpearsonuk: @prmukherj It's simple--I already created an issue. I had missed that any of this code was done like this.

seanpearsonuk: > This was done to prevent circular import I guess.. Will try to fix this.

Do it as a separate piece of work that touches several mods.

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -187,6 +195,9 @@ def __init__(
         """
         import ansys.fluent.core as pyfluent
 
+        self.certificates_folder = certificates_folder
+        self.insecure_mode = insecure_mode
```
mkundu1, line 208: How are these arguments passed to Fluent or used in the client during the channel construction?

PR #4660, opened 2025-11-26
feat: Minor improvements of the gRPC feature

mkundu1:
## Context
1. There was no easy way to use the insecure mode while running the gallery examples in github without changing the example scripts or overriding library methods.
2. While connecting to a local Fluent server in linux, it was not possible to pass the UDS path to `connect_to_fluent`. User had to pass the server-info filepath.

## Change Summary
1. Launching Fluent in container mode from PyFluent will use the insecure mode when the new environment variable `PYFLUENT_CONTAINER_INSECURE_MODE` is set.
2. A new `address` argument has been added to `connect_to_fluent`.

## Rationale
1. The change 1 is not meant for users. Hence the environment variable approach has been taken, instead of other user-facing approach like adding new config variable.
2. User-facing change so they can pass the server-info file content (similar to passing ip+port) to `connect_to_fluent`.

## Impact
1. It will be easier to configure the CI scripts to use insecure mode during github runs. No complicated workaroud like overriding `launch_fluent` behaviour is necessary.
2. Will make user-workflow consistent with other modes where user can pass ip+port to `connect_to_fluent`.

### DIFF
--- a/.github/workflows/doc-build-dev-nightly.yml
+++ b/.github/workflows/doc-build-dev-nightly.yml
@@ -134,7 +134,7 @@ jobs:
 
   deploy_dev_docs:
     name: "Deploy Documentation"
-    runs-on: ubuntu-latest
+    runs-on: public-ubuntu-latest-8-cores
     needs: [build_dev_docs]
     if: github.ref == 'refs/heads/main'
     permissions:
--- a/.github/workflows/doc-build-release.yml
+++ b/.github/workflows/doc-build-release.yml
@@ -96,7 +96,7 @@ jobs:
 
   deploy_release_docs:
     name: "Deploy Release Documentation"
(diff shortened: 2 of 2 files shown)


hpohekar (approved):

.github/workflows/doc-build-dev-nightly.yml
```diff
@@ -73,30 +73,35 @@ jobs:
           python examples/00-fluent/exhaust_system_settings_api.py
         env:
           FLUENT_IMAGE_TAG: ${{ env.FLUENT_STABLE_IMAGE }}
+          PYFLUENT_CONTAINER_INSECURE_MODE: 1
```
prmukherj, line 76: 👍

prmukherj (approved):

PR #4690, opened 2025-12-03
feat: Remove old Fluent version support

mkundu1:
## Context
Fluent versions before 24.2 are not currently supported by Ansys. So, we don't need to support those old versions in PyFluent. Also, as patches are not released for those Fluent versions, we shouldn't point users to use those versions.

## Change Summary
1. Restrict the supported Fluent versions to 24.2 - 26.2 in code. I've included next dev version as it has been requested.
2. Update the minimum supported Fluent version to 24.2 in doc.
3. Added a CI script `.ci\docker_data_cleaner.py` to remove all docker data except the supported Fluent images. The script will be run before and after every job that starts a Fluent container. Currently, the self-hosted runners don't have the sufficient space to cache all supported Fluent images, but that won't be a problem anymore. I've also deleted other docker-related cleanup scripts.

## Rationale
We have aligned the supported Fluent version range with the versions supported by Ansys.

## Impact
Package size will be reduced. CI speedup is expected.

### DIFF
--- a/doc/source/getting_started/faqs.rst
+++ b/doc/source/getting_started/faqs.rst
@@ -101,6 +101,7 @@ all PyFluent packages in a Python *virtual environment*:
    For Python version information, see the answer to the next question.
 
 #. Install Fluent 2024 R2 or later.
+
 #. Set the environment variable for your installed release to point to
    the appropriate Ansys installation folder.
 
@@ -136,6 +137,10 @@ all PyFluent packages in a Python *virtual environment*:
       python -m pip install ansys-fluent-core   # Use Fluent’s core capabilities (mesh, solve, postprocess)
       python -m pip install ansys-fluent-parametric   # Use Fluent’s parametric capabilities (optional)
       python -m pip install ansys-fluent-visualization   # Use Fluent's postprocessing capabilities with pyvista and matplotlib (optional)
+
+.. note::
+   Ansys Fluent versions prior to 2024 R2 were supported by PyFluent version 0.37 and earlier.
+   These versions are no longer supported in the current PyFluent releases.
 
 
 Which version of Python should you use?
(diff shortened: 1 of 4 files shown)


src/ansys/fluent/core/services/field_data.py
```diff
@@ -993,7 +993,7 @@ def get_response(self) -> BatchFieldData:
             Data is returned as dictionary of dictionaries in the following structure:
             tag int | Tuple-> surface_id [int] -> field_name [str] -> field_data[np.array]
 
-            The tag is a tuple for Fluent 2023 R1 or later.
+            The tag is a tuple for Fluent 2024 R2 or later.
```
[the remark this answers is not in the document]
seanpearsonuk: @mkundu1 Is moving the qualification to point to the earliest supported version redundant? Should we not just remove the qualification about versions here?

PyPablo (approved):

doc/source/getting_started/faqs.rst
```diff
@@ -100,12 +100,12 @@ all PyFluent packages in a Python *virtual environment*:
 
    For Python version information, see the answer to the next question.
 
-#. Install Fluent 2022 R2 or later.
+#. Install Fluent 2024 R2 or later.
```
raph-luc, line 103: Any concerns about consumers which which may be (for whatever reason) still using old Ansys releases? As far as I know, that is actually quite common.

Would the consumers be able to "easily" find old PyFluent releases that they could try out with their old Ansys release, if they wanted?

Perhaps we could maintain a list in the documentation indicating what was the last released PyFluent version that still supported each Ansys release?

seanpearsonuk: > Perhaps we could maintain a list in the documentation indicating what was the last released PyFluent version that still supported each Ansys release?

@raph-luc  Yes, that's something we have discussed as well.

@prmukherj 
@mkundu1

seanpearsonuk (approved):

gyeole (approved):

doc/source/getting_started/faqs.rst
mkundu1: Added doc note for the same

PR #4707, opened 2025-12-08
build: Bump version 0.38.dev2

mkundu1:
Releasing 0.38.dev2


src/ansys/fluent/core/__init__.py
raph-luc, line 1: So `0.38.dev1` did not work?

mkundu1: No, the previous changelog task got messed up beyond repair, hence creating a new tag

PR #4713, opened 2025-12-08
feat: Make sure slurm launcher works even if Slurm is not available in client

mkundu1:
## Context
In the original design of slurm launcher it was assumed that the client machine is also within the same cluster where Slurm for Fluent is running and a few client-side features for querying/controlling Slurm jobs were added based on that assumption. Although these client-side features provide usability, they are not necessary for the basic functionality of PyFluent's Slurm launcher: launch and connect to a Fluent session within a Slurm environment. OTOH, it is a common use case that PyFluent is running outside a Slurm cluster where user wants to launch Fluent from PyFluent.

## Change Summary
In this PR, the strict requirement of Slurm availability on the client machine has been revoked. If Slurm is available in the client machine, those client-side features for job query/control will work. Otherwise, they will throw a RuntimeError. A config var is also added to avoid using Slurm in the client machine even when it is available considering the scenario where the client machine is part of a different Slurm cluster than where Fluent would be run.

## Rationale
The basic functionality will be available irrespective of the client machine being part of Fluent's Slurm cluster. If the client machine is in Fluent's Slurm cluster, additional usability feature will be available.

## Impact
It will be possible to use the Slurm launcher from outside a Slurm environment.

### DIFF
--- a/src/ansys/fluent/core/launcher/slurm_launcher.py
+++ b/src/ansys/fluent/core/launcher/slurm_launcher.py
@@ -145,6 +145,23 @@ class _SlurmWrapper:
         return config.use_slurm_from_current_machine and _SlurmWrapper.is_available()
 
     @staticmethod
+    def check_and_raise_slurm_not_used_error(msg: str) -> None:
+        """Raise RuntimeError if Slurm is not used from the current machine.
+
+        Parameters
+        ----------
+        msg : str
+            Error message prefix.
+
+        Raises
+        ------
+        RuntimeError
+            If Slurm is not used from the current machine.
+        """
+        if not _SlurmWrapper.use_slurm():
+            raise RuntimeError(msg + _slurm_unavailable_in_current_machine_clause)
+
+    @staticmethod
     def list_queues() -> list[str]:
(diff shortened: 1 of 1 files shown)


hpohekar (approved):

src/ansys/fluent/core/launcher/slurm_launcher.py
```diff
@@ -203,7 +223,16 @@ def cancel(self, timeout: int = 60) -> bool:
...
+
+        Raises
+        ------
+        RuntimeError
+            If Slurm job cannot be cancelled from client
         """
+        if not _SlurmWrapper.use_slurm():
+            raise RuntimeError(
+                f"Cannot cancel Slurm job from client {_slurm_unavailable_in_current_machine_clause}"
+            )
```
seanpearsonuk: There is some opportunity to factor out duplicated code. It seems that this code repeats a few times: equivalent though not always in identical form.

mkundu1: moved the code to a common method

Issue #4351, opened 2025-08-06
Add tests for integration-level intellisense checks


mkundu1:
Integration-level IntelliSense checks are about verifying what symbols, methods, properties, and type information are actually visible to external tools (like IDEs, LSPs, or static analyzers) when someone uses a Python library. Since IntelliSense is not part of Python's runtime, the idea is to simulate the perspective of a developer using the Python library in an IDE or editor. These checks are most commonly done using tools like [jedi](https://jedi.readthedocs.io/en/latest/).

PR #4725, opened 2025-12-10
test: Add first intellisense test

mkundu1:
https://github.com/ansys/pyfluent/issues/4351

### DIFF
--- a/doc/changelog.d/4725.test.md
+++ b/doc/changelog.d/4725.test.md
@@ -1,0 +1,1 @@
+Add first intellisense test


Gobot1234:
Is jedi what pyconsole uses? Otherwise I'd be worried that we're testing something that very few people use nowadays as most things are on something much newer

Gobot1234:
https://github.com/astral-sh/ruff/blob/ad3de4e4881a15545fc85fdfaf5374df7793538d/crates/ty_completion_eval/truth/object-attr-instance-methods/main.py#L13 I've spoken to some people online and they suggested maybe taking a look at this code for ty

PR #4762, opened 2025-12-18
fix: Updated the dependabot.yml file

mayankansys:
This pull request focuses on dependency management improvements and minor documentation updates. The main changes include refining how dependencies are handled by Dependabot, updating the version constraint for an optional dependency.

Dependency management improvements:

* Updated `.github/dependabot.yml` to ignore updates for several main and optional dependencies, preventing Dependabot from creating PRs for these packages. This helps control which dependencies are automatically updated.

* Relaxed the version constraint for the `h5py` package in the `reader` optional dependency, allowing any version greater than or equal to 3.15.1 instead of requiring exactly 3.15.1 in `pyproject.toml`.

## Context
Dependabot was creating update PRs for dependencies that do not need automatic updates, adding unnecessary noise. Some optional dependencies were also pinned to exact versions, which limited flexibility and caused avoidable constraints.

## Change Summary
Updated the Dependabot configuration to ignore selected dependencies and reduce unwanted PRs. Relaxed version constraints for optional dependencies by allowing minimum versions instead of exact pins, and updated documentation to reflect these changes.

## Impact
This pull request primarily updates dependency management and documentation. The most significant changes include refining the dependencies handled by Dependabot, relaxing version constraints for optional dependencies, and updating documentation.

Dependency management improvements:

* Updated `.github/dependabot.yml` to ignore updates for a list of specified dependencies, preventing Dependabot from creating PRs for these packages.
* Relaxed version constraints for optional dependencies in `pyproject.toml`, changing pinned versions to minimum version requirements for `h5py`, `ipywidgets`, and `panel`.


.github/dependabot.yml
```diff
@@ -8,6 +8,24 @@ updates:
     labels:
       - "Maintenance"
       - "Dependencies"
+    ignore:
+      # Exclude main project dependencies and optional dependencies (like reader)
+      - dependency-name: "ansys-api-fluent"
```
prmukherj, line 13: I am just curious that why these were not part of dependabot PRs earlier even though it was not part of ignore list. Specifically the ansys-* ones.

@mkundu1, any thoughts on this?

prmukherj:
Looks Good 👍

seanpearsonuk (approved):

pyproject.toml
```diff
@@ -40,9 +40,9 @@ dynamic = ["version"]
 include = ["src/ansys/fluent/core/generated/"]
 
 [project.optional-dependencies]
-reader = ["h5py==3.15.1"]
-ui-jupyter = ["ipywidgets"]
-ui = ["panel"]
+reader = ["h5py>=3.15.1"]
```
mkundu1, line 43: Should we relax this to something `>=3.13.0`? I feel `3.15.1`, which has been released in [Oct 2025](https://pypi.org/project/h5py/#history), is unnecessarily strict. 

@prmukherj

.github/dependabot.yml
mkundu1: I don't think the ignore list is necessary. It used to work before without the ignore list.

pyproject.toml
mayankansys: Agreed. What you think @prmukherj.

prmukherj: yes, sounds like the right thing to do.

PR #4763, opened 2025-12-18
fix: Some issues <session>.is_active +1385523

prmukherj:
## Context

These are completely separate issues (just for the benefit of all readers):

1. 'is_active' was not checking for the health status of the server. Suppose a scenario, where we connect to an existing solver and that solver crashes, the connected solver would still be active, that is resolved here.

2. 1385523 - Fluent side bug was also resolved here. 

## Change Summary
1. 'is_server_healthy' is deprecated. 'is_active' does all the chechking.
2. 'after_execute' is always run if 'before_execute' is run.

## Impact
1. Users to use <session>.is_active() instead of <session>.is_server_healthy. A deprecation message will be shown in the latter case.
2. Such issues will be avoided:
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,12 +42,12 @@ jobs:
     name: "Add license headers"
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           persist-credentials: false
 
       - name: Set up Python
-        uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6.0.0
+        uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # v6.1.0
         with:
           python-version: ${{ env.MAIN_PYTHON_VERSION }}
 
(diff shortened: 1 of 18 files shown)


doc/deprecated_pyfluent_apis.py
```diff
@@ -21,7 +21,7 @@
     (
         ":py:meth:`ansys.fluent.core.session.BaseSession.health_check <ansys.fluent.core.session.BaseSession.health_check>`",
         "0.32",
-        ":py:meth:`ansys.fluent.core.session.BaseSession.is_server_healthy <ansys.fluent.core.session.BaseSession.is_server_healthy>`",
+        ":py:meth:`ansys.fluent.core.session.BaseSession.is_active <ansys.fluent.core.session.BaseSession.is_active>`",
```
mkundu1, line 24: I think there should be now 2 entries: `health_check` and `is_server_healthy`. Both have been replaced by `is_active`.

mkundu1 (approved):

doc/deprecated_pyfluent_apis.py
prmukherj: Done, Thank you.

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -2278,6 +2278,9 @@ def _process_cls_names(info_dict, names, write_doc=False):
         commands = info.get("commands")
         if commands:
             commands.pop("exit", None)
+            if version >= "261" and parent is not None:
```
mkundu1: @prmukherj We need to rework this change to address the Fluent bug 1385445. Some named-objects can be special implementations for list and list_properties which should get priority over the common implementations.

prmukherj: yes, I will get back to you on this soon.
Thank you

PR #4778, opened 2026-01-02
fix: Remove dry_run from PIM launcher

hpohekar:
## Context
`dry_run=True` is not returning configuration dictionary.

## Change Summary
Removed `dry_run` argument.

## Rationale
`dry_run` is supported for the container mode only.

## Impact
PIM launcher signature is updated.

### DIFF
--- a/src/ansys/fluent/core/launcher/pim_launcher.py
+++ b/src/ansys/fluent/core/launcher/pim_launcher.py
@@ -136,7 +136,7 @@ class PIMLauncher:
 
         Returns
         -------
-        Union[Meshing, PureMeshing, Solver, SolverIcing, dict]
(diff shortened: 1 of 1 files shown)


src/ansys/fluent/core/launcher/pim_launcher.py
```diff
@@ -140,7 +137,7 @@ def __init__(
         Returns
         -------
         Union[Meshing, PureMeshing, Solver, SolverIcing, dict]
```
Gobot1234:
```suggestion
        Union[Meshing, PureMeshing, Solver, SolverIcing, SolverAero]
```

prmukherj (approved):

PR #4779, opened 2026-01-02
feat: Update server based enhanced meshing workflow.

prmukherj:
## Context
1. "parts" and "parts-files" should only be accessible from Fault Tolerant workflows. It should not appear in the dir of the other workflows and also should raise Attribute error if tried to access from other workflow types.
2. "create_workflow" is introduced with insertable task list for initial set of tasks that can be inserted.

```python
created_workflow = meshing_session.create_workflow()
created_workflow.insertable_tasks.import_geometry.insert()
created_workflow.import_geometry.file_name.set_state(import_file_name)
created_workflow.import_geometry.length_unit.set_state('in')
created_workflow.import_geometry()
```

Test has been added for the above case.

3. Added test for "load_workflow".

## Change Summary
Existing logic and tests were migrated.

## Rationale
Simplest approach to make the existing tests work.

## Impact
No direct impact to users.

### DIFF
--- a/src/ansys/fluent/core/workflow_new.py
+++ b/src/ansys/fluent/core/workflow_new.py
@@ -535,7 +535,7 @@ class Workflow:
         """
 
         def __init__(self, workflow):
-            """Initialize an ``FirstTask`` instance."""
+            """Initialize a ``FirstTask`` instance."""
             self._workflow = workflow
             self._insertable_tasks: list = []
             # Map: server command name -> python-friendly task name


src/ansys/fluent/core/workflow_new.py
```diff
@@ -509,8 +511,67 @@ def delete_tasks(self, list_of_tasks: list[TaskObject]):
...
+                command_obj = getattr(workflow._command_source, command)
+                if isinstance(command_obj, PyCommand):
+                    command_obj_instance = command_obj.create_instance()
+                    if not command_obj_instance.get_attr("requiredInputs"):
+                        help_str = command_obj_instance.get_attr(
+                            "APIName"
+                        ) or command_obj_instance.get_attr("helpString")
+                        if help_str:
+                            self._initial_task_map[help_str] = command
+                    del command_obj_instance
```
seanpearsonuk: @prmukherj 
- The code appears to be a workaround for an issue where the server API does not provide the needed information directly. When doing something like this, we should try to keep the code looking clean: this _special_ code can be moved to a separate function and comments can be added to explain that it's a special, temporary workaround.
- Is the workaround, involving access to attributes handled in the back-end mappings ("APIName" and "helpString"), command instantiations, and access to the "requiredInputs" attribute fully necessary? Is it possible to use the API in a simpler manner, thereby keeping the layers more separated? For instance, command attributes should be accessible via the API.

prmukherj: @seanpearsonuk, you are right. It is missing a server side implementation. I have the server side implementation ready but I don't think so it will be appropriate to get it in now. Anyways, this is only a PyFluent feature as of now. I will modify the code as per your comments, but firstly for the enhanced meshing workflow I will hard-code the insertable tasks and following my server side additions, I will add the generic code. As of now this PR will contain both.

prmukherj: Done, Thank you

src/ansys/fluent/core/workflow_new.py
```diff
@@ -509,8 +511,60 @@ def delete_tasks(self, list_of_tasks: list[TaskObject]):
...
+            self._workflow = workflow
+            self._insertable_tasks = []
+            self._initial_task_map = {}
+            try:
+                initial_tasks = self._workflow.general.get_new_insertable_tasks()
+            except AttributeError:
+                # For Fluent Version 26R1 or before.
+                initial_tasks = ["ImportGeometry", "PartManagement", "RunCustomJournal"]
+            for command in initial_tasks:
+                self._initial_task_map[command] = command_name_to_task_name(
```
seanpearsonuk, line 552: Overall, there needs to be some clarification (via naming) about what is private and what is meant to be public, as it's a bit mixed up presently.  We also need to pay heed to what's plural and what's singular (e.g. `_FirstTask` is used where a collection is expected, which is compounded by a missing type hint). There is also a conflict between what is commented as being specific to <=26.1 and what is actually called for versions <=26.1 versus what is called for later versions. Either the comments are wrong or the code is wrong, so we need to rectify at least one of them (in truth, later versions should not need any client-side mapping so it ought to be the code that is wrong). We also should clarify why `initial_tasks` is correct (your earlier implementation used metadata which appeared plausible but this version is probably OK as long as we comment it). I think that an overarching issue here is that the code is trying to do a lot in one layer, and also that it's using clever tricks that shouldn't be required (a property returning an object with a call method etc, but you can leave the design unchanged (it's not a good investment) and deal with the localized issues).

prmukherj: _FirstTask -> make public and document it.

src/ansys/fluent/core/workflow_new.py
```diff
@@ -169,11 +169,13 @@ def command_name_to_task_name(meshing_root, command_name: str) -> str:
     This is a workaround for Fluent 26R1.
     """
     # TODO: This is a fix only for 26R1 as the server lacks the mechanism to return mapped values
```
prmukherj: Remove 26R1 comment.

src/ansys/fluent/core/workflow_new.py
prmukherj: Done, Added more comments for clarity. "get_new_insertable_tasks" name will be updated once the server side code is updated.

PR #4783, opened 2026-01-02
fix: Update create_launcher function

hpohekar:
## Context
The default mode was None and error was not handled properly.

## Change Summary
The default mode is set to Standalone and error is raised for an incorrect launch mode.

## Rationale
The return was None in the current case.

## Impact
Launcher utils and testing.

### DIFF
--- a/src/ansys/fluent/core/launcher/launcher.py
+++ b/src/ansys/fluent/core/launcher/launcher.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2021 - 2025 ANSYS, Inc. and/or its affiliates.
+# Copyright (C) 2021 - 2026 ANSYS, Inc. and/or its affiliates.
 # SPDX-License-Identifier: MIT
 #
 #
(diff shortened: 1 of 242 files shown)


src/ansys/fluent/core/launcher/launcher.py
```diff
@@ -68,25 +69,30 @@
...
+    fluent_launch_mode: LaunchMode | None = None, **kwargs
+) -> DockerLauncher | PIMLauncher | SlurmLauncher | StandaloneLauncher:
     """Use the factory function to create a launcher for supported launch modes.
 
     Parameters
     ----------
     fluent_launch_mode: LaunchMode
         Supported Fluent launch modes. Options are ``"LaunchMode.CONTAINER"``,
         ``"LaunchMode.PIM"``, ``"LaunchMode.SLURM"``, and ``"LaunchMode.STANDALONE"``.
+        The default is ``None``, in which case ``"LaunchMode.STANDALONE"`` is used.
```
Gobot1234: Should this not just be the default in the function then?

prmukherj: yes, already being done by @hpohekar

Gobot1234: Just to clarify I meant set it to STANDALONE

PR #4790, opened 2026-01-08
fix: dimensional check for pim not supporting ints

Gobot1234:
## Change Summary
Currently PIM doesn't support dimension=2. This fixes support for that

### DIFF
--- a/tests/test_launcher.py
+++ b/tests/test_launcher.py
@@ -443,12 +443,12 @@ def test_fluent_launchers():
         assert pim_solver_session
         pim_solver_session.exit()
 
-        pim_solver_launcher = create_launcher(
+        two_d_pim_meshing_launcher = create_launcher(
             LaunchMode.PIM, mode=FluentMode.MESHING, **kwargs, dimension=2
         )
-        pim_solver_session = pim_solver_launcher()
-        assert pim_solver_session.dimension == pyfluent.Dimension.TWO
-        pim_solver_session.exit()
+        two_d_pim_meshing_session = two_d_pim_meshing_launcher()
+        assert two_d_pim_meshing_session.dimension == pyfluent.Dimension.TWO
+        two_d_pim_meshing_session.exit()
 
 
 @pytest.mark.parametrize(


src/ansys/fluent/core/launcher/pim_launcher.py
```diff
@@ -293,7 +293,7 @@ def launch_remote_fluent(
     instance = create_fluent_instance(
         pim=pim,
         mode=mode,
-        dimensionality=dimensionality,
+        dimensionality=Dimension(dimensionality),
```
hpohekar, line 296: 👍 Looks good. Thank you.

src/ansys/fluent/core/launcher/pim_launcher.py
```diff
@@ -333,7 +333,7 @@ def launch_remote_fluent(
 
 
 def create_fluent_instance(
-    pim, mode: FluentMode, dimensionality: Dimension | int, product_version: str | None
+    pim, mode: FluentMode, dimensionality: Dimension, product_version: str | None
```
prmukherj, line 336: is it better to change line 343 to Dimension(dimensionality) == Dimension.TWO, and keep rest of the code untouched? In that case we don't need to change signature of this method and keep it generic.

seanpearsonuk: The question is: who is calling `create_fluent_instance()`?

It's not mentioned in `tests` or `doc` so we can consider it to be internal. Tightening this interface is only defensive and doesn't impose any problematic restrictions.

prmukherj: only 'launch_remote_fluent' for now.

seanpearsonuk: Therefore this change is only positive and it's already generic given that `Dimension` is constructible from `int`.

seanpearsonuk (approved):

src/ansys/fluent/core/launcher/pim_launcher.py
prmukherj: got it, Thanks

hpohekar:
@Gobot1234 Code Style job is failing. We need to fix this.

cc: @prmukherj

tests/test_launcher.py
```diff
@@ -438,18 +438,17 @@ def test_fluent_launchers():
...
-        )
+        pim_solver_launcher = create_launcher(LaunchMode.PIM, **kwargs, dimension=2)
         pim_solver_session = pim_solver_launcher()
         assert pim_solver_session
         pim_solver_session.exit()
 
-        pim_meshing_launcher = create_launcher(
-            LaunchMode.PIM, mode=FluentMode.MESHING, **kwargs, dimension=2, dry_run=True
+        pim_solver_launcher = create_launcher(
+            LaunchMode.PIM, **kwargs, dimension=2
```
Gobot1234:
```suggestion
        pim_solver_launcher = create_launcher(
            LaunchMode.PIM, mode=FluentMode.MESHING, **kwargs, dimension=2
```

tests/test_launcher.py
```diff
@@ -438,18 +438,17 @@ def test_fluent_launchers():
...
         pim_solver_session.exit()
 
-        pim_meshing_launcher = create_launcher(
-            LaunchMode.PIM, mode=FluentMode.MESHING, **kwargs, dimension=2, dry_run=True
+        pim_solver_launcher = create_launcher(
+            LaunchMode.PIM, mode=FluentMode.MESHING, **kwargs, dimension=2
         )
-        args = pim_meshing_launcher()
-        assert args[0] == "fluent-2ddp"
+        pim_solver_session = pim_solver_launcher()
```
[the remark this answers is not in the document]
seanpearsonuk: @Gobot1234 The comment is technically correct so we should probably make the requested change.

PR #4809, opened 2026-01-14
chore: Use argument to use legacy interface.

prmukherj:
## Context
1. Only way to switch to the new and legacy meshing interface was the environment variable.
2. The docstrings were not much helpful either.

## Change Summary
1. All the user exposed methods have a 'legacy' argument to support launching in the legacy interface for Fluent version since 26R1. For versions before 26R1, it'll always be the legacy interface.

2. Sensible docstrings attached to user exposed methods.

## Rationale
To allow users to use legacy meshing interface without the use of setting environment variable.

## Impact
If user needs the legacy interface after version Fluent 26R1, they just need to pass the argument "legacy=True", for example, watertight = meshing.watertight(legacy=True)

For other versions it is automatically handled.

Please refer to the updated tests as well.

### DIFF
--- a/src/ansys/fluent/core/session_base_meshing.py
+++ b/src/ansys/fluent/core/session_base_meshing.py
@@ -27,7 +27,6 @@ import os
 
 from ansys.fluent.core._types import PathType
 from ansys.fluent.core.fluent_connection import FluentConnection
-from ansys.fluent.core.meshing.meshing_workflow import name_to_identifier_map
 from ansys.fluent.core.pyfluent_warnings import PyFluentUserWarning
 from ansys.fluent.core.session_shared import (
     _make_datamodel_module,
@@ -416,13 +415,6 @@ class BaseMeshing:
         )
         return self._current_workflow
 
-    def _check_workflow_type(self, name: str):
-        return getattr(self.meshing.GlobalSettings, name_to_identifier_map[name])()
-
-    def _get_current_workflow(self, name: str):
-        if self._current_workflow and self._current_workflow._name == name:
-            return self._current_workflow
-
     def current_workflow(self, legacy: bool | None = None):
         """Get the currently active meshing workflow.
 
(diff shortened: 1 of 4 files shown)


src/ansys/fluent/core/session_base_meshing.py
```diff
@@ -131,14 +132,76 @@ def meshing_workflow(self):
...
-        else:
+    def _fallback_check(self, legacy: bool | None):
+        """Determine whether to use legacy workflow implementation.
+
+        This method handles backward compatibility by automatically selecting the
+        appropriate workflow implementation based on Fluent version and user preference.
+
+        Parameters
+        ----------
+        legacy : bool or None
```
mkundu1, line 142: This documentation of the `legacy` parameter is missing in the public methods.

src/ansys/fluent/core/session_base_meshing.py
```diff
@@ -131,14 +132,76 @@ def meshing_workflow(self):
...
+
+        - ``legacy=False``: Request new mode
+        - Returns False for Fluent >= 26R1 (as requested)
+        - Returns True for Fluent < 26R1 (fallback with warning)
+
+        - ``legacy=True``: Force legacy mode
+        - Returns True regardless of version
+        """
+        fluent_version = self.get_fluent_version()
+        is_legacy_only = fluent_version < FluentVersion.v261
```
seanpearsonuk: @prmukherj The meaning of a name like `must_be_legacy` or `only_legacy_allowed` is more obvious.

src/ansys/fluent/core/session_base_meshing.py
prmukherj: "legacy" is documented in the PureMeshing session exposure which is the public facing method. Will add it here as well.

src/ansys/fluent/core/session_base_meshing.py
prmukherj: Done, Thank you

src/ansys/fluent/core/session_base_meshing.py
prmukherj: Done, Thank you

mkundu1 (approved):

PR #4811, opened 2026-01-15
docs: Update meshing workflow docs and some dostrings [skip tests]

prmukherj:
## Context
The existing documentation was mostly out of date with the updated version of the enhanced meshing workflow in place.
Some docstrings needed improvements.
Some earlier minor comments needed to be worked on.

## Change Summary
Mostly the docs were updated along with some updates in docstrings of individual methods.

Some scripts that have been modified, have their earlier counterparts in another doc in legacy section.

## Impact
Updated the user-facing meshing workflow docs.

This is how the doc page looks:

image

### DIFF
--- a/doc/source/user_guide/legacy/meshing_workflow.rst
+++ b/doc/source/user_guide/legacy/meshing_workflow.rst
@@ -9,7 +9,7 @@ distinct from the classic meshing workflow (:ref:`ref_meshing_workflows_guide`).
 Supported versions
 ------------------
 - Available in PyFluent with Ansys Fluent versions prior to 2026 R1.
-- Also supported when using PyFluent with Ansys Fluent 2026 R1 and 2027 R1.
+- Also supported when using PyFluent with Ansys Fluent 2026 R1 or later version.
 
 Enable the legacy interface
 ---------------------------
--- a/doc/source/user_guide/meshing/new_meshing_workflows.rst
+++ b/doc/source/user_guide/meshing/new_meshing_workflows.rst
@@ -5,29 +5,12 @@ Meshing workflows
 
 PyFluent provides access to Fluent’s meshing workflows.
 
-Definitions
------------
+Overview
+--------
 - Enhanced Meshing Workflows: A PyFluent API available only when using PyFluent with Ansys Fluent 2026 R1 and later.
   It provides clearer task organization, easier navigation, and strongly typed, well-documented arguments.
-- Legacy Meshing Workflows: The PyFluent meshing API used prior to Ansys Fluent 2026 R1
-  (previously branded “Enhanced Meshing Workflows”). It remains available by passing
-  ``legacy=True`` when initializing a workflow. You can pass ``legacy=True`` unconditionally to avoid version checks.
-  See :ref:`ref_legacy_meshing_workflow` for details.
-
-Choosing an interface
----------------------
-- Default (recommended): Use Enhanced Meshing Workflows.
-- Existing scripts: Initialize with ``legacy=True`` to use Legacy Meshing Workflows.
-  See :ref:`ref_legacy_meshing_workflow`.
-
-For example:
-
-  **Initialize a legacy workflow:**
-
-  ``watertight = meshing_session.watertight(legacy=True)``
-  ``fault_tolerant = meshing_session.fault_tolerant(legacy=True)``
-
-- Backward compatibility: Most behavior is preserved. For earlier behaviors and examples, see :ref:`ref_legacy_meshing_workflow`.
+- Legacy Meshing Workflows: The PyFluent meshing API used prior to Ansys Fluent 2026 R1 remains available.
+  For information on how to enable and use it, see :ref:`ref_legacy_meshing_workflow`.
 
 Watertight geometry meshing workflow
 ------------------------------------


doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -281,6 +341,32 @@ Update regions settings
...
+    )
+    fault_tolerant.setup_size_controls.object_selection_list = [
+        "inlet-1",
+        "inlet-2",
+        "inlet-3",
+    ]
+    fault_tolerant.setup_size_controls.add_child_and_update(defer_update=False)
+
+Choose mesh control options
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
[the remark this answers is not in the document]
prmukherj: This is addressed. Will push next.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -566,48 +705,118 @@ Current workflow
...
 
 .. code:: python
 
     watertight.import_geometry.mark_as_updated()
 
 
-Sample use of ``arguments``
+Renaming tasks in workflow
+--------------------------
+After a task is renamed, the new display name change its Python attribute name.
```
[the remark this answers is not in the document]
prmukherj: 👍

doc/changelog.d/4811.documentation.md
```diff
@@ -0,0 +1 @@
+Update meshing workflow docs and some dostrings
```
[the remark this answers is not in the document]
prmukherj: 👍

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -2,7 +2,33 @@
...
 Meshing workflow
 ================
-You can use PyFluent to access the new, enhanced meshing workflows.
+
+PyFluent provides access to Fluent's meshing workflows. A new, enhanced meshing workflow
+has existed for prior releases; in Fluent 26R1 the API was further refined and extended.
+This refinement is referred to as ``enhanced_api_261`` and is the default in Fluent 26R1
+and later.
+
+- To use the legacy workflow interface in Fluent 26R1 and later,
```
mkundu1: We already have separate pages for classic (legacy) and new meshing workflow in the guide. The `legacy=True` option in Fluent 26.1 is basically the classic meshing workflow. So, the mention of the `legacy=True` option can be removed from the top of the current page and optionally be introduced near the bottom pointing to the classic meshing workflow page. It can also be moved altogether to the classic meshing workflow page.

This page should focus solely on the new meshing workflow, without distracting users with other, suboptimal alternatives.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -566,48 +705,118 @@ Current workflow
...
+
+Known limitations
+-----------------
+- In ``enhanced_api_261``, switching between workflows and re-initializing workflows
+  within the same session are currently blocked.
+- Workaround: start a new session and initialize the required workflow again.
+
+.. Note::
+
+   This is a beta feature; user feedback is welcome.
```
mkundu1: Is the new API really beta? If not, we should remove this line as it will give indication that the new API is unstable (which I think is not).

prmukherj: yes, right. Will get rid of this one.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Actually this is an issue I was thinking that the docs will clear but it seems not to. 

The legacy means, the earlier implementation of the enhanced meshing workflow and not the classical one. From 26R1, we are defaulting to the new one, with server based infrastructure in the back-end. However, if we are using an earlier Fluent version we would default to the earlier implementation of the enhanced meshing workflow.

Strictly speaking the legacy argument makes sense only for 26R1 and above as here we can switch to the server based workflow or the client based (both being enhanced workflow) via. this argument.

Do you have any idea how to convey this via the docs?

@mkundu1, @seanpearsonuk

prmukherj: We don't want to mention server or client based to users as it is just implementation details with no effect on end usage.

mkundu1: Thanks for correcting me as there are actually 3 versions. Then, I think we can even skip mentioning the `legacy=True` option in the user-guide unless it provides something important which is not possible with the default option.

mkundu1: We have a [legacy API section](https://fluent.docs.pyansys.com/version/stable/user_guide/legacy/legacy_contents.html) in the user-guide. We can also consider moving the `legacy=True` doc in that section.

seanpearsonuk: Agreed.

doc/source/user_guide/meshing/new_meshing_workflows.rst
seanpearsonuk: This is a smart catch.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -2,7 +2,33 @@
...
+- To use the legacy workflow interface in Fluent 26R1 and later,
+  pass ``legacy=True`` when initializing a workflow.
+
+  For example:
+
+    **Initializing a legacy watertight workflow with Fluent version 26R1:**
+
+    ``watertight = meshing.watertight(legacy=True)``
+
+- Most functionality remains backward compatible; some APIs have updated names or structure.
```
seanpearsonuk: I'm not fond of the words, "some APIs"...

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -2,7 +2,33 @@
...
+and later.
+
+- To use the legacy workflow interface in Fluent 26R1 and later,
+  pass ``legacy=True`` when initializing a workflow.
+
+  For example:
+
+    **Initializing a legacy watertight workflow with Fluent version 26R1:**
+
+    ``watertight = meshing.watertight(legacy=True)``
```
seanpearsonuk: I might have comments on this part but I believe that there will be some movement here due to @mkundu1's comment(?) The basic point is that users wishing to use the old interface _can_ pass `legacy=True` also for earlier Fluent versions, but it's absolutely necessary for 26R1 in order to discriminate between API versions. It's a small point to let them know that their scripts are as portable as they should be.

prmukherj: yes, this will be moved.

In earlier versions, passing legacy=True, does not change any behavior as by default legacy=True, till 25R2. In 26R1, the non-legacy is default. Just to re-iterate.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -2,7 +2,33 @@
...
+
+- Most functionality remains backward compatible; some APIs have updated names or structure.
+  Differences are noted in the relevant sections below.
+
+
+Terminology and versioning
+--------------------------
+
+- ``enhanced_api_261``: The Fluent 26R1 enhancement of the enhanced workflow, offering
+  clearer task organization, improved traversal, and updated object names.
```
seanpearsonuk: Also, every item in the interface, including specific tasks and their arguments, is strongly-typed and documented in detail, which offers a new level of discoverability and usability.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Could you please suggest an appropriate title for it?
Thank you.

src/ansys/fluent/core/meshing/meshing_workflow.py
```diff
@@ -378,14 +378,15 @@ def get_current_workflow(
...
                 initialize=False, legacy=True
             )
 
-        # Check if this workflow type is active (returns False when active)
-        if _check_workflow_type(meshing_root, workflow_name):
+        if workflow_name == "Create New":
+            # Default to create_workflow if no specific type matches
             return _get_current_workflow(current_workflow, workflow_name) or factory(
                 initialize=False, legacy=True
             )
```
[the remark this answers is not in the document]
prmukherj: It is handled.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -2,7 +2,33 @@
...
+
+- Legacy workflow: The pre-26R1 interface. Use ``legacy=True`` to opt in on 26R1+.
```
seanpearsonuk: This reads a bit like internal memorandum-style documentation rather than user-facing docs. We need to anticipate a readership that includes more casual or first-time users.

Terms such as “pre-26R1 interface” and “26R1+” rely on internal shorthand and assume background knowledge that many readers won’t have. In particular, users are expected to infer that “26R1” corresponds to the Ansys 2026 R1 release, and then extrapolate the meaning of pre- and + from that. This creates an unnecessary mental leap, especially given that the versioning format used here doesn’t align with the official product naming used elsewhere in the documentation.

Even once that leap is made, “26R1+” remains ambiguous: it isn’t clear whether this is intended to include 2026 R1 itself or only subsequent releases.

It would be clearer to use the official release naming explicitly (e.g. “versions prior to Ansys 2026 R1” vs “Ansys 2026 R1 and later”) rather than relying on internal shorthand.

Additionally, the mention of legacy=True feels disconnected from the relevant code examples, which limits its usefulness. Showing this directly in context, or clearly linking to an example, would make it more actionable for readers.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -25,16 +51,24 @@ Import geometry
     )
     watertight = meshing_session.watertight()
     watertight.import_geometry.file_name.set_state(import_file_name)
-    watertight.import_geometry.length_unit.set_state('in')
+    watertight.import_geometry.length_unit.set_state("in")
     watertight.import_geometry()
 
+.. Note::
+   Fluent 26R1 and later use ``enhanced_api_261`` by default. To use the legacy interface:
```
seanpearsonuk: I can only find the format, **2026 R1** elsewhere in the documentation. We should avoid casual abbreviations.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -50,7 +84,7 @@ Describe geometry
 .. code:: python
 
     watertight.describe_geometry.update_child_tasks(setup_type_changed=False)
-    watertight.describe_geometry.setup_type.set_state("The geometry consists of only fluid regions with no voids")
+    watertight.describe_geometry.setup_type = "fluids"
```
seanpearsonuk: Surely, this change is wrong.

doc/source/user_guide/meshing/new_meshing_workflows.rst
mkundu1: 26.1 is not yet released to users. So, why we need to document the `legacy=True` + Fluent 26.1 combination? With Fluent 26.1, users will always use the enhanced, non-legacy PyFluent API.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -131,13 +167,42 @@ Import CAD and part management
...
+
     fault_tolerant.import_cad_and_part_management.context.set_state(0)
     fault_tolerant.import_cad_and_part_management.create_object_per.set_state("Custom")
     fault_tolerant.import_cad_and_part_management.fmd_file_name.set_state(import_file_name)
     fault_tolerant.import_cad_and_part_management.file_loaded.set_state("yes")
     fault_tolerant.import_cad_and_part_management.object_setting.set_state("DefaultObjectSetting")
     fault_tolerant.import_cad_and_part_management()
 
+.. Note::
+   Fluent 26R1 and later use ``enhanced_api_261`` by default. To use the legacy interface:
```
seanpearsonuk: 2026 R1

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -566,48 +705,118 @@ Current workflow
...
+
+The refined API enables straightforward traversal of tasks within a workflow:
+
+.. code:: python
+
+    watertight = meshing.watertight()
+
+    task_1 = watertight.first_child()
+    task_1.has_parent() is True
+    task_1.parent().__class__.__name__ == "WatertightMeshingWorkflow"
```
seanpearsonuk: The class-name string here feels brittle and unnecessary. Wouldn’t this be clearer?
```python
    task_1.parent() is watertight
```
Also, the bare `==` / `is True` expressions read like assertions but don’t actually evaluate to anything in the doc. If the intent is to validate behaviour, we should either use `assert` or show output. Otherwise, I think it would be clearer to demonstrate the API (e.g. via identity checks or `help()` output) rather than listing test-like predicates.

doc/source/user_guide/meshing/new_meshing_workflows.rst
mkundu1: Maybe - `Meshing workflows with Fluent 25.2 or earlier`

There we describe the enhanced, legacy API and how to reuse those scripts with Fluent 26.1 by passing `legacy=True`.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -566,48 +705,118 @@ Current workflow
...
+    task_6.name() == "Add Boundary Layers"
+
+This enables navigation without relying on Python attribute names.
+
+
+Known limitations
+-----------------
+- In ``enhanced_api_261``, switching between workflows and re-initializing workflows
+  within the same session are currently blocked.
+- Workaround: start a new session and initialize the required workflow again.
```
seanpearsonuk: I find this a bit unclear as a user.

* “blocked” feels stronger than what’s actually happening here: these operations are simply not supported within a single session.
* Two distinct operations (“switching workflows” vs “re-initializing a workflow”) are grouped together in a way that makes it unclear whether this is one limitation or two.
* It’s also ambiguous whether the workaround applies to both operations or only the latter.

Since these are separate commands with the same constraint and the same workaround, I think this would be clearer if they were listed explicitly as separate limitations, with the shared workaround made unambiguous.

seanpearsonuk:
```suggestion
Known limitations
-----------------
In ``enhanced_api_261``, the following operations are not supported within a single meshing session:
- Switching from one meshing workflow to another.
- Re-initializing a meshing workflow after it has already been initialized.

To perform either operation, start a new meshing session and initialize the required workflow.
```

src/ansys/fluent/core/meshing/meshing_workflow.py
```diff
@@ -378,14 +378,15 @@ def get_current_workflow(
 
     # Find active workflow type
     for workflow_name, factory in workflow_factories.items():
-        if workflow_name == "Create New":
-            # Default to create_workflow if no specific type matches
+
+        # Check if this workflow type is active (returns True when active)
```
seanpearsonuk: This comment could be rendered unnecessary by renaming the called function.

doc/source/user_guide/meshing/new_meshing_workflows.rst
seanpearsonuk: I’m not suggesting that `legacy=True` changes behaviour in earlier versions. In fact, I agree that it doesn’t.

My point is about *how we guide users to write portable scripts*. Users who want legacy semantics can safely pass `legacy=True` for **all** Fluent versions, which avoids version checks and keeps scripts working unchanged across 25R2 → 26R1+.

The current wording frames `legacy=True` as something that is only relevant for 26R1+, which may inadvertently encourage users to write version-dependent code when they don’t need to.

A short note that `legacy=True` may be passed unconditionally for portability would make this clearer.

doc/source/user_guide/meshing/new_meshing_workflows.rst
seanpearsonuk, line 1: Aliasing is underused in the code snippets. Hence the repetitive, long object chains. It's not a great look.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Got the point. Thank you @seanpearsonuk.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Improved aliasing.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Implemented the suggestion. Thank you.

src/ansys/fluent/core/meshing/meshing_workflow.py
```diff
@@ -378,14 +378,15 @@ def get_current_workflow(
...
+        if _check_workflow_type(meshing_root, workflow_name):
             return _get_current_workflow(current_workflow, workflow_name) or factory(
                 initialize=False, legacy=True
             )
 
-        # Check if this workflow type is active (returns False when active)
-        if _check_workflow_type(meshing_root, workflow_name):
+        if workflow_name == "Create New":
```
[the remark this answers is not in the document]
prmukherj: It has been handled via if else.

src/ansys/fluent/core/meshing/meshing_workflow.py
```diff
@@ -333,7 +333,7 @@ def __init__(
             self._activate_dynamic_interface(dynamic_interface=True)
 
 
-def _check_workflow_type(meshing_root, name: str):
+def _is_workflow_active(meshing_root, name: str):
```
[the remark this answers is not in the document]
prmukherj: No, it returns as intended.

src/ansys/fluent/core/meshing/meshing_workflow.py
```diff
@@ -371,21 +371,19 @@ def get_current_workflow(
...
                 initialize=False, legacy=True
             )
-
-        # Check if this workflow type is active (returns False when active)
-        if _check_workflow_type(meshing_root, workflow_name):
+        else:
+            # Default to create_workflow if no specific type matches
             return _get_current_workflow(current_workflow, workflow_name) or factory(
                 initialize=False, legacy=True
             )
```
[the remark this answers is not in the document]
prmukherj: This is a fail safe check.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Done.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Improved.

src/ansys/fluent/core/meshing/meshing_workflow.py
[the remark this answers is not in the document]
prmukherj: Please ignore this, it is a failsafe and will hit the else if there is no matching instance found.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -547,16 +617,50 @@ Duplicate tasks
...
+       >>> watertight.import_boi_geometry[2]
+       task < import_boi_geometry: 2 >
+
+   Index 0 returns the first instance; calling the task or indexing with 0 are equivalent.
+
+   After inserting the tasks above, the workflow contains:
+
+   .. code:: python
+
+       >>> watertight.children()
```
prmukherj, line 622: Please note that the indentation issue here has already been resolved. Will be pushed.

doc/source/user_guide/legacy/meshing_workflow.rst
```diff
@@ -0,0 +1,103 @@
...
+This page documents the legacy meshing workflow interface that existed in PyFluent
+versions prior to using Ansys Fluent 2026 R1. Use the enhanced workflow for all new scripts.
+If you need to keep older scripts working without changes, you can initialize a workflow
+with legacy semantics by passing ``legacy=True``.
+
+Portability note
+----------------
+You can pass ``legacy=True`` unconditionally in PyFluent using any Fluent version, including
+Ansys Fluent 2026 R1 and later. This lets you avoid version checks and keep scripts
+written for earlier releases working as-is.
```
seanpearsonuk: I have a few concerns about clarity and scope on this page:

* The title “Legacy meshing workflows (prior to Ansys Fluent 2026 R1)” is misleading since the legacy workflow interface is still available in 26R1 and will continue to be available in 27R1. This page documents a legacy *interface*, not something that only exists prior to 26R1.
* The sentence “Use the enhanced workflow for all new scripts” is vague and not actionable. If the intent is to steer users away from this page, it should use the full, proper name of the preferred API and include a hyperlink.
* The term “legacy” is historically overloaded here. What is now called “legacy” was originally branded as “enhanced”, and predates the current enhanced workflow interface. None of that context is clear from the text, and users could easily confuse this with Fluent’s classic workflows.
* More generally, this page feels like the wrong place to give an overview of enhanced vs legacy workflows. That’s a higher-level concern that would fit better in an overview or migration section. This page should probably focus on documenting the legacy interface itself and how to opt into it.
* The text refers to `legacy=True` in multiple places without stating **where** it is passed. It’s not obvious that this is an argument to workflow initialization, or which APIs accept it.
* The phrase “initialize a workflow with legacy semantics” is unclear and potentially misleading. The workflow itself and Fluent backend are unchanged. What differs is the PyFluent interface exposed to the user. Using “semantics” here feels vague and may confuse readers.

I think tightening the scope and adjusting the title and guidance would make this much clearer for users.

doc/source/user_guide/legacy/meshing_workflow.rst
```diff
@@ -0,0 +1,103 @@
...
+In the legacy fault-tolerant workflow, Part and File Management use camel-case
+attributes and methods accessed directly on the meshing session like
+``meshing_session.PartManagement.*``,  ``meshing_session.PMFileManagement.FileManager.*``,
+``InputFileChanged`` and ``LoadFiles``.
+
+.. Note::
+   Backward compatible: These camel-case Part/File Management entries are still
+   supported in Ansys Fluent 2026 R1 for legacy scripts. The recommended approach
+   in 2026 R1 and later is to use the enhanced workflow attributes, for example
+   ``fault_tolerant.parts.*`` and ``fault_tolerant.parts_files.file_manager.*``.
```
seanpearsonuk: I find this paragraph quite hard to follow as a reader, especially in a section that is meant to document the *legacy* fault-tolerant workflow:

* Referring to the entries as “camel-case” feels unnecessarily abstract. The concrete names (`PartManagement`, `PartManagementFileManagement`) are what users actually see and search for, and could just be named directly.
* This legacy subsection repeatedly references the enhanced workflow API and “recommended approaches”, which breaks separation of concerns. It’s not clear from the wording whether `fault_tolerant.parts.*` and `fault_tolerant.parts_files.file_manager.*` are being mentioned for comparison only, or whether they are expected to exist in the legacy interface.
* Version-specific framing (“in Ansys Fluent 2026 R1”, “in 2026 R1 and later”) is being reintroduced here even though the legacy vs enhanced domains have already been established earlier on the page, which makes the mental model harder rather than clearer.
* Overall, I think less would be more here. Focusing this subsection purely on what is supported in the legacy interface, without cross-referencing newer APIs or versions, would make it much clearer.

Cross-API guidance and recommendations would fit better in a higher-level overview or migration section rather than inside a legacy reference page.

doc/source/user_guide/legacy/meshing_workflow.rst
seanpearsonuk, line 1: Some need for aliasing in the code snippets.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -2,14 +2,46 @@
 
 Meshing workflow
 ================
-You can use PyFluent to access the new, enhanced meshing workflows.
+
+PyFluent provides access to Fluent’s meshing workflows. When using PyFluent with
+Ansys Fluent 2026 R1, the enhanced workflow API is refined and extended for clearer
+task organization, easier navigation, and stronger typing.
+
+Choosing an interface
```
seanpearsonuk: You are correct to position this guidance here. The legacy documentation meanwhile is clearly positioned in the legacy section with legacy labelling, so it should stand alone and not call back to the enhanced workflow whatsoever. In doing so, it only creates confusion.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -2,14 +2,46 @@
 
 Meshing workflow
 ================
-You can use PyFluent to access the new, enhanced meshing workflows.
+
+PyFluent provides access to Fluent’s meshing workflows. When using PyFluent with
+Ansys Fluent 2026 R1, the enhanced workflow API is refined and extended for clearer
+task organization, easier navigation, and stronger typing.
```
seanpearsonuk: For maxium clarity, we must first define Enhanced Meshing Workflows categorically as an API that is only available with Ansys Fluent 2026 R1 onward. We then state its characteristics as you have done. We then state that the API previously branded as Enhanced Meshing Workflows prior to the release of Ansys Fluent 2026 R1 is now branded as Legacy Meshing Workflows. Given such clear definitions here, we need never mention them again anywhere in the documentation.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -2,14 +2,46 @@
...
+
+Choosing an interface
+---------------------
+- Default (recommended): Use the enhanced workflow in PyFluent using Ansys Fluent 2026 R1 and later.
```
seanpearsonuk: using &rarr; with

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -2,14 +2,46 @@
...
+  It offers clearer task organization, improved traversal, and strongly typed,
+  well-documented arguments.
+- Legacy (for existing scripts): If you want legacy semantics, pass ``legacy=True`` when
+  initializing a workflow. You can pass ``legacy=True`` unconditionally in any Fluent version,
```
seanpearsonuk: Pass it where?

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -2,14 +2,46 @@
...
+PyFluent provides access to Fluent’s meshing workflows. When using PyFluent with
+Ansys Fluent 2026 R1, the enhanced workflow API is refined and extended for clearer
+task organization, easier navigation, and stronger typing.
+
+Choosing an interface
+---------------------
+- Default (recommended): Use the enhanced workflow in PyFluent using Ansys Fluent 2026 R1 and later.
+  It offers clearer task organization, improved traversal, and strongly typed,
+  well-documented arguments.
+- Legacy (for existing scripts): If you want legacy semantics, pass ``legacy=True`` when
```
seanpearsonuk: What are "legacy semantics"?

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -2,14 +2,46 @@
...
+  initializing a workflow. You can pass ``legacy=True`` unconditionally in any Fluent version,
+  including Ansys Fluent 2026 R1 and later, to keep legacy scripts running without version checks.
```
seanpearsonuk: The wording needs attention here. Where is this text coming from? Readers will have difficulty with this documentation as it stands.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -2,14 +2,46 @@
...
+    ``watertight = meshing.watertight(legacy=True)``
+
+- Backward compatibility: The enhanced workflow preserves most behavior. For examples and features
+  that existed in versions prior to Ansys Fluent 2026 R1, see :ref:`ref_legacy_meshing_workflow`.
+
+
+Terminology and versioning
+--------------------------
+
+- ``enhanced_api_261``: PyFluent using Ansys Fluent 2026 R1 with an enhanced workflow interface, offering
```
seanpearsonuk: - The fact that Ansys Fluent 2026 R1 is needed is already established at this point.
- The reader knows that they're on the _enhanced_ workflow page, so no need to keep mentioning it.
- In general, define the scope at the top, and then just document the usage.
- There's a lot of text, apparently employed for clarification, which is only causing confusion and wasting the reader's time.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -2,14 +2,46 @@
...
+  clearer task organization, improved traversal, and updated object names. Every item in the
+  interface (tasks and their arguments) is strongly typed and documented for discoverability and usability.
+
+- Legacy workflow: The interface used in PyFluent versions prior to using Ansys Fluent 2026 R1. It remains available
+  in Ansys Fluent 2026 R1 and later by passing ``legacy=True`` when initializing a workflow.
```
seanpearsonuk: Pass it where?

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Just in one place the legacy one refers this as of now. I think it makes sense. Shall we remove it?

doc/source/user_guide/legacy/meshing_workflow.rst
```diff
@@ -0,0 +1,126 @@
+.. _ref_legacy_meshing_workflow:
+
+Legacy meshing interface (opt-in)
```
seanpearsonuk: (opt-in) not helpful in title.

doc/source/user_guide/legacy/meshing_workflow.rst
```diff
@@ -0,0 +1,126 @@
...
+=================================
+
+This page documents the legacy PyFluent meshing interface. It remains available when using
```
seanpearsonuk: That "This page documents the legacy PyFluent meshing interface" is already indicated by the title.

doc/source/user_guide/legacy/meshing_workflow.rst
prmukherj: Right, this can be removed now as we have defined the terms in the new meshing workflow docs.

doc/source/user_guide/legacy/meshing_workflow.rst
prmukherj: Done.

doc/source/user_guide/legacy/meshing_workflow.rst
prmukherj: Done.

doc/source/user_guide/legacy/meshing_workflow.rst
```diff
@@ -0,0 +1,126 @@
...
+PyFluent with Ansys Fluent 2026 R1 and later (for example, Ansys Fluent 2027 R1). Use it to
+run existing scripts without changes.
+
```
seanpearsonuk, line 8: The information about versioning is too indirect. We can easily just specify the supported version range.

Regarding "Use it to run existing scripts without changes", either the reader knows why they are looking at this page or they don't. If they do then the statement is superfluous. If they don't then the statement is confusing.

There are still too many words on this page.

doc/source/user_guide/legacy/meshing_workflow.rst
```diff
@@ -0,0 +1,126 @@
...
+.. Note::
+    This page documents the legacy PyFluent meshing interface for the API-based meshing infrastructure.
```
seanpearsonuk: This is practically a duplication of a few lines ago, which was already superfluous.

The meshing workflow landing page is enough to provide the defintions all in one go.

prmukherj: Resolved.

doc/source/user_guide/legacy/meshing_workflow.rst
prmukherj: Made this one short and to the point as well.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: No reference in the old file to the new.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -1,15 +1,39 @@
...
+Meshing workflows
+=================
+
+PyFluent provides access to Fluent’s meshing workflows.
+
+Definitions
+-----------
+- Enhanced Meshing Workflows: A PyFluent API available only when using PyFluent with Ansys Fluent 2026 R1 and later.
+  It provides clearer task organization, easier navigation, and strongly typed, well-documented arguments.
+- Legacy Meshing Workflows: The PyFluent meshing API used prior to Ansys Fluent 2026 R1
```
mkundu1: I feel information under this section (till line 30) have duplicates in doc/source/user_guide/legacy/meshing_workflow.rst. Here we can say:
```
The PyFluent meshing API used prior to Ansys Fluent 2026 R1 remains available in Ansys Fluent 2026 R1 and later 
releases. For information on how to enable and use it, see :ref:ref_legacy_meshing_workflow.
```
We can even move the above text to a Note from the bullet point.

(I assumed that the script from line 32 onwards in the current page works only for 2026 R1 and later. Please ignore my comment if that assumption is not correct.)

prmukherj: Thanks @mkundu1, The choosing an interface was mostly redundant and have removed it. The re-writing that you provided is applied as a section in the definition now changed to overview.

prmukherj: Done.

PR #4824, opened 2026-01-20
build: limit major version on ansys-tools-filetransfer project

RobPasMue:
## Context
The previous limit impedes installing the main branch of the filetransfer repo together with PyFluent, without any real motive. I understand limiting the major version but minor versions should remain without breaking changes in that repository.

## Change Summary
Change the major version

## Rationale
Compatibility

## Impact
None

### DIFF
--- a/doc/changelog.d/4826.maintenance.md
+++ b/doc/changelog.d/4826.maintenance.md
@@ -1,0 +1,1 @@
+Update the usage of Variable Descriptors in svars.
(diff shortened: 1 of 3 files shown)


prmukherj (approved):

hpohekar:
> Any reason why the workflows are not being triggered on this PR?

@RobPasMue This is a common issue as status check is not propagated correctly from the workflow run.

RobPasMue:
See https://github.com/ansys/pyfluent/actions/runs/21163417330

Documentation build fails for an unrelated error

PR #4832, opened 2026-01-21
chore: update battery pack and transient compressible nozzle examples [skip tests]

prmukherj:
## Context
The required arguments were not passed to Monitor in couple of examples and the GraphicsWindow was wrongly initialized in one of them.

## Change Summary
1. Default arguments passed to Monitor
2. GraphicsWindow() initialized without argument. It might take renderer string or renderer object as input, not the solver object.

## Impact
Corrected user facing examples.

### DIFF
--- a/doc/changelog.d/4826.maintenance.md
+++ b/doc/changelog.d/4826.maintenance.md
@@ -1,0 +1,1 @@
+Update the usage of Variable Descriptors in svars.
--- a/src/ansys/fluent/core/services/solution_variables.py
+++ b/src/ansys/fluent/core/services/solution_variables.py
@@ -608,6 +608,11 @@ class SolutionVariableData:
         This array can be populated  with values to set SVAR data.
         """
         self._update_solution_variable_info()
+        variable_name = self._allowed_solution_variable_names.valid_name(
+            variable_name,
+            [zone_name],
+            domain_name,
+        )
 
         zones_info = self._solution_variable_info.get_zones_info()
         if zone_name in zones_info.zone_names:
@@ -720,6 +725,11 @@ class SolutionVariableData:
         None
         """
         self._update_solution_variable_info()
+        variable_name = self._allowed_solution_variable_names.valid_name(
+            variable_name,
+            list(zone_names_to_data.keys()),
+            domain_name,
+        )
         domain_id = self._allowed_domain_names.valid_name(domain_name)
         zone_ids_to_svar_data = {
             self._allowed_zone_names.valid_name(zone_name): solution_variable_data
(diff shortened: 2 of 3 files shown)


examples/00-fluent/battery_pack.py
```diff
@@ -385,12 +385,12 @@
 # Display monitor plots
 # ---------------------
 
-plot_window = GraphicsWindow(solver)
+plot_window = GraphicsWindow()
```
prmukherj, line 388: @MohammedAnsys , please be aware that the "GraphicsWindow" either takes renderer object or renderer string as input and not the solver object.

Please note the recent updates on the visualization repo regarding required arguments like 'monitor_set_name'

Thank you.

examples/00-fluent/battery_pack.py
```diff
@@ -385,12 +385,12 @@
...
 
-voltage_rplot = Monitor(solver, monitor_set_name="voltage_surface_areaavg-rplot")
+voltage_rplot = Monitor(solver=solver, monitor_set_name="voltage_surface_areaavg-rplot")
 plot_window.add_plot(voltage_rplot, position=(0, 0))
 
-temp_rplot = Monitor(solver, monitor_set_name="volume_max_temp-rplot")
+temp_rplot = Monitor(solver=solver, monitor_set_name="volume_max_temp-rplot")
 plot_window.add_plot(temp_rplot, position=(1, 1))
 
 plot_window.renderer = "matplotlib"
```
prmukherj: We know that matplotlib is not supported in PyConsole (Fluent) because matplotlib is in an older version there. Shall we document it anywhere?

Like in PyConsole docs we can mention that changing renderer is not allowed in PyConsole and user needs to use default renderer.

Please let me know your thoughts @seanpearsonuk and @mkundu1

seanpearsonuk: @prmukherj Is there a specific reason why `matplotlib` was chosen here?

prmukherj: Not sure, maybe @MohammedAnsys can answer this better, but I can confirm that it works fine with PyVista as well.

seanpearsonuk: > Not sure, maybe @MohammedAnsys can answer this better, but I can confirm that it works fine with PyVista as well.

We can just remove that line then.

prmukherj: yes, @seanpearsonuk. I agree as well

prmukherj: Done.

Gobot1234 (approved):

Gobot1234:
This doesn't update catalytic_converter which I believe has a similar issue.

I can see `monitor = Monitor(solver_session)` in its code at the very least

PR #4833, opened 2026-01-21
docs: Example on ventilation flow simulation [skip tests]

MohammedAnsys:
## Context
No existing example demonstrated ventilation system analysis with multiple inlets and exhaust fans for indoor air quality applications.

## Change Summary
Added ventilation_workflow.py example simulating airflow in a mechanically ventilated poultry house with 32 inlet vents and 6 exhaust fans. Includes k-epsilon turbulence model, SIMPLEC solver, pathline/contour visualizations, and quantitative velocity analysis. Added 3 supporting images.
## Rationale
Provides users with a practical reference for HVAC/ventilation CFD workflows using PyFluent

## Impact
additive, no modifications to existing code

The example tries to follow the provided guidelines with https://github.com/ansys/pyfluent/discussions/4378 and uses the new API implementation.

Thanks.

### DIFF
--- a/doc/changelog.d/4843.documentation.md
+++ b/doc/changelog.d/4843.documentation.md
@@ -1,0 +1,1 @@
+Update radiation headlamp example due to no backwards compatibility of reference_direction [skip tests]
--- a/examples/00-fluent/radiation_headlamp.py
+++ b/examples/00-fluent/radiation_headlamp.py
@@ -303,7 +303,7 @@ models.viscous.model = "laminar"
 
 radiation = models.radiation
 radiation.model = "monte-carlo"
-radiation.monte_carlo.number_of_histories = 1e7
+radiation.solve_frequency.number_of_histories = 10000000
 radiation.solve_frequency.iteration_interval = 20
 
 ###############################################################################
@@ -366,7 +366,9 @@ plastic.refractive_index.value = 1
 # ~~~~~~~~~~~~~~~~~~~~
 # Set the cell zone conditions for the bezel and the lens.
 
-solver_session.settings.setup.cell_zone_conditions.solid["bezel"].material = "plastic"
+solver_session.settings.setup.cell_zone_conditions.solid["bezel"].general.material = (
+    "plastic"
+)
 solver_session.settings.setup.cell_zone_conditions.copy(
     from_="bezel",
     to=[
@@ -380,7 +382,7 @@ solver_session.settings.setup.cell_zone_conditions.copy(
 )
 
 lens_cellzone_conds = solver_session.settings.setup.cell_zone_conditions.solid["lens"]
-lens_cellzone_conds.material = "glass"
+lens_cellzone_conds.general.material = "glass"
 lens_cellzone_conds.general.participates_in_radiation = True
 
 ###########################################################################################################
@@ -396,7 +398,7 @@ lens_cellzone_conds.general.participates_in_radiation = True
 
 bezel_enc_bc = solver_session.settings.setup.boundary_conditions.wall["bezel-enclosure"]
 bezel_enc_bc.thermal.material = "plastic"
-bezel_enc_bc.radiation.radiation_bc = "Opaque"
+bezel_enc_bc.radiation.bc_type = "Opaque"
 bezel_enc_bc.radiation.internal_emissivity = 1
 bezel_enc_bc.radiation.diffuse_irradiation_settings.diffuse_fraction_band = {"s-": 1}
 
@@ -420,7 +422,7 @@ solver_session.settings.setup.boundary_conditions.copy(
 
 enc_lens_bc = solver_session.settings.setup.boundary_conditions.wall["enclosure-lens"]
 enc_lens_bc.thermal.material = "glass"
-enc_lens_bc.radiation.radiation_bc = "Semi Transparent"
+enc_lens_bc.radiation.bc_type = "Semi Transparent"
 enc_lens_bc.radiation.diffuse_irradiation_settings.diffuse_fraction_band = {"s-": 0}
 
 # Copy enclosure-lens BC to other lens boundary
@@ -439,7 +441,7 @@ enc_rim_bezel_bc = solver_session.settings.setup.boundary_conditions.wall[
     "enclosure-rim-bezel"
 ]
 enc_rim_bezel_bc.thermal.material = "plastic"
-enc_rim_bezel_bc.radiation.radiation_bc = "Opaque"
+enc_rim_bezel_bc.radiation.bc_type = "Opaque"
 enc_rim_bezel_bc.radiation.internal_emissivity = 0.16
 enc_rim_bezel_bc.radiation.diffuse_irradiation_settings.diffuse_fraction_band = {
     "s-": 0.1
@@ -479,7 +481,7 @@ rad_input_bc.radiation.boundary_source = True
 rad_input_bc.radiation.direct_irradiation_settings.direct_irradiation = {
     "Full-spectrum": {"option": "value", "value": 1200}
 }
-rad_input_bc.radiation.direct_irradiation_settings.reference_direction = [
+rad_input_bc.radiation.direct_irradiation_settings.beam_direction = [
     -0.848,
     0,
     -0.53,


examples/00-fluent/ventilation_workflow.py
```diff
@@ -0,0 +1,287 @@
...
+Wind Flow Through A Mechanically Ventilated Poultry House
+-------------------------------------------------------------
+"""
+# %%
+# Objective
+# ---------
+#
+# The primary objective of this study is to investigate airflow patterns, air distribution,
+# and ventilation effectiveness inside a mechanically ventilated poultry house
+# using PyFluent. The simulation represents a typical poultry housing configuration with
```
seanpearsonuk: @MohammedAnsys This is a great set of introductory paragraphs and exactly what we need for maximum impact and accessibility. Nice work.

One small wording suggestion below: it slightly restructures the first sentence to avoid a momentary stumble for the reader, but the intent and content are unchanged.

```suggestion
# The primary objective of this PyFluent study is to investigate airflow patterns, air distribution,
# and ventilation effectiveness inside a mechanically ventilated poultry house. 
# The simulation represents a typical poultry housing configuration with
```

examples/00-fluent/ventilation_workflow.py
```diff
@@ -0,0 +1,287 @@
...
+# Setup
+# -----
+#
+# Configure the simulation settings including turbulence model, boundary conditions,
+# and solution methods.
+#
+# Viscous Model
+# ^^^^^^^^^^^^^
+viscous = Viscous(solver)
+viscous.model = "k-epsilon"
```
hpohekar: We can use an enum as follows:

`viscous.model = viscous.model.K_EPSILON`

examples/00-fluent/ventilation_workflow.py
```diff
@@ -0,0 +1,287 @@
...
+# %%
+# Launch Fluent in solver mode
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# .. note::
+#   To enable the graphical user interface (GUI), import the ``UIMode`` class
+#   and set the UI mode before launching the solver::
+#
+#       from ansys.fluent.core import UIMode
+#       ui_mode = UIMode.GUI
```
seanpearsonuk: @MohammedAnsys This is a thoughtful addition 👍

As written, though, the note suggests that assigning `ui_mode = UIMode.GUI` “before launching the solver” would have an effect on its own, which isn’t the case. My assumption is that the intent is to show that `ui_mode=UIMode.GUI` can be passed as a `launch_fluent()` argument.

Since this kind of guidance is mainly for inexperienced users, it needs to be either complete and correct, or omitted. An alternative would be to point readers to the PyFluent documentation for available launch options, which is more flexible and avoids maintenance issues. If we agree on that approach, the same pattern could be applied consistently across examples.

examples/00-fluent/ventilation_workflow.py
```diff
@@ -0,0 +1,287 @@
...
+# %%
+# Boundary conditions
+# ^^^^^^^^^^^^^^^^^^^
+
+boundary_conditions = BoundaryConditions(solver)
+boundary_conditions.set_zone_type(new_type="inlet-vent", zone_list=["*inlet_vent_*"])
+
+pressure_outlets = PressureOutlets(solver)
+# %%
+# The target mass flow rate option is enabled to ensure that each exhaust outlet (fan) removes a specified and controlled amount of air
```
seanpearsonuk: @MohammedAnsys Recognizing that the comment helps to provide additional clarification in this case 👍

examples/00-fluent/ventilation_workflow.py
```diff
@@ -0,0 +1,287 @@
...
+pressure_outlets["outlet_1"] = {
+    "momentum": {
+        "target_mass_flow_rate": True,
+        "target_mass_flow": {
+            "value": 12.25  # kg/s  Represents the design airflow capacity of one exhaust fan
+        },
+    }
+}
+
+boundary_conditions.copy(from_="outlet_1", to=["*outlet_*"])
```
seanpearsonuk: @MohammedAnsys It's great to show this capability, partly because it forces the data transfer to happen within the server, which is a performance gain especially for large problems :+1:.

Just out of interest, I believe that `pressure_outlets["outlet_*"] = {...}` also works unless I'm misremembering.

examples/00-fluent/ventilation_workflow.py
MohammedAnsys: I agree. It would be clearer and more robust to point readers directly to the PyFluent documentation for the available ``launch_fluent()`` options.
Will update accordingly

examples/00-fluent/ventilation_workflow.py
seanpearsonuk: @hpohekar Great catch, thanks.

examples/00-fluent/ventilation_workflow.py
```diff
@@ -0,0 +1,287 @@
...
+# making it less sensitive to mesh irregularities and more stable than node-based methods
+
+# %%
+monitor_residuals = Residual(solver)
+monitor_residuals.equations["continuity"].absolute_criteria = 1e-06
+monitor_residuals.equations["x-velocity"].absolute_criteria = 1e-06
+monitor_residuals.equations["y-velocity"].absolute_criteria = 1e-06
+monitor_residuals.equations["z-velocity"].absolute_criteria = 1e-06
+monitor_residuals.equations["k"].absolute_criteria = 1e-06
+monitor_residuals.equations["epsilon"].absolute_criteria = 1e-06
```
seanpearsonuk: @MohammedAnsys Some would certainly argue that `dict` based assignment is preferable here. It's certainly faster, which becomes significant in remote contexts. It can also be a slightly smoother read (the difference is exaggerated here by the duplication of `monitor_residuals.equations`).

For the sake of reducing duplication further, does something like this have the same effect:
```python
monitor_residuals.equations =
    {
        eqn: {"absolute_criteria": 1e-6}
        for eqn in (
            "continuity",
            "x-velocity",
            "y-velocity",
            "z-velocity",
            "k",
            "epsilon",
        )
    }
```

examples/00-fluent/ventilation_workflow.py
```diff
@@ -0,0 +1,287 @@
...
+
+pressure_outlets = PressureOutlets(solver)
+# %%
+# The target mass flow rate option is enabled to ensure that each exhaust outlet (fan) removes a specified and controlled amount of air
+
+pressure_outlets["outlet_1"] = {
+    "momentum": {
+        "target_mass_flow_rate": True,
+        "target_mass_flow": {
+            "value": 12.25  # kg/s  Represents the design airflow capacity of one exhaust fan
```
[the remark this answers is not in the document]
Gobot1234: Can I suggest using `from ansys.units.common import kg, s` and then this should work which I think is nicer 
```suggestion
            "value": 12.25 * kg / s  # Represents the design airflow capacity of one exhaust fan
```

examples/00-fluent/ventilation_workflow.py
```diff
@@ -0,0 +1,287 @@
...
+# Residual convergence: to ensure a high level of numerical accuracy and solution stability in the simulation
+
+# %%
+# Initialize and run
+# ^^^^^^^^^^^^^^^^^^
+solution_initialization = Initialization(solver)
+solution_initialization.initialization_type = "hybrid"
+solution_initialization.initialize()
+
+calculation = RunCalculation(solver)
```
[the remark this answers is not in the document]
seanpearsonuk: @MohammedAnsys This is a very pertinent comment.

In practice, `iter_count` varies widely across our PyFluent examples (from very small values up to 1000), and in most cases there’s little or no explanation for why a particular value was chosen. Where comments *do* exist, they’re sometimes ambiguous; for example, references to “the tutorial”, when for our users these Python scripts *are* the tutorials. That kind of wording can create uncertainty rather than clarity.

More broadly, our examples appear to be a mixture of fully converged calculations and intentionally unconverged, demonstration-style runs, but this distinction isn’t made explicit or applied consistently. That makes it hard for users to know what level of physical validity to expect from a given example.

This goes beyond this PR, so I’ll raise a GitHub issue to address it systematically. For this example specifically, I’d prefer either a clear statement of intent (converged vs demonstrative) or a more general reference to convergence criteria rather than a hard justification of a specific iteration count.

examples/00-fluent/ventilation_workflow.py
Gobot1234: Maybe better still
```py
dict.fromkeys({
        "continuity",
        "x-velocity",
        "y-velocity",
        "z-velocity",
        "k",
        "epsilon",
    }, {"absolute_criteria": 1e-6}
)
```

examples/00-fluent/ventilation_workflow.py
```diff
@@ -0,0 +1,287 @@
...
+velocity_pathline = Pathlines(solver)
+
+velocity_pathline["pathlines-1"] = {
+    "field": "velocity-magnitude",
+    "release_from_surfaces": ["*inlet_vent_*"],
+}
+velocity_pathline["pathlines-1"].display()
+
+graphics.picture.x_resolution = 650  # Horizontal resolution for clear visualization
+graphics.picture.y_resolution = 450  # Vertical resolution matching typical aspect ratio
```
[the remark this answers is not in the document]
seanpearsonuk: We should mention it here too: https://github.com/ansys/pyfluent/discussions/4378

examples/00-fluent/ventilation_workflow.py
```diff
@@ -0,0 +1,309 @@
...
+# The target mass flow rate option is enabled to ensure that each exhaust outlet
+# (fan) removes a specified and controlled amount of air.
+
+pressure_outlets["outlet_*"] = {
+    "momentum": {
+        "target_mass_flow_rate": True,
+        "target_mass_flow": {
+            "value": 12.25
+            * kg
+            / s  # Represents the design airflow capacity of one exhaust fan
```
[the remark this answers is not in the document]
Gobot1234: Goodness me that gets formatted awfully by the linter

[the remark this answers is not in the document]
seanpearsonuk: 😒

examples/00-fluent/ventilation_workflow.py
```diff
@@ -0,0 +1,306 @@
+# Copyright (C) 2021 - 2025 ANSYS, Inc. and/or its affiliates.
```
prmukherj: @MohammedAnsys, please update the license header for this file. Please refer to other files.
2026.

PR #4853, opened 2026-01-23
chore: change ansys-units range

hpohekar:
## Context
The range was too broader.

## Change Summary
The range is removed.

## Rationale
Compatible version only allowed.

## Impact
pyproject.toml is updated.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ dependencies = [
     "ansys-platform-instancemanagement~=1.1",
     "ansys-tools-common>=0.4.0",
     "ansys-tools-filetransfer>=0.2,<1.0",
-    "ansys-units==0.10.0",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -22,7 +22,7 @@ dependencies = [
     "ansys-platform-instancemanagement~=1.1",
     "ansys-tools-common>=0.4.0",
     "ansys-tools-filetransfer>=0.2,<1.0",
-    "ansys-units>=0.10.0,<1.0",
+    "ansys-units==0.10.0",
```
mkundu1: `ansys-units~=0.10.0` is better as that will allow patch versions (0.10.*)

hpohekar:
```suggestion
    "ansys-units~=0.10.0",
```

PR #4867, opened 2026-01-28
fix: Remove insecure options

hpohekar:
## Context
Insecure options are for internal testing only.

## Change Summary
Insecure  options replaced with MTLS options.

## Rationale
MTLS options work on all platforms.

## Impact
File transfer service workflow.


tests/test_file_transfer_service.py
```diff
@@ -28,14 +28,19 @@
...
 
 
-@pytest.mark.skip(reason="https://github.com/ansys/pyfluent/issues/4649")
+def _get_certs_folder():
+    return os.path.join(os.getcwd(), "certs")
+
+
+@pytest.mark.skip(
+    reason="Works fine locally and on Test Custom Run workflow, fails on CI."
+)
```
seanpearsonuk, line 43: @hpohekar 

I have a couple of thoughts that arise due to the current `@pytest.mark.skip` usage here:

1. The skip reason only states the symptom (“works locally, fails on CI”) but not our current level of understanding of the failure (unknown / partially understood / understood but unresolved). Even an explicit statement that the root cause is unknown would help future maintainers who don’t have today’s offline context.

2. We have seen sets of near-duplicate skip reasons, including this one, across the test suite. For consistency, searchability, and maintainability, it would be good to converge on a small set of canonical skip reasons (possibly as string constants), with optional adjacent comments for test-specific context.

I think tightening this up will save time and confusion later.

**The work will be assigned separately in a new issue.**

hpohekar: Yes, I agree. Done. Thank you.

seanpearsonuk (approved):

PR #4883, opened 2026-02-02
docs: Example on pipe flow [skip tests]

MohammedAnsys:
## Context
No existing example demonstrated thermal energy transport with constant wall temperature boundary conditions and validation against analytical heat transfer solutions.

## Change Summary
Added pipe_flow example simulating oil flow through a pipe submerged in icy lake water. Includes laminar flow setup, energy equation, custom oil material properties, constant wall temperature (0°C), temperature profile visualization, and analytical validation using Nusselt number correlations. Added 2 supporting images.

## Rationale
Provides a validated benchmark for conjugate heat transfer in PyFluent

## Impact
Purely additive, no modifications to existing code

### DIFF
--- a/doc/changelog.d/4883.documentation.md
+++ b/doc/changelog.d/4883.documentation.md
@@ -1,0 +1,1 @@
+Example on pipe flow [skip tests]


examples/00-fluent/pipe_flow.py
```diff
@@ -0,0 +1,318 @@
...
+# %%
+# Materials
+# ^^^^^^^^^
+
+fluid_materials = FluidMaterials(solver)
+
+fluid_materials.rename(new="oil", old="air")
+fluid_materials["oil"] = {
+    "density": 888.1,
+    "viscosity": 0.0009429,
```
seanpearsonuk, line 161: @MohammedAnsys The resolved Copilot comments regarding viscosity here are justified.

examples/00-fluent/pipe_flow.py
```diff
@@ -0,0 +1,318 @@
...
+# ^^^^^^^^^^^^^^^^^^
+
+solution_initialization = Initialization(solver)
+solution_initialization.initialization_type = "hybrid"
+solution_initialization.initialize()
+
+calculation = RunCalculation(solver)
+calculation.iterate(iter_count=300)
+
+# %%
```
seanpearsonuk, line 201: The comment should go **before** the `iterate()` call and it should **not** comment **concretely** on what **has happened**.

E.g.,
```py
# Allow up to 300 iterations. The solver will terminate early if
# convergence criteria are met, so the actual iteration count may
# be lower and can vary between runs.
calculation.iterate(iter_count=300)
```

examples/00-fluent/pipe_flow.py
```diff
@@ -0,0 +1,318 @@
...
+# Discussion
+# ----------
+#
+# This example validates PyFluent's capability to accurately simulate convective heat transfer
+# in internal pipe flows involving high-Prandtl-number fluids in thermally developing flow regimes.
+# The PyFluent simulation was compared with the analytical solution for thermally developing laminar flow in a pipe with constant wall temperature.
+#
+# Analytical outlet temperature: :T_e = 292.74 K
+#
+# PyFluent outlet temperature: :T_e = 291.79 K
```
seanpearsonuk, line 314: We might refer to this as a _typical_ result for PyFluent, given that the user is not guaranteed to get the same result in future.

examples/00-fluent/pipe_flow.py
```diff
@@ -0,0 +1,318 @@
...
+#
+# Both approaches show only a small temperature drop over the 200-m pipe, consistent with the very long thermal entry length.
```
seanpearsonuk, line 316: I have concerns about how this validation is being presented.

1. **Significance of the discrepancy**

The discussion compares two *absolute* outlet temperatures and concludes that “both show only a small temperature drop.”
However, the relevant quantity for validation is not the absolute temperature but the **temperature drop relative to the inlet**:

* Inlet: 293.15 K
* Analytical outlet: 292.74 K → ΔT ≈ 0.41 K
* PyFluent outlet: 291.79 K → ΔT ≈ 1.36 K

The *difference between the two predicted drops* (~0.95 K) is more than **twice the magnitude of the analytical drop itself**.
That is not a minor deviation—it corresponds to substantially different thermal behaviour over the pipe length.

As written, the text risks giving the impression that the two approaches are in close agreement when, in terms of the physics being evaluated (heat transfer rate), they are not.

2. **Model vs. model vs. reality**

The analytical solution is itself a **model** based on idealisations (fully developed laminar flow, constant properties, etc.).
The PyFluent result is another model with different assumptions and discretisation.

What we are doing here is:

> model A ↔ model B

not

> model ↔ physical reality.

The discussion should acknowledge that a discrepancy reflects a gap between two modelling approaches, and that either (or neither) may be closer to reality. Calling this a “validation of PyFluent’s capability” feels too strong without experimental reference data or a more rigorous error analysis.

3. **Viscosity specification**

There is also the unresolved issue noted earlier: the example appears to supply **kinematic viscosity** where Fluent expects **dynamic viscosity**.
The material properties used in the simulation are inconsistent which would directly affect the Prandtl number and the predicted thermal entry behaviour.

Before drawing conclusions about agreement with the analytical solution, we should correct the viscosity definition.

PR #4887, opened 2026-02-04
chore: Update examples to use modern API

Gobot1234:
## Context
The examples are currently not following best practices as described by @seanpearsonuk now they do (I hope).

They should also type check better once all the PRs for these things are merged (units, vis and pyfluent), there are a couple of examples which locally have less that 10 issues (I copied over the typing changes from the aformentioned libraries and turned off a couple of checks that currently haven't been fixed) which is a massive improvement IMO

Mostly untested and not currently functional due to various settings API and other github side pull requests not being merged or fixed

## Change Summary
- Use the much more descriptive object oriented API with ORM style `create()` and `get()` methods that are more clear in intent
- I've tried to get rid of TUI based examples and where not possible I've at least described what the parameters are
- I've colocated the units next to their values
- Removed the dictionary state assignment based APIs in favour of direct attribute access as it will leave users in a more recoverable state and is easier to type (both type hinting and on the keyboard)
- Used tuples for vectors rather than lists to allow for better type hinting in the future checking 2D vs 3D cases
- Used `pathlib` instead of `os.path`
- Added calls to solver.upload(filename) after the example is downloaded if the mode is Meshing for based workflows
- Changed all the `lanuch_fluent` to it's appropriate `ModeClass.from_install`
- Stored the results of `create()` and use that to apply next variable assignments if appropriate rather than performing another lookup
- I've tried to reduce the stringly-typedness of the API by reusing objects (where things are explicitly suffixed with name I've tried to use the `name()` method though. In practice though i don't like this and we should be instead passing round the full objects e.g. `custom_field_function_name=omega_r.name()` where `omega_r` is a `custom_field_function`. I think it makes more sense to pass a str | custom_field_function to a parameter called `custom_field_function` than passing the same type to `custom_field_function_name` which would work if you use another one of my PRs)
- I've tried to update fields to use VariableCatalog though many are missing due to the fields not existing in VariableCatalog or VariableCatalog.fluent
- Improved general pythonicness of code/algorithms/loops etc

## Rationale
We want the examples to be the best things for customers to learn from

NB I've updated some of the examples @MohammedAnsys has PRs open for that looked like they were ready for merging 
NBB I 've not switched the save locations for all these to examples_path, though that might be a good idea in the future


examples/00-fluent/battery_pack.py
```diff
@@ -110,23 +123,22 @@
 mesh_file = examples.download_file(
     "1P3S_battery_pack.msh",
     "pyfluent/battery_pack",
-    save_path=os.getcwd(),
+    save_path=Path.cwd(),
 )
 
-solver.settings.file.read_mesh(file_name=mesh_file)
+solver.setting.file.read_mesh(file_name=mesh_file)
```
Gobot1234:
```suggestion
solver.settings.file.read_mesh(file_name=mesh_file)
```

examples/00-fluent/battery_pack.py
```diff
@@ -179,78 +191,50 @@
...
-    "uds_diffusivity": {
-        "option": "value",
-        "value": 3.541e7,  # S/m  (Copper-like conductivity)
-    },
-}
+busbar_material = SolidMaterial.create(
+    solver, name="busbar_material", chemical_formula="bus"
+)
+busbar_material.uds_diffusivity.option = "value"
+busbar_material.uds_diffusivity.value = 3.541e7 * S / m  # Copper-like conductivity
```
Gobot1234:
```suggestion
busbar_material = SolidMaterial.create(
    solver, 
    name="busbar_material",
    chemical_formula="bus",
    uds_diffusivity=3.541e7 * S / m,  # Copper-like conductivity
)
```
honestly would be surprised this didn't work

examples/00-fluent/battery_pack.py
```diff
@@ -179,78 +191,50 @@
...
-# Copy to cell_2 and cell_3
-cell_zone_conditions.copy(from_="cell_1", to="cell_*")
-
-# Assign busbar_material to bar1
-cell_zone_conditions.solid["bar1"] = {"general": {"material": "busbar_material"}}
-
-# Copy to all passive zones
-cell_zone_conditions.copy(from_="bar1", to="*bar*|*tabzone*")
-# '*bar*' matches any zone containing 'bar',
-# '|*tabzone*' matches any zone containing 'tabzone' → OR logic
```
Gobot1234, line 225: Deleted because it should be explained further up what this does

examples/00-fluent/battery_pack.py
```diff
@@ -264,34 +248,35 @@
...
+    cell_zones=["*cell|*bar*|*tabzone*"],
+    create_report_file=True,
+    create_report_plot=True,
+)
 
-# Format plot axes
-report_plot_1 = solver.settings.solution.monitor.report_plots["volume_max_temp-rplot"]
-report_plot_1.axes.x.number_format.precision = 0
-report_plot_1.axes.y.number_format.precision = 2
+volume_max_temp = ReportPlot.get(solver, name="volume_max_temp-rplot")
```
Gobot1234:
```suggestion
# Format plot axes
volume_max_temp = ReportPlot.get(solver, name="volume_max_temp-rplot")
```

examples/00-fluent/battery_pack.py
```diff
@@ -302,24 +287,28 @@
...
-Transient_controls.time_step_size = 30  # s  (30 s per step)
+transient_controls = calculation.transient_controls
+# Use typed quantities for time step settings
+transient_controls.time_step_count = 50  # Number of time steps
+transient_controls.time_step_size = 30  # 30s per step
 
-calculation = RunCalculation(solver)
-calculation.calculate()  # Run transient simulation
+# Run transient simulation
+calculation.calculate()
```
Gobot1234:
```suggestion
calculation.calculate()  # Run transient simulation
```

examples/00-fluent/catalytic_converter_workflow.py
```diff
@@ -88,39 +88,44 @@
...
 # sphinx_gallery_thumbnail_path = '_static/catalytic_converter/catalytic_converter_cad_geo.png'
 
-import os
+from pathlib import Path
 import platform
 
+from ansys.units import VariableCatalog
+
 import ansys.fluent.core as pyfluent
 from ansys.fluent.core import (
```
Gobot1234: Should try to get rid of all of these except examples I think

examples/00-fluent/catalytic_converter_workflow.py
```diff
@@ -88,39 +88,44 @@
...
     Dimension,
-    FluentMode,
-    FluentVersion,
     Precision,
     UIMode,
     examples,
 )
+from ansys.fluent.core.generated.solver.settings_builtin import IsoSurface
+from ansys.fluent.core.generated.solver.settings_builtin_261 import write_case_data
```
Gobot1234: These autoimports should be moved to the below

examples/00-fluent/catalytic_converter_workflow.py
```diff
@@ -370,12 +363,14 @@
...
-mesh.surfaces_list = all_walls
+all_walls = WallBoundaries(solver).get_object_names()
+mesh = Mesh(
+    solver,
+    new_instance_name="mesh-1",
+    surfaces_list=all_walls,
+)
 mesh.options.edges = True
 mesh.display()
-os.makedirs("out", exist_ok=True)
```
Gobot1234, line 378: This being made every single time seemed excessive

examples/00-fluent/catalytic_converter_workflow.py
```diff
@@ -436,30 +420,32 @@
 
 # Configure first substrate zone as porous media
```
Gobot1234, line 372:
```suggestion
# Configure both substrate zones as porous media
```

examples/00-fluent/catalytic_converter_workflow.py
```diff
@@ -572,18 +552,18 @@
...
-    ("z=230", "z-coordinate", [0.23]),
-    ("z=280", "z-coordinate", [0.28]),
-    ("z=330", "z-coordinate", [0.33]),
-    ("z=375", "z-coordinate", [0.375]),
+    ("y=-425", "y-coordinate", [-0.425] * m),
+    ("z=185", "z-coordinate", [0.185] * m),
+    ("z=230", "z-coordinate", [0.23] * m),
+    ("z=280", "z-coordinate", [0.28] * m),
+    ("z=330", "z-coordinate", [0.33] * m),
+    ("z=375", "z-coordinate", [0.375] * m),
```
Gobot1234, line 511: These should be mm maybe, seems to be configured earlier in the example

examples/00-fluent/catalytic_converter_workflow.py
```diff
@@ -617,55 +597,50 @@
...
 # Configure scene appearance and display settings
-solver_session.settings.results.scene["scene-1"] = {
-    "graphics_objects": {
-        "vector-vel": {"name": "vectors"},  # Label for the vector plot
-        "mesh-1": {"transparency": 75},  # Semi-transparent mesh (75% transparent)
-    }
+scene_1= Scene.create(solver, name="scene-1")
+scene_1.graphics_objects = {
+    "vector-vel": {"name": "vectors"},  # Label for the vector plot
+    "mesh-1": {"transparency": 75},  # Semi-transparent mesh (75% transparent)
```
Gobot1234, line 591: Missed conversion to attribute assignment

examples/00-fluent/conjugate_heat_transfer.py
```diff
@@ -523,243 +533,140 @@
...
-
-solver_session.settings.solution.monitor.report_files["wall-shear-int-file"] = {}
-solver_session.settings.solution.monitor.report_files["wall-shear-int-file"] = {
-    "report_defs": ["wall-shear-int"],
-    "file_name": r"wall-shear-int.out",
-}
+monitor.report_plots.create(report_defs=[vol_max])
+monitor.report_files.create(report_defs=[vol_max], file_name="max-vel-louvers4.out")
+monitor.report_plots.create(report_defs=[wall_shear])
+monitor.report_files.create(report_defs=[wall_shear], file_name="wall-shear-int.out")
```
Gobot1234, line 498: The diff here is very large for some reason

examples/00-fluent/DOE_ML.py
```diff
@@ -62,20 +62,28 @@
...
-import plotly.express as px
-import plotly.graph_objects as go
-import seaborn as sns
-import tensorflow as tf
-from tensorflow import keras
+# import plotly.express as px
+# import plotly.graph_objects as go
+# import seaborn as sns
+# import tensorflow as tf
+# from tensorflow import keras
```
Gobot1234: obvious

examples/00-fluent/DOE_ML.py
```diff
@@ -62,20 +62,28 @@
...
 
 import ansys.fluent.core as pyfluent
 from ansys.fluent.core import examples
+from ansys.fluent.core.generated.solver.settings_builtin import SurfaceIntegrals
+from ansys.fluent.core.generated.solver.settings_builtin_261 import iterate, read_case
+from ansys.fluent.core.solver import Initialization, VelocityInlet, RunCalculation
```
Gobot1234: Imports need merging

examples/00-fluent/DOE_ML.py
```diff
@@ -95,69 +103,67 @@
...
+cold_velocities = [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7] * m / s
+
+hot_velocities = [0.8, 1.0, 1.2, 1.4, 1.6, 1.8, 2.0] * m / s
+
 
 # Allocate a results array. Entry (i, j) will hold the Mass‑Weighted
-# Average Temperature at the outlet for coldVelArr[i] and hotVelArr[j].
-resArr = np.zeros((coldVelArr.size, hotVelArr.size))
-
-for idx1, coldVel in np.ndenumerate(coldVelArr):
```
Gobot1234, line 124: ndenumerate is the same as enumerate cause it's a vector

examples/00-fluent/DOE_ML.py
```diff
@@ -95,69 +103,67 @@
...
+    hot_inlet.momentum.velocity = hot_value
+
+    initialize = Initialization(solver)
+    initialize.initialization_type = "standard"
+    initialize.standard_initialize()
+
+    iterate(solver,iter_count=200)
+
+    temperatures = SurfaceIntegrals(solver).get_mass_weighted_avg(
+        surface_names=["outlet"], report_of="temperature"
```
Gobot1234, line 153: Can this use variable catalog?

examples/00-fluent/DOE_ML.py
```diff
@@ -282,33 +291,26 @@ def fit_and_predict(model):
...
-        test_list.append("Test")
-
-    com_train_set["Result"] = train_predictions.tolist()
-    com_train_set["Set"] = train_list
-    com_test_set["Result"] = test_predictions.tolist()
-    com_test_set["Set"] = test_list
+    com_train_set["result"] = train_predictions.tolist()
+    com_train_set["set"] = train_list
+    com_test_set["result"] = test_predictions.tolist()
+    com_test_set["set"] = test_list
```
Gobot1234, line 311: Honestly this code kinda confused me so much IDK if it's right

examples/00-fluent/Electrolysis_Modeling_workflow.py
```diff
@@ -136,153 +147,115 @@
...
-}
+SolidMaterial.create(solver, name="collector-default", electric_conductivity = 20000 * S / m)
 
-# Porous layer
-materials.solid["porous-default"] = {"electric_conductivity": {"value": 20000}}  # S/m
+SolidMaterial.create(solver, name="porous-default", electric_conductivity = 20000 * S / m)
 
-# Catalyst layer: dual conductivity
-materials.solid["catalyst-default"] = {
-    "electric_conductivity": {"value": 5000},  # S/m  Electronic
```
Gobot1234, line 227: Comments were lost

examples/00-fluent/Electrolysis_Modeling_workflow.py
```diff
@@ -136,153 +147,115 @@
...
 # %%
 # Initialize solution
 # -------------------
 initialize = Initialization(solver)
 
 initialize.initialization_type = "standard"
-initialize.defaults = {
-    "temperature": 333.15,  # K
-    "phase-2-mp": 1,  # Initial volume fraction of liquid
-}
```
Gobot1234, line 268: comments lost again

examples/00-fluent/exhaust_system_settings_api.py
```diff
@@ -86,22 +91,22 @@
...
 # Initialize workflow
 # ~~~~~~~~~~~~~~~~~~~
 # Initialize the fault-tolerant meshing workflow.
 
-meshing_session.workflow.InitializeWorkflow(WorkflowType="Fault-tolerant Meshing")
+meshing.workflow.InitializeWorkflow(WorkflowType="Fault-tolerant Meshing")
 
 ###############################################################################
-# Fault-folerant meshing workflow
+# Fault-tolerant meshing workflow
```
Gobot1234, line 120: Slightly concerned these weren't picked up at the time can someone check the spell checker to see if this is fixable?

examples/00-fluent/external_compressible_flow.py
```diff
@@ -78,6 +78,16 @@
 
 import ansys.fluent.core as pyfluent
 from ansys.fluent.core import examples
+from ansys.fluent.core.generated.solver.settings_builtin import RunCalculation
+from ansys.fluent.core.generated.solver.settings_builtin_261 import write_case
```
Gobot1234: merge imports

examples/00-fluent/frozen_rotor_workflow.py
```diff
@@ -80,10 +80,42 @@
...
+from ansys.fluent.core.generated.solver.settings_builtin_261 import write_case, write_case_data
+from ansys.fluent.core.solver import (
+    FluidCellZone,
+    General,
+    MassFlowOutletBoundary,
+    PressureInletBoundary,
+    Viscous,
+    WallBoundary,
+)
+from ansys.fluent.core.solver import RunCalculation
```
Gobot1234: merge imports again

examples/00-fluent/frozen_rotor_workflow.py
```diff
@@ -94,57 +126,60 @@
...
-viscosity_water = 0.001002  # kg/(m.s)
-g = 9.81  # m/s^2
-# Impeller speed
+density_water = 998.2 * kg / m**3
+viscosity_water = 0.001002 * Pa * s
+g = 9.81 * m / s**2
 impeller_speed = 1450  # rpm
-# Convert to rad/s
-impeller_speed_rad = impeller_speed * 2 * math.pi / 60  # rad/s
+# Convert to rad/s (numeric)
```
Gobot1234:
```suggestion
# Convert to rad/s
```

examples/00-fluent/frozen_rotor_workflow.py
```diff
@@ -183,204 +218,157 @@
...
 # p-blade
-blade_pressure_report_def = (
-    solver_session.settings.solution.report_definitions.surface.create("p-blade")
-)
-blade_pressure_report_def.report_type = "surface-massavg"
-blade_pressure_report_def.surface_names = ["blade"]
-blade_pressure_report_def.field = "pressure"
-blade_pressure_report_def.per_surface = False
-
+blade_pressure_report_def = report_definitions.surface.create("p-blade", report_type = "surface-massavg", surface_names = ["blade"], field = "pressure", per_surface = False)
```
Gobot1234: VariableCatalog the field?

examples/00-fluent/frozen_rotor_workflow.py
```diff
@@ -183,204 +218,157 @@
...
+pump_head_report_def = report_definitions.single_valued_expression.create("pump-head")
 
 pump_head_report_def.definition = "head"
 
 # report plot
-pump_head_report_plot = solver_session.settings.solution.monitor.report_plots.create(
-    "pump-head-rplot"
-)
-pump_head_report_plot.report_defs = "pump-head"
+pump_head_report_plot = monitor.report_plots.create("pump-head-rplot", report_defs=pump_head_report_def)
```
Gobot1234: stick defs in a list

examples/00-fluent/frozen_rotor_workflow.py
```diff
@@ -183,204 +218,157 @@
...
 
 # report file
-pump_head_report_file = solver_session.settings.solution.monitor.report_files.create(
-    "pump-head-rfile"
-)
-pump_head_report_file.report_defs = "pump-head"
+pump_head_report_file = monitor.report_files.create("pump-head-rfile", report_defs=pump_head_report_def)
```
Gobot1234: also stick in list

examples/00-fluent/frozen_rotor_workflow.py
```diff
@@ -389,24 +377,14 @@
...
-)
-pressure_contour.field = "pressure"
-pressure_contour.surfaces_list = ["z_mid_plane"]
-
-# Display the contour and save the image
-
-graphics.views.restore_view(view_name="front")
+# Define and display the contour for static pressure using typed API
+graphics = Graphics(solver)
+pressure_contour = Contour.create(solver=solver, field=VariableCatalog.PRESSURE, surfaces=["z_mid_plane"])
```
Gobot1234: Comments lost

examples/00-fluent/fsi_1way_workflow.py
```diff
@@ -67,29 +67,34 @@
...
 
-import os
+from pathlib import Path
 
 import ansys.fluent.core as pyfluent
-from ansys.fluent.core import FluentMode, Precision, examples
+from ansys.fluent.core import Precision, examples
+from ansys.fluent.core.generated.solver.settings_242 import read_case
+from ansys.fluent.core.generated.solver.settings_builtin import Controls
+from ansys.fluent.core.generated.solver.settings_builtin_261 import BoundaryCondition, Materials, iterate, write_case, write_case_data
```
Gobot1234: similar import issues to before

examples/00-fluent/fsi_1way_workflow.py
```diff
@@ -98,16 +103,16 @@
...
+read_case(solver, file_name=mesh_file)
 
 # %%
 # Configure solver settings for fluid flow
 # ----------------------------------------
 
-velocity_inlet = VelocityInlet(solver, name="velocity_inlet")
-velocity_inlet.momentum.velocity_magnitude = 100.0  # High-speed inlet flow (m/s)
+velocity_inlet = VelocityInlet.get(solver, name="velocity_inlet")
+velocity_inlet.momentum.velocity_magnitude = 100.0 * m / s
```
Gobot1234:
```suggestion
velocity_inlet.momentum.velocity_magnitude = 100.0 * m / s  # High-speed inlet flow
```

examples/00-fluent/lunar_lander_thermal.py
```diff
@@ -95,18 +95,43 @@
...
+    CellZoneConditions,
+    Monitor,
+    RunCalculation,
+)
+from ansys.fluent.core.generated.solver.settings_builtin_261 import (
+    CellZoneCondition,
+    calculate,
+)
+from ansys.fluent.core.session_solver import Solver
+from ansys.fluent.core.solver import (
```
Gobot1234: Import issues

examples/00-fluent/lunar_lander_thermal.py
```diff
@@ -322,23 +347,15 @@ def get_surf_mean_temp(
...
-    "start": 0,
-    "end": 2.8,
-}
-# Define range of thermal IR wavelengths
-radiation.multiband["thermal-ir"] = {
-    "start": 2.8,
-    "end": 100,
-}
-
-# Solve radiation once per timestep
```
Gobot1234, line 341: lost comments

examples/00-fluent/lunar_lander_thermal.py
```diff
@@ -351,38 +368,25 @@ def get_surf_mean_temp(
...
-vacuum = solver_session.settings.setup.materials.solid.create("vacuum")
-vacuum.chemical_formula = ""
-vacuum.thermal_conductivity.value = 0
-vacuum.absorption_coefficient.value = 0
-vacuum.refractive_index.value = 1
-
-# --- Properties of fluff (see ref. [2]) ---
-# Density: 1000 [kg m^-3]
-# Specific heat capacity: 1050 [J kg^-1 K^-1]
-# Thermal conductivity: 9.22e-4*(1 + 1.48*(temperature/350 K)^3) [W m^-1 K^-1]
```
Gobot1234, line 366: lost comment

examples/00-fluent/lunar_lander_thermal.py
```diff
@@ -351,38 +368,25 @@ def get_surf_mean_temp(
...
-fluff.specific_heat.value = 1050
+fluff = SolidMaterial.create(
+    solver, name="fluff", density=1000 * kg / m**3, specific_heat=1050 * J / (kg * K)
+)
 fluff.thermal_conductivity.option = "expression"
 fluff.thermal_conductivity.expression = (
     "9.22e-4[W m^-1 K^-1]*(1 + 1.48*(StaticTemperature/350[K])^3)"
 )
 
-# --- Properties of regolith (see ref. [2]) ---
```
Gobot1234, line 377: lost comment

examples/00-fluent/lunar_lander_thermal.py
```diff
@@ -411,15 +416,9 @@ def get_surf_mean_temp(
 # used to represent the geothermal heat from the Moon's interior that heats
 # the regolith from the bottom.
 
-# --- Regolith BC ---
-# Thickness of layers: 0.02, 0.04, 0.08, 0.16, 0.32 [m]
-# Heating at base: 0.031 [W m^-2]
-# Surface absorptivity: 0.87
-# Surface emissivity: 0.97
```
Gobot1234, line 418: Lost comments

examples/00-fluent/lunar_lander_thermal.py
```diff
@@ -500,13 +493,7 @@ def get_surf_mean_temp(
 # walls, but the emissivity is left unset as it will be changed dynamically by
 # our PyFluent script depending on its temperature during the simulation.
 
-# --- Set up spacecraft radiator boundary condition ---
-# Thickness: 0.03 [m]
-# Material: aluminum
-# Absorptivity: 0.17
-# Emissivity: 0.09 below 273 K, 0.70 otherwise
-
```
Gobot1234, line 508: lost comments

examples/00-fluent/lunar_lander_thermal.py
```diff
@@ -451,38 +450,31 @@ def get_surf_mean_temp(
...
 # Spacecraft walls boundary condition
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 # The spacecraft is covered in reflective MLI (multilayer insulation) and heat
 # transfer through the aluminum shell can be simulated using shell conduction.
 
-# --- Set up spacecraft shell boundary condition ---
-# Thickness: 0.03 [m]
-# Material: aluminum
-# Absorptivity: 0.05
-# Emissivity: 0.05
```
Gobot1234, line 483: lost comments

examples/00-fluent/lunar_lander_thermal.py
```diff
@@ -525,82 +512,88 @@ def get_surf_mean_temp(
...
-    report_name = f"regolith-layer-{i}-temp"
-    surf_report_defs[report_name] = {
-        "report_type": "surface-facetavg",
-        "field": "temperature",
-    }
-    regolith_report_names.extend([report_name])
+regolith_reports = [
+    report_defs.surface.create(
+        name=f"regolith-layer-{i}-temp",
+        report_type="surface-facetavg",
```
Gobot1234, line 568: NB check if there's enums for these cause I don't think there are, do we need to auto gen them? (#4891)

examples/00-fluent/mixing_elbow_settings_api.py
```diff
@@ -53,16 +53,29 @@
...
 
-import os
+from pathlib import Path
 
 # sphinx_gallery_thumbnail_path = '_static/mixing_elbow_settings.png'
 import ansys.fluent.core as pyfluent
 from ansys.fluent.core import examples
+from ansys.fluent.core.generated.solver.settings_builtin import Fluxes, Vector
+from ansys.fluent.core.generated.solver.settings_builtin_261 import Materials, read_case
+from ansys.fluent.core.solver import (
```
Gobot1234: import merge needed

examples/00-fluent/mixing_tank_workflow.py
```diff
@@ -75,25 +75,49 @@
...
+from ansys.fluent.core.generated.solver.settings_builtin import (
+    BoundaryCondition,
+    Mesh,
+    PlaneSurface,
+    ReportFile,
+    ReportPlot,
+    Residual,
+)
+from ansys.fluent.core.generated.solver.settings_builtin_261 import write_case, write_case_data
+from ansys.fluent.core.solver import (
```
Gobot1234: merge imports again

examples/00-fluent/mixing_tank_workflow.py
```diff
@@ -195,115 +217,97 @@
...
 # ===========================================================================================================
-solver_session.settings.setup.general.operating_conditions.gravity.enable = True
-solver_session.settings.setup.general.operating_conditions.gravity.components = [
-    0.0,
-    0.0,
-    -g,
-]
+general = General(solver)
+general.operating_conditions.gravity.enable = True
+general.operating_conditions.gravity.components = [0.0, 0.0, -g]
```
Gobot1234: should be tuple

examples/00-fluent/mixing_tank_workflow.py
```diff
@@ -195,115 +217,97 @@
...
+solution.controls.under_relaxation["turb-viscosity"] = 0.6
 
 ##############################################################################################################
 #  Residuals Criteria & Initialization
 # ===========================================================================================================
 
-# Residuals criteria
-residuals_options = solver_session.settings.solution.monitor.residual
-residuals_options.equations["continuity"].absolute_criteria = 0.0001
-residuals_options.equations["continuity"].monitor = True  # Enable continuity residuals
```
Gobot1234, line 297: lost comments

Gobot1234 (commented):
Will finish review later

examples/00-fluent/ahmed_body_workflow.py
```diff
@@ -61,18 +61,36 @@
...
+)
 from ansys.fluent.visualization import Contour, GraphicsWindow
+from ansys.units.common import kg, m, s
 
 #######################################################################################
 # Launch Fluent session with meshing mode and print Fluent version
 # =====================================================================================
-session = pyfluent.launch_fluent(mode="meshing", cleanup_on_exit=True)
-print(session.get_fluent_version())
+meshing = pyfluent.Meshing.from_install(cleanup_on_exit=True, fluent_path=r"C:\ANSYSDev\v261\fluent\ntbin\win64\fluent.exe")
```
seanpearsonuk: @Gobot1234 

Other team members can comment here before you change anything due to the following observation.

One reason to avoid `meshing =` / `solver =` is that `meshing` and `solver` are protected global names in Fluent's Python console. Code that assigns to them would attempt (and fail) to overwrite those globals if the script were reused there, making the script unusable in that environment.

examples/00-fluent/ahmed_body_workflow.py
```diff
@@ -92,70 +110,63 @@
 geometry_filename = examples.download_file(
     filenames.get(platform.system(), filenames["Other"]),
     "pyfluent/examples/Ahmed-Body-Simulation",
-    save_path=os.getcwd(),
+    save_path=Path.cwd(),
 )
+meshing.upload(geometry_filename)
```
seanpearsonuk, line 118: @Gobot1234 

Seeing `upload()` calls like this one is surprising given that the upload should be taken care of seamlessly due to the geometry import operation. Either the script's explicit upload is redundant or you're aware of a limitation.

examples/00-fluent/ahmed_body_workflow.py
Gobot1234: Ah right, I was thinking it'd be clearer for using pyconsole because you can just skip that step but then things work for you without having to change variable names, maybe it should be clearer though that this is optional in pyconsole using a separate section

examples/00-fluent/ahmed_body_workflow.py
Gobot1234: This was for PIM based environments so customers only need to change the launch method refs TFS/1365452

examples/00-fluent/ahmed_body_workflow.py
```diff
@@ -243,107 +252,110 @@
...
+inlet_area = 0.11203202 * m**2
 
 #######################################################################################
 # Define Materials
 # =====================================================================================
-session.tui.define.materials.change_create("air", "air", "yes", "constant", density)
-session.settings.setup.models.viscous.model = "k-epsilon"
-session.settings.setup.models.viscous.k_epsilon_model = "realizable"
-session.settings.setup.models.viscous.options.curvature_correction = True
+air = FluidMaterial.get(solver, name="air")
```
seanpearsonuk: @Gobot1234 

As you'll already know, this change depends on another PR you have open. In that PR, we looked into adding this (preferred) style: `air = FluidMaterial(solver).get(name="air")`. Is that something we're still looking into?

Gobot1234: I didn't realise that style was preferred by you for all situations I thought that was for cases where Fluid Material was repeated instantiated

examples/00-fluent/ahmed_body_workflow.py
```diff
@@ -185,10 +194,10 @@
...
 # Describe Geometry, Update Boundaries, Update Regions
 # =====================================================================================
-workflow.TaskObject["Describe Geometry"].Arguments = dict(
-    CappingRequired="Yes",
-    SetupType="The geometry consists of only fluid regions with no voids",
-)
+workflow.TaskObject["Describe Geometry"].Arguments = {
+    "CappingRequired": "Yes",
+    "SetupType": "The geometry consists of only fluid regions with no voids",
+}
```
seanpearsonuk: Horrific API 😆

examples/00-fluent/ahmed_body_workflow.py
seanpearsonuk: ...unfortunate because I much prefer these names.

examples/00-fluent/ahmed_body_workflow.py
Gobot1234: I'm not on testing the new workflow so probably don't have the time to fix these 😔

examples/00-fluent/ahmed_body_workflow.py
```diff
@@ -243,107 +252,110 @@
 #######################################################################################
 # Define Constants
 # =====================================================================================
-density = 1.225
-inlet_velocity = 30
-inlet_area = 0.11203202
+density = 1.225 * kg / m**3
```
seanpearsonuk, line 232: This is not only more readable: it adds safety on the client side, where the quantity type is pre-validated.

examples/00-fluent/ahmed_body_workflow.py
```diff
@@ -243,107 +252,110 @@
...
+residual = Residual(solver)
+for monitor in (
+    "continuity",
+    "x-velocity",
+    "y-velocity",
+    "z-velocity",
+    "k",
+    "epsilon",
+):
+    residual.equations[monitor].absolute_criteria = 1e-4
```
seanpearsonuk, line 288: :+1:

examples/00-fluent/ahmed_body_workflow.py
```diff
@@ -243,107 +252,110 @@
...
-session.settings.solution.monitor.residual.equations["epsilon"].absolute_criteria = (
-    0.0001
-)
+methods = Methods(solver)
+methods.p_v_coupling.flow_scheme = "Coupled"
+
+discretization_scheme = methods.spatial_discretization.discretization_scheme
+discretization_scheme["pressure"] = "second-order"
+discretization_scheme["k"] = "second-order-upwind"
+discretization_scheme["epsilon"] = "second-order-upwind"
```
seanpearsonuk, line 275: @Gobot1234 For some of these, perhaps we could add a comment, either `# Alternatively, an enum is available in place of this string value` or `# Using a string as no enum is currently available`. In the latter case, an issue could be raised. Why I'm commenting this is that, as a reader, I am not sure of the reason for using strings here, but I would take some (avoidable) time to try using enums in their place.

examples/00-fluent/ahmed_body_workflow.py
```diff
@@ -243,107 +252,110 @@
...
+initialization = Initialization(solver)
+initialization.defaults.k = 0.000001
```
seanpearsonuk:
```suggestion
initialization.defaults.k = 1e-6
```

examples/00-fluent/ahmed_body_workflow.py
```diff
@@ -61,18 +61,36 @@
...
 
+from ansys.units import VariableCatalog
+
 import ansys.fluent.core as pyfluent
 from ansys.fluent.core import examples
+from ansys.fluent.core.solver import (
+    Methods,
+    ReferenceValues,
+    Residual,
+ iterate, write_case,
```
seanpearsonuk: @Gobot1234 
1. Indentation issue. 
2. Is this snake_case format new here?

Gobot1234: 1. I'm aware of the general import problems in this I need to go through them again.
2. The snake case version of this was around before I started touching the object oriented API, here I'm using it because these are used like functions as suggested by PEP 8
> The naming convention for functions may be used instead in cases where the interface is documented and used primarily as a callable.

https://peps.python.org/pep-0008/#class-names

examples/00-fluent/battery_pack.py
```diff
@@ -89,18 +89,31 @@
...
+    read_mesh,
 )
 from ansys.fluent.visualization import GraphicsWindow, Monitor
+from ansys.units.common import K, W, m, ohm
+
+# %%
+# Define constants
+# ----------------
+
+S = 1 / ohm  # a Siemen
```
seanpearsonuk:
```suggestion
S = 1 / ohm  # = 1 siemens
```
When surnames are used as units the name is always lowercase whereas the symbol is always uppercase. In this case, the unit is the siemens. 

The library should probably expose it directly as a common unit.

examples/00-fluent/battery_pack.py
```diff
@@ -110,23 +123,22 @@
...
 graphics = Graphics(solver)
-mesh = Mesh(solver, new_instance_name="mesh-1")
-
-all_walls = mesh.surfaces_list.allowed_values()
+mesh = Mesh.create(solver, name="mesh-1")
 
 graphics.picture.x_resolution = 650  # Horizontal resolution for clear visualization
 graphics.picture.y_resolution = 450  # Vertical resolution matching typical aspect ratio
 
+all_walls = mesh.surfaces_list.allowed_values()
```
seanpearsonuk: @Gobot1234 

What do you think about adding `all()` for string types as an alias for `allowed_values()`, with the intention to deprecate the latter over time? This would align with the new `<NamedType>.all()` (albeit the semantics mismatch slightly) and eliminate a relatively verbose method name.

Gobot1234: I like that yes, will update the PR soonish

examples/00-fluent/battery_pack.py
```diff
@@ -179,78 +191,50 @@
...
-}
+e_material = SolidMaterial.create(solver, name="e_material", chemical_formula="e")
+e_material.thermal_conductivity = 20 * W / (m * K)
+e_material.uds_diffusivity.option = "defined-per-uds"
+e_material.uds_diffusivity.uds_diffusivities["uds-0"] = (
+    1e6 * S / m
+)  # Electronic conductivity
+e_material.uds_diffusivity.uds_diffusivities["uds-1"] = (
+    1e6 * S / m
+)  # Ionic conductivity
```
seanpearsonuk: I have a couple of general questions about this section, i.e. as it already exists, and therefore **not directed at** @Gobot1234:

**1. Terminology:**
The quantities provided under `uds_diffusivity` have units of **S/m (electrical conductivity)** rather than any form of diffusivity (m²/s).
Is the term “diffusivity” here only a consequence of re-using the generic UDS diffusion operator internally? From a user perspective this is confusing, as the physical meaning is conductivity rather than diffusivity. The API user is **not interested in Fluent's implementation details**.

**2. API structure:**
The keys `"uds-0"` and `"uds-1"` appear to be created implicitly inside the `uds_diffusivities` container, with their semantics defined only by comments (electronic vs ionic conductivity).
This raises a few concerns:

* these objects are not discoverable in the public schema/docs;
* their meaning is position-based rather than name-based;
* nothing prevents them being swapped accidentally.

It would obviously make sense to expose a more explicit structure with named fields such as `electronic_conductivity` and `ionic_conductivity`, so that the API reflects the physical intent rather than relying on implicit indices?

seanpearsonuk: Referenced in https://github.com/ansys/pyfluent/issues/4892

Gobot1234: You'll see a similar sort of index based API with the multiphase stuff which I also wasn't a big fan of

seanpearsonuk: @Gobot1234 It would be more defensible if the names weren't utterly meaningless. It shows you the level of thinking that goes into some of this stuff.

examples/00-fluent/battery_pack.py
seanpearsonuk: 😃

examples/00-fluent/ahmed_body_workflow.py
seanpearsonuk: @Gobot1234 In the new version, the values are simply `true` and `"fluid"`.

examples/00-fluent/battery_pack.py
Gobot1234: Yeah... as someone with just the code you wouldn't even know what a UDS was which I think is a pretty massive naming failure

examples/00-fluent/tyler_sofrin_modes.py
```diff
@@ -188,49 +186,45 @@
 for angle in range(0, 360, d_theta):
     x = math.cos(math.radians(angle)) * r
     y = math.sin(math.radians(angle)) * r
-    pt_name = "point-" + str(angle)
-    session.settings.results.surfaces.point_surface[pt_name] = {}
-    session.settings.results.surfaces.point_surface[pt_name].point = [x, y, z]
+    PointSurface(solver, new_instance_name=f"point-{angle}"). point=(x, y, z)
```
Gobot1234:
```suggestion
    PointSurface.create(solver, name=f"point-{angle}"), point=(x, y, z))
```

examples/00-fluent/battery_pack.py
Gobot1234: Just found this online 
> “Siemens” is the singular and plural; “1 siemen” is wrong.

examples/00-fluent/tyler_sofrin_modes.py
seanpearsonuk, line 1: This script is full of obscurely abbreviated variable names.

Now referenced in https://github.com/ansys/pyfluent/issues/4944. Resolving comment.

seanpearsonuk (approved):

examples/00-fluent/conjugate_heat_transfer.py
Gobot1234, line 44:
```suggestion
# As a result, the temperature fields evolve together.
```

examples/00-fluent/conjugate_heat_transfer.py
```diff
@@ -473,25 +483,25 @@
...
 
-solver_session = meshing_session.switch_to_solver()
+solver = meshing.switch_to_solver()
 
 #############################################################################
 # Auto-create Mesh Interfaces
 # ===========================
 
-solver_session.tui.define.mesh_interfaces.create("int", "yes", "no")
+MeshInterfaces(solver).create(si_name="int", all_bnd=True)
```
Gobot1234, line 354: Should really explain what these parameters are because looking at this I have no clue, it's somehow only slightly better than the TUI call.

PR #4888, opened 2026-02-04
test: Standardize test skip reasons across test suite

mayankansys:
## Context
Several tests were skipped using `@pytest.mark.skip`, but the skip reasons were inconsistent and often unclear. Many reasons only described symptoms (for example, “works locally but fails on CI”) without stating whether the root cause was known, partially understood, or unknown. Similar situations were described using different wording across the test suite, making maintenance and investigation harder.

## Change Summary
- Introduced a small set of canonical skip reason string literals in `conftest.py`.
- Replaced ad-hoc skip reason strings in test files with these shared constants.
- Updated existing skipped tests to use the standardized reasons.

## Impact
- Test code only: no impact on runtime or production behavior.
- Improves readability, searchability, and maintainability of skipped tests.
- Makes it easier to audit and clean up skipped tests in the future.

### DIFF
--- a/doc/changelog.d/4888.test.md
+++ b/doc/changelog.d/4888.test.md
@@ -1,0 +1,1 @@
+Standardize test skip reasons across test suite


tests/test_datamodel_api.py
```diff
@@ -203,7 +204,9 @@ def cb():
     assert called == 1
 
 
-@pytest.mark.skip(reason="https://github.com/ansys/pyfluent/issues/4298")
+@pytest.mark.skip(
+    reason=f"{SKIP_BLOCKED} https://github.com/ansys/pyfluent/issues/4298"
```
prmukherj: Skipping with github issues linked to them should be marked "SKIP_INVESTIGATING"?

cc. @seanpearsonuk, @mkundu1

tests/conftest.py
```diff
@@ -41,6 +41,15 @@
...
 
+# 1. Root cause completely unknown
+SKIP_UNKNOWN = "Skipped (root cause unknown)"
+
+# 2. Partially understood but unresolved
+SKIP_INVESTIGATING = "Skipped (partially understood, under investigation)"
+
+# 3. Understood but blocked for other reasons
+SKIP_BLOCKED = "Skipped (understood but blocked)"
+
```
seanpearsonuk, line 73: :+1:

tests/test_datamodel_api.py
seanpearsonuk: That's a good point, @prmukherj. To avoid any more confusion around that, we should comment what BLOCKED means in a little more detail:

```
# 3. Root cause understood but test cannot be enabled due to an external blocker
#    (e.g., upstream bug, CI/environment limitation, product constraint).
#    An issue link may be present for tracking, but this does NOT imply active investigation.
```

seanpearsonuk: @prmukherj @mayankansys Can you confirm that the three categories make sense... most importantly, do they work practically?

tests/conftest.py
```diff
@@ -41,6 +41,36 @@
...
+# ============================================================================
+
+# 1. Root cause completely unknown
+SKIP_UNKNOWN = "Skipped (root cause completely unknown)"
+
+# 2. Partially understood but unresolved
+SKIP_INVESTIGATING = "Skipped (partially understood but unresolved)"
+
+# 3. Understood but blocked for other reasons
+SKIP_BLOCKED = "Skipped (root cause understood but test cannot be enabled due to an external blocker)"
```
hpohekar, line 72: Using an Enum class for skip reasons could provide both clarity and IDE/editor support, reducing risk of typos:

```
from enum import Enum

class SkipReason(Enum):
    UNKNOWN = "Skipped (root cause completely unknown)"
    INVESTIGATING = "Skipped (partially understood but unresolved)"
    BLOCKED = "Skipped (root cause understood but blocked)"
```

I think we can remove `SKIP_` prefix as `@pytest.mark.skip` decorator conveys the message that we are skipping a particular test.

prmukherj (approved):

PR #4916, opened 2026-02-10
chore: update bandit checks

prmukherj:
## Context
Some bandit checks were commented out without appropriate reasons. Those are addressed here.

## Change Summary
Either alternatives were applied to avoid vulnerability issues, or they were skipped with appropriate reasons where required.

## Rationale
Simplest way to get round the justifications required for skipping the checks.

## Impact
No user impact.

Pre-commit checks will now include bandit so that no new vulnerabilities are pushed in future.

Please note, there are 6 more vulnerabilities skipped in the pyproject.toml file. Those will be addressed in the next iteration as they are all related to subprocess checks that needs careful investigation for detecting and identifying suitable alternatives that does not break any existing features.

### DIFF
--- a/doc/source/user_guide/meshing/new_meshing_workflows.rst
+++ b/doc/source/user_guide/meshing/new_meshing_workflows.rst
@@ -238,6 +238,7 @@ Identify regions
     identify_regions.show_coordinates = True
     identify_regions.material_points_name.set_state("void-region-1")
     identify_regions.new_region_type.set_state("void")
+    identify_regions.selection_type = "object"
     identify_regions.object_selection_list.set_state(["inlet-1", "inlet-2", "inlet-3", "main"])
     identify_regions.x.set_state(374.722045740589)
     identify_regions.y.set_state(-278.9775145640143)


src/ansys/fluent/core/codegen/allapigen.py
```diff
@@ -24,7 +24,10 @@
 
 import argparse
 from pathlib import Path
-import pickle
+
+# pickle is used to serialize internally-generated API tree data only.
+# No untrusted data is deserialized, so there are no security risks.
+import pickle  # nosec: B403
```
prmukherj: @RobPasMue, @clatapie 
Do we need to provide justifications to this individually via comments in each place where we skip these checks?

RobPasMue: Hi @prmukherj ! If it were just 1 or 2 files, yes I'd agree - place the comment in every file. But in this case it seems like you are skipping it in plenty of places. If you prefer to skip globally this rule, you can provide the justification in #4911 and the @ansys/pyansys-core team can approve it (if deemed correct).

prmukherj: I have provided the justification for this one. Let's see how it goes and I will update the others as well.

src/ansys/fluent/core/search.py
```diff
@@ -29,7 +29,7 @@
 import logging
 import os
 from pathlib import Path
-import pickle
+import pickle  # nosec: B403
```
prmukherj:
```suggestion
import pickle
```

doc/source/conf.py
```diff
@@ -139,8 +141,6 @@ def _stop_fluent_container(gallery_conf, fname):
         for container_name in container_names:
             print(f"Container still running for script {fname}")
             subprocess.run(f"docker stop {container_name}", shell=is_linux)
-    except Exception:
-        pass
```
clatapie, line 143: Another option would be to print the raised exception:
```
   except Exception as e:
        log.warning(e)
```

src/ansys/fluent/core/services/datamodel_se.py
```diff
@@ -1075,14 +1075,14 @@ def _get_cached_attr(self, attrib: str) -> Any:
         cached_val = self._cached_attrs.get(attrib)
         if cached_val is None:
             cached_val = self._get_remote_attr(attrib)
-            try:  # will fail for Fluent 23.1 or before
+            # Use suppress to ignore exceptions during task lookup without triggering B110
+            with suppress(Exception):
```
clatapie: The same approach could be applied everywhere.

src/ansys/fluent/core/filereader/lispy.py
```diff
@@ -135,11 +135,11 @@ def next_token(self):
                             break
             if self.line == "":
                 return eof_object
-            token, self.line = re.match(InputPort.tokenizer, self.line).groups()
-            if token != "" and not token.startswith(";"):
+            tok, self.line = re.match(InputPort.tokenizer, self.line).groups()
```
clatapie, line 138: Can you explain what are those tokens please?
Feel free to answer internally!

src/ansys/fluent/core/services/datamodel_se.py
prmukherj: Wherever we are not concerned about the exception, like for example in workflow.py (since it will be deprecated soon), etc. we will suppress. In other places we will either log the error and add TODOs to investigate into these cases later.

Thank you.

src/ansys/fluent/core/filereader/lispy.py
prmukherj: @clatapie, here the token is not related to any passwords, etc. It is just picking up separated parts of a text based on some specific characters. For example you could refer to a public test in this repository:
test_casereader.py -> test_lispy_for_quotes

For ease of understanding I have just printed the tokens from the test:

Input:
'(define x "\n(format \\"\n-------------------------\nRunning Original Settings\n------------------------\n\\")")'

Output: 
```
Token:  (
Token:  define
Token:  x
Token:  "
(format \"
-------------------------
Running Original Settings
------------------------
\")"
Token:  )
```

Thank you.

prmukherj: The regex returns two groups: the matched token and the remaining line text

doc/source/conf.py
RobPasMue: I'd advocate for @clatapie's approach as well if possible, but if you want to ignore the exception overall this might be acceptable

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1994,10 +1994,10 @@ def __getitem__(self, name):
         """Get a child object."""
         for cname in self.child_names:
             cobj = getattr(self, cname)
-            try:
+            # Use suppress to ignore exceptions during task lookup without triggering B110
+            # TODO: Investigate why this exception handling is required?
+            with suppress(Exception):
```
RobPasMue, line 1999: Same comment as before - but I see you already created TODOs for this

src/ansys/fluent/core/utils/networking.py
```diff
@@ -160,8 +171,21 @@ def get_url_content(url: str) -> str:
...
-    with urllib.request.urlopen(url) as response:
+    # Validate URL scheme to prevent security issues
+    parsed_url = urllib.parse.urlparse(url)
+    if parsed_url.scheme not in ("http", "https"):
+        raise ValueError(
+            f"Invalid URL scheme: {parsed_url.scheme}. Only http and https are allowed."
+        )
+    with urllib.request.urlopen(
+        url
+    ) as response:  # nosec B310 (Already validated url scheme)
```
RobPasMue, line 188: Nice pre-validation

src/ansys/fluent/core/solver/flobject.py
prmukherj: yes, I have already added TODOs to check later.

src/ansys/fluent/core/filereader/lispy.py
clatapie: Sounds good, thank you for the explanation!

src/ansys/fluent/core/utils/dump_session_data.py
```diff
@@ -131,7 +131,7 @@ def __init__(self, file_name: str):
             str(Path(file_name).resolve()),
             "rb",
         ) as pickle_obj:
-            self._session_data = pickle.load(pickle_obj)
+            self._session_data = pickle.load(pickle_obj)  # nosec B301
```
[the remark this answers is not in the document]
clatapie: If the reason proposed by Copilot is correct, I would recommend to keep it as it explains why this check can be skipped safely.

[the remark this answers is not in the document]
prmukherj: @clatapie, please ignore dump_session_data.py as we will be removing this file completely in this PR only.

Thank you

PR #4921, opened 2026-02-11
refactor: Exposure & User guide of using context manager

mayankansys:
## Context
The `using` context manager was implemented in `utils` and exposed through `ansys.fluent.core.solver`, but it was not directly accessible from `ansys.fluent.core`. It was also not properly documented in the user guide, even though it works for both meshing and solver sessions.

## Change Summary
The `using` context manager was re-exported so it can be imported directly from `ansys.fluent.core`. The user guide was updated to document its usage for both meshing and solver workflows.

## Impact
Users can now import `using` directly from `ansys.fluent.core`. Documentation now reflects its general-purpose usage, improving clarity for both meshing and solver workflows.

### DIFF
--- a/src/ansys/fluent/core/__init__.py
+++ b/src/ansys/fluent/core/__init__.py
@@ -82,7 +82,7 @@ from ansys.fluent.core.utils.context_managers import using  # noqa: F401
 from ansys.fluent.core.utils.fluent_version import FluentVersion  # noqa: F401
 from ansys.fluent.core.utils.setup_for_fluent import setup_for_fluent  # noqa: F401
 
-__version__ = "0.38.dev5"
+__version__ = "0.38.dev6"
 
 _VERSION_INFO = None
 """
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -446,7 +446,7 @@ jobs:
     name: Unit Testing
     if: ${{ always() }}
     needs: build
-    runs-on: [public-ubuntu-latest-8-cores]
+    runs-on: [public-ubuntu-latest-16-cores]
     strategy:
       fail-fast: false
       matrix:
--- a/doc/changelog.d/4873.maintenance.md
+++ b/doc/changelog.d/4873.maintenance.md
@@ -1,0 +1,1 @@
+Include 'api_help_text' in documentation.
--- a/doc/changelog.d/4934.dependencies.md
+++ b/doc/changelog.d/4934.dependencies.md
@@ -1,0 +1,1 @@
+Bump version to v0.38.dev6
--- a/doc/rstgen.py
+++ b/doc/rstgen.py
@@ -309,12 +309,6 @@ def _get_reference(menu: type, menu_path: str, mode: str, is_datamodel: bool):
         if menu["name"].__name__ == "Root":
             reference = f".. _ref_{mode}_datamodel_{menu_path}:\n\n"
         else:
-            temp_path = _get_task_object_name(menu["name"])
-            menu_path = (
-                menu_path.removesuffix("/_" + temp_path)
-                if temp_path and menu_path.endswith("/_" + temp_path)
-                else menu_path
-            )
             reference = f".. _ref_{mode}_datamodel_{menu_path.rstrip('/').replace('/', '_')}:\n\n"
     else:
         if menu["name"].__name__ == "main_menu":
@@ -353,20 +347,6 @@ def _get_title(mode: str, menu_path: str, menu: type, is_datamodel: bool):
     return title
 
 
-def _get_task_object_name(menu_obj):
-    temp_path_list = menu_obj.__qualname__.split(".")
-    if (
-        len(temp_path_list) > 3
-        and temp_path_list[1] in ["task_object", "parts", "parts_files"]
-        and (
-            menu_obj.__name__.startswith("_")
-            or menu_obj.__qualname__.split(".")[-2].startswith("_")
-            or menu_obj.__qualname__.split(".")[-3].startswith("_")
-        )
-    ):
-        return temp_path_list[2]
-
-
 def _write_doc(menu: type, mode: str, is_datamodel: bool):
     """Write RST file for each menu.
 
@@ -380,23 +360,11 @@ def _write_doc(menu: type, mode: str, is_datamodel: bool):
         Whether to generate datamodel RST files.
     """
     menu_name, menu_path = _get_menu_name_path(menu["name"], is_datamodel)
-    temp_task_obj_name = _get_task_object_name(menu["name"])
-    if temp_task_obj_name:
-        temp_folder_path = _get_docdir(mode, menu_path, is_datamodel)
-        full_folder_path = Path(
-            *(p for p in temp_folder_path.parts if p != f"_{temp_task_obj_name}")
-        )
-    else:
-        full_folder_path = _get_docdir(mode, menu_path, is_datamodel)
+    full_folder_path = _get_docdir(mode, menu_path, is_datamodel)
     Path(full_folder_path).mkdir(parents=True, exist_ok=True)
     folder = Path(full_folder_path)
     index_file = folder / f"{folder.name}_contents.rst"
     title = _get_title(mode, menu_path, menu, is_datamodel)
-    title = (
-        title.removeprefix("_")
-        if temp_task_obj_name and title == "_" + temp_task_obj_name
-        else title
-    )
     with open(index_file, "w", encoding="utf8") as f:
         f.write(_get_reference(menu, menu_path, mode, is_datamodel))
         f.write(f"{title}\n")
@@ -415,15 +383,8 @@ def _write_doc(menu: type, mode: str, is_datamodel: bool):
         if menu["with_members"]:
             f.write(".. toctree::\n")
             f.write("   :hidden:\n\n")
-            for member in menu["with_members"]:
-                temp_member_name = _get_task_object_name(member)
-                if (
-                    temp_member_name
-                    and temp_member_name == member.__name__.removeprefix("_")
-                ):
-                    f.write(f"   {temp_member_name}/{temp_member_name}_contents\n")
-                else:
-                    f.write(f"   {member.__name__}/{member.__name__}_contents\n")
+            for member in _get_sorted_members(menu["with_members"]):
+                f.write(f"   {member}/{member}_contents\n")
 
 
 def _generate_all_attribute_classes(all_menus: list, main_menu: type):
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ classifiers = [
     "Operating System :: OS Independent",
 ]
 dependencies = [
-    "ansys-api-fluent>=0.3.36",
+    "ansys-api-fluent>=0.3.37",
     "ansys-platform-instancemanagement~=1.1",
     "ansys-tools-common>=0.4.0",
     "ansys-tools-filetransfer>=0.2,<1.0",
--- a/src/ansys/fluent/core/codegen/datamodelgen.py
+++ b/src/ansys/fluent/core/codegen/datamodelgen.py
@@ -137,13 +137,33 @@ def _build_parameter_docstring(name: str, t: str):
     return f"Parameter {name} of value type {_PY_TYPE_BY_DM_TYPE[t]}."
 
 
+def _get_api_help_text(info: Dict[str, Any], default: str) -> str:
+    """Prefer attrs.api_help_text, else helpstring, else default."""
+    # Newer servers (Ansys Fluent 2026 R1 and later) may attach a richer help payload under 'attrs' -> 'api_help_text'.
+    # Older entries may only have a plain 'helpstring'. If neither exists, use 'default'.
+    attrs = info.get("attrs") or {}
+    # Get the richer help payload if present; otherwise None.
+    api_help_text = attrs.get("api_help_text")
+    # Return precedence:
+    # 1) api_help_text['stringState'] if available (richer text from server).
+    # 2) info['helpstring'] if present (legacy help).
+    # 3) default (provided by caller) when nothing else exists.
+    return (
+        api_help_text.get("stringState")
+        if api_help_text
+        else info.get("helpstring", default)
+    )
+
+
 def _build_command_query_docstring(
     name: str, static_info: Any, indent: str, is_command: bool
 ):
     doc = StringIO()
     info = static_info["commandinfo"] if is_command else static_info["queryinfo"]
-    if static_info.get("helpstring"):
-        for line in static_info["helpstring"].splitlines():
+
+    top_text = _get_api_help_text(static_info, "")
+    if top_text:
+        for line in top_text.splitlines():
             doc.write(f"{indent}{line}\n")
     elif info.get("docstring"):
         for line in info["docstring"].split("."):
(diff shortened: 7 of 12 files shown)


doc/source/user_guide/session/session.rst
```diff
@@ -173,6 +173,69 @@ switched to solution mode directly. The existence of the pure session type promo
...
+  >>> import ansys.fluent.core as pyfluent
+  >>> from ansys.fluent.core.examples import download_file
+  >>> from ansys.fluent.core.solver import using
+  >>> solver = pyfluent.launch_fluent()
+  >>> case_file = download_file("mixing_elbow.cas.h5", "pyfluent/mixing_elbow")
+  >>> with using(solver):
+  ...     # Call file I/O and settings without passing solver explicitly
+  ...     read_case(file_name=case_file)
+  ...     # Access models directly
+  ...     viscous_model = Viscous()
```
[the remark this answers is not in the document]
seanpearsonuk: That's right. All snippets are fully usable if I recall correctly.

doc/source/user_guide/session/session.rst
```diff
@@ -173,6 +173,69 @@ switched to solution mode directly. The existence of the pure session type promo
...
+-----------------------------------
+
+When working with generated settings and functions, you often need to pass the current session explicitly via the
+``settings_source`` or a session argument. The ``using(session)`` context manager lets you set an "active session"
+so you can call top-level settings objects and functions without wiring the session each time. This works for both
+:obj:`~ansys.fluent.core.session_solver.Solver` and :obj:`~ansys.fluent.core.session_meshing.Meshing` sessions and is
+thread-safe.
+
+Note: You can import the context manager as ``from ansys.fluent.core import using``. It also remains available via
+``from ansys.fluent.core.solver import using`` for backward compatibility.
```
prmukherj: @mayankansys, backwards compatibility needs not be documented.

doc/source/user_guide/session/session.rst
```diff
@@ -173,6 +173,69 @@ switched to solution mode directly. The existence of the pure session type promo
 of minimal server images, which becomes significant in the context of containerization.
 
 
+Context manager for active sessions
+-----------------------------------
+
+When working with generated settings and functions, you often need to pass the current session explicitly via the
+``settings_source`` or a session argument. The ``using(session)`` context manager lets you set an "active session"
```
prmukherj: Is it better to mention the source term more generically here?

doc/source/user_guide/session/session.rst
```diff
@@ -173,6 +173,69 @@ switched to solution mode directly. The existence of the pure session type promo
...
+Solver context manager
+~~~~~~~~~~~~~~~~~~~~~~
+
+Use ``using(solver_session)`` to make a solver the active session inside a ``with`` block:
+
+.. code:: python
+
+  >>> import ansys.fluent.core as pyfluent
+  >>> from ansys.fluent.core.examples import download_file
+  from ansys.fluent.core.solver import Viscous, Read_case
```
prmukherj: @mayankansys, please check the code execution once before finally pushing all changes.

Here, Read_case -> ReadCase

doc/source/user_guide/session/session.rst
```diff
@@ -173,6 +173,69 @@ switched to solution mode directly. The existence of the pure session type promo
...
+  >>> from ansys.fluent.core.solver import using
+  >>> solver = pyfluent.launch_fluent()
+  >>> case_file = download_file("mixing_elbow.cas.h5", "pyfluent/mixing_elbow")
+  >>> with using(solver):
+  ...     # Call file I/O and settings without passing solver explicitly
+  ...     Read_case(file_name=case_file)
```
prmukherj:
```suggestion
  ...     ReadCase(file_name=case_file)
```

doc/source/user_guide/session/session.rst
```diff
@@ -173,6 +173,69 @@ switched to solution mode directly. The existence of the pure session type promo
...
+
+Use ``using(meshing_session)`` to make a meshing session active and interact with workflows without passing the session:
+
+.. code:: python
+
+  >>> import ansys.fluent.core as pyfluent
+  >>> from ansys.fluent.core import using
+  >>> meshing = pyfluent.launch_fluent(mode=pyfluent.FluentMode.MESHING)
+  >>> with using(meshing):
+  ...     wt = watertight()
```
prmukherj: This example is not correct. 

As of the current design, watertight does not support running with context manager like this.

doc/source/user_guide/session/session.rst
prmukherj: @seanpearsonuk, could you please guide on this?

doc/source/user_guide/session/session.rst
```diff
@@ -173,6 +173,48 @@ switched to solution mode directly. The existence of the pure session type promo
...
+  from ansys.fluent.core.solver import Viscous, ReadCase
+  >>> from ansys.fluent.core.solver import using
+  >>> solver = pyfluent.launch_fluent()
+  >>> case_file = download_file("mixing_elbow.cas.h5", "pyfluent/mixing_elbow")
+  >>> with using(solver):
+  ...     # Call file I/O and settings without passing solver explicitly
+  ...     ReadCase()(file_name=case_file)
+  ...     # Access models directly
+  ...     viscous_model = Viscous()
+  ...     viscous_model.model()  # returns the current viscous model state 'k-omega'
```
prmukherj: @mayankansys, if we call it like this we need to provide the output here. It is better to show the output explicitly here instead of asserting (another option), as it is a document. Please show the output as you would see if writing the same code in a command window.

src/ansys/fluent/core/__init__.py
```diff
@@ -78,6 +78,7 @@
 )
 from ansys.fluent.core.streaming_services.events_streaming import *  # noqa: F401, F403
 from ansys.fluent.core.utils import fldoc
+from ansys.fluent.core.utils.context_managers import using  # noqa: F401
```
prmukherj, line 81: 👍

doc/source/user_guide/session/session.rst
```diff
@@ -173,6 +173,48 @@ switched to solution mode directly. The existence of the pure session type promo
 of minimal server images, which becomes significant in the context of containerization.
 
 
+Context manager for active sessions
```
hpohekar, line 176: This can contain a description of actual change without technical terms something as follows:

Before, you typically had to pass an active session object into every API call.

Now you can do:

```
from ansys.fluent.core import using

with using(session):
    # API calls no longer need explicit session
```

This significantly reduces boilerplate, improves readability, and supports thread safety where each thread can set its own contextual active session.

Then subsequent sections can explain examples related to a solver and meshing sesion.

doc/changelog.d/4921.miscellaneous.md
```diff
@@ -0,0 +1 @@
+Exposure & User guide of using context manager
```
hpohekar, line 1: Expose `using` context manager in public API and document its usage

doc/source/user_guide/session/session.rst
```diff
@@ -173,6 +173,48 @@ switched to solution mode directly. The existence of the pure session type promo
...
+  >>> import ansys.fluent.core as pyfluent
+  >>> from ansys.fluent.core.examples import download_file
+  from ansys.fluent.core.solver import Viscous, ReadCase
+  >>> from ansys.fluent.core.solver import using
+  >>> solver = pyfluent.launch_fluent()
+  >>> case_file = download_file("mixing_elbow.cas.h5", "pyfluent/mixing_elbow")
+  >>> with using(solver):
+  ...     # Call file I/O and settings without passing solver explicitly
+  ...     ReadCase()(file_name=case_file)
+  ...     # Access models directly
```
hpohekar: The added doc shows basic usage; you can extend it with:

- Multiple sessions in one script
- How context manager interacts with asynchronous tasks
- Comparing before/after for common workflows

I think we can add a migration section or an example that will mention the following points:

- Show a real setup + solve script with the context manager
- Provide a section about migrating older scripts

This helps users who come from plain script use to context-manager style.

doc/source/user_guide/session/session.rst
prmukherj: I agree, in user facing documentation, terms like "generated_settings" and "functions", etc. does not make much use. Please use something generic.

doc/source/user_guide/session/session.rst
```diff
@@ -173,6 +173,48 @@ switched to solution mode directly. The existence of the pure session type promo
...
+  ...     viscous_model.model()  # returns the current viscous model state 'k-omega'
+
+Thread safety: The active session is maintained per thread, so each thread can safely set and use its own session:
+
+.. code:: python
+
+  >>> import threading
+  >>> def work(session):
+  ...     with using(session):
+  ...         assert Setup() == session.setup
```
hpohekar: This assert statement has been written for testing purpose. This can be part of the migration section as explained in the previous comment, however we can give some other example here considering user point of view.

doc/source/user_guide/session/session.rst
seanpearsonuk: This statement:

When working with generated settings and functions, you often need to pass the current session explicitly via the settings_source or a session argument.

requires an example. Otherwise, the reader won't understand it. We should lead with examples.

doc/source/user_guide/session/session.rst
```diff
@@ -173,6 +173,69 @@ switched to solution mode directly. The existence of the pure session type promo
...
 
 
+Context manager for active sessions
+-----------------------------------
+
+When working with generated settings and functions, you often need to pass the current session explicitly via the
+``settings_source`` or a session argument. The ``using(session)`` context manager lets you set an "active session"
+so you can call top-level settings objects and functions without wiring the session each time. This works for both
+:obj:`~ansys.fluent.core.session_solver.Solver` and :obj:`~ansys.fluent.core.session_meshing.Meshing` sessions and is
+thread-safe.
```
seanpearsonuk: This reference to thread-safety could give false confidence:

Current: “…and is thread-safe.”

Proposed: "The using() context manager sets a thread-local active session. Sessions activated in one thread are invisible to code running in other threads. In multi-threaded applications, always pass the session explicitly instead of relying on the implicit active session."

doc/source/user_guide/session/session.rst
seanpearsonuk: I made some related comments below, prior to seeing these.

doc/source/user_guide/session/session.rst
```diff
@@ -173,6 +173,69 @@ switched to solution mode directly. The existence of the pure session type promo
...
+  ...     viscous_model = Viscous()
+  ...     viscous_model.model()  # returns the current viscous model state
+  'k-omega'
+
+Thread safety: The active session is maintained per thread, so each thread can safely set and use its own session:
+
+.. code:: python
+
+  >>> import threading
+  >>> def work(sess):
```
seanpearsonuk: I believe that we always use the full name, `session` elsewhere in the documentation. The problem with `sess` is that it loses meaning.

doc/source/user_guide/session/session.rst
```diff
@@ -173,6 +173,69 @@ switched to solution mode directly. The existence of the pure session type promo
...
+Meshing context manager
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Use ``using(meshing_session)`` to make a meshing session active and interact with workflows without passing the session:
+
+.. code:: python
+
+  >>> import ansys.fluent.core as pyfluent
+  >>> from ansys.fluent.core import using
+  >>> meshing = pyfluent.launch_fluent(mode=pyfluent.FluentMode.MESHING)
```
seanpearsonuk: I believe that we have a policy to avoid `meshing` and `solver` as session variable names, since they are protected globals in the Fluent Python console. Trying to shadow them makes the scripts unusable in the console.

doc/source/user_guide/session/session.rst
```diff
@@ -173,6 +173,69 @@ switched to solution mode directly. The existence of the pure session type promo
...
+  >>> with using(meshing):
+  ...     wt = meshing.watertight()
+  ...     import_geometry = wt.import_geometry
+  ...     # Set states and execute tasks here
+  ...     # e.g., import_geometry.file_name.set_state(<geometry_file>); import_geometry()
+
+Outside of a ``with using(...)`` block, you can continue to pass sessions explicitly when you need to operate on
+multiple sessions in the same scope or prefer explicit control.
```
seanpearsonuk: It might be worth mentioning the multhreading scenario here, as long as we agree on that.

doc/source/user_guide/session/session.rst
mayankansys: Image


Does this looks correct ?

mayankansys: @hpohekar @prmukherj I have changed the description to be more user facing. Also, added sections like 'Before and after comparison',  'Multiple sessions in one script' and 'Thread-local behavior'. 
Kindly check.

doc/source/user_guide/session/session.rst
```diff
@@ -173,6 +173,130 @@ switched to solution mode directly. The existence of the pure session type promo
...
+
+.. code:: python
+
+  >>> from ansys.fluent.core import using
+  >>> from ansys.fluent.core.solver import Viscous
+  >>> with using(solver_session):
+  ...     print(Viscous().model())
+  k-omega
+
+This significantly reduces boilerplate and improves readability.
```
seanpearsonuk: @mayankansys Thanks for the update, a couple of suggestions on this part:

The opening here ("Before, you had to...") reads a bit like a changelog rather than reference documentation. Most readers will come to this page cold and don't need the historical context to understand the feature; it's enough to describe what `using()` does and show the example. If the migration context is useful for existing users, it might sit better as a brief note at the end rather than leading with it.

The closing sentence ("This significantly reduces boilerplate and improves readability") is a bit evaluative for reference docs; the example makes the point well enough on its own.

doc/source/user_guide/session/session.rst
```diff
@@ -173,6 +173,130 @@ switched to solution mode directly. The existence of the pure session type promo
...
+
+.. code:: python
+
+  >>> from ansys.fluent.core import using
+  >>> from ansys.fluent.core.solver import Setup, Viscous
+  >>> with using(solver_session):
+  ...     Setup() == solver_session.setup
+  True
+  ...     print(Viscous().model())
+  k-omega
```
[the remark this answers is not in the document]
seanpearsonuk: I do not believe that the `==` evaluations are needed, and we don't need a "before and after" comparison (see previous comment).

doc/source/user_guide/session/session.rst
```diff
@@ -173,6 +173,130 @@ switched to solution mode directly. The existence of the pure session type promo
...
+
+Solver context manager
```
seanpearsonuk: Examples are repetitive. THe reader will get frustrated by this. We also repeat the point about threads.

doc/source/user_guide/session/session.rst
```diff
@@ -173,6 +173,130 @@ switched to solution mode directly. The existence of the pure session type promo
...
+  >>> from ansys.fluent.core import using
+  >>> from ansys.fluent.core.solver import Viscous
+  >>> with using(solver_session_1):
+  ...     print(Viscous().model())
+  k-omega
+  >>> with using(solver_session_2):
+  ...     print(Viscous().model())
+  k-omega
+
+Thread-local behavior
```
seanpearsonuk, line 225: If including this, remove the first mention of threeading.

doc/source/user_guide/session/session.rst
mayankansys: @seanpearsonuk Based on the suggestions. I have updated the file.

doc/source/user_guide/session/session.rst
mayankansys: Thanks @seanpearsonuk. I have removed the repetitive examples and tried to make the docs simpler for the reader.

PR #4943, opened 2026-02-20
docs: clarify pyfluent version compatibility statement

mayankansys:
## Context
The documentation said:

`“All versions of PyFluent support Fluent 2022 R2 and later.”`

This sounds like a strong promise. It may make users think PyFluent guarantees support for certain Fluent versions forever. But PyFluent is open-source and does not give product-level guarantees. Also, this message was placed too low on the page, so users might miss it.

## Change Summary

- Reworded the statement to:

    `“PyFluent supports the Fluent versions that were officially supported at the time of its release.”`
 
- Removed language implying guaranteed or perpetual support.

## Rationale

- The new wording is clearer and more honest.
- It avoids sounding like a long-term support promise.
- It also helps users understand version compatibility before installing.

## Impact

- Only documentation text changed.
- No code changes.
- Users now have clearer information about version compatibility.


src/ansys/fluent/core/field_data_interfaces.py
```diff
@@ -367,9 +369,42 @@ def is_active(self, field_name):
...
+        list[VariableDescriptor]
+            List of VariableDescriptor objects for all allowed fields.
+            Fields without a corresponding VariableDescriptor are excluded.
+        """
+        result = []
+        for field_name in self._available_field_names():
+            descriptor = _naming_strategy_instance.to_variable_descriptor(field_name)
+            if descriptor is not None:
+                result.append(descriptor)
+        return result
```
hpohekar: Please could you link the issue which requires this change also updated description will help to understand the requirements. I think we can add tests to verify this.

src/ansys/fluent/core/field_data_interfaces.py
seanpearsonuk, line 1: @mayankansys Inadvertent file removal.

Otherwise looks good.

doc/source/getting_started/installation.rst
```diff
@@ -84,7 +84,7 @@ Fluent Installation
 -------------------
 
 To benefit fully from using PyFluent, you must have a licensed copy of Ansys Fluent installed.
-All versions of PyFluent support Fluent 2024 R2 and later. 
+PyFluent always supports the full set of Fluent versions officially supported at its release date.
```
prmukherj: Each PyFluent release is compatible with the complete set of Fluent versions that are officially supported as of the PyFluent release date.

Clearer?

Otherwise, looks good as well. Just please rebase with main branch to bring back the removed file.

seanpearsonuk: The issue did intend to say "is compatible with" as implied in the surrounding text. The use of "supports" there is a typo. We only need to say "PyFluent is compatible with the full set of Fluent versions officially supported at its release date." That's actually what the issue said originally.

@prmukherj @mayankansys

doc/source/index.rst
```diff
@@ -39,7 +39,7 @@ extraction.
 
 Compatibility
 -------------
-PyFluent supports **Ansys Fluent 2024 R2 and later**.
+PyFluent always supports the full set of Fluent versions officially supported at its release date.
```
seanpearsonuk: Following text, "PyFluent is *not supported* when using a **Fluent Pro** license." needs to be addressed. Apologies, just noticed.

@mayankansys

doc/source/getting_started/installation.rst
seanpearsonuk: Original text prior to @prmukherj's edit: "All versions of PyFluent are compatible with all officially supported releases of Ansys Fluent."

PR #4958, opened 2026-02-26
feat: Create RP Vars.

prmukherj:
## Context
Previously, PyFluent only supported accessing and modifying pre-existing rpvars in Fluent sessions. There was no way to create new custom user-defined rpvars programmatically from Python. Users who needed custom rpvars had to create them manually through Fluent's TUI or scheme commands, which was cumbersome and not pythonic.

## Change Summary
This PR introduces the `create()` method to the `RPVars` class, enabling users to create custom rpvars with type safety and validation.

#### Key Features

1. **Type-safe rpvar creation** with support for both Python types and `RPVarType` enum
2. **Automatic type validation** to ensure values match declared types
3. **Flexible custom-typed rpvars** using `var_type=None` for dynamic typing
4. **Clear error messages** for type mismatches and duplicate names

#### Example
```python

solver = pyfluent.launch_fluent()

# Create integer rpvar
solver.rp_vars.create(name="my-int-var", value=55, var_type=int)
assert solver.rp_vars("my-int-var") == 55

# Modify it
solver.rp_vars("my-int-var", 60)
assert solver.rp_vars("my-int-var") == 60
```

## Impact
Users can create rp-vars from PyFluent

### DIFF
--- a/src/ansys/fluent/core/rpvars.py
+++ b/src/ansys/fluent/core/rpvars.py
@@ -175,6 +175,8 @@ class RPVars:
             If the rpvar name already exists.
         TypeError
             If the value type doesn't match the specified var_type.
+        RuntimeError
+            If 'make-new-rpvar' returns #f for a reason other than the variable name already matching an existing entry.
         """
         if var_type is None:
             var_type = RPVarType.CUSTOM
@@ -219,8 +221,10 @@ class RPVars:
         if returned_val is False:
             if name in self.allowed_values():
                 raise NameError(f"'{name}' already exists as an rpvar.")
-        else:
-            return returned_val
+            raise RuntimeError(
+                f"Failed to create rpvar '{name}': make-new-rpvar returned #f."
+            )
+        return returned_val
 
     def _execute(self, cmd: str):
         scheme_val = self._eval_fn(cmd)
--- a/doc/source/user_guide/legacy/rpvars.rst
+++ b/doc/source/user_guide/legacy/rpvars.rst
@@ -76,7 +76,7 @@ is enforced to prevent mismatches:
    
    >>> # This raises NameError: rpvars already exists
    >>> solver_session.rp_vars.create(name="my-int-var", value=55, var_type=int)
-   NameError: 'my-int-var' is already a pre-defined rpvar.
+   NameError: 'my-int-var' already exists as an rpvar.
 
 **Note:** Custom-typed rpvars (created with ``var_type=None``) do not enforce type restrictions 
 and can be reassigned to any value type.
--- a/tests/test_rp_vars.py
+++ b/tests/test_rp_vars.py
@@ -138,24 +138,14 @@ def test_create_rp_vars(new_solver_session) -> None:
     assert isinstance(list_var_value, (list, tuple))
     assert list(list_var_value) == [1, 2, 3]
 
-    # Create RPVar with a tuple value and verify it is stored and retrieved correctly.
-    solver.rp_vars.create(name="my-tuple-var", value=(4, 5, 6), var_type=None)
-    tuple_var_value = solver.rp_vars("my-tuple-var")
-    assert isinstance(tuple_var_value, (list, tuple))
-    assert list(tuple_var_value) == [4, 5, 6]
-
     # REAL (float) RP var
-    solver.rp_vars.create(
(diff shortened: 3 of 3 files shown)


src/ansys/fluent/core/rpvars.py
```diff
@@ -116,7 +152,61 @@ def _get_vars(self):
...
+        elif isinstance(var_type, type):
+            python_type = var_type
+            var_type = RPVarType.from_python_type(var_type)
+        else:
+            type_map = {
+                RPVarType.INTEGER: int,
+                RPVarType.REAL: float,
+                RPVarType.BOOLEAN: bool,
+                RPVarType.STRING: str,
+                RPVarType.LIST: (list, tuple),
```
seanpearsonuk: Not sure if `'list` is actually supported as a static type of RPVar on the server-side in the same way as the other types. That would make including it redundant. Can you please check if the list type is properly protected on the server-side via appropriate unit-testing. Otherwise that type should be dropped from the list, and users would just use dynamic typing explicitly or lists.

src/ansys/fluent/core/rpvars.py
```diff
@@ -116,7 +152,61 @@ def _get_vars(self):
...
+            }
+            python_type = type_map.get(var_type)
+
```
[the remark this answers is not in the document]
prmukherj: Done

src/ansys/fluent/core/rpvars.py
```diff
@@ -91,19 +131,15 @@ def allowed_values(self) -> List[str]:
...
 
     def _get_var(self, var: str):
         if var not in self.allowed_values():
             raise RuntimeError(
                 allowed_name_error_message(
                     context="rp-vars",
                     trial_name=var,
-                    allowed_values=RPVars._allowed_values,
+                    allowed_values=self.allowed_values(),
                 )
```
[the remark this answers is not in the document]
prmukherj: Done.

src/ansys/fluent/core/rpvars.py
prmukherj: yes, list is not one of the protected ones. Removed it.

src/ansys/fluent/core/rpvars.py
```diff
@@ -116,7 +150,73 @@ def _get_vars(self):
...
+
+        Raises
+        ------
+        NameError
+            If the rpvar name already exists.
+        TypeError
+            If the value type doesn't match the specified var_type.
+        """
+        if name in self.allowed_values():
+            raise NameError(f"'{name}' already exists as an rpvar.")
```
seanpearsonuk: In principle, this check is not needed until the `make-` call returns `#f`/`false`. Worth considering, given that it might be expensive to pull the full list up front every time.

prmukherj: yes, I agree that it is expensive. However this is the server behavior:

image

seanpearsonuk: Yes, that's consistent with my point. The second call returns `#f` ( = `false`), and we can act on that on the Python side. In other words, we can do the allowed-values check only after finding that the `make-` call failed.

prmukherj: Done.

src/ansys/fluent/core/rpvars.py
```diff
@@ -116,9 +150,82 @@ def _get_vars(self):
...
+                )
+        prefix = "'" if isinstance(value, (list, tuple)) else ""
+        if var_type == RPVarType.STRING:
+            cmd = f'(make-new-rpvar {RPVars._var(name)} "{prefix}{lispy.to_string(value)}" \'{lispy.to_string(var_type.value)})'
+        else:
+            cmd = f"(make-new-rpvar {RPVars._var(name)} {prefix}{lispy.to_string(value)} '{lispy.to_string(var_type.value)})"
+        returned_val = self._execute(cmd)
+        if returned_val is False:
+            if name in self.allowed_values():
+                raise NameError(f"'{name}' already exists as an rpvar.")
```
prmukherj, line 223: I thought it gives a clear error message to users in this case, so that they know what they are doing wrong. For the other cases where make... returns False we just raise a RuntimeError.

PR #4978, opened 2026-03-06
ci: use build artifact in release testing job

mayankansys:
## Context
The "Release Testing" job was building and installing pyfluent from source (including running API codegen) instead of using the wheel artifact from the "Build" job. This meant the release tests were not testing the same package that gets published to PyPI.

## Change Summary
- Added `needs: build` to the Release Testing job
- Replaced source build + codegen steps with downloading the `PyFluent-packages` artifact and installing from the wheel
- Removed now-unnecessary steps: pip cache, `make version-info`, `make install`, API codegen, version info printing, and post-codegen reinstall

## Impact
Only the "Release Testing" CI job (triggered on release tag push) is affected. No changes to the build, unit testing, or release/publish jobs.

### DIFF
--- a/doc/changelog.d/4938.added.md
+++ b/doc/changelog.d/4938.added.md
@@ -1,0 +1,1 @@
+Add VariableDescriptor support to field data interfaces
--- a/doc/changelog.d/4984.maintenance.md
+++ b/doc/changelog.d/4984.maintenance.md
@@ -1,0 +1,1 @@
+Update lunar_lander_thermal.py to use newer syntax for auto_save
--- a/examples/00-fluent/lunar_lander_thermal.py
+++ b/examples/00-fluent/lunar_lander_thermal.py
@@ -600,7 +600,7 @@ surf_report_files["regolith-temps-rfile"] = {
 # ~~~~~~~~
 # Set the case to save only the data file at each timestep for post-processing.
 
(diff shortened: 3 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -553,20 +554,14 @@ jobs:
...
-          lookup-only: false # zizmor: ignore[cache-poisoning]
-
-      - name: Add version information
-        run: make version-info
+          name: PyFluent-packages
+          path: dist
 
       - name: Install pyfluent
-        run: make install
+        run: pip install -q --force-reinstall dist/*.whl > /dev/null
```
prmukherj, line 564: you could use make install only here

.github/workflows/ci.yml
```diff
@@ -586,21 +581,6 @@ jobs:
         env:
           FLUENT_IMAGE_TAG: ${{ vars.FLUENT_STABLE_IMAGE_DEV }}
 
-      - name: Run API codegen
-        run: make api-codegen
-        env:
-          FLUENT_IMAGE_TAG: ${{ vars.FLUENT_STABLE_IMAGE_DEV }}
-
-      - name: Print Fluent version info
```
prmukherj, line 594: I am not sure how tests will run without codegen.

PR #4990, opened 2026-03-12
fix: couple of issues in launch fluent

### LINKED ISSUE #4265: Couple of issues in launch_fluent
Following issues are reported by ACE:
1. `case_file_name`/`case_data_file_name` should be processed before `journal_file_names`.
2. The output Python journal filename is malformed when more than one filepaths are passed to `journal_file_names` (Fluent side issue).

mayankansys:
## Context:
`launch_fluent()` had sequencing issues in startup flow:

- **case/case-data** handling was not guaranteed to run before journal processing, and
- In **lightweight mode**, sync could begin before journal execution completed.
- This could lead to inconsistent initialization behavior depending on launch inputs.

## Change summary:
This PR refactors the launch flow in `launch_fluent()` to enforce deterministic order:

- Case / case-data files are processed first
- Journal files are processed after case inputs
- In lightweight mode, sync is deferred until journal execution fully completes 

## Impact:

- Improves correctness and reliability of Fluent startup flow
- Reduces risk of order-related runtime issues in lightweight launches
- Strengthens confidence with automated test coverage for these launch pathways
- Affects behavior where users provide case files, case-data files, journals, and/or use lightweight mode


mkundu1:
@mayankansys Is the following happening in the lightweight-mode?

bg session: read case + settings -> read journals -> make this as the main session -> apply changes that user made in the fg session
fg session: read settings -> start accepting changes from user -> shutdown after the main session is transferred

The journals should be read in the background session which reads the full mesh before it. 

Additionally, I think the journals should be read in the foreground session also, because use will expect the journal is already read when they start making change in the activated foreground session. But I'm not sure if the journal has, e.g., a solve call, whether that will be a no-op when only settings are read.

mayankansys:
> @mayankansys Is the following happening in the lightweight-mode?
> 
> bg session: read case + settings -> read journals -> make this as the main session -> apply changes that user made in the fg session fg session: read settings -> start accepting changes from user -> shutdown after the main session is transferred
> 
> The journals should be read in the background session which reads the full mesh before it.
> 
> Additionally, I think the journals should be read in the foreground session also, because use will expect the journal is already read when they start making change in the activated foreground session. But I'm not sure if the journal has, e.g., a solve call, whether that will be a no-op when only settings are read.

Thanks @mkundu1 for the question and to make sure I understand before I rework on this: 

The PR fixes the ordering but you're pointing at a separate problem which was, In lightweight mode the journals currently run on the foreground settings whihc doesn't have mesh. 
I agree that's wrong for the mesh-dependent journals. A journal that calls the solve needs the mesh, so should I run on the background session that reads the full mesh matching the order which you described. 

If the journal has solver call then I expect a solve in settings only session to either error or no-op which is more evidence the journals belong on the bg session. Kindly let me know what's you take on it.

One issue while implementation can be the bg session is inside `read_case_lightweight()` but journal reading currently happens in the launchers `process_case_data_and_journals` which has no handle to the bg session. 
For doing this properly I think we need to move the journal reading into the lightweight read path. 

Before I move forwards kindly let me know: 
* Am I understading the thing properly and am I on the correct path ? 
* Should Journals run only on the bg session or is there a case where they also need to run in the fg session ?

mkundu1:
@mayankansys For the case/data+journal combination we can pass the case file name via the `-case` command line flag of Fluent (there is also `-data` command line flag for data files), Fluent will figure out how to read the case/data+journal combination and we can avoid any client-side logic/assumptions.

You can start Fluent with `-help` command line flag to see all supported command line flags.

src/ansys/fluent/core/launcher/error_warning_messages.py
```diff
@@ -62,3 +62,9 @@
     "This mode is not recommended. For more details on the implications "
     "and usage of insecure mode, refer to the Fluent documentation."
 )
+
+LIGHTWEIGHT_MODE_IGNORED_WITH_JOURNAL = (
+    "``lightweight_mode`` cannot be combined with ``journal_file_names`` and "
+    "will be ignored. There is no consistent way to read a case in lightweight "
+    "mode while also executing a journal, so the case is read normally instead."
```
mkundu1: > There is no consistent way to read a case in lightweight "
>     "mode while also executing a journal, so the case is read normally instead."

I think we can remove this part as this is a low-level details which would not be of user's interest.

Also, can make the message bit more direct:
```
``lightweight_mode`` will be ignored because ``journal_file_names`` has been provided.
```

mkundu1:
> @mayankansys For the case/data+journal combination we can pass the case file name via the `-case` command line flag of Fluent (there is also `-data` command line flag for data files), Fluent will figure out how to read the case/data+journal combination and we can avoid any client-side logic/assumptions.
> 
> You can start Fluent with `-help` command line flag to see all supported command line flags.

Was this considered? My primary intention is to avoid any client-side duplication of logic for the read-sequence.

We can use the command-line flags all the time even if only case or only journal name is provided. We can get rid of all the defered_journal_read logic in the current PR.

src/ansys/fluent/core/launcher/error_warning_messages.py
mayankansys: Understood, let me keep it very clear and simple.

mkundu1:
<table>
  <thead>
    <tr>
      <th colspan="3" style="text-align: center;">Launch argument</th>
      <th rowspan="2" style="text-align: left; vertical-align: bottom;">Implementation</th>
    </tr>
    <tr>
      <th style="text-align: left;">case/data</th>
      <th style="text-align: left;">journals(s)</th>
      <th style="text-align: left;">lightweight_mode</th>
    </tr>
  </thead>
  <tbody>
    <tr><td>FALSE</td><td>FALSE</td><td>FALSE</td><td>No-op</td></tr>
    <tr><td>FALSE</td><td>FALSE</td><td>TRUE</td><td>Warn and ignore lightweight_mode, then no-op</td></tr>
    <tr><td>FALSE</td><td>TRUE</td><td>FALSE</td><td>Pass -i to Fluent</td></tr>
    <tr><td>FALSE</td><td>TRUE</td><td>TRUE</td><td>Warn and ignore lightweight_mode, then pass -i to Fluent</td></tr>
    <tr><td>TRUE</td><td>FALSE</td><td>FALSE</td><td>Pass -case/-data to Fluent</td></tr>
    <tr><td>TRUE</td><td>FALSE</td><td>TRUE</td><td>Keep current implementation with fg/bg sessions</td></tr>
    <tr><td>TRUE</td><td>TRUE</td><td>FALSE</td><td>Pass -case/-data and -i to Fluent</td></tr>
    <tr><td>TRUE</td><td>TRUE</td><td>TRUE</td><td>Warn and ignore lightweight_mode, then pass -case/-data and -i to Fluent</td></tr>
  </tbody>
</table>
Above is the implementation plan in my mind.

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -225,6 +230,14 @@ def __init__(
...
         if self.argvals.get("lightweight_mode") is None:
             self.argvals["lightweight_mode"] = False
+        # lightweight_mode reads the case in a background session while a
+        # journal would run in the foreground session; there is no consistent
+        # ordering for that combination, so ignore lightweight_mode (issue #4265).
+        if self.argvals.get("lightweight_mode") and self.argvals.get(
+            "journal_file_names"
+        ):
+            warn(LIGHTWEIGHT_MODE_IGNORED_WITH_JOURNAL, UserWarning)
+            self.argvals["lightweight_mode"] = False
```
seanpearsonuk, line 247: Similar to the earlier conversations on this method (see below), this continues a pattern I've flagged before: accumulation of low-level, untested branching logic. I understand this follows the existing style of the method, but each addition makes the eventual refactor larger and the method harder to review as a whole.

I'd like to see argument-validation branches like this (and the timeout-string construction from before) extracted into small, testable methods rather than added inline. Given this has come up multiple times now, including in internal meetings, please treat this as a standing convention going forward, not a case-by-case judgment call.

Image

Final changes to the same method in same PR:

Image

seanpearsonuk (changes requested):
As per my comment.

seanpearsonuk (approved):

src/ansys/fluent/core/launcher/launcher_utils.py
```diff
@@ -241,3 +241,169 @@ def _impl(
         return fluent_jou_arg
 
     return _impl(topy, journal_file_names)
+
+
+def _read_journals(session, journal_file_names: None | str | list[str]) -> None:
```
mkundu1: Why we still have this? When both case and journal are provided in the launch command, we'll just pass `-case <case filename> -i <journal filename>` to Fluent.

mkundu1:
@mayankansys I feel the logic is still quite complicated with the introduction of file strategy. Can we break it down into following tasks (feel free to create separate PRs if that makes things easier)?
1. Replace the existing API read-case/data call during launch with passing `-case / -data` to Fluent command line.
2. Show warning for lightweight_mode + journal combination.

What else do we need?

seanpearsonuk:
We're using the term _refactor_ liberally in the branch name, PR title and PR description. Please reserve that term for situations where it actually applies. As I understand it, this is a **bug-fix** and the title itself refers to a _couple of issues_.

I also note the PR display five labels, many of which don't appear relevant here.

src/ansys/fluent/core/launcher/fluent_container.py
```diff
@@ -476,6 +476,7 @@ def start_fluent_container(
     container_dict: dict | None = None,
     start_timeout: int = 100,
     compose_config: ComposeConfig | None = None,
+    cleanup_on_exit: bool = True,
```
mkundu1: Just noticing that, there is already an argument `remove_server_info_file` that can be passed via `container_dict`. So, no change is needed in fluent_container.py as there is an existing way to persist the server-info file for container case.

src/ansys/fluent/core/launcher/launcher_utils.py
```diff
@@ -241,3 +241,60 @@ def _impl(
...
+    """
+    fluent_case_data_arg = ""
+    if case_file_name:
+        # Convert Path to str if necessary
+        case_file_name = str(case_file_name)
+        fluent_case_data_arg += f' -case "{case_file_name}"'
+    if case_data_file_name:
+        # Convert Path to str if necessary
+        case_data_file_name = str(case_data_file_name)
+        fluent_case_data_arg += f' -data "{case_data_file_name}"'
```
mkundu1:
```suggestion
        fluent_case_data_arg += f' -case "{case_data_file_name}" -data'
```

src/ansys/fluent/core/launcher/slurm_launcher.py
```diff
@@ -614,9 +628,26 @@ def _prepare(self):
...
+
+        # Add journal files via CLI unless lightweight_mode is enabled
+        if not self._argvals.get("lightweight_mode"):
+            launch_cmd += _build_journal_argument(
+                self._argvals.get("topy"),
+                self._argvals.get("journal_file_names"),
+            )
+        else:
+            # Lightweight mode: only pass topy conversion if needed
+            launch_cmd += _build_journal_argument(
```
mkundu1: This argument combination `(self._argvals.get("topy"), None)` is not supported. We can keep the old code 617-618.

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -258,6 +271,22 @@ def __init__(
         )
         if self.argvals.get("cwd"):
             self._kwargs.update(cwd=self.argvals.get("cwd"))
+
+        # For lightweight_mode with case file, defer case reading to post-connection
```
mkundu1, line 292: Does the `lightweight_mode` + `case_data_file_name` combination already work via the fg/bg sessions?

mayankansys: It's not working via the fg/bg. I need to pass both as None. 
`self._launch_string += _build_case_data_arguments(
                None, None)`

mkundu1: Yes, blocking `lightweight_mode` + `case_data_file_name` the combination (as done in the latest revision) is appropriate

src/ansys/fluent/core/launcher/launcher_utils.py
```diff
@@ -241,3 +244,91 @@ def _impl(
...
+
+    Returns
+    -------
+    tuple[bool, str | None]
+        A tuple where:
+        - First element (bool): True if lightweight_mode should be disabled, False otherwise.
+        - Second element (str | None): Warning message if lightweight_mode should be disabled, None otherwise.
+    """
+    from ansys.fluent.core.launcher.error_warning_messages import (
+        LIGHTWEIGHT_MODE_IGNORED_WITH_CASE_DATA,
```
mkundu1: can import at top level

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -258,6 +288,24 @@ def __init__(
...
+        # to support background session orchestration. Otherwise pass via CLI.
+        # Note: case_data_file_name + lightweight_mode is not supported and is
+        # already disabled with a warning during argvals validation above.
+        if self.argvals.get("lightweight_mode") and self.argvals.get("case_file_name"):
+            # Don't add case via CLI for lightweight mode
+            self._launch_string += _build_case_data_arguments(
```
mkundu1: I don't get this part. What are we appending to `self._launch_string` in this case?

src/ansys/fluent/core/launcher/launcher_utils.py
mayankansys: Yeah.

src/ansys/fluent/core/launcher/standalone_launcher.py
mayankansys: An empty string ("").

_build_case_data_arguments(None, None) returns "" because in lightweight mode, the case is not added via CLI. Instead, it's read later post-connection using session.read_case_lightweight().

mkundu1: Then do we need to keep the true branch?

We can just keep the false branch.

mkundu1 (approved):

PR #4994, opened 2026-03-16
feat: add BaseSession solver properties

mayankansys:
## Context
Before this change, validation for launch settings (especially in tests) relied more on launch construction/dry-run behavior than on querying the live Fluent session state directly.

## Change Summary

- Added runtime properties on BaseSession:
    > - precision
    > - dimension
    > - processor_count
- Implemented these using live Fluent queries so values come from the active session.
- Updated launcher test coverage to assert these properties directly from a running session.

## Impact
- Affects session metadata access and launch validation tests.
- Improves developer/test workflows by providing direct, readable runtime checks.
- No breaking API changes; this is additive `(new BaseSession properties)`.

### DIFF
--- a/src/ansys/fluent/core/session.py
+++ b/src/ansys/fluent/core/session.py
@@ -312,6 +312,33 @@ class BaseSession:
         """Return the session ID."""
         return self._fluent_connection._id
 
+    @property
+    def precision(self) -> int:
+        """Return the current solver precision.
+
+        Returns
+        -------
+        int
+            ``1`` for single precision, ``2`` for double precision.
+        """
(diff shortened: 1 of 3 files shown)


src/ansys/fluent/core/session.py
```diff
@@ -312,6 +312,33 @@ def id(self) -> str:
...
+    @property
+    def dimension(self) -> int:
+        """Return the current dimensionality.
+
+        Returns
+        -------
+        int
+            ``2`` for 2D, ``3`` for 3D.
+        """
+        return 3 if self.scheme.eval("(rp-3d?)") else 2
```
Gobot1234, line 335: These should return the enums no?

PR #4995, opened 2026-03-17
feat: Document supported patch versions of Fluent for gRPC connection.

mkundu1:
## Context
Before this PR changes, low-level errors were shown up when an unsupported Fluent versions is used.

## Change Summary
1. Document supported Fluent versions including the patch versions required for gRPC connection.
2. If an unsupported Fluent version is used on local connection on Linux, PyFluent receives ip+port instead of UDS path from the server-info file. Raise a RuntimeError in that case.
3. If an unsupported Fluent version is used on local connection on Windows, it is not possible to detect that condition from PyFluent as the local connection mechanism on Windows (WNUA) is purely a server-side implementation. The connection will be successful in that case.

## Rationale
The PyFluent documentation should reflect the list of supported Fluent versions including the patch versions.

## Impact
Improvement in the documentation and error message.

### DIFF
--- a/src/ansys/fluent/core/fluent_connection.py
+++ b/src/ansys/fluent/core/fluent_connection.py
@@ -342,7 +342,9 @@ def _get_channel(
                 # User is most likely using an unsupported Fluent server version that uses TCP for local connections.
                 # The supported Fluent versions on Linux should always use UDS for local connections.
                 raise RuntimeError(
-                    "Fluent version is not supported. Please refer to the documentation to see the list of supported Fluent versions."
+                    "Unexpected transport mode for a local connection on this platform. "
+                    "This may indicate that the Fluent version is not supported by PyFluent. "
+                    "Please check the PyFluent documentation for supported Fluent versions."
                 )
 
 


src/ansys/fluent/core/fluent_connection.py
```diff
@@ -328,12 +328,22 @@ def _get_channel(
...
+                    host=ip,
+                    port=port,
+                    grpc_options=options,
+                )
+            else:
+                # Got non-uds transport mode on non-Windows system on local connection.
+                # User is most likely using an unsupported Fluent server version that uses TCP for local connections.
+                # The supported Fluent versions on Linux should always use UDS for local connections.
+                raise RuntimeError(
+                    "Fluent version is not supported. Please refer to the documentation to see the list of supported Fluent versions."
```
seanpearsonuk: I think there’s a bit of a mismatch here between the comment and the exception message.

The comment describes this as a *heuristic* (“most likely using an unsupported Fluent version”), but the error message presents it as a definite classification (“Fluent version is not supported”). That feels a bit too strong given we’re inferring the cause from transport behaviour rather than directly checking version compatibility.

Two suggestions:

1. **Align certainty level**
   Either:

   * Make the message slightly more probabilistic (e.g. “This may indicate that the Fluent version is not supported”), or
   * Strengthen the detection logic if we really want to make a definitive claim.

2. **Be more specific about documentation**
   “the documentation” is a bit ambiguous here. Users could reasonably look at multiple sources (Fluent docs across versions/SPs, PyFluent docs, etc.). Since PyFluent compatibility is the key concern, it would be more helpful to explicitly point to *PyFluent documentation*.

3. **Actionable guidance**
   It might help to briefly explain *what to check*, not just *where*. For example:

   * Confirm Fluent version compatibility with PyFluent
   * Ensure local connections use UDS on Linux

4. **Optional: include diagnostic detail**
   Including the detected transport mode / platform in the error message could make debugging much easier without needing to reproduce.

Something along the lines of:

```
raise RuntimeError(
    "Unexpected transport mode for a local connection on this platform. "
    "This may indicate that the Fluent version is not supported by PyFluent. "
    "Please check the PyFluent documentation for supported Fluent versions "
    "and ensure that local connections use UDS on Linux."
)
```

This keeps the message accurate, actionable, and avoids overcommitting on the root cause while still guiding the user effectively.

mkundu1: Thanks @seanpearsonuk for the error message suggestion. Users would not be able to check whether UDS is used on Linux as they will need to look inside the server-info file which is temporarily created in a temp location. So, I'll avoid that part.

PR #4996, opened 2026-03-17
feat: PRE_POST_DOC accuracy

mayankansys:
## Context
PRE_POST mode documentation could be misunderstood.

It said `calculations are not allowed, but in PyFluent this mode maps to the Solver session type`, which caused confusion.

## Change Summary
- Updated PRE_POST description in launcher docs to clearly state it maps to Solver session type.
- Updated the user guide wording for PRE_POST mode to match.

## Rationale
Keep documentation accurate and aligned with actual behavior so users understand what PRE_POST does in PyFluent.

## Impact
Affects documentation clarity for launch mode behavior.

### DIFF
--- a/src/ansys/fluent/core/launcher/launch_options.py
+++ b/src/ansys/fluent/core/launcher/launch_options.py
@@ -112,7 +112,7 @@ class FluentMode(FluentEnum):
     - ``SOLVER``: The default Ansys Fluent full solution mode allows you to set up, solve, and postprocess a problem.
     - ``SOLVER_ICING``: Fluent Icing allows you to simulate airflow, particles and ice accretion on aircraft surfaces.
     - ``SOLVER_AERO``: Fluent Aero allows you to easily explore the aerodynamic performance of aircraft from a wide range of flight regimes, from subsonic to hypersonic conditions.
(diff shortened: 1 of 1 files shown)


doc/source/user_guide/session/launching_ansys_fluent.rst
```diff
@@ -222,7 +222,7 @@ This example shows how to launch Fluent in meshing mode:
 
 Pre/post mode
 ~~~~~~~~~~~~~
-Run Ansys Fluent with only the setup and postprocessing capabilities available. It does not allow you to perform calculations.
+Run Ansys Fluent in pre/post mode. In PyFluent, this mode maps to the ``Solver`` session type, but when Fluent is running in ``PRE_POST`` mode, solve commands are not available.
```
seanpearsonuk, line 225: ~Aside from appending a brief note that the same `Solver` sesson type is used,~ the original text is fine. Just one correction to the original: post-processing should be hyphenated.

seanpearsonuk, line 225:
```suggestion
Run Ansys Fluent with only the setup and post-processing capabilities available, without the ability to perform calculations.
```

PR #4999, opened 2026-03-17
docs: Add docstrings for Field Requests.

prmukherj:
## Context
Attributes of "ScalarFieldDataRequest", "SurfaceFieldDataRequest", "VectorFieldDataRequest" and "PathlinesFieldDataRequest" were missing docstrings resulting in improper docs rendering.

## Change Summary
Docstrings added for "ScalarFieldDataRequest", "SurfaceFieldDataRequest", "VectorFieldDataRequest" and "PathlinesFieldDataRequest".

## Impact
Proper docs rendering in docs page.

image

### DIFF
--- a/src/ansys/fluent/core/field_data_interfaces.py
+++ b/src/ansys/fluent/core/field_data_interfaces.py
@@ -118,7 +118,7 @@ _set_namedtuple_field_docs(
     ScalarFieldDataRequest,
     {
         "field_name": "Scalar field name to request.",
-        "surfaces": "Surface identifiers, names, or objects exposing a ``name()`` method.",
+        "surfaces": "A sequence of valid Fluent surfaces, each identified by either an integer ID, a name string, or a settings API surface object (or any object with a name() -> str method).",
         "node_value": "Whether to request nodal values. If ``False``, element values are requested.",
         "boundary_value": "Whether to request boundary values when supported.",
     },
@@ -128,7 +128,7 @@ _set_namedtuple_field_docs(
     VectorFieldDataRequest,
     {
         "field_name": "Vector field name to request.",
-        "surfaces": "Surface identifiers, names, or objects exposing a ``name()`` method.",
+        "surfaces": "A sequence of valid Fluent surfaces, each identified by either an integer ID, a name string, or a settings API surface object (or any object with a name() -> str method).",
     },
 )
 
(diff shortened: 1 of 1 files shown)


src/ansys/fluent/core/field_data_interfaces.py
```diff
@@ -94,6 +94,66 @@ class PathlinesFieldDataRequest(NamedTuple):
...
+    "Alias for field number N" in attribute/member tables.
+    """
+    for field_name, field_doc in field_docs.items():
+        getattr(cls, field_name).__doc__ = field_doc
+
+
+_set_namedtuple_field_docs(
+    SurfaceFieldDataRequest,
+    {
+        "data_types": "Surface data entries to request, such as vertices, face connectivity, face normals, and face centroids.",
```
seanpearsonuk: Exhaustive, is this list? If so, a colon replaces 'such as'. If not exhaustive it is, should it not be?

prmukherj: Done.

src/ansys/fluent/core/field_data_interfaces.py
```diff
@@ -94,6 +94,66 @@ class PathlinesFieldDataRequest(NamedTuple):
...
+        "surfaces": "Surface identifiers, names, or objects exposing a ``name()`` method.",
```
seanpearsonuk:
```suggestion
        "surfaces": "A sequence of valid Fluent surfaces, each identified by either an integer ID, a name string, or a settings API surface object (or any object with a name() -> str method).",
```
Other strings can be updated following a similar pattern.

mayankansys (approved):

src/ansys/fluent/core/field_data_interfaces.py
seanpearsonuk: @prmukherj Was this suggestion applied? I.e. was it applied here, and also was the same pattern applied elsewhere?

prmukherj: @seanpearsonuk, sorry, forgot to comment. Yes, it was applied here and elsewhere as well. It was in auto merge so was merged without your approval.

seanpearsonuk: @prmukherj Thanks I saw the refreshed version eventually.

PR #5002, opened 2026-03-18
refactor: added the test for the preprocessing doc

mayankansys:
## Context
To reduce ambiguity and prevent regressions in the pre/post, a focused test was needed.

## Change Summary
Added one test function only in `test_launcher.py`

## Impact
Affects launcher test coverage in `test_launcher.py`

### DIFF
--- a/doc/changelog.d/5010.added.md
+++ b/doc/changelog.d/5010.added.md
@@ -1,0 +1,1 @@
+Keep record of Fluent sessions
--- a/src/ansys/fluent/core/launcher/launcher_utils.py
+++ b/src/ansys/fluent/core/launcher/launcher_utils.py
@@ -26,8 +26,10 @@ import logging
 import os
 from pathlib import Path
 import platform
+import shutil
 import socket
 import subprocess
+import sys
 import time
 from typing import Any, Dict
 import warnings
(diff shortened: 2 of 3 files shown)


doc/source/user_guide/session/launching_ansys_fluent.rst
```diff
@@ -222,7 +222,7 @@ This example shows how to launch Fluent in meshing mode:
 
 Pre/post mode
 ~~~~~~~~~~~~~
-Run Ansys Fluent with only the setup and postprocessing capabilities available. It does not allow you to perform calculations.
+Run Ansys Fluent with only the setup and post-processing capabilities available. When Fluent is running in the pre-post mode it does not allow you to perform calculations.
```
seanpearsonuk:
```suggestion
Run Ansys Fluent with only setup and post-processing capabilities available. When Fluent is running in pre-post mode it does not allow you to perform calculations.
```

tests/test_launcher.py
```diff
@@ -92,6 +92,12 @@ def test_mode():
         )
 
 
+def test_pre_post_mode_maps_to_solver_session_type():
+    assert (
+        FluentMode.PRE_POST.get_fluent_value() is FluentMode.SOLVER.get_fluent_value()
+    )
```
seanpearsonuk: This test only asserts a mapping that's baked into the design: it doesn't verify any behavioural constraint of pre-post mode. We should already have a test covering the actual constraints (no solver execution, etc.) from when the mode was introduced. If this assertion is worth keeping, it fits naturally there rather than as a standalone test.

doc/source/user_guide/session/launching_ansys_fluent.rst
seanpearsonuk: This change hasn't actually been applied -- according to what's visible in GitHub.

PR #5010, opened 2026-03-23
feat: Keep record of Fluent sessions

hpohekar:
## Context
It is possible to list already running Fluent sessions launched by PyFluent but it is not possible to connect them without connection details like ip, port and password. In the existing workflow PyFluent deletes server info file after connecting to Fluent.

## Change Summary
Once the server info file is generated, it is first copied to AppData/Roaming folder then pid is appended to it. Fluids One also follows similar approach.

## Rationale
To list and connect to running Fluent sessions launched using PyFluent.

## Impact
The existing workflow remains unchanged; the original server info file will still be deleted from the user-specified or current working directory as expected. The only addition is a secondary copy stored in the background for session management purposes.

### DIFF
--- a/src/ansys/fluent/core/launcher/launcher_utils.py
+++ b/src/ansys/fluent/core/launcher/launcher_utils.py
@@ -171,9 +171,7 @@ def _update_server_info_file(server_info_file_name: str, pid: int | None = None)
             with open(si_file, "a", encoding="utf-8") as f:
                 f.write(f"\n{pid}")
     except PermissionError:
-        raise PermissionError(
-            "Insufficient permissions to copy and update server info."
-        )
+        logger.warning("Insufficient permissions to update server info file. Skipping.")
 
 
 def _await_fluent_launch(


src/ansys/fluent/core/launcher/launcher_utils.py
```diff
@@ -149,14 +151,38 @@ def _get_subprocess_kwargs_for_fluent(env: Dict[str, Any], argvals) -> Dict[str,
     return kwargs
 
 
+def _get_app_data_root() -> Path:
+    if sys.platform == "win32":
+        return Path(os.environ.get("APPDATA", Path.home() / "AppData" / "Roaming"))
+    elif sys.platform == "darwin":
+        return Path.home() / "Library" / "Application Support"
+    else:
```
seanpearsonuk: Written as a catch-all, but this branch is really intended for Unix-like systems (Linux, XDG-compliant). Shouldn't we check the platform explicitly? That would leave a genuine else to handle unrecognised platforms intentionally.

src/ansys/fluent/core/launcher/launcher_utils.py
```diff
@@ -149,14 +151,38 @@ def _get_subprocess_kwargs_for_fluent(env: Dict[str, Any], argvals) -> Dict[str,
...
+
+def _update_server_info_file(server_info_file_name: str, pid: int | None = None):
+    servers_dir = _get_app_data_root() / "pyfluent" / "servers"
+    servers_dir.mkdir(parents=True, exist_ok=True)
+    si_file = servers_dir / Path(server_info_file_name).name
+    shutil.copy2(server_info_file_name, si_file)
+
+    if pid is not None:
+        with open(si_file, "a", encoding="utf-8") as f:
+            f.write(f"\n{pid}")
```
seanpearsonuk: Isn't this a potential unhandled `PermissionError` (for example)?

src/ansys/fluent/core/launcher/launcher_utils.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/launcher/launcher_utils.py
hpohekar: Done. Thank you.

src/ansys/fluent/core/launcher/launcher_utils.py
```diff
@@ -149,14 +151,41 @@ def _get_subprocess_kwargs_for_fluent(env: Dict[str, Any], argvals) -> Dict[str,
...
+    try:
+        servers_dir = _get_app_data_root() / "pyfluent" / "servers"
+        servers_dir.mkdir(parents=True, exist_ok=True)
+        si_file = servers_dir / Path(server_info_file_name).name
+        shutil.copy2(server_info_file_name, si_file)
+        if pid is not None:
+            with open(si_file, "a", encoding="utf-8") as f:
+                f.write(f"\n{pid}")
+    except PermissionError:
+        raise RuntimeError("Insufficient permissions to copy and update server info.")
```
seanpearsonuk:
```suggestion
        logger.warning("Insufficient permissions to update server info file. Skipping.")
```
Otherwise, updating the file becomes a do-or-die mission, which is presumably not the intent.

hpohekar: Yes, I agree.

hpohekar: Done. Thank you.

PR #5037, opened 2026-04-01
Fix mutation of data_types causing order-dependent behavior

Adir-html:
Fix mutation of data_types causing order-dependent behavior

The current implementation mutates the `data_types` list while iterating over it, which can lead to order-dependent behavior and inconsistent return types.

This PR replaces the mutation logic with a list comprehension that preserves order and avoids in-place modification:

    data_types = [
        d_type if isinstance(d_type, SurfaceDataType)
        else SurfaceDataType(d_type)
        for d_type in data_types
    ]

This ensures consistent behavior regardless of input ordering.

Tested using the reproduction steps from the issue.

### DIFF
--- a/src/ansys/fluent/core/file_session.py
+++ b/src/ansys/fluent/core/file_session.py
@@ -689,10 +689,11 @@ class FileFieldData(FieldDataSource):
         overset_mesh: bool | None = False,
         flatten_connectivity: bool = False,
     ):
-        for d_type in data_types:
-            if isinstance(d_type, str):
-                data_types.remove(d_type)
-                data_types.append(SurfaceDataType(d_type))
+        data_types = [
+    d_type if isinstance(d_type, SurfaceDataType)
+    else SurfaceDataType(d_type)
+    for d_type in data_types
+]
 
         surface_ids = self.get_surface_ids(surfaces=surfaces)
 


src/ansys/fluent/core/file_session.py
```diff
@@ -689,10 +689,11 @@ def _get_surface_data(
...
     ):
-        for d_type in data_types:
-            if isinstance(d_type, str):
-                data_types.remove(d_type)
-                data_types.append(SurfaceDataType(d_type))
+        data_types = [
+    d_type if isinstance(d_type, SurfaceDataType)
+    else SurfaceDataType(d_type)
+    for d_type in data_types
+]
```
seanpearsonuk, line 696: Great. As well as not mutating the container while iterating, it also uses a positive type check.

seanpearsonuk (approved):

src/ansys/fluent/core/file_session.py
Gobot1234: You don't really need the type check here as suggested in the original issue as Enum(Enum.member) -> Enum.member

Gobot1234:
Thank you for this but we have made a large refactor to the functionality of the original function so this isn't required any more.

PR #5067, opened 2026-04-16
feat: Update Client code w.r.t. updated datamodel grpc service.

prmukherj:
## Context

This PR updates the Fluent Python client to support the updated v1 datamodel gRPC/proto APIs, selecting v0 vs v1 implementations based on server capability.

## Change Summary

Added v1 implementations for StateEngine datamodel service + TUI datamodel service, and v1 datamodel streaming/event-streaming adapters.
Refactored datamodel event streaming to share a base implementation across v0/v1 variants and updated session(s) to select the correct implementation at runtime.
Updated DataModelCache variant-vector handling to tolerate proto field-name differences (item vs items)  <<-- only for this scenario.

## Impact
Users will be able to use new datamodel grpc service layer via PyFluent, though no impact to end user interface.


src/ansys/fluent/core/data_model_cache.py
```diff
@@ -232,6 +232,13 @@ def _update_cache_from_variant_state(
         rules_str: str,
         version,
     ):
+        def get_variant_vector_items(variant_state):
+            return getattr(
+                variant_state.variant_vector_state,
+                "items",
+                getattr(variant_state.variant_vector_state, "item", None),
```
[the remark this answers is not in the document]
prmukherj: Only for this case this issue is there and test is there that is using this in the backen.

src/ansys/fluent/core/services/datamodel_se_v1.py
```diff
@@ -0,0 +1,817 @@
...
+
+from google.protobuf.json_format import MessageToDict, ParseDict
+import grpc
+
+from ansys.api.fluent.v1 import datamodel_se_pb2 as DataModelProtoModule
+from ansys.api.fluent.v1 import datamodel_se_pb2_grpc as DataModelGrpcModule
+from ansys.api.fluent.v1.variant_pb2 import Variant
+from ansys.fluent.core.data_model_cache import DataModelCache
+from ansys.fluent.core.module_config import config
+from ansys.fluent.core.services import datamodel_se as _v0
```
seanpearsonuk: I think maybe this should be commented to explain why it's used here (similarly for the TUI module).

prmukherj: Done, Thank you @seanpearsonuk.

mkundu1:
@prmukherj Have we generally evaluated the strategy pattern to swap between the v0 and v1 versions of the implementation classes?

prmukherj:
> @prmukherj Have we generally evaluated the strategy pattern to swap between the v0 and v1 versions of the implementation classes?



> @prmukherj Have we generally evaluated the strategy pattern to swap between the v0 and v1 versions of the implementation classes?

@mkundu1, we are assuming if v1 is present in server use it, else use v0. Anything else is needed?

mkundu1 (approved):

seanpearsonuk (approved):

PR #5090, opened 2026-04-29
test: Test command return types at flobject level

mkundu1:
Test command return types at test_flobject level as there are only few instances of return-type attribute in the settings API commands.


seanpearsonuk (approved):

prmukherj (approved):

mkundu1:
All passed - https://github.com/ansys/pyfluent/actions/runs/25116390307

PR #5121, opened 2026-05-11
fix: introduce DomainError for invalid domain name (#5044)

### LINKED ISSUE #5044: TypeError in _AllowedDomainNames.valid_name
### 🔍 Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### 🐞 Description of the bug

ZoneError doesn't take a domain_name, it only takes zone_name parameter
https://github.com/ansys/pyfluent/blob/d5af341bdecdf2cf7cbdf32de9d2a6594b541d38/src/ansys/fluent/core/services/solution_variables.py#L420-L433

### 📝 Steps to reproduce

...

### 💻 Which operating system are you using?

Windows

### 📀 Which ANSYS version are you using?

27.1

### 🐍 Which Python version are you using?

3.13

### 📦 Installed packages

```shell
N/A
```

SAY-5:
## Context

`_AllowedDomainNames.valid_name` in `solution_variables.py` raised `ZoneError(domain_name=..., allowed_values=...)`, but `ZoneError.__init__` only accepts `zone_name`. Passing an invalid domain name therefore raised an unrelated `TypeError` instead of the intended validation error.

## Change Summary

- Add a `DomainError` class that mirrors `ZoneError` but uses the `domain` context in the error message.
- Switch `_AllowedDomainNames.valid_name` to raise `DomainError`.
- Re-export `DomainError` from `solution_variables_v1` alongside `ZoneError`.

## Rationale

`ZoneError` keeps its zone-specific signature (back-compatible) and a separate `DomainError` carries the domain context cleanly, matching the existing pattern.

## Impact

Any caller that passes an invalid `domain_name` to `_AllowedDomainNames.valid_name` will now receive a clear `DomainError` with the list of allowed domains, instead of a `TypeError`. `ZoneError` behavior is unchanged.

Resolves #5044.

### DIFF
--- a/src/ansys/fluent/core/services/solution_variables_v1.py
+++ b/src/ansys/fluent/core/services/solution_variables_v1.py
@@ -45,6 +45,7 @@ _to_field_name_str = _v0._to_field_name_str
 
 InvalidSolutionVariableNameError = _v0.InvalidSolutionVariableNameError
 ZoneError = _v0.ZoneError
+DomainError = _v0.DomainError
 _AllowedNames = _v0._AllowedNames
 _AllowedSvarNames = _v0._AllowedSvarNames
 _AllowedZoneNames = _v0._AllowedZoneNames
--- a/src/ansys/fluent/core/services/solution_variables.py
+++ b/src/ansys/fluent/core/services/solution_variables.py
@@ -331,6 +331,19 @@ class ZoneError(ValueError):
         )
 
 
+class DomainError(ValueError):
+    """Exception class for errors in Domain name."""
+
+    def __init__(self, domain_name: str, allowed_values: List[str]):
+        """Initialize DomainError."""
+        self.domain_name = domain_name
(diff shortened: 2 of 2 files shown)


src/ansys/fluent/core/services/solution_variables_v1.py
```diff
@@ -45,6 +45,7 @@
 
 InvalidSolutionVariableNameError = _v0.InvalidSolutionVariableNameError
 ZoneError = _v0.ZoneError
+DomainError = _v0.DomainError
```
prmukherj, line 48: 👍

prmukherj: I will cherry-pick this and add test case for this as well and create another PR once we have enough approvals.

prmukherj (approved):

PR #5133, opened 2026-05-15
refactor: disable code owners  [skip tests]

mayankansys:
Disabled auto review from codeowners.


hpohekar (approved):

seanpearsonuk (approved):

seanpearsonuk:
Hi @mayankansys 

For future reference, please try to keep wording in branch names and PR titles reasonably accurate. Here we have branch name, test/test1 and a PR category, refactor. Each gives a different and wrong signal for this change.

Note that refactor has a narrow definition, so that we would normally expect quite a small proportion of PRs to fall into that category:

> Refactoring is the process of restructuring existing code to improve its internal structure without changing its external behavior.

@mkundu1 Does our contributing doc point to a recommended set of prefixes for branch names and PR titles, which might include definitions, so that contributors are able to select accurately?

cc: @hpohekar @prmukherj

PR #5149, opened 2026-05-22
chore: Revert temporary fix for blocking get_attrs for Action items.

### LINKED ISSUE #5138: Solver settings command object issues
The following was demoed by @swilson-work in a meeting.

- `solver_session.settings.results.report.forces.option` appears to be a string argument (it has a allowed values) on the `forces` command, but `option.get_state()` returns `False`
- `option.set_state()` raises an `AttributeError`. If arguments are read-only then `AttributeError` might not be the most appropriate response.
- `forces.get_attrs(['active?'], recursive=True)` errors in an unclear way, though a similar call on its parent is effective

```python
>>> solver_session.settings.results.report.forces.option = "forces"

  File "<python-input-71>", line 1, in <module>

    solver_session.settings.results.report.forces.option = "forces"

    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "D:\AnsysDev\CODE\pyfluent\src\ansys\fluent\core\solver\flobject.py", line 490, in __setattr__

    raise AttributeError(name)

AttributeError: option

 
>>> solver_session.settings.results.report.forces.option.get_state()

False
 
>>> solver_session.settings.results.report.forces.get_attrs(['active?'], recursive=True)
 
Error: wta(1st) to string_ci_eq

Error Object: #f

  File "<python-input-132>", line 1, in <module>

    solver_session.settings.results.report.forces.get_attrs(['active?'], recursive=True)

    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "D:\AnsysDev\CODE\pyfluent\src\ansys\fluent\core\solver\flobject.py", line 432, in get_attrs

    return self.flproxy.get_attrs(self.path, attrs, recursive)

           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "D:\AnsysDev\CODE\pyfluent\src\ansys\fluent\core\services\settings.py", line 150, in _fn

    return fn(self, *args, **kwds)

  File "D:\AnsysDev\CODE\pyfluent\src\ansys\fluent\core\services\settings.py", line 418, in get_attrs

    response = self._service_impl.get_attrs(request)

  File "D:\AnsysDev\CODE\pyfluent\src\ansys\fluent\core\services\settings.py", line 129, in get_attrs

    return self._stub.GetAttrs(request, metadata=self._metadata)

           ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "D:\AnsysDev\CODE\pycfx01\.venv\Lib\site-packages\grpc\_interceptor.py", line 276, in __call__

    response, ignored_call = self._with_call(

                             ~~~~~~~~~~~~~~~^

        request,

        ^^^^^^^^

        compression=compression,

        ^^^^^^^^^^^^^^^^^^^^^^^^

    )

 ```

### PR DESCRIPTION
**Pull request by prmukherj**
## Context
Earlier recursive = True was being blocked for Action (command and query) items. Ideal scenario is that recursive=True on action items like commands should list the command arguments attributes.

## Change Summary
The ideal scenario is working.

```python
pprint.pprint(
solver.settings.solution.run_calculation.iterate.get_attrs(["active?"], recursive=True))
{'attrs': {'active?': True},
 'group_children': {'iter-count': {'attrs': {'active?': True}}}}
```

## Rationale
The server side was behaving correctly, only there were some issues with some commands like:
```python
solver_session.settings.results.report.forces.get_attrs(['active?'], recursive=True)
```

There are a few of them that are being corrected at the server level.

## Impact
Users to get the expected behavior that they were getting earlier.

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -1949,32 +1949,6 @@ class Action(Base):
                     raise allowed_values_error(name, value, allowed) from ex
             raise
 
-    def get_attrs(self, attrs, recursive=False) -> Any:
-        """Get the requested attributes for the object.
-
-        Parameters
-        ----------
-        attrs : list
-            List of attribute names to retrieve.
-        recursive : bool, optional
-            Whether to retrieve attributes recursively. Recursive queries are
-            not supported for command/query objects; if ``True`` is passed it
-            is ignored and a warning is issued, by default False.
-
-        Returns
-        -------
-        Any
-            Requested attributes.
-        """
-        if recursive:
-            warnings.warn(
-                f"Recursive attribute queries are not supported for command/query "
-                f"objects. Ignoring recursive=True for '{self.python_path}'.",
-                PyFluentUserWarning,
-            )
-            recursive = False
-        return self.flproxy.get_attrs(self.path, attrs, recursive)
-
 
 class BaseCommand(Action):
     """Executes command."""
--- a/doc/changelog.d/5149.maintenance.md
+++ b/doc/changelog.d/5149.maintenance.md
@@ -1,0 +1,1 @@
+Revert temporary fix for blocking get_attrs for Action items.
--- a/tests/test_settings_api.py
+++ b/tests/test_settings_api.py
@@ -906,7 +906,7 @@ def test_action_behavior(mixing_elbow_case_session):
     )
     solver.settings.solution.run_calculation.iterate.iter_count = 55
     assert solver.settings.solution.run_calculation.iterate.iter_count() == 55
-    with pytest.warns(PyFluentUserWarning, match="command/query object"):
-        solver.settings.solution.run_calculation.iterate.get_attrs(
-            ["active?"], recursive=True
-        )
+    result = solver.settings.solution.run_calculation.iterate.get_attrs(
+        ["active?"], recursive=True
+    )
+    assert "iter-count" in result["group_children"]


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -1949,32 +1949,6 @@ def __setattr__(self, name: str, value):
                     raise allowed_values_error(name, value, allowed) from ex
             raise
```
seanpearsonuk, line 1951: Do we still need to do something here? E.g., put try-except around the `flproxy.get_attrs()` call, and if it throws, re-try with `recursive = False`.

prmukherj: According to the settings-api design recursive=True should be allowed at all nodes and if not then there is some design issue with that api. Do you still think we need a layer like you mentioned, or better just file a bug if we get errors like this in future?

seanpearsonuk: I'm thinking about it from the users' viewpoint who would want this to just work. PyFluent still connects with Fluent versions that permanently have this issue at some paths. I believe that there should be a way to write this code so that it works transparently for any version for any path. We could still output the warning but it seems somewhat redundant since such warnings probably won't be reported.

prmukherj: @seanpearsonuk, yes this sounds right for me. However do we want to do this only for command and queries only or more generically, like you had mentioned initially "put try-except around the flproxy.get_attrs() call, and if it throws, re-try with recursive = False."

Generic is better right then?

seanpearsonuk: It should re-throw for anything with internal structure, where recursion is meaningful.

mkundu1 (approved):

seanpearsonuk (approved):

PR #5153, opened 2026-05-26
chore: Weight not to be a property of Reduction.

### LINKED ISSUE #5152: New doc issues
The API reference must be reserved for user-facing interfaces and it must be complete. 

Are all the things listed under "services" user-facing interfaces? Please check the reduction interface.

The `Reduction` class has a misplaced `weight` attribute.

prmukherj:
## Context
'weight' was a property of Reduction class. It was causing it to be improperly documented in the Api docs.

## Change Summary
'weight' is just an attribute of Reduction class now. Added a bit more context to docstring.

## Impact
The api docs to be updated now correctly.

### DIFF
--- a/src/ansys/fluent/core/services/reduction.py
+++ b/src/ansys/fluent/core/services/reduction.py
@@ -280,6 +280,21 @@ class Reduction:
 
     _proto_module = ReductionProtoModule
 
+    weight = Weight
+    """Enum of available weighting options for :meth:`sum` and :meth:`sum_if`.
+
+    Provides access to the :class:`~ansys.fluent.core.solver.function.reduction.Weight`
+    enum without requiring a separate import. Use a member of this enum as the
+    ``weight`` argument to :meth:`sum` or :meth:`sum_if`.
+
+    Available members: ``AREA``, ``VOLUME``, ``MASS``, ``MASS_FLOW_RATE``,
+    ``ABS_MASS_FLOW_RATE``.
+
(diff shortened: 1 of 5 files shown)


src/ansys/fluent/core/services/reduction.py
```diff
@@ -280,6 +280,8 @@ class Reduction:
 
     _proto_module = ReductionProtoModule
 
+    weight = Weight
```
mkundu1: Why an alias is needed? Can't we expose to user it like other enums in the library like `UIMode`?

prmukherj: since this is particular to only reduction service we had thought about an implementation like this. We can expose it like other enums as well. Please let me know your thoughts about this and will update accordingly.

This was basically done to make the below usage possible:
```python
>>> solver.fields.reduction.weight.AREA
```

mkundu1: I think we are discussing to move the user-facing part of the services.reduction module to the top level. So it can become:
```
from ansys.fluent.core.reduction import Weight
x = Weight.AREA
```

prmukherj: Yes agreed, this is consistent with the repo. I will update this. Thanks.

cc. @seanpearsonuk

seanpearsonuk (approved):

mkundu1 (approved):

PR #5168, opened 2026-06-04
feat: Add 'get_completer_info' for datamodel.

prmukherj:
## Context
Implementation of 'get_completer_info' was missing from datamodel code. Hence there was a server side patch that was applied and can be removed now.

## Change Summary
The 'get_completer_info' was available in solver via flobject.py. Followed the same implementation here.

Moved implementation for both datamodel and settings objects to a single place.

## Impact
Users would be able to query 'get_completer_info' on datamodel objects like meshing or meshing_workflow objects.

Please refer tests for details.


src/ansys/fluent/core/utils/get_completer_info.py
```diff
@@ -0,0 +1,60 @@
...
+from typing import Callable, Iterable
+
+
+def get_completer_info(
+    obj,
+    base_class: type,
+    prefix: str = "",
+    excluded: Iterable = None,
+    predicate: Callable = None,
+    get_type_for_completer_info: Callable = None,
```
mkundu1: maybe rename:
predicate -> filter_function, 
get_type_for_completer_info - > type_name_map (this can just be a map containing the custom type names?)

prmukherj: Agreed and resolved. Thanks @mkundu1

mkundu1 (approved):

seanpearsonuk (approved):

PR #5172, opened 2026-06-04
feat: Include UDS path in connection properties.

### LINKED ISSUE #5165: connection_properties are None in linux
Image

mkundu1:
## Context
For local-connection on Linux, ip+port is replaced by UDS path which should be available within session's connection properties.

## Change Summary
Include the UDS full-path in session's connection properties.

image

## Rationale
The connection properties are useful to connect another client session to the same server and the UDS path is required to connect  locally in Linux.

## Impact
The UDS path info will be available within session's connection properties.

### DIFF
--- a/src/ansys/fluent/core/fluent_connection.py
+++ b/src/ansys/fluent/core/fluent_connection.py
@@ -248,11 +248,34 @@ class ErrorState:
         self._details = ""
 
 
-@dataclass(frozen=True)
+@dataclass(frozen=True, kw_only=True)
 class FluentConnectionProperties:
-    """Stores Fluent connection properties, including connection IP, port and password;
+    """Stores Fluent connection properties, including connection IP/port or UDS path and password;
     Fluent Cortex working directory, process ID and hostname; and whether Fluent was
     launched in a docker container.
+
+    Attributes
+    ----------
+    ip : str | None
+        IP address used to connect to Fluent.
+    port : int | None
+        Port used to connect to Fluent.
+    uds_fullpath : str | Path | None
+        Full path to the Unix Domain Socket used for local connections on Linux.
+    password : str | None
+        Password used for authentication when calling Fluent services.
+    cortex_pwd : str | None
+        Fluent Cortex working directory.
+    cortex_pid : int | None
+        Process ID of the Cortex process.
+    cortex_host : str | None
+        Hostname of the Cortex process.
+    fluent_host_pid : int | None
+        Process ID of the Fluent solver host process.
+    inside_container : bool | ContainerT | None
+        Container context for Fluent.
+        Returns ``False`` when not running in a container,
+        and the container instance when running in one.
 
     Examples
     --------
@@ -268,6 +291,7 @@ class FluentConnectionProperties:
 
     ip: str | None = None
     port: int | None = None
+    uds_fullpath: str | Path | None = None
     password: str | None = None
     cortex_pwd: str | None = None
     cortex_pid: int | None = None
@@ -584,14 +608,14 @@ class FluentConnection:
         self._channel_str = None
         self._slurm_job_id = None
         self.finalizer_cbs = []
+        self._uds_fullpath = None
         if channel is not None:
             self._channel = channel
         else:
-            uds_fullpath = None
             if address is not None:
                 self._channel_str = address
-                uds_fullpath = get_uds_path(address)
-                if uds_fullpath is None:
+                self._uds_fullpath = get_uds_path(address)
+                if self._uds_fullpath is None:
                     ip, port = address.rsplit(":", 1)
                     port = int(port)
             else:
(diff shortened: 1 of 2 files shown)


prmukherj (approved):

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -247,12 +247,35 @@ def clear(self):
...
     Fluent Cortex working directory, process ID and hostname; and whether Fluent was
     launched in a docker container.
 
+    Attributes
+    ----------
+    ip : str | None
+        IP address used to connect to Fluent.
+    port : int | None
+        Port used to connect to Fluent.
+    uds_fullpath : str | Path | None
```
seanpearsonuk, line 263:
```suggestion
    uds_full_path : str | Path | None
```

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -267,6 +290,7 @@ class FluentConnectionProperties:
 
     ip: str | None = None
     port: int | None = None
+    uds_fullpath: str | Path | None = None
```
seanpearsonuk, line 294:
```suggestion
    uds_full_path: str | Path | None = None
```

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -583,14 +607,14 @@ def __init__(
         self._channel_str = None
         self._slurm_job_id = None
         self.finalizer_cbs = []
+        self._uds_fullpath = None
```
seanpearsonuk, line 611:
```suggestion
        self._uds_full_path = None
```

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -583,14 +607,14 @@ def __init__(
...
         if channel is not None:
             self._channel = channel
         else:
-            uds_fullpath = None
             if address is not None:
                 self._channel_str = address
-                uds_fullpath = get_uds_path(address)
-                if uds_fullpath is None:
+                self._uds_fullpath = get_uds_path(address)
```
seanpearsonuk, line 617:
```suggestion
                self._uds_full_path = get_uds_path(address)
```

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -583,14 +607,14 @@ def __init__(
...
+                if self._uds_fullpath is None:
```
seanpearsonuk, line 618:
```suggestion
                if self._uds_full_path is None:
```

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -599,7 +623,7 @@ def __init__(
             self._channel = _get_channel(
                 ip=ip,
                 port=port,
-                uds_fullpath=uds_fullpath,
+                uds_fullpath=self._uds_fullpath,
```
seanpearsonuk, line 627:
```suggestion
                uds_full_path=self._uds_full_path,
```

src/ansys/fluent/core/fluent_connection.py
```diff
@@ -664,14 +688,15 @@ def __init__(
...
-            port,
-            password,
-            cortex_pwd,
-            cortex_pid,
-            cortex_host,
-            fluent_host_pid,
-            inside_container,
+            ip=ip,
+            port=port,
+            uds_fullpath=self._uds_fullpath,
```
seanpearsonuk, line 694:
```suggestion
            uds_full_path=self._uds_full_path,
```

src/ansys/fluent/core/fluent_connection.py
mkundu1: @seanpearsonuk The naming followed the parameter name in the [cyberchannel module documentation](https://tools.docs.pyansys.com/version/stable/api/ansys/tools/common/cyberchannel/index.html#module-detail). The `create_channel` within cyberchannel is exposed to users and we also support passing the channel within the `FluentConnection` constructor while directly initializing the session classes (although this is not a mainstream usage pattern). Should we change the parameter name in PyFluent or be consistent with the `cyberchannel` module?

seanpearsonuk: @RobPasMue I think that the name of the parameter in `cyberchannel` could be updated. WDYT?

RobPasMue: Hi @seanpearsonuk! We can update the cyberchannel if needed - but the update would imply accepting both options in fact, at least for a while. We could then move to deprecating the old version.

Nonetheless - what's the main driver? I'm guessing readability but just in case =)

mkundu1: Moved to https://github.com/ansys/ansys-tools-common/issues/288

hpohekar (approved):

PR #5181, opened 2026-06-10
chore: Update new meshing workflow to enable switching.

prmukherj:
## Context
The meshing workflow docs were organized in an unintuitive fashion. The capabilities were listed later and the common workflow types before.

## Change Summary
1. Restructured meshing workflow docs.
2. Get rid of the limitation that stopped re-initialization of workflows.
3. Added tests for the same.

## Impact
1. Updated user docs.
2. Possible to re-initialize workflows.


prmukherj:
Additionally create rp_vars tests have been skipped here.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -12,6 +12,250 @@ Overview
 - Legacy Meshing Workflows: The PyFluent meshing API used prior to Ansys Fluent 2026 R1 remains available.
   For information on how to enable and use it, see :ref:`ref_legacy_meshing_workflow`.
```
seanpearsonuk, line 14: Leading with custom workflows is surprising. Most users will be most familiar with wtm/ftm and would prefer to start there, so the the docs would normally be organised similarly. Presumably, writing the code to build a custom workflow is quite challenging, too, which is a further reason to defer this section to later.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -12,6 +12,250 @@ Overview
...
 
+Creating new meshing workflow
```
seanpearsonuk: The heading grammar is a bit off, due to a missing article.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -12,6 +12,250 @@ Overview
...
+-----------------------------
+Use ``create_workflow()`` to build a custom workflow.
+The example below demonstrates how to create and populate a workflow.
```
seanpearsonuk: AFAIK, we're encouraged to use "you". I don't know to what extent that's true, but I'm supposing that the following is preferred according to the guidelines:
```
The following example shows you how to use ``create_workflow()`` to build a custom workflow.
```
I have also combined the two sentences because the original has two very short sentences with overlapping meaning. The lack of connectives makes for awkward reading there, too.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -12,6 +12,250 @@ Overview
...
+.. code:: python
+
+    import ansys.fluent.core as pyfluent
+    from ansys.fluent.core import examples
+
+    import_file_name = examples.download_file('mixing_elbow.pmdb', 'pyfluent/mixing_elbow')
+    meshing_session = pyfluent.launch_fluent(
+        mode=pyfluent.FluentMode.MESHING, precision=pyfluent.Precision.DOUBLE, processor_count=2
+    )
+    created_workflow = meshing_session.create_workflow()
```
seanpearsonuk: For the returned object name, `workflow` is sufficient, while `custom_workflow` adds context. The `created_` suffix isn't very interesting.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -12,6 +12,250 @@ Overview
...
+
+.. code:: python
+
+    created_workflow.import_geometry.insertable_tasks.add_local_sizing.insert()
+    created_workflow.add_local_sizing_wtm()
+
+
+Loading a saved meshing workflow
+--------------------------------
+Use ``load_workflow()`` to load and execute a previously saved workflow.
```
seanpearsonuk: The reader should encounter `save_workflow()` first. That will help to give meaning and context to this method.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -12,6 +12,250 @@ Overview
...
+Tasks can be inserted into a workflow using an object-oriented approach.
+
+.. code:: python
+
+    import ansys.fluent.core as pyfluent
+
+    meshing_session = pyfluent.launch_fluent(
+        mode=pyfluent.FluentMode.MESHING, precision=pyfluent.Precision.DOUBLE, processor_count=2
+    )
+    watertight = meshing_session.watertight()
```
seanpearsonuk: Watertight meshing is first encountered incidentally inside a general subsection about inserting new tasks, but it should be given first class status. We already had information about inserting new tasks earlier and I'm confused about the structure here.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -12,6 +12,250 @@ Overview
...
+    )
+    meshing_session = pyfluent.launch_fluent(
+        mode=pyfluent.FluentMode.MESHING, precision=pyfluent.Precision.DOUBLE, processor_count=2
+    )
+    loaded_workflow = meshing_session.load_workflow(file_path=saved_workflow_path)
+
+
+Insert new task
+---------------
+Tasks can be inserted into a workflow using an object-oriented approach.
```
seanpearsonuk: We shouldn't try to preach about what style of programming something is. If the reader cares about such things, they might notice it for themselves (or they might decide that it's a poor implementation of an object-oriented interface). Otherwise, it's just extra text to process. The real danger here is naming unfamiliar concepts that could confuse.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -12,6 +12,250 @@ Overview
...
+
+.. code:: python
+
+    ig = watertight.import_geometry
+    ig.insertable_tasks.import_boi_geometry.insert()
+    ig.insertable_tasks.import_boi_geometry.insert()
+    ig.insertable_tasks.import_boi_geometry.insert()
+    assert watertight.import_boi_geometry.arguments()
+    assert watertight.import_boi_geometry_1.arguments()
+    assert watertight.import_boi_geometry_2.arguments()
```
seanpearsonuk: This block deserves some explanatory text. The more straightforward blocks were explained. We're asking the reader to parse the code to follow what's going on.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -12,6 +12,250 @@ Overview
...
+
+.. Note::
+   **Enhanced Meshing Workflows** also supports indexed access to duplicate tasks:
```
seanpearsonuk: No need to repeat "**Enhanced**".

Again though, we can orient the wording in terms of what "**you**" can do.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -12,6 +12,250 @@ Overview
...
+Use the ``mark_as_updated()`` to explicitly mark a task as updated.
+
+.. code:: python
+
+    watertight.import_geometry.mark_as_updated()
+
+
+Renaming tasks in workflow
+--------------------------
+In **Enhanced Meshing Workflow** the display name update is decoupled from the Python attribute access:
```
seanpearsonuk: This statement is unnecessarily complicated. What meaning do you want to convey?

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -12,6 +12,250 @@ Overview
...
+       >>> watertight.import_boi_geometry
+       task < import_boi_geometry: 0 >
+       >>> watertight.import_boi_geometry[0]
+       task < import_boi_geometry: 0 >
+       >>> watertight.import_boi_geometry[1]
+       task < import_boi_geometry: 1 >
+       >>> watertight.import_boi_geometry[2]
+       task < import_boi_geometry: 2 >
+
+   Index 0 returns the first instance; calling the task or indexing with 0 are equivalent.
```
seanpearsonuk: It's unclear what is meant by "calling the task" without an example.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -12,6 +12,250 @@ Overview
...
+
+   After inserting the tasks above, the workflow contains:
```
seanpearsonuk: This sentence lacks an object after "contains". Only a code block follows, so the reader will find this confusing.

What I mean is: the sentence should completely explain the block that follows - the block must not act as a stand-in for the last part of the sentence.

What you want to say is that you can call `children()` to confirm the task list.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -12,6 +12,250 @@ Overview
...
+
+Current workflow
+~~~~~~~~~~~~~~~~
+
+.. code:: python
+
+    meshing_session.current_workflow
+
+.. Note::
+   ``current_workflow`` raises an attribute error if no workflow has been initialized.
```
seanpearsonuk: **Really?!**

What about reflection? That would be a bug.

What's wrong with a `None` value?

What's the value provided by raising an exception here?

doc/source/user_guide/meshing/new_meshing_workflows.rst
seanpearsonuk: Omit "**Enhanced**" and no need to keep saying meshing workflow anyway.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -12,6 +12,250 @@ Overview
...
+
+    watertight.import_geometry.mark_as_updated()
+
+
+Renaming tasks in workflow
+--------------------------
+In **Enhanced Meshing Workflow** the display name update is decoupled from the Python attribute access:
+
+.. Note::
+   Behavior change. Display name changes do not affect attribute access.
```
seanpearsonuk: This note only reiterates the above point, but now we're saying it more clearly.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -12,6 +12,250 @@ Overview
...
+
+       >>> watertight.import_geometry.rename(new_name="IG")
+       >>> watertight.import_geometry["IG"]
+       task < import_geometry: 0 >
+       >>> watertight.import_geometry
+       task < import_geometry: 0 >
+       >>> watertight.import_geometry[0]
+       task < import_geometry: 0 >
+
+    This allows non-Pythonic display names (for example, "I-G") without affecting attribute access.
```
seanpearsonuk: Could just use a non-Python name in the example code. And in the sentence before the code block say "...allowing non-Pythonic naming".

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -12,6 +12,250 @@ Overview
...
+
+
+Deleting tasks from workflow
+----------------------------
+Tasks can be deleted individually or in groups. In **Enhanced Meshing Workflow**,
+pass task objects to ``list_of_tasks``:
```
seanpearsonuk: No, `list_of_tasks` is the argument, and the method has not been named here. As such, the sentence is unclear.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -12,6 +12,250 @@ Overview
...
+
+       del watertight.import_boi_geometry[1]
+       watertight.import_boi_geometry.delete()
+       del watertight.create_regions
+
+
+Workflow navigation enhancements
+--------------------------------
+
+The refined API enables straightforward traversal of tasks within a workflow:
```
seanpearsonuk: ~refined~

Simply tell them what they can do and how.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -12,6 +12,250 @@ Overview
...
+
+.. Note::
+   New in **Enhanced Meshing Workflow**. This capability is not available in the legacy interface.
```
seanpearsonuk: This comment will get old, so omit. It's only new if the version is specified.

mkundu1:
@prmukherj Shall we also change the order in the section navigation? I assume that "Enhanced" is now default and "Classic" should be the 2nd item here.

image

seanpearsonuk:
> @prmukherj Can we also change the order in the section navigation? Classic should be below the standard meshing workflow.
> 
> image

In the API reference, classic is listed as a legacy API. For consistency, we'll do the same in the user guide.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -12,6 +12,250 @@ Overview
 - Legacy Meshing Workflows: The PyFluent meshing API used prior to Ansys Fluent 2026 R1 remains available.
```
mkundu1: I prefer placing this item in a "Notes" section to lower its importance compared to the "Enhanced" one. Then, we can also avoid using the word "Enhanced". User will understand it as the best available API as we present it as the default choice to the users.

doc/source/user_guide/meshing/new_meshing_workflows.rst
mkundu1: Is `current_workflow` mapped from Fluent? If not, I think we can get rid of this API as the initialization methods already return the workflow?

seanpearsonuk: I believe that the main use case would be a client connecting to Fluent where a workflow is already initialised -possibly by another client, or directly.

doc/source/user_guide/meshing/new_meshing_workflows.rst
mkundu1: Yes, I also think we should avoid mentioning "attribute access". That is not a feature or point of discussion with respect to the "Enhanced" API. We need to assume that the "Enhanced" API users will not be aware of or need not read the legacy docs to understand the "Enhanced" docs.

The legacy API has a limitation  that non-Pythonic names cannot be set. That can be mentioned in notes here or in the legacy doc.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -12,6 +12,250 @@ Overview
...
+
+
+Deleting tasks from workflow
+----------------------------
+Tasks can be deleted individually or in groups. In **Enhanced Meshing Workflow**,
+pass task objects to ``list_of_tasks``:
+
+.. Note::
+   Behavior change. Delete-by-name (strings) is replaced by passing task objects.
+   Calling ``task.delete()`` still works. See :ref:`ref_legacy_meshing_workflow` for earlier usage.
```
mkundu1: I suppose this note is meant to be a migration guide. We need to say only what is changed, i.e. the `delete_task` methods now accepts a list of task objects instead of a list of task names.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Done. Thank you.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Done, Thank you.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Done, Thank you.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Done, Thank you.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Yes, will move the familiar workflows to the top like it was and inserting the new task has been kept only in here. Create workflows only deals with inserting the first task to a workflow now.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Done, Thank you.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Done, Thank you.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: I have updated the behavior to return None now as it makes more sense. Thank you.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Done. Thank you.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Restructured.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Done. Thank you.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Done. Thank you.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Rephrased the sentence to be clearer.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Done. Thank you.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Done. Thank you.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Done. Thank you.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Rephrased the sentence.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Done. Thank you.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Done. Thank you.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Done. Thank you.

doc/source/user_guide/legacy/legacy_contents.rst
```diff
@@ -8,6 +8,7 @@ Legacy interfaces
    :hidden:
 
    meshing_workflow
+   meshing_workflows
```
seanpearsonuk, line 10: What are the two separate legacy items here, meshing_workflow &  meshing_workflows?

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -7,13 +7,16 @@ PyFluent provides access to Fluent’s meshing workflows.
 
 Overview
 --------
-- Enhanced Meshing Workflows: A PyFluent API available only when using PyFluent with Ansys Fluent 2026 R1 and later.
-  It provides clearer task organization, easier navigation, and strongly typed, well-documented arguments.
-- Legacy Meshing Workflows: The PyFluent meshing API used prior to Ansys Fluent 2026 R1 remains available.
-  For information on how to enable and use it, see :ref:`ref_legacy_meshing_workflow`.
+Meshing workflow API: A PyFluent API available only when using PyFluent with Ansys Fluent 2026 R1 and later.
```
seanpearsonuk: This is not a sentence.

The overview would be a brief summary of what you can achieve using this framework. Mentioning the 2026 R1 is correct  but we're not really saying much more than that - we're practically assuming that the reader already knows exactly what this is.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -7,13 +7,16 @@ PyFluent provides access to Fluent’s meshing workflows.
...
+It provides clearer task organization, easier navigation, and strongly typed, well-documented arguments.
```
seanpearsonuk: This statement is no longer made in the context of any other (legacy) frameworks, so saying that it is "clearer" doesn't land well.

Saying it is clearer is a bit like saying that it is object-oriented, as in the previous iteration of these changes. The point is that the reader will make up their own mind about whether it is clear or not when they read on.

Overall, this "overview" section has only taught the reader that this is a PyFluent API that can be used since 26R1.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -7,13 +7,16 @@ PyFluent provides access to Fluent’s meshing workflows.
...
 
-Watertight geometry meshing workflow
-------------------------------------
+Notes
+-----
+    The PyFluent meshing API used prior to Ansys Fluent 2026 R1 remains available.
+    For information on how to enable and use it, see :ref:`ref_legacy_meshing_workflow`.
```
seanpearsonuk:
```suggestion
    Alternatively, you can use Fluent's native meshing workflows framework which is exposed directly in PyFluent.
    See :ref:`ref_legacy_meshing_workflow`.
```

doc/source/user_guide/legacy/legacy_contents.rst
prmukherj: One is the classic meshing workflow with the task based approach. The other is object based one that we had achieved with the python wrapper. The later contains only some behavior changes.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -372,10 +374,9 @@ Switch to solution mode
 
     solver_session = meshing_session.switch_to_solver()
 
-
-2-dimensional meshing workflow
-------------------------------
-Use the **2D** meshing workflow to mesh specific two-dimensional geometries.
+2D workflow
+-----------
+Use the **2D** workflow to mesh specific two-dimensional geometries.
```
seanpearsonuk: Do we actually mean "specific" here or is the word superfluous? If it is needed, we need to add a qualification, saying what the specific quality is.

Is the text copied from Ansys help?

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -659,33 +663,30 @@ Use the ``mark_as_updated()`` to explicitly mark a task as updated.
...
-       task < import_geometry: 0 >
+   >>> watertight.import_geometry.rename(new_name="I-G")
+   >>> watertight.import_geometry["I-G"]
+   task < import_geometry: 0 >
+   >>> watertight.import_geometry
+   task < import_geometry: 0 >
 
-    This allows non-Pythonic display names (for example, "I-G") without affecting attribute access.
+.. Note::
+   The legacy meshing API does not support non-Pythonic display names.
```
seanpearsonuk: Perhaps drop this note.

General points:
- Avoid overuse of technical language. API is a prime example of that. It's possible to describe software interfaces without over-reliance on such technical terms.
- Don't shorten meshing workflow to meshing, as it loses important meaning.
- It's tempting to cross-reference repeatedly with the "legacy" framework but (for new readers especially) that's adding extra cognitive load. Let's allow them to discover this interface without outside distractions.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -659,33 +663,30 @@ Use the ``mark_as_updated()`` to explicitly mark a task as updated.
...
-Tasks can be deleted individually or in groups. In **Enhanced Meshing Workflow**,
-pass task objects to ``list_of_tasks``:
+You can delete tasks individually or in groups. To delete multiple tasks at once,
+pass task objects to the ``list_of_tasks`` argument of ``delete_tasks()``:
 
 .. Note::
-   Behavior change. Delete-by-name (strings) is replaced by passing task objects.
-   Calling ``task.delete()`` still works. See :ref:`ref_legacy_meshing_workflow` for earlier usage.
+   ``delete_tasks()`` now accepts a list of task objects instead of a list of task names.
+   See :ref:`ref_legacy_meshing_workflow` for earlier usage.
```
seanpearsonuk: My instinct is to say eliminate this. The reference is to the "legacy" Fluent framework, right - rather than an earlier version of this one? If so, it doesn't feel relevant at all.

doc/source/user_guide/meshing/new_meshing_workflows.rst
```diff
@@ -713,10 +714,9 @@ Duplicate tasks can also be deleted via indexing:
 Workflow navigation enhancements
 --------------------------------
 
-The refined API enables straightforward traversal of tasks within a workflow:
-
-.. Note::
-   New in **Enhanced Meshing Workflow**. This capability is not available in the legacy interface.
+You can traverse tasks within a workflow using navigation methods such as
+``first_child()``, ``last_child()``, ``next()``, ``previous()``, ``parent()``,
+and their corresponding ``has_*()`` predicates:
```
seanpearsonuk: Glob-style notation is potentially difficult here for our readership as is use of the term _predicates_. Keep it simple.
```suggestion
and corresponding boolean checks (``has_parent()``, ``has_next()``, etc.):
```

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: This is there in Ansys help: "The 2D Meshing workflow is available from the list of meshing workflows for perform common meshing operations on 2D geometries."

So, let's do "Use the **2D** workflow to perform common meshing operations on two-dimensional geometries."

cc. @seanpearsonuk

doc/source/user_guide/legacy/legacy_contents.rst
seanpearsonuk: OK, probably OK then

doc/source/user_guide/meshing/new_meshing_workflows.rst
seanpearsonuk: So actually the reference here is not to the Fluent framework, but to our client-side implementation?

doc/source/user_guide/meshing/new_meshing_workflows.rst
seanpearsonuk: Yes, "specific" was doing nothing.

doc/source/user_guide/legacy/legacy_contents.rst
prmukherj: In the tree and docs they appear as "Legacy meshing interface" and "Classic meshing workflow".

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: yes @seanpearsonuk. Because before 26R1 we did not have this framework in place in the server side, so we are trying to tell the user this doc is valid for PyFluent using 26R1 and beyond. If they have an older version some of the capabilities may differ.

doc/source/user_guide/legacy/legacy_contents.rst
seanpearsonuk: Got it. In the sequence v1 -> v2 -> v3, we're including v1 & v2 in _legacy_. We're including v2 because we think we're being fair to users when actually we might only be confusing them (and I'm sure that's an understatement). The fairest, kindest thing might be to erase v2 from history altogether, asserting that it was only ever an early alpha framework, leading to v3 - and that's a pretty honest assessment anyway.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: This will not be relevant after: https://github.com/ansys/pyfluent/pull/5181#discussion_r3395327216

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Done.

doc/source/user_guide/meshing/new_meshing_workflows.rst
prmukherj: Done.

prmukherj:
@seanpearsonuk, Addressed the last set of comments. Thanks.

seanpearsonuk (approved):

doc/styles/config/vocabularies/ANSYS/accept.txt
```diff
@@ -11,6 +11,7 @@
 [Ss]treaming
 API
 batch_ops
+boolean
```
prmukherj:
```suggestion
Boolean
```

mkundu1 (approved):

PR #5190, opened 2026-06-16
chore: Separate out Scalar and Vector Variable Descriptors.

### LINKED ISSUE #4876: Separate out Variable descriptor access for Vectors and Scalars
For example for field data.

prmukherj:
## Context
Separate out Variable descriptor access for Vectors and Scalars. 

Earlier:
You could use a scalar where vector is expected and vice-versa.

## Change Summary
Runtime type checking has been enabled to provide `TypeError` in exceptional cases like mentioned.

## Impact
Users will get early, self-explanatory error messages if they try to use a scalar where vector is expected and vice-versa.


src/ansys/fluent/core/field_data_interfaces.py
```diff
@@ -59,36 +64,75 @@ class SurfaceDataType(Enum):
...
+
+
+@dataclasses.dataclass(frozen=True)
+class SurfaceFieldDataRequest(BaseDataRequest):
     """Container storing parameters for surface data request."""
 
     data_types: list[SurfaceDataType] | list[str]
-    surfaces: list[int | str | object]
     overset_mesh: bool | None = False
     flatten_connectivity: bool = False
```
seanpearsonuk, line 101: Template method seems more solid than requiring derived classes to call super. The change is worth considering though not a big deal.

prmukherj: Sure @seanpearsonuk, will do. Thanks

prmukherj: Done @seanpearsonuk, Thank you.

src/ansys/fluent/core/field_data_interfaces.py
```diff
@@ -103,18 +149,31 @@ class PathlinesFieldDataRequest(NamedTuple):
     zones: list | None = None
     flatten_connectivity: bool = False
 
+    def _validate_inputs(self) -> None:
+        if not isinstance(self.field_name, (str, ScalarVariableDescriptor)):
```
Gobot1234, line 157: This I don't think works in the way you're thinking it does sadly. This is what my comment in the original issue was about.

You should use `__orig_class__` if it exists and then `typing.get_args` on that (it should be `Literal[EnumMember]` then `typing.get_args` again to check if it's a `DescriptorType.SCALAR`

prmukherj: @Gobot1234, I have made a small change in the ansys-units to support this concrete type checking.
https://github.com/ansys/pyansys-units/pull/537

seanpearsonuk (approved):

src/ansys/fluent/core/field_data_interfaces.py
```diff
@@ -59,36 +64,77 @@ class SurfaceDataType(Enum):
...
+        """Serialize dataclass fields dynamically."""
+        return {field: getattr(self, field) for field in self.__dataclass_fields__}
+
+    def _replace(self, **changes) -> "BaseDataRequest":
+        """Replicate NamedTuple._replace behavior for frozen dataclasses."""
+        return dataclasses.replace(self, **changes)
+
+    def __post_init__(self):
+        """Validate shared attributes."""
+        if not isinstance(self.surfaces, Iterable):
```
mkundu1: This is not exactly correct as string is also `Iterable`.

prmukherj: Correct, Thanks. Let's update to list?

mkundu1:
```
if not isinstance(obj, Iterable) or isinstance(obj, (str, bytes)):
```
Above is probably a good way without constructing complicated typevars.

prmukherj:
```suggestion
        if not isinstance(self.surfaces, Iterable)  or isinstance(self.surfaces, (str, bytes)):
```

src/ansys/fluent/core/field_data_interfaces.py
```diff
@@ -59,36 +64,77 @@ class SurfaceDataType(Enum):
...
+        """Serialize dataclass fields dynamically."""
+        return {field: getattr(self, field) for field in self.__dataclass_fields__}
+
+    def _replace(self, **changes) -> "BaseDataRequest":
+        """Replicate NamedTuple._replace behavior for frozen dataclasses."""
+        return dataclasses.replace(self, **changes)
+
+    def __post_init__(self):
+        """Validate shared attributes."""
+        if not isinstance(self.surfaces, Iterable)  or isinstance(self.surfaces, (str, bytes)):
```
prmukherj:
```suggestion
        if not isinstance(self.surfaces, Iterable) or isinstance(self.surfaces, (str, bytes)):
```

src/ansys/fluent/core/field_data_interfaces.py
```diff
@@ -59,36 +64,77 @@ class SurfaceDataType(Enum):
...
+class SurfaceFieldDataRequest(BaseDataRequest):
     """Container storing parameters for surface data request."""
 
     data_types: list[SurfaceDataType] | list[str]
-    surfaces: list[int | str | object]
     overset_mesh: bool | None = False
     flatten_connectivity: bool = False
 
+    def _validate_inputs(self) -> None:
+        if not isinstance(self.data_types, Iterable):
```
prmukherj:
```suggestion
        if not isinstance(self.data_types, Iterable) or isinstance(self.data_types, (str, bytes)):
```

mkundu1 (approved):

PR #5192, opened 2026-06-16
chore: Restructure some grpc services.

prmukherj:
## Context
The grpc service layer and python implementation layers were closely couple together for the grpc services.

## Change Summary
Cleanly segregate the grpc service layer and python implementation layer for 5 grpc services.
Implemented a factory layer which takes care of version checking logic, etc.

## Impact
Users to see a cleaner and well documented interface.

***********
* Design: 
***********
_grpc_services
**************
	- application_runtime_service_v0.py / application_runtime_service.py
	- health_check_service_v0.py / health_check_service.py
	- reduction_service_v0.py / reduction_service.py
	- scheme_interpreter_service_v0.py / scheme_interpreter_service.py
	- settings_service_v0.py / settings_service.py

services
********
	- application_runtime (abstract_application_runtime)
		- ApplicationRuntime (application_runtime_service)
		- ApplicationRuntimeV261 (application_runtime_service_v0)
		- ApplicationRuntimeV252 (application_runtime_service_v0, scheme_interpreter_service_v0)
		- ApplicationRuntimeOld (scheme_interpreter_service_v0)

	- health_check (abstract_health_check)
		- HealthCheck (health_check_service / health_check_service_v0)

	- reduction (abstract_reduction)
		- Reduction (reduction_service / reduction_service_v0)

	- scheme_interpreter (abstract_scheme_interpreter)
		- SchemeInterpreter (scheme_interpreter_service / scheme_interpreter_service_v0)

	- settings (abstract_settings)
		- Settings (settings_service)
		- SettingsV261 (settings_service_v0, application_runtime_service_v0, scheme_interpreter_service_v0)
		- SettingsV251 (settings_service_v0, scheme_interpreter_service_v0)
		

* The Factory layer in `_grpc_services` handles all version checking, etc.

From FluentConnection or session layer:
***************************************

	self._grpc_factory = GRPCFactory(channel, metadata, error_state)
	self.application_runtime = self._grpc_factory.application_runtime
	self.health_check = self._grpc_factory.health_check
	self.reduction = self._grpc_factory.reduction
	self.scheme = self._grpc_factory.scheme_interpreter
	self.settings = self._grpc_factory.settings


src/ansys/fluent/core/abstract_application_runtime.py
```diff
@@ -0,0 +1,106 @@
...
+    build_id: str
+    vcs_revision: str
+    vcs_branch: str
+
+
+class AbstractApplicationRuntime(ABC):
+    """Abstract base class for the Application runtime."""
+
+    @abstractmethod
+    def get_product_version(self) -> str:
```
seanpearsonuk: Can this be defined to return a version object?

prmukherj: yes. Will do

src/ansys/fluent/core/application_runtime.py
```diff
@@ -0,0 +1,379 @@
...
+service.  The raw service stubs live in:
+
+* ``ansys.fluent.core.services.application_runtime`` (v0 and v1 proto API)
+
+Class hierarchy
+---------------
+``ApplicationRuntime``
+    gRPC-based implementation (v1 proto API).
+
+``ApplicationRuntimeV261(ApplicationRuntimeV261V252)``
```
seanpearsonuk: It would help to clarify if `ApplicationRuntimeV261V252` were listed and defined as a specific item here.

src/ansys/fluent/core/application_runtime.py
```diff
@@ -0,0 +1,379 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+"""High-level application runtime wrappers.
+
+This module owns the business-logic layer on top of the ApplicationRuntime gRPC
+service.  The raw service stubs live in:
+
+* ``ansys.fluent.core.services.application_runtime`` (v0 and v1 proto API)
+
```
seanpearsonuk: Both "raw service stubs" and "v0 and v1" differ from what's in the module.

src/ansys/fluent/core/services/application_runtime.py
```diff
@@ -0,0 +1,161 @@
...
+
+        Raises
+        ------
+        ValueError
+            If app mode is unknown.
+        """
+        from ansys.fluent.core import FluentMode
+
+        request = application_runtime_pb2.GetAppModeRequest()
+        response = self._stub.GetAppMode(request, metadata=self._metadata)
```
seanpearsonuk: This method was renamed to `GetMode`.

src/ansys/fluent/core/application_runtime.py
```diff
@@ -0,0 +1,379 @@
...
+    BuildInfo,
+    ProcessInfo,
+)
+from ansys.fluent.core.streaming_services.events_streaming import SolverEvent
+
+
+class ApplicationRuntime(AbstractApplicationRuntime):
+    """Application runtime backed by the ApplicationRuntime gRPC service."""
+
+    def __init__(self, service):
```
seanpearsonuk: The `service` already abstracts away gRPC. Do we need this extra layer?

src/ansys/fluent/core/session.py
```diff
@@ -241,19 +241,21 @@ def _build_from_fluent_connection(
         if self._start_transcript:
             self.transcript.start()
 
-        self._app_utilities = _AppUtilitiesFactory._create_app_utilities(
-            self.scheme, self._fluent_connection
+        self.application_runtime = (
```
seanpearsonuk: A factory would hide those dependencies.

src/ansys/fluent/core/session.py
```diff
@@ -241,19 +241,21 @@ def _build_from_fluent_connection(
...
+            _ApplicationRuntimeFactory._create_application_runtime(
+                self.scheme, self._fluent_connection
+            )
         )
 
-        self.journal = Journal(self._app_utilities)
+        self.journal = Journal(self.application_runtime)
 
         self._datamodel_service_tui = service_creator(
```
seanpearsonuk, line 235: This code is unnecessarily dealing with multiple layers. The design is severely broken.

src/ansys/fluent/core/session.py
```diff
@@ -241,19 +241,21 @@ def _build_from_fluent_connection(
...
             "tui", supports_v1=fluent_connection._server_supports_v1
```
seanpearsonuk, line 236: We have "creator" and "factory" objects - supposed creational patterns, but we're asking version questions at the call site.

src/ansys/fluent/core/session.py
```diff
@@ -241,19 +241,21 @@ def _build_from_fluent_connection(
...
         ).create(
             fluent_connection._channel,
             fluent_connection._metadata,
             self._error_state,
-            self._app_utilities,
+            self.application_runtime,
             self.scheme,
```
seanpearsonuk, line 242: The create interface shouldn't look like this. This object should have received a simple factory. It's quite simple to fix this.

src/ansys/fluent/core/session.py
```diff
@@ -315,15 +315,27 @@ def _build_from_fluent_connection(
...
+            self._settings_service = service_creator(
+                "settings",
+                supports_v1=False,
+            ).create(
+                fluent_connection._channel,
+                fluent_connection._metadata,
+                self.application_runtime,
+                self.scheme,
+                self._error_state,
+            )
```
seanpearsonuk: Please be aware of the issues contributing to the broken design of this class - it's vital that we don't compound existing design weaknesses.
These are the signs in some of the code you've added:
- If you have a factory object, it feels wrong to have to pass every dependency through the method interface. The factory should already hide those for us, simplifying our interactions and removing direct coupling with things we don't need to care about.
- It feels wrong to be using multiple factories to create one object. It means you're reaching into lower layers and creating new couplings between unrelated modules, and complicating the code.
- It feels wrong to be doing version queries to write conditional code around factory calls. The factory should be taking care of that. Consider that the business of a factory is already to apply predicates in order to decide what to create, so that business should live there.

seanpearsonuk: When I refer to the broken design of this class, I also mean `FluentConnection` in combination with this. It should be fairly straightforward business to simply both the `*Session*` classes and `FluentConnection` with an effective factory layer over the service layer.

src/ansys/fluent/core/services/application_runtime.py
prmukherj: This will be done as part of another PR after the restructuring.

prmukherj:
Nightly tests are all passing.
https://github.com/ansys/pyfluent/actions/runs/28530589368

mkundu1 (approved):

src/ansys/fluent/core/services/abstract_settings.py
seanpearsonuk, line 1: this mod nam should end with s

src/ansys/fluent/core/_grpc_services/__init__.py
seanpearsonuk, line 1: The factory doesn't belong in this package - putting it here forces client code to reach through layers.

src/ansys/fluent/core/_grpc_services/__init__.py
```diff
@@ -0,0 +1,188 @@
...
+        method_desc = service_desc.FindMethodByName("GetProductVersion")
+        return (
+            method_desc.full_name
+            == "ansys.api.fluent.v1.application_runtime.ApplicationRuntime.GetProductVersion"
+        )
+    except KeyError:
+        return False
+
+
+class GRPCFactory:
```
seanpearsonuk: It's a service factory

src/ansys/fluent/core/_grpc_services/__init__.py
```diff
@@ -0,0 +1,188 @@
...
+    """Factory for getting instantialted public facing grpc methods."""
```
seanpearsonuk: _instantialted_, plus the description is off.

src/ansys/fluent/core/_grpc_services/__init__.py
prmukherj: Renamed to 'GRPCServiceFactory'

src/ansys/fluent/core/_grpc_services/__init__.py
prmukherj: Detailed description provided.

src/ansys/fluent/core/_grpc_services/__init__.py
prmukherj: @seanpearsonuk agreed. For now refactored the code to be independent. Later will move this out into a completely different package, probably into ansys-api-fluent.

seanpearsonuk (approved):

PR #5213, opened 2026-06-24
fix: Boolean to Allowed Strings [skip tests]

### LINKED ISSUE #4554: frozen_rotor_workflow incorrect boolean assignment
### Description of the modifications

the example https://fluent.docs.pyansys.com/version/stable/examples/00-fluent/frozen_rotor_workflow.html#sphx-glr-examples-00-fluent-frozen-rotor-workflow-py contains two issues;

`inblock_shroud.relative = False`
and 
`impeller_hub.relative = True`

but these should be;
`['Relative to Adjacent Cell Zone', 'Absolute']`

### Useful links and references

https://fluent.docs.pyansys.com/version/stable/examples/00-fluent/frozen_rotor_workflow.html#sphx-glr-examples-00-fluent-frozen-rotor-workflow-py

abhishekchitwar:
## Context
The existing implementation of wall motion used Boolean values (True/False) for the relative attribute in boundary condition setup:

``` python

impeller_hub.relative = True
inblock_shroud.relative = False

```
However, this approach does not align with the latest PyFluent API design. The relative field is not a Boolean, but a string-based setting with restricted allowed values. Using Booleans obscures the intended reference frame selection and can lead to incorrect or ambiguous physical interpretation in MRF/Frozen Rotor simulations.

## Change Summary
Updated the relative attribute from Boolean values to their corresponding explicit API options "Define wall motion relative to adjacent cell zone."

["Relative to Adjacent Cell Zone", "Absolute"]

## Rationale
The update aligns with the latest PyFluent API specification, where:

relative is defined as a String with AllowedValuesMixin
Valid values include:

"Relative to Adjacent Cell Zone"
"Absolute"

Reference: https://fluent.docs.pyansys.com/version/stable/api/solver/_autosummary/settings/relative.html#relative

## Impact
✅ Removes ambiguity introduced by Boolean mapping, earlier not transparent.
✅ Makes the code consistent with Fluent GUI terminology.
✅ Improves readability and maintainability.
✅ Prevents incorrect interpretation of reference frames in rotating machinery simulations.

### DIFF
--- a/examples/00-fluent/frozen_rotor_workflow.py
+++ b/examples/00-fluent/frozen_rotor_workflow.py
@@ -76,6 +76,12 @@ Impeller-Volute simulation using the Frozen Rotor Approach
 # *Saving the case file
 # *Closing the solver
 
+
+# .. note::
+# This example has been verified and validated using Ansys Fluent 2026 R1.
+# It is recommended to run this example with the same version.
+
+
 ################################################################################################################
 # Import required libraries/modules
 # ==============================================================================================================
--- a/examples/00-fluent/frozen_rotor_workflow.py
+++ b/examples/00-fluent/frozen_rotor_workflow.py
@@ -76,6 +76,10 @@ Impeller-Volute simulation using the Frozen Rotor Approach
 # *Saving the case file
 # *Closing the solver
 
+# .. note::
+# This example has been verified and validated using Ansys Fluent 2026 R1.
+# It is recommended to run this example with the same version.
+
 ################################################################################################################
 # Import required libraries/modules
 # ==============================================================================================================
--- a/examples/00-fluent/frozen_rotor_workflow.py
+++ b/examples/00-fluent/frozen_rotor_workflow.py
@@ -77,8 +77,9 @@ Impeller-Volute simulation using the Frozen Rotor Approach
 # *Closing the solver
 
 # .. note::
-# This example has been verified and validated using Ansys Fluent 2026 R1.
-# It is recommended to run this example with the same version.
+# This example has been verified and validated using Ansys Fluent 2025 R2.
+# It requires Ansys Fluent version 2025 R2 or later, as it uses features that are not available in 
+# earlier versions.
 
 ################################################################################################################
 # Import required libraries/modules
--- a/examples/00-fluent/frozen_rotor_workflow.py
+++ b/examples/00-fluent/frozen_rotor_workflow.py
@@ -232,7 +232,7 @@ impeller_hub = solver_session.settings.setup.boundary_conditions.wall[
     "impeller-hub"
 ].momentum
 impeller_hub.wall_motion = "Moving Wall"
-impeller_hub.relative = True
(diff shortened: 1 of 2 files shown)


prmukherj:
@abhishekchitwar, did you check for backwards-compatibility?

abhishekchitwar:
> @abhishekchitwar, did you check for backwards-compatibility?

Verified for backward compatibility with version 2024 R2. The concerned API functions correctly without issues. 

However, the momentum boundary condition lacks the velocity_spec attribute, which was introduced in later versions. Additionally, there are some API signature changes, particularly related to the turbo mesh interface.

Kindly note that the example runs with versions 2025 R2 and 2026 R2, as expected with the updates mentioned with PR. 

A note has been added to the example to inform users accordingly.

prmukherj (approved):

examples/00-fluent/frozen_rotor_workflow.py
```diff
@@ -76,6 +76,10 @@
 # *Saving the case file
 # *Closing the solver
 
+# .. note::
+# This example has been verified and validated using Ansys Fluent 2026 R1.
+# It is recommended to run this example with the same version.
```
prmukherj: @abhishekchitwar, I think it is better to avoid this sentence?

@seanpearsonuk please guide on this. Thank you.

seanpearsonuk: @abhishekchitwar Before we decide what to do here, could you please add some more information on this? Specifically:
Do we expect this example to fail with older versions? Has it been tested against older versions? If there is a specific concern related to older versions then we could phrase it in those terms rather than presenting this as a single-version example that might become obsolete with future versions.

cc: @prmukherj

abhishekchitwar: > @abhishekchitwar Before we decide what to do here, could you please add some more information on this? Specifically: Do we expect this example to fail with older versions? Has it been tested against older versions? If there is a specific concern related to older versions then we could phrase it in those terms rather than presenting this as a single-version example that might become obsolete with future versions.
> 
> cc: @prmukherj

Following are observations while doing backward compatibility test.

- I tested this example with version 2024 R2 to verify backward compatibility, and it fails because the velocity_spec attribute is not available for the momentum wall boundary condition.

- For testing purposes, I removed the velocity_spec attribute and reran the workflow; however, it then fails at the Turbo mesh interface stage.

seanpearsonuk: @abhishekchitwar Thanks! Let's be transparent and say that 2026 R1 or later is needed because the example uses features unavailable in older versions.

cc: @prmukherj

abhishekchitwar: > @abhishekchitwar Thanks! Let's be transparent and say that 2026 R1 or later is needed because the example uses features unavailable in older versions.
> 
> cc: @prmukherj

Note updated!

hpohekar (approved):

seanpearsonuk (approved):

PR #5222, opened 2026-07-01
chore: Refactor Field Data.

prmukherj:
## Context
The grpc service layer and python implementation layers were closely couple together for the Field Data grpc services.

## Change Summary
Cleanly segregate the grpc service layer and python implementation layer for the Field Data grpc service.
Implemented a factory layer which takes care of version checking logic, etc.

## Impact
Users to see a cleaner and well documented interface.

***********
* Design: 
***********
_grpc_services
**************
	- field_data_v0.py / field_data.py

services
********
	- field_data(abstract_field_data)
		- FieldData
		- FieldDataV261
		- FieldDataV252

* The Factory layer in `_grpc_services` handles all version checking, etc.

From FluentConnection or session layer:
***************************************

	self.field_data = self._grpc_factory.field_data


prmukherj:
https://github.com/ansys/pyfluent/actions/runs/28606828071

src/ansys/fluent/core/services/abstract_setting.py
seanpearsonuk, line 1: `abstract_settings.py`?

prmukherj: Right @seanpearsonuk, Thanks. During merging conflicts this was changed.

prmukherj: This was a duplicate. Removed now.

src/ansys/fluent/core/_grpc_services/field_data_service.py
```diff
@@ -0,0 +1,484 @@
...
+    }
+    np_data_type_to_proto_field_type = {
+        np.int32: field_data_pb2.FieldType.FIELD_TYPE_INT_ARRAY,
+        np.int64: field_data_pb2.FieldType.FIELD_TYPE_LONG_ARRAY,
+        np.float32: field_data_pb2.FieldType.FIELD_TYPE_FLOAT_ARRAY,
+        np.float64: field_data_pb2.FieldType.FIELD_TYPE_DOUBLE_ARRAY,
+    }
+    chunk_size = 256 * 1024
+    bytes_stream = True
+    payloadTags = {
```
mkundu1, line 474: I don't think this map is needed
```
>>> x = field_data_pb2.PayloadTag.PAYLOAD_TAG_OVERSET_MESH
>>> x
1
>>> type(x)
<class 'int'>
```

src/ansys/fluent/core/_grpc_services/field_data_service.py
```diff
@@ -0,0 +1,484 @@
...
+class _FieldDataConstants:
+    """Defines constants for Fluent field data."""
+
+    proto_field_type_to_np_data_type = {
+        field_data_pb2.FieldType.FIELD_TYPE_INT_ARRAY: np.int32,
+        field_data_pb2.FieldType.FIELD_TYPE_LONG_ARRAY: np.int64,
+        field_data_pb2.FieldType.FIELD_TYPE_FLOAT_ARRAY: np.float32,
+        field_data_pb2.FieldType.FIELD_TYPE_DOUBLE_ARRAY: np.float64,
+    }
+    np_data_type_to_proto_field_type = {
```
mkundu1, line 469: We can avoid duplicating the data:
```
np_data_type_to_proto_field_type = {v : k for k, v in proto_field_type_to_np_data_type.items()}
```

src/ansys/fluent/core/services/solution_variables.py
```diff
@@ -54,6 +51,32 @@
 _to_field_name_str = naming_strategy().to_string
 
 
+class _FieldDataConstants:
```
mkundu1, line 54: I suppose this is temporary and will be cleaned up when the svar service is decomposed?

prmukherj: yes, this will be done as part of another PR after Field Data.

src/ansys/fluent/core/_grpc_services/field_data_service.py
prmukherj: The proto enum values are sequential (1, 2, 3, 4) but the map converts them to bit positions (1, 2, 4, 8).

It can actually be replaced with a formula, are you suggesting that? If so will do it in later stage. I am focusing on only the refactoring as part of this PR.

mkundu1 (approved):

seanpearsonuk (approved):

PR #5240, opened 2026-07-08
chore: Update license header via upgrading pre-commit hook.

### LINKED ISSUE #5208: License statement change
Please see pyansys group message for full details

prmukherj:
As title says.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -102,4 +102,4 @@ repos:
   hooks:
     - id: add-license-headers
       args:
-      - --start_year=2023
+      - --start_year=2021
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -75,7 +75,7 @@ repos:
 #    ]
 
 - repo: https://github.com/pre-commit/pre-commit-hooks
-  rev: v5.0.0
+  rev: 3e8a8703264a2f4a69428a0aa4dcb512790b2c8c # frozen: v6.0.0
   hooks:
     - id: check-yaml
 
@@ -98,9 +98,8 @@ repos:
       args: []
 
 - repo: https://github.com/ansys/pre-commit-hooks
-  rev: v0.5.1
+  rev: 55c5b6a5439abd847613579c6924da750e8b78b2 # frozen: v0.8.0
   hooks:
     - id: add-license-headers
-      stages: [manual]
       args:
-      - --start_year=2021
+      - --start_year=2023
--- a/examples/00-fluent/ahmed_body_workflow.py
+++ b/examples/00-fluent/ahmed_body_workflow.py
@@ -1,11 +1,4 @@
-# /// script
-# dependencies = [
-#   "ansys-fluent-core",
-#   "ansys-fluent-visualization",
-# ]
-# ///
-
-# Copyright (C) 2021 - 2026 ANSYS, Inc. and/or its affiliates.
+# Copyright (C) 2023 - 2026 Synopsys, Inc. and ANSYS, Inc. All rights reserved.
 # SPDX-License-Identifier: MIT
 #
 #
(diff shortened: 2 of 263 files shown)


examples/00-fluent/ahmed_body_workflow.py
```diff
@@ -1,11 +1,4 @@
-# /// script
-# dependencies = [
-#   "ansys-fluent-core",
-#   "ansys-fluent-visualization",
-# ]
-# ///
-
-# Copyright (C) 2021 - 2026 ANSYS, Inc. and/or its affiliates.
+# Copyright (C) 2023 - 2026 Synopsys, Inc. and ANSYS, Inc. All rights reserved.
```
prmukherj: @abhishekchitwar, just to let you know that if I upgrade the license statement via pre-commit hook, it will get rid of the dependency information that you have with the examples. Actually the convention is to have the license as the first statement and thus the pre-commit hook is performing correctly. If we want to preserve the dependency information then we need to put it in some other place below the license statement. Please let me know what to do and will anyone from your side be able to do the refactoring if the dependency information in examples are needed to be retained?

cc. @mkundu1, @seanpearsonuk

prmukherj: @Gobot1234, the dependencies were added as part of https://github.com/ansys/pyfluent/pull/4579.

I don't think so it is following PyAnsys guidelines to add dependencies with license headers. We might add it somewhere else? @Revathyvenugopal162, please let us know if you have suggestions here. Thank you.

Revathyvenugopal162: We normally don't add dependencies along with license headers. 

However, if that's a requirement, one option could be to keep certain dependency groups, such as examples, so that users can install them only when needed:

```pip install ansys-fluent-core[examples]```

Most of these dependencies may overlap with the documentation dependencies, although we could still keep the documentation dependencies separate to avoid unnecessary installations.

If the main goal is to inform users about the required dependencies, another approach would be to add a **Dependencies section** in the docstrings. This would be rendered automatically in the documentation and would also remain available to users if they download the example as a notebook or script. You can add a **Dependencies** section in the same way that the **Objective** section is added in this example. I think that would provide good visibility without having to manage the dependencies through license headers.

File: .pre-commit-config.yaml [hunk 2]
```diff
@@ -98,9 +98,8 @@ repos:
...
 
 - repo: https://github.com/ansys/pre-commit-hooks
-  rev: v0.5.1
+  rev: 55c5b6a5439abd847613579c6924da750e8b78b2 # frozen: v0.8.0
   hooks:
     - id: add-license-headers
-      stages: [manual]
       args:
-      - --start_year=2021
+      - --start_year=2023
```
Revathyvenugopal162:
```suggestion
      - --start_year=2021
```
Shouldn't this be 2021? Just checking whether this was intentional.

seanpearsonuk (approved):

Revathyvenugopal162 (approved):

.pre-commit-config.yaml
prmukherj: Right, Thanks @Revathyvenugopal162

examples/00-fluent/ahmed_body_workflow.py
prmukherj: I agree, the last option makes most sense. @Gobot1234, could you please confirm.

Gobot1234: I think a section on it would be the best alternative if we can't use PEP 723 style dependencies

prmukherj: I have added the dependency section in all examples.

Gobot1234 (approved):

mkundu1 (approved):

PR #5242, opened 2026-07-08
refactor: avoid deleting server-info file 

### LINKED ISSUE #5145: Avoid deleting server-info file if cleanup_on_exit=False is passed to launch_fluent
I think the server-info file should not be deleted if the argument `cleanup_on_exit=False` is passed to `launch_fluent`. This will help triaging issues related to port-setup when the client-server connection fails after launching Fluent. When Fluent is manually exited after launched via `launch_fluent(..., cleanup_on_exit=False)`, the server-info file will be automatically deleted by Fluent (please confirm the behaviour).

mayankansys:
## Context
When using `launch_fluent(..., cleanup_on_exit=False)`, the server-info file was being deleted during cleanup or workflow execution even though the user explicitly requested to avoid cleanup. This caused avoidable issues when that file was still needed for triaging or debugging, particularly for troubleshooting client-server connection failures related to port setup. The server-info file is essential for understanding how a Fluent process was initialized and should be preserved when cleanup is disabled.

## Change Summary
This change refactors the cleanup logic to respect the `cleanup_on_exit=False` parameter and avoid deleting the server-info file when requested. The refactoring updates:
- The server-info file deletion logic in the `StandaloneLauncher` finally block to conditionally delete based on the `cleanup_on_exit` parameter
- The deletion logic changed from unconditionally deleting to using `if self.argvals.get("cleanup_on_exit", True)` to check the parameter first
- Related tests in `tests/test_launcher.py` to verify the new behavior with three new test cases covering cleanup_on_exit=False, cleanup_on_exit=True, and default behavior

## Rationale
Preserving the server-info file when `cleanup_on_exit=False` is critical for debugging and issue triaging. This file contains port configuration and connection metadata that becomes inaccessible once deleted, making it impossible to diagnose connection failures after they occur. When Fluent is manually exited after being launched with `cleanup_on_exit=False`, the Fluent process itself will handle cleanup. PyFluent should not preemptively delete resources the user explicitly opted to preserve.

## Impact
- **Improved Stability:** Debugging workflows are no longer disrupted by premature resource cleanup
- **Better Traceability:** Port-setup issues and connection failures can now be investigated using the preserved server-info file
- **Backward Compatible:** User-facing behavior remains unchanged; only the internal cleanup logic is refactored to be more respectful of user settings
- **Affected Workflows:** This change affects all cleanup workflows involving the server-info file, ensuring they honor the `cleanup_on_exit` parameter consistently


mkundu1:
@mayankansys Does the issue exist also in standalone launcher?

seanpearsonuk (changes requested):
The PR description misses out all the important information about: what was wrong; the rationale for the change; what is changed now. The key points are buried in the original issue and none of that information will be visible in the log. The reader is simply left guessing.

mayankansys:
> The PR description misses out all the important information about: what was wrong; the rationale for the change; what is changed now. The key points are buried in the original issue and none of that information will be visible in the log. The reader is simply left guessing.

Thanks @seanpearsonuk for pointing this. I have updated the PR description. Kindly have a look

mayankansys:
> @mayankansys Does the issue exist also in standalone launcher?

Thanks @mainakk that file is missed due to some conflicts. I have updated it, kindly have a look.

prmukherj (approved):

src/ansys/fluent/core/launcher/fluent_container.py
```diff
@@ -606,5 +610,9 @@ def start_fluent_container(
         logger.error(f"Exception caught - {type(ex).__name__}: {ex}")
         raise LaunchFluentError(launch_string) from ex
     finally:
-        if remove_server_info_file and host_server_info_file.exists():
+        if (
+            remove_server_info_file
+            and cleanup_on_exit
+            and host_server_info_file.exists()
```
seanpearsonuk: `unlink()` checks `exists()` - no need to duplicate the call.

seanpearsonuk (changes requested):
https://github.com/ansys/pyfluent/pull/5242/changes#r3602456035

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -378,6 +378,7 @@ def __call__(
             logger.error(f"Exception caught - {type(ex).__name__}: {ex}")
             raise LaunchFluentError(self._launch_cmd) from ex
         finally:
-            server_info_file = Path(self._server_info_file_name)
-            if server_info_file.exists():
-                server_info_file.unlink()
+            if self.argvals.get("cleanup_on_exit", True):
+                server_info_file = Path(self._server_info_file_name)
+                if server_info_file.exists():
```
seanpearsonuk: Again `exists()` + `unlink()`

tests/test_launcher.py
```diff
@@ -678,6 +678,133 @@ def test_respect_driver_is_not_null_in_linux():
...
+            ):
+                host_server_info_file.unlink()
+
+            assert (
+                server_info_file_path.exists()
+            ), "Server-info file should be preserved when cleanup_on_exit=False"
+
+        finally:
+            if server_info_file_path.exists():
+                server_info_file_path.unlink()
```
seanpearsonuk: just unlink

tests/test_launcher.py
```diff
@@ -678,6 +678,133 @@ def test_respect_driver_is_not_null_in_linux():
...
+            # Simulate the finally block logic with cleanup_on_exit=True
+            remove_server_info_file = True
+            cleanup_on_exit = True
+            host_server_info_file = server_info_file_path
+
+            # This is the actual condition from fluent_container.py
+            if (
+                remove_server_info_file
+                and cleanup_on_exit
+                and host_server_info_file.exists()
```
seanpearsonuk: This also and any others you find

tests/test_launcher.py
```diff
@@ -799,3 +926,231 @@ def test_idle_timeout(monkeypatch):
...
+        finally:
+            if true_path.exists():
+                true_path.unlink()
+
+        # Test with cleanup_on_exit=False
+        with tempfile.NamedTemporaryFile(
+            suffix=".txt", prefix="serverinfo-", delete=False, dir=None
+        ) as tmp_file:
+            false_path = Path(tmp_file.name)
+            tmp_file.write(b"cleanup_false_test")
```
seanpearsonuk: `tmp_file` already has the same API. The way to reason when you see code like this is that it would be surprising if Python were forcing you to write this code. The same thought process applies for `exists` + `unlink` - it's simply that certain patterns stick out as being surprising.

Noting that there are more of these elsewhere in this module.

tests/test_launcher.py
mayankansys: Noted.

src/ansys/fluent/core/launcher/fluent_container.py
mkundu1: @seanpearsonuk I think we should check for existence before calling unlink.
```
Python 3.14.5 (tags/v3.14.5:5607950, May 10 2026, 10:43:50) [MSC v.1944 64 bit (AMD64)] on win32
Ctrl click to launch VS Code Native REPL
>>> from pathlib import Path
>>> Path("xyz.abc").unlink()
    Path("xyz.abc").unlink()
    ~~~~~~~~~~~~~~~~~~~~~~^^
  File "C:\Users\mkundu\AppData\Local\Programs\Python\Python314\Lib\pathlib\__init__.py", line 1042, in unlink
    os.unlink(self)
    ~~~~~~~~~^^^^^^
FileNotFoundError: [WinError 2] The system cannot find the file specified: 'xyz.abc'
>>> Path("xyz.abc").exists()
False
```

mayankansys: @mkundu1  @seanpearsonuk  file_path.unlink(missing_ok=True) 

This would be the correct approach ?

mkundu1: Yes, that's better

mayankansys: Thanks and Noted.

src/ansys/fluent/core/launcher/fluent_container.py
```diff
@@ -610,9 +610,5 @@ def start_fluent_container(
         logger.error(f"Exception caught - {type(ex).__name__}: {ex}")
         raise LaunchFluentError(launch_string) from ex
     finally:
-        if (
-            remove_server_info_file
-            and cleanup_on_exit
-            and host_server_info_file.exists()
-        ):
-            host_server_info_file.unlink()
+        if remove_server_info_file and cleanup_on_exit:
```
seanpearsonuk: This logic is wrong. These two variables (one is from the config `dict`) are saying exactly the same thing (one needs to be deprecated). This logic overlooks that fact.

**IF** either is `False` **AND** either is `True` -> `raise` (cannot proceed on contradictory input)
**ELSE IF** at least one is `True` -> `unlink()`

Make sure to fail early.

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -379,6 +379,4 @@ def __call__(
             raise LaunchFluentError(self._launch_cmd) from ex
         finally:
             if self.argvals.get("cleanup_on_exit", True):
```
seanpearsonuk, line 381: Just a comment on what is presumably the pre-existing status here. I find this odd. If `cleanup_on_exit` is an optional argument, it's intuitive to think that if I don't pass that argument then nothing gets cleaned up, but it's actually the other way around. As such, it feels like an optional argument like `preserve_info_file: bool` (defaulting to False) makes more sense. Then if I want to preserve the file, I explicitly request it. We have it the other way round where I might ignore `cleanup_on_exit` because I don't want to do that but then it just happens anyway.

tests/test_launcher.py
```diff
@@ -928,205 +801,183 @@ def test_idle_timeout(monkeypatch):
     )
```
seanpearsonuk, line 808: Do we know the history of this class? Why are PyFluent tests checking the mechanics of Python's documented file handling functionality?

We should just delete this - and anything else with a similar smell.

mayankansys: Noted.

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -173,9 +183,9 @@ def __init__(
             Additional command-line arguments for Fluent, formatted as they would be on the command line.
         env : dict[str, str], optional
             A mapping for modifying environment variables in Fluent. Defaults to ``None``.
-        cleanup_on_exit : bool, optional
```
mkundu1, line 176: How shall I call the launcher now if I don't want to shut down the Fluent process on `session.exit()`?

mayankansys: @mainakk If I understood properly your point then we can preserve the server info file and avoid cleanup on session.exit() , call the launcher with 

`session = pyfluent.launch_fluent(preserve_info_file = True)`
`session.exit()`

Kindly let me know, if I misunderstood your question or any further query or any change needed.

seanpearsonuk: @mkundu1 is intended above.

seanpearsonuk: @mayankansys is responding to [my comment above](https://github.com/ansys/pyfluent/pull/5242#discussion_r3613261140) but that comment does not request any changes. If we agree to follow up on that idea, we would create a separate issue.

prmukherj (approved):

prmukherj:
LGTM.

src/ansys/fluent/core/launcher/container_launcher.py
```diff
@@ -181,9 +181,9 @@ def __init__(
         dry_run : bool, optional
             If True, does not launch Fluent but prints configuration information instead. If dry running a
             container start, this method will return the configured ``container_dict``. Defaults to False.
-        cleanup_on_exit : bool
-            Determines whether to shut down the connected Fluent session upon exit or when calling
-            the session's `exit()` method. Defaults to True.
+        cleanup_on_exit : bool, optional
+            If True (default), the server-info file will be deleted when the session exits.
+            If False, the server-info file will be preserved for debugging. Defaults to True.
```
seanpearsonuk: @mayankansys I didn't see this radical change here previously. Is this correct? Is this the result of internal discussions with @mkundu1, @prmukherj et al during my absence?

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -174,8 +174,8 @@ def __init__(
         env : dict[str, str], optional
             A mapping for modifying environment variables in Fluent. Defaults to ``None``.
         cleanup_on_exit : bool, optional
-            Determines whether to shut down the connected Fluent session when exiting PyFluent or calling
-            the session's `exit()` method. Defaults to True.
+            If True (default), the server-info file will be deleted when the session exits.
+            If False, the server-info file will be preserved for debugging. Defaults to True.
```
seanpearsonuk: @mayankansys The same question applies to all cases ("I didn't see this radical change here previously. Is this correct? Is this the result of internal discussions ... during my absence?").

@prmukherj Are you aware of these changes since your approval?

@mkundu1

src/ansys/fluent/core/launcher/container_launcher.py
prmukherj: Right @seanpearsonuk, noticed it now. We should preserve the info about `Fluent` session exit as well. I think along with the existing docstring we can just append the extra info. I will revert my approval and wait for this change.

seanpearsonuk: @mayankansys 
The prior information about optionally avoiding Fluent session termination was **significant** and that is now lost. The new information implies that the session is terminated unconditionally and that the file is salvaged only for inspection purposes. Retaining the file allows reconnection to the still-running session.

Am additional point: "Defaults to True" is appended awkwardly to the end of the paragraph that is supposed to deal specifically with one topic: what happens in the `False` case.

mkundu1:
Copying my comment from the other PR:

As there is already a way to persist the server-info file for container case, via the `remove_server_info_file` key of `container_dict`, we do not need any immediate change in `fluent_container.py`.

mkundu1 (approved):

seanpearsonuk (approved):
The PR contains minimal changes to address the issue at hand along with comprehensive tests. This is the ideal outcome that we should all aim for.

PR #5248, opened 2026-07-09
fix(types): more specific returns for app related enums

Gobot1234:
I just noticed these weren't typed thought it would be useful to be more specific for typing+testing purposes

### DIFF
--- a/src/ansys/fluent/core/services/application_runtime.py
+++ b/src/ansys/fluent/core/services/application_runtime.py
@@ -47,9 +47,9 @@ Class hierarchy
     Scheme-based fallback used for Fluent versions before 25R2.
 """
 
-from enum import Enum
 import os
 import platform
+from typing import TYPE_CHECKING
 
 from ansys.fluent.core._types import PathType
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/services/application_runtime.py
```diff
@@ -102,19 +112,19 @@ def get_solver_process_info(self) -> ProcessInfo:
             working_directory=working_directory,
         )
 
-    def get_app_mode(self) -> Enum:
+    def get_app_mode(self) -> "FluentMode":
         """Get app mode."""
         from ansys.fluent.core import FluentMode
```
mkundu1, line 117: @prmukherj Can we move these to global imports? I think we can import from `ansys.fluent.core.launcher.launch_options` to avoid any dependency issues coming from the top-level imports.

prmukherj: yes, will do, Thanks. @Gobot1234, I will take care of this file.

PR #5252, opened 2026-07-10
feat: expr builder

seanpearsonuk:
## Pull request overview

This PR introduces a new Fluent expression builder subsystem (AST + registry + parser + discovery + builder facade) and integrates it with the solver settings API so expression objects can be assigned directly to `setup.named_expressions[...].definition`. It also adds a comprehensive set of unit/grammar/discovery/integration tests plus a minimal Fluent smoke test.

**Changes:**
- Added `ansys.fluent.core.expressions` package providing an AST, function registry, builder facade, discovery layer, and a recursive-descent parser with round-tripping.
- Extended `Textual.set_state()` to accept “expression objects” (duck-typed via `__fluent_expr__`) at whitelisted settings paths.
- Added tests covering parsing, registry coverage, discovery via settings/field-info fakes, flobject integration, and a live Fluent smoke test.


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -103,6 +103,64 @@
...
+    """If ``state`` looks like an expression object, return its rendered string.
+
+    Duck-typed on ``__fluent_expr__``.  Returns ``None`` when ``state`` is not
+    an expression object, so callers can fall through to their existing logic.
+    """
+    render = getattr(state, "__fluent_expr__", None)
+    if callable(render):
+        rendered = render()
+        if isinstance(rendered, str):
+            return rendered
```
hpohekar, line 160: I think there's no sanitization of the rendered string.

Can we update it as follows?

```
rendered = render()
if isinstance(rendered, str):
    # Validate through the parser
    from ansys.fluent.core.expressions._parser import parse as _parse
    _parse(rendered)  # Will raise ExpressionBuildError if malformed
    return rendered
```

Gobot1234:
Related https://github.com/ansys/pyfluent/discussions/4516

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -831,6 +893,14 @@ def set_state(self, state: StateT | None = None, **kwargs):
         TypeError
             If state is not a string.
         """
+        # Expression-object support: convert an Expr-like input to its
+        # Fluent-side string form when this leaf is a recognised
+        # expression-definition slot.
+        if _matches_expression_definition_path(self):
+            rendered = _try_render_expression(state)
+            if rendered is not None:
+                return self.base_set_state(state=rendered, **kwargs)
```
hpohekar: I think `chain = _python_name_chain(obj)` walks up to root every call in `_matches_expression_definition_path`.

Can we invert the order? 

check for expression object first, then only walk the path if it looks like an expression object.

```
def set_state(self, state=None, **kwargs):
    render = getattr(state, "__fluent_expr__", None)
    if callable(render) and _matches_expression_definition_path(self):
        rendered = render()
        if isinstance(rendered, str):
            return self.base_set_state(state=rendered, **kwargs)
    ...
```

hpohekar (approved):

src/ansys/fluent/core/solver/flobject.py
seanpearsonuk: @hpohekar 

Good question. 

The rendered string comes from the AST's `__fluent_expr__`, whose nodes validate at construction. Parsing here would repeat work the builder already did, and make flobject depend on `expressions._parser`, breaking the deliberate one-way duck-typed boundary.

I already have a CFX version of this built on the same core code. One consequence of that is that the `__fluent_expr__` name automatically emerges from the shared AST layer via the CFX builder. I need to rework that later: the AST layer needs to be generic while we want to expose a strongly typed interface to ensure that the expression family fits the context.

hpohekar: Thanks for the details.

src/ansys/fluent/core/solver/flobject.py
seanpearsonuk: @hpohekar Thanks, that's reasonable. An update has been pushed.

prmukherj (approved):

src/ansys/fluent/core/solver/flobject.py
hpohekar: Yes `hasattr(state, "__fluent_expr__")` is straight forward.

tests/expressions/test_parser.py
```diff
@@ -0,0 +1,255 @@
...
+        "Dot(MakeVec(1,0,0),MakeVec(0,1,0))",
+        "sqrt(VelocityMagnitude)",
+        "pow(VelocityMagnitude,2)",
+        "IF((VelocityMagnitude > 10),VelocityMagnitude,0)",
+        "(AreaAve(AbsolutePressure,['inlet1']) - AreaAve(AbsolutePressure,['outlet']))",
+        "((2 * VelocityMagnitude) + 1)",
+        "(-VelocityMagnitude)",
+    ],
+)
+def test_round_trip_canonical(canonical):
```
prmukherj, line 193: 👍

prmukherj: We need to document the use cases as well right? This test gave a very good idea.
Thanks

PR #5256, opened 2026-07-13
chore: Unify schema attrs in flobject.py

prmukherj:
## Context

## Change Summary
What changes were made?

## Rationale
Why was this approach taken?

## Impact

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -2450,12 +2450,8 @@ def get_cls(name, info, parent=None, version=None, parent_taboo=None):
                 else:
                     dct["__doc__"] = f"'{pname.strip('_')}' child."
 
-        include_child_named_objects = info.get(
-            "include-child-named-objects?", False
-        ) or info.get("include_child_named_objects", False)
-        user_creatable = info.get("user-creatable?", False) or info.get(
-            "user_creatable", False
-        )
+        include_child_named_objects = info.get("include_child_named_objects", False)
+        user_creatable = info.get("user_creatable", False)
 
         if version == "222":
             user_creatable = True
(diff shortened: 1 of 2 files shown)


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -2625,7 +2619,7 @@ def _process_cls_names(info_dict, names, write_doc=False):
                 )
             cls._allowed_values = allowed_values
 
-        has_migration_adapter = info.get("has-migration-adapter?", False)
+        has_migration_adapter = info.get("has_migration_adapter", False)
```
prmukherj: Only this we need to add. I am not sure how it was getting propagated from settings side as I could not find this attribute in the schema (proto of in the server code.)
@mkundu1

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -2615,7 +2609,7 @@ def _process_cls_names(info_dict, names, write_doc=False):
                     k,
                 )
 
-        allowed_values = info.get("allowed-values") or info.get("allowed_values", [])
+        allowed_values = info.get("allowed_values", [])
```
```suggestion
        allowed_values = info.get("allowed-values", [])
```

src/ansys/fluent/core/solver/flobject.py
```suggestion
        has_migration_adapter = info.get("has-migration-adapter", False)
```

mkundu1: @prmukherj Please see https://github.com/ansys-internal/ansys-api-fluent/pull/180/changes how the settings schema can be restructured to avoid the dynamic attrs and the scheme names like `has-migration-adapter?` etc. 

Please also note: https://github.com/ansys-internal/ansys-api-fluent/issues/121

mkundu1: `has-migration-adapter?`

PR #5274, opened 2026-07-20
feat: FUTURE service layer

mayankansys:
## Context

## Change Summary
What changes were made?

## Rationale
Why was this approach taken?

## Impact


src/ansys/fluent/core/services/__init__.py
```diff
@@ -131,6 +138,16 @@ def settings(self):
...
 
+    @cached_property
+    def rest_settings(self):
+        """REST-based solver settings service.
+
+        Returns ``None`` if REST client is not available.
+        """
+        if self._rest_client is not None:
+            return RestSettings(self._rest_client)
+        return None
```
mayankansys: Never return the None, we have to handle it via error or warning.

mayankansys:
Closing the PR as I have created a new PR.

PR #5286, opened 2026-07-28
feat: Use 'download_file'  method directly from ansys-tools-common.

### LINKED ISSUE #5031: add retries to download file
### 📝 Description of the feature

currently when using `examples.download_file` there is no option to retry the download.
Request to make this available

### 💡 Steps for implementing the feature

_No response_

### 🔗 Useful links and references

_No response_

prmukherj:
## Context
A copy of the `download_file`'s implementation was in the PyFluent repo that can be got rid of. The PyFluent method just redirects to the method in ansys-tools-common directly.

It exposes some arguments like:
`force`, `timeout`, `max_retries`

## Change Summary
The PyFluent method just redirects to the method in ansys-tools-common directly.
No changes to the import path, etc., since it just changes internal implementation without affecting the users.

## Rationale
So that users can use the full capability of the method in ansys-tools-common.

## Impact
Users to use some documented methods from the download_manager's version directly:

force, timeout, max_retries

They have been documented in the PyFluent as well.


mkundu1 (approved):

src/ansys/fluent/core/examples/downloads.py
```diff
@@ -182,10 +112,23 @@ def download_file(
...
+            file_name,
+            directory,
+            destination=save_path,
+            force=force,
+            timeout=timeout,
+            max_retries=max_retries,
+        )
+    except Exception as ex:
+        raise RemoteFileNotFoundError(
+            f"{file_name} does not exist in the {directory} directory of the Ansys example data repository."
```
mkundu1: The final message will be malformed as it will be wrapped within `f"{url} does not exist."`

src/ansys/fluent/core/examples/downloads.py
```diff
@@ -182,10 +112,23 @@ def download_file(
...
+        local_path = download_manager.download_file(
+            file_name,
+            directory,
+            destination=save_path,
+            force=force,
+            timeout=timeout,
+            max_retries=max_retries,
+        )
+    except Exception as ex:
+        raise RemoteFileNotFoundError(
```
seanpearsonuk: - shouldn't arbitrarily narrow the type
- should trust the library to raise the right exception directly without interference
- something is wrong if we have to create the message string
- perhaps we should be proposing changes  to the library instead of working around any deficiencies

prmukherj: @seanpearsonuk, the library already has it's own error handling logic as you mentioned in point 2. We have removed from the PyFluent code.

seanpearsonuk: Thanks for updating. The original version would make a useful interview snippet.

src/ansys/fluent/core/examples/downloads.py
```diff
@@ -143,11 +59,12 @@ def download_file(
         Assuming the Fluent inside the container has its working directory set to the path that was mounted from
         the host, and that the example files are being made available by the host through this same path,
         only the file name is required for Fluent to find and open the file.
-
-    Raises
-    ------
-    RemoteFileNotFoundError
-        If remote file does not exist.
```
seanpearsonuk, line 150: Yesterday we discussed this function and you indicated that the underlying `download_file()` raises an exception if the file is not found. Now, the exception documentation is completely stripped out even though we allow the underlying exception to escape!

If you look at the underlying implementation, you will see that it emits an undocumented `ValueError` as well as any number of exception types from lower layers. That lack of documentation (as well as incorrect exception type) is itself an issue that should be addressed.

### Outstanding issues here
- The function emits undocumented exceptions.
- The underlying exception type is too broad.
- We have broken our previous contract to raise `RemoteFileNotFoundError`. That would have been an interface change if allowed to stand.
### Solution
- Reinstate the above exception documentation.
- `try`/`except` and raise `RemoteFileNotFoundError` upon exception. Reuse the exact prior code to raise the exception in the same way.
### Additional step
- Create an issue in https://github.com/ansys/ansys-tools-common, citing two points
  - The exception is undocumented.
  - The exception type is inaccurate.
### Final point
- Given the current lack of exception documention and uncertainty of exception types from the underlying library, we can either:
  - Insert an additional layer now that satisfies the contract of a corrected underlying implementation, or
  - Not use the library until it is corrected.

@prmukherj @mkundu1

prmukherj: @seanpearsonuk, trying and excepting a `ValueError` will not work always for example when we have an incorrect directory name it does raise a `RuntimeError`. That's why I had to broaden the exception type. I think to properly achieve this task we need to get properly documented error types from the ansys-tools-common - download_manager.download_file. Please let me know if my thinking is correct.

Another issue I have found is that if the file name matches the cached file name then it just returns that file even if we try to download a file with same name but from a different directory. This behavior needs correctio as well.

Shall we raise these as issues in https://github.com/ansys/ansys-tools-common as well? Please let me know your thoughts. Thanks

cc. @mkundu1

seanpearsonuk: @prmukherj I had updated my comment while you were writing yours. See my final point. We should stick with what we have until the underlying library is corrected - the issues are quite severe on their side.

prmukherj: yes @seanpearsonuk. I just saw and I agree completely. We were trying to work around things and that will make our life more difficult. I will just raise these as issue in https://github.com/ansys/ansys-tools-common


Will convert this PR to draft for now.

Thank you.

prmukherj:
Improvement task created in https://github.com/ansys/ansys-tools-common/issues/365

PR #5287, opened 2026-07-28
feat: Strongly typed settings schema.

### LINKED ISSUE #5250: `flobject.get_cls()` is confusing
`get_cls()` references attribute names containing a mixture of hyphens and underscores. Can someone provide an explanation for this - that explanation should be added as code comments. E.g., `exposure_level_str = info.get("api_exposure_level")` versus `user_creatable = info.get("user-creatable?", False)`. Note that the code behind the API references `api-exposure-level` and not `api_exposure_level`.

Is the gRPC settings schema clear, discoverable and documented in detail? Is attribute naming uniform - i.e. is there a single naming convention?

prmukherj:
## Context
Earlier a lot of the attributes were categorized under optional attributes. They were not documented and lacked strong typing as well. That issue has been resolved.

## Change Summary
Move the attributes in the settings schema from optional to strongly typed ones.

## Impact
Users to query strongly typed attributes from the schema.

### DIFF
--- a/src/ansys/fluent/core/_grpc_services/settings_service.py
+++ b/src/ansys/fluent/core/_grpc_services/settings_service.py
@@ -206,11 +206,37 @@ class SettingsService(ServiceProtocol):
 
     def _extract_static_info(self, info) -> dict[str, Any]:
         ret = {}
-        ret["type"] = info.type
-        for key, value in sorted(info.attrs.items()):
-            ret[key] = self._get_state_from_value(value)
+
+        # 1. Basic Primitive Fields (Strings, Ints, Bools)
+        if info.type:
+            ret["type"] = info.type
+        if info.help:
+            ret["help"] = info.help
         if info.has_allowed_values:
             ret["has-allowed-values"] = info.has_allowed_values
+        if info.include_child_named_objects:
+            ret["include_child_named_objects"] = info.include_child_named_objects
+        if info.list_size:
+            ret["list_size"] = info.list_size
+
+        # user_creatable is a boolean; check directly as it exists in the schema
+        if info.user_creatable:
+            ret["user_creatable"] = info.user_creatable
+
+        if info.return_type:
+            ret["return_type"] = info.return_type
+        if info.deprecated_version:
+            ret["deprecated_version"] = info.deprecated_version
+        if info.api_exposure_level:
+            ret["api_exposure_level"] = info.api_exposure_level
+        if info.file_purpose:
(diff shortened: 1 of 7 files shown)


src/ansys/fluent/core/_grpc_services/settings_service.py
```diff
@@ -231,26 +257,24 @@ def _extract_static_info(self, info) -> dict[str, Any]:
...
-        except AttributeError:
-            pass
-
-        try:
-            if info.user_creatable:
-                ret["user_creatable"] = info.user_creatable
-        except AttributeError:
-            ret["user_creatable"] = True
+
+        # 4. Repeated StringsPair Fields (Aliases)
```
seanpearsonuk:
```suggestion
        # 4. Repeated StringPair Fields (Aliases)
```

seanpearsonuk:
- Re keeping the low-level string-based layer over the higher-level proto layer, is that temporary or a long-term intention?
- Note that the unit test code also tries to construct these `dict`s independently.
- Is it ok to simply skip all the falsy values in the response as we're doing here? `list_size` for instance.
- Are all the keys mapped correctly (hyphens versus underscores) for v0?

prmukherj:
> * Re keeping the low-level string-based layer over the higher-level proto layer, is that temporary or a long-term intention?
> * Note that the unit test code also tries to construct these `dict`s independently.
> * Is it ok to simply skip all the falsy values in the response as we're doing here? `list_size` for instance.
> * Are all the keys mapped correctly (hyphens versus underscores) for v0?

1. It is a temporary solution @seanpearsonuk.
2. yes, fixing them as well.
3. I need to investigate that, just migrating as of now.
4. That's the part I am currently fixing.

Actually this PR is yet to be ready. I need to fix a few compatibility issues for v0. I might merge the strongly typed settings and flobject clean up in separate PRs.

@seanpearsonuk

prmukherj:
> * Re keeping the low-level string-based layer over the higher-level proto layer, is that temporary or a long-term intention?

The low-level string-based layer is an intentionally scoped to the lifetime of v0 server support. The long-term design is exactly what v1 already has — no map, no translation, direct access to strongly-typed proto fields.

prmukherj:
> * Are all the keys mapped correctly (hyphens versus underscores) for v0?

There is a well constructed backwards compatibility map for it now.

mkundu1 (approved):

seanpearsonuk (approved):

PR #5290, opened 2026-07-29
docs: connect to fluent user guide [skip tests]

### LINKED ISSUE #5279: `connect_to_fluent` user guide a bit strange
https://fluent.docs.pyansys.com/version/stable/user_guide/session/launching_ansys_fluent.html#connect-to-an-existing-session

The guidance for connection shouldn't depend on launching via PyFluent, as is shown in the doc. Launching Fluent in the normal way is easier and more flexible than using PyFluent, so users may find this frustrating. In many cases, they may be using the connect functionality precisely to avoid having to use PyFluent to launch.

The guidance also shows users how to access the individual connection data from the launched session object. It would be more user-riendly if they were simply directed to specify the server info file path which is more encapsulated.

The example is very long while only a tiny proportion of it is central to the task.

The example invokes `solver.exit()` while `solver_connected` is still connected. The example doesn't specify whether or not that call has any side effects for `solver_connected`.

The example includes a call to `check_health()` from each `session`. If control is returned from launching or connecting without an exception raised, can it be assumed that the connection is healthy (do we check it before returning? That was the design at one time at least)?

mayankansys:
## Context
The `connect_to_fluent` user guide documentation had several usability issues:

- The connection guidance unnecessarily depended on launching Fluent via PyFluent, which was frustrating for users who preferred to launch Fluent normally
- Users were shown how to access individual connection data from the launched session object instead of being directed to the simpler server info file path approach
- The example was overly long with only a small portion relevant to the core connection task
- The example had unclear behavior around calling `solver.exit()` while `solver_connected` was still active
- It included unnecessary `check_health()` calls when the connection health is already validated before returning

## Change Summary
Updated the `connect_to_fluent` user guide documentation to provide clearer, more user-friendly guidance for connecting to existing Fluent sessions. The documentation now focuses on the more encapsulated approach using server info file paths and streamlines the examples to be more concise and relevant.

## Impact

- User guide documentation for connecting to existing Fluent sessions
- User onboarding experience for developers connecting PyFluent to existing Fluent instances
- Documentation examples are now more focused and easier to follow

### DIFF
--- a/doc/source/user_guide/session/launching_ansys_fluent.rst
+++ b/doc/source/user_guide/session/launching_ansys_fluent.rst
@@ -88,22 +88,12 @@ Use this method when:
 
 **Prerequisites**
 
-Before connecting, you must launch Fluent externally with the gRPC server enabled. You can do this in several ways:
-
-1. **Command line with server info file**:
-
-   - **Linux/WSL**: ``ansys_inc/v261/fluent/bin/fluent 3ddp -sifile=server.txt``
-   - **Windows**: ``ANSYS Inc\v261\fluent\ntbin\win64\fluent.exe 3ddp -sifile=server.txt``
-
-2. **Within Fluent GUI**: Execute ``File`` → ``Applications`` → ``Server`` → ``Start...`` in solution mode.
-
-3. **Via TUI command**: Execute ``server/start-server`` in solution mode.
+Before connecting, you must launch Fluent externally with the gRPC server enabled.
 
 After starting the gRPC server, Fluent writes connection information to the server info file. The file contains the IP address, port, and password needed for connection.
+The recommended approach is to connect using the server info file path, which encapsulates all connection information.
 
 **Example:**
-
-The recommended approach is to connect using the server info file path, which encapsulates all connection information:
 
 .. code-block:: python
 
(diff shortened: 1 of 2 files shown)


seanpearsonuk (approved):

prmukherj (approved):

doc/source/user_guide/session/launching_ansys_fluent.rst
```diff
@@ -85,26 +85,25 @@ Use this method when:
...
-
-   ip = solver.connection_properties.ip
-   password = solver.connection_properties.password
-   port = solver.connection_properties.port
-
-   # Connect to the session
-   solver_connected = pyfluent.Solver.from_connection(ip=ip, password=password, port=port)
-   print(solver_connected.health_check.check_health())
+   # Connect to the session using the server info file
+   solver = pyfluent.Solver.from_connection(server_info_file_name="server.txt")
```
prmukherj: A placeholder is better here?

`<server-info-file>`

mayankansys:
```suggestion
   solver = pyfluent.Solver.from_connection(server_info_file_name=<server-info-file>)
```

PR #4925, opened 2026-02-11
fix: type hints for meshing workflows

Gobot1234:
## Context
Fixes #4772 by adding intellisense for the meshing classes by making their proxying behaviour typed.

## Change Summary
Adds types and tests for the meshing workflows

## Impact
No user noticeable runtime changes only changed class names to stop interfering with type checking, added tests and a couple imports

NB I've intentionally picked the base to be my branch to make changes easier to review, should be merged onto main once the base is merged.

Issue #4772, opened 2025-12-29
Bug located in ...most command suggestions not appearing in VS-Code


rploconnor:
### 🔍 Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### 🐞 Description of the bug

I'm working in an ipynb file within VS-Code. Usually when creating a command I can tab through a popup menu of suggestions to complete the command. The menu appears but it only contains a few options.

### 📝 Steps to reproduce

`import ansys.fluent.core as pyfluent
meshing = pyfluent.launch_fluent(mode="meshing",ui_mode="gui")
watertight = meshing.watertight()
watertight.[I press tab here and see the following]
`
Menu options:
.children
.deletetasks
.first_child
.last_child
.load_state
.save_workflow
.task_names
.tasks

### 💻 Which operating system are you using?

Windows

### 📀 Which ANSYS version are you using?

26R1

### 🐍 Which Python version are you using?

3.13

### 📦 Installed packages

```shell
ansys-api-fluent==0.3.36
ansys-api-platform-instancemanagement==1.1.3
ansys-api-tools-filetransfer==0.1.2
-e git+https://github.com/ansys/pyfluent.git@673466e10089302ed60fc5f481e651b1e523d9d6#egg=ansys_fluent_core
ansys-platform-instancemanagement==1.1.2
ansys-tools-common==0.4.0
ansys-tools-filetransfer==0.2.1
ansys-units==0.9.1
appdirs==1.4.4
asttokens==3.0.1
certifi==2025.11.12
charset-normalizer==3.4.4
click==8.3.1
colorama==0.4.6
comm==0.2.3
debugpy==1.8.19
decorator==5.2.1
defusedxml==0.7.1
Deprecated==1.3.1
docker==7.1.0
executing==2.2.1
googleapis-common-protos==1.72.0
grpcio==1.76.0
grpcio-health-checking==1.62.3
grpcio-status==1.62.3
idna==3.11
importlib_metadata==8.7.1
ipykernel==7.1.0
ipython==9.8.0
ipython_pygments_lexers==1.1.1
jedi==0.19.2
joblib==1.5.3
jupyter_client==8.7.0
jupyter_core==5.9.1
matplotlib-inline==0.2.1
nest-asyncio==1.6.0
nltk==3.9.2
numpy==2.4.0
packaging==25.0
pandas==2.3.3
parso==0.8.5
platformdirs==4.5.1
prompt_toolkit==3.0.52
[...]

PR #5293, opened 2026-07-30
feat: Add dir for Workflow objects

Gobot1234:
## Context
Currently trying to use autocomplete in python console inside fluent won't show the `task_names()`

## Change Summary
Added `task_names()` to the `dir()`.

## Impact
Improves the usability of the new meshing workflow inside of fluent for engineers

### DIFF
--- a/doc/changelog.d/5293.added.md
+++ b/doc/changelog.d/5293.added.md
@@ -1,0 +1,1 @@
+Add dir for Workflow objects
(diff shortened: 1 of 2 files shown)


prmukherj:
#4772, might be actually fixed with this.

Gobot1234:
I'd be doubtful that it would. Most LSPs (pyconsole excluded) don't call dir() on the objects, I believe the correct solution is #4925 still. I might be able to fix that up once I get back to my computer if no one from the team is able to fix it.

PR #5298, opened 2026-07-31
feat: REST settings services

### LINKED ISSUE #4959: Demonstrate that PyFluent can run transparently over REST or gRPC for solver settings

mayankansys:
## Context
PyFluent previously supported communication with Fluent solvers exclusively through gRPC-based transport, which requires additional infrastructure and dependencies. There was a need to demonstrate that PyFluent can run seamlessly over both REST and gRPC, allowing users to choose their preferred transport mechanism.

## Change Summary
This PR introduces comprehensive REST/HTTP transport support to PyFluent with the following key additions:

- **New RestSettings class** — A settings service wrapper that implements the AbstractSettings interface by delegating all operations to a FluentRestClient instance. Includes schema normalization to convert REST's native hyphenated Scheme keys (e.g., object-type, user-creatable?) to underscore format matching **gRPC** conventions.

- **New HttpSolver class** — A standalone, lightweight solver session that communicates exclusively over REST, completely independent of gRPC infrastructure. Settings classes are built at runtime from get_static_info() with no need for pre-generated modules.

- **Factory method**  — Convenience classmethod to create an HttpSolver instance from a URL and authentication token, making REST-based sessions accessible through the standard Solver interface.

- **Comprehensive tests** — Unit and integration tests covering REST settings wildcard detection, static-info key normalization, nested schema recursion, and end-to-end REST-based workflows.

## Rationale

- **Transport flexibility**: REST/HTTP is stateless and works in cloud, containerized, and restricted network environment. It gives the user access to choose the REST/gRPC as required.
- **Runtime schema building**: The HttpSolver derives settings structure dynamically from server responses via get_static_info(), eliminating the need for hand-maintained pre-generated settings modules.
- **API consistency**: The RestSettings class implements AbstractSettings, ensuring compatibility with existing PyFluent abstractions designed for any settings transport backend.
- **Minimal dependencies**: HttpSolver has no dependency on gRPC libraries or session management infrastructure, making it suitable for lightweight client applications.
- **Server compatibility**: Available from Fluent 27.1 onward; client-side schema corrections handle known gaps in REST server responses.

## Impact

- **User workflows**: Users can now connect via REST using Solver.from_http(url, token), in addition to the existing gRPC-based Solver() constructors. Both transports coexist without interference.
- **Transparent operation**: Existing code using the settings API (e.g., solver.settings.setup.models.energy.enabled()) works identically over REST and gRPC.
- **Testing**: New REST fixtures enable testing on live Fluent 27.1+ REST servers, complementing existing gRPC-based test infrastructure (controlled via FLUENT_REST_URL and FLUENT_REST_TOKEN environment variables).
- **Schema normalization**: Automatic conversion of REST's hyphenated schema keys to underscore format hides transport-specific details from downstream code.


src/ansys/fluent/core/session_http_solver.py
```diff
@@ -0,0 +1,193 @@
...
+                flproxy=self._rest_settings_service,
+                version=self._version,
+            )
+        return self._settings
+
+    def is_active(self) -> bool:
+        """Whether the REST connection is still usable."""
+        return self._rest_client is not None
+
+    def exit(self) -> None:
```
prmukherj, line 91: We can later take this exit into finalizer, like with the session exit. I think inheritance will be a good option here going forwards as well to remove these duplicate code. First we need the integration to work.

src/ansys/fluent/core/session_http_solver.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.exit()
+        return False
+
+    # ------------------------------------------------------------------
+    # REST CRUD Operations for Named Objects
+    # ------------------------------------------------------------------
+
+    def create_named_object(
```
prmukherj: Why are these methods exposed at the solver level? Any particular requirement?

src/ansys/fluent/core/session_solver.py
```diff
@@ -128,6 +131,47 @@ def __init__(
             fluent_connection, scheme_eval, launcher_args=launcher_args
         )
 
+    @classmethod
+    def from_http(
```
prmukherj, line 177: This should ideally live in `src\ansys\fluent\core\session_utilities.py`

tests/test_settings_api.py
```diff
@@ -46,6 +46,238 @@
 from ansys.fluent.core.utils.execution import timeout_loop
 from ansys.fluent.core.utils.fluent_version import FluentVersion
 
+# ============================================================================
+# Helpers for transport-parametrized tests (gRPC vs REST)
+# ============================================================================
+
+
+def _is_rest(session_fixture_name: str) -> bool:
```
mayankansys: Put it in the `conftest.py`

tests/test_settings_api.py
```diff
@@ -381,10 +613,26 @@ def test_deprecated_settings_with_custom_aliases(new_solver_session):
...
+        ),
+    ],
+)
+def test_deprecated_settings_with_settings_api_aliases(session_fixture_name, request):
+    """Test deprecated settings API aliases."""
+    solver = request.getfixturevalue(session_fixture_name)
+    adapter = SettingsTestAdapterFactory.create(session_fixture_name, solver)
+
+    # Create iso_clip object
+    adapter.create_named_object("results/surfaces/iso_clip", "clip-1")
```
prmukherj: This is the reason why those methods are exposed at high level that I have commented above. These extra hacks are unnecessary and if the original tests don't work without these hacks we need to look deeper.

tests/test_settings_api.py
```diff
@@ -381,10 +613,26 @@ def test_deprecated_settings_with_custom_aliases(new_solver_session):
...
+        pytest.param(
+            "mixing_elbow_case_data_session_rest",
+            marks=[pytest.mark.rest_server, pytest.mark.fluent_version(">=27.1")],
+        ),
+    ],
+)
+def test_deprecated_settings_with_settings_api_aliases(session_fixture_name, request):
+    """Test deprecated settings API aliases."""
+    solver = request.getfixturevalue(session_fixture_name)
+    adapter = SettingsTestAdapterFactory.create(session_fixture_name, solver)
```
mayankansys: adapter is not required.  In the whole file.

prmukherj:
@mayankansys, in general you could just re-use our existing tests by just parameterizing them. You can avoid any kind of duplication via the adapters, etc. The tests should be the existing ones just by parameterizing them so as to avoid duplication. A few of the extra tests that you have written like over-writing named objects, etc. can be kept but similarly parameterize them for grpc as well.

Thank you.

src/ansys/fluent/core/services/rest_settings.py
```diff
@@ -0,0 +1,256 @@
...
+                for child_name, child in container.items()
+            }
+    object_type = corrected.get("object_type")
+    if isinstance(object_type, dict):
+        corrected["object_type"] = _apply_schema_corrections(object_type, name=name)
+
+    return corrected
+
+
+class RestSettings(AbstractSettings):
```
prmukherj: Will it be better to inherit from `BaseSettings`? It will reduce a huge amount of duplication.

tests/test_settings_api.py
prmukherj: `solver.settings.results.surfaces.iso_clip["clip-1"] = {}`, this should work after the integration.

src/ansys/fluent/core/services/rest_settings.py
seanpearsonuk: This is correct.

Are the wildcard methods implemented in `Settings` out of scope with the REST backend? I think we only added those for gRPC.

mayankansys: Yeah. That's why handled I handled it client side it is needed.

src/ansys/fluent/core/session_http_solver.py
mayankansys: Noted. Let me have a look in that.

src/ansys/fluent/core/session_http_solver.py
mayankansys: No, while testing I have done it.

src/ansys/fluent/core/session_solver.py
mayankansys: Sure. Let me put it in the sesson_utilities.py

tests/test_settings_api.py
```diff
@@ -46,6 +46,238 @@
 from ansys.fluent.core.utils.execution import timeout_loop
 from ansys.fluent.core.utils.fluent_version import FluentVersion
 
+# ============================================================================
+# Helpers for transport-parametrized tests (gRPC vs REST)
+# ============================================================================
+
+
```
seanpearsonuk: The settings API can be tested through either REST or gRPC interchangeably. Therefore, given any such API test, one can pass in any solver session object (or its `settings` attribute).

If you need all the following complications, please explain. Thanks.

tests/test_settings_api.py
```diff
@@ -419,15 +460,31 @@ def test_deprecated_settings_with_settings_api_aliases(mixing_elbow_case_data_se
             "bground_color": "#CCD3E2",
             "title_elements": "Variable and Object Name",
         },
-        "range_option": {
-            "option": "auto-range-off",
-            "auto_range_off": {
+    }
+    if isinstance(solver, HttpSolver):
```
prmukherj: use same as existing. Don't update paths as part of this PR.

tests/conftest.py
```diff
@@ -365,11 +365,46 @@ def mixing_elbow_watertight_pure_meshing_session(
...
+
+        solver = Solver.from_http(url=rest_url, token=rest_token)
+    else:
+        solver = create_session()
     yield solver
-    solver.exit()
+    # HttpSolver (REST transport) has no .exit()/close mechanism yet - guard
+    # so REST-parametrized tests don't fail during fixture teardown.
+    if hasattr(solver, "exit"):
+        solver.exit()
```
prmukherj: We need to incorporate the exit as part of this PR only, but we can do it as a last change after finalizing this PR with other changes first.

tests/conftest.py
```diff
@@ -365,11 +365,46 @@ def mixing_elbow_watertight_pure_meshing_session(
     return meshing
 
 
-@pytest.fixture
-def new_solver_session():
-    solver = create_session()
+@pytest.fixture(
```
prmukherj, line 395: 👍

tests/conftest.py
```diff
@@ -365,11 +365,46 @@ def mixing_elbow_watertight_pure_meshing_session(
...
+    params=[
+        "grpc",
+        pytest.param(
+            "rest",
+            marks=[pytest.mark.rest_server, pytest.mark.fluent_version(">=27.1")],
+        ),
+    ]
+)
+def new_solver_session(request):
```
prmukherj: Create a new one named "solver_session_grpc_rest", and use that for all the tests that needs to use the parameterized fixture testing both grpc and rest server. Keep the "new_solver_session" as it is.

tests/test_session.py
```diff
@@ -52,6 +52,7 @@
 from ansys.fluent.core.launcher.error_handler import LaunchFluentError
 from ansys.fluent.core.pyfluent_warnings import PyFluentDeprecationWarning
 from ansys.fluent.core.session import BaseSession
+from ansys.fluent.core.session_http_solver import HttpSolver
```
prmukherj: This file will not require any changes after the comment on conftest is addressed.

prmukherj: https://github.com/ansys/pyfluent/pull/5298/changes#r3840562166

tests/test_settings_api.py
```diff
@@ -438,14 +495,22 @@ def test_deprecated_settings_with_settings_api_aliases(mixing_elbow_case_data_se
...
+    solver.settings.solution.report_definitions.surface["surface-1"] = {
+        "surface_names": ["cold-inlet"]
+    }
+    try:
+        ret = solver.settings.solution.report_definitions.compute(
+            report_defs=["surface-1"]
+        )
+    except Exception:
+        if isinstance(solver, HttpSolver):
+            pytest.xfail("REST server bug: string-list command args fail with HTTP 500")
```
seanpearsonuk: @mayankansys Does this refer to a known server-side issue/limitation?

@mkundu1 @prmukherj

tests/test_session.py
```diff
@@ -1037,6 +1042,16 @@ def test_error_raised_for_beta_feature_access_for_older_versions(
...
+
+def test_error_raised_for_beta_feature_access_for_older_versions_solver(
+    new_solver_session,
+):
+    solver = new_solver_session
+    if isinstance(solver, HttpSolver):
+        pytest.skip(
+            "enable_beta_features()/get_fluent_version() are not applicable to REST transport"
+        )
+
```
seanpearsonuk: @mayankansys 

I prefer to simply define what is to be tested fully _from the top_ without any internal conditional skipping. That provides a clean, well defined view of what is and is not covered.

This simple example gets the basic idea across by defining what is to be run from the top and contains no conditionals. It lets the code decide what happens.

```
# define a test function that only works with gRPC backend:
def do_test_some_solver_features(solver): # ...

# define the test only for gRPC:
def test_some_solver_features(grpc_solver_session):
    do_test_some_solver_features(grpc_solver_session)

# that inflates the test code so we can just do it in one go:
def test_some_solver_features(grpc_solver_session): # ...

# @pytest.mark is also possible above but unnecessary

# define a test function that works with either backend:
def do_test_some_other_solver_features(solver): # ...

# define both tests:
# gRPC:
def test_some_other_solver_features(grpc_solver_session):
    do_test_some_other_solver_features(grpc_solver_session)

# http:
def test_some_other_solver_features(http_solver_session):
    do_test_some_other_solver_features(http_solver_session)

# again, that inflates the test code so we can make use of parametrised fixture selection:
@pytest_fixture(params=...)
def solver_session(...)
    return { ...

# run the test with either backend, gRPC or http:
def test_some_other_solver_features(solver_session): # ...
```
@mkundu1 @prmukherj

prmukherj: Thanks @seanpearsonuk, This is exactly what we should so.

tests/test_settings_api.py
mayankansys: No it's not a server-side issue. I handled it wrong.

src/ansys/fluent/core/launcher/fluent_container.py
```diff
@@ -223,41 +223,67 @@ def _setup_volumes(container_dict, mount_source, mount_target, certificates_fold
 
 
 def _resolve_port_mapping(port, container_dict):
```
mayankansys: Fluent_container can handle it without any major change.

src/ansys/fluent/core/rest/client.py
```diff
@@ -58,16 +58,21 @@ class FluentRestClient:
     strategy : RequestStrategy
         Responsible for executing HTTP requests (real or fake).
     component : str, optional
-        DataModel component name.  Defaults to ``"fluent_1"`` (solver).
-        Use ``"fluent_meshing_1"`` for a meshing session.
+        DataModel component name. If ``None``, auto-resolves based on
+        the ``rest_api_component`` configuration (``"fluent_1"`` for
+        standalone, ``"solver"`` for container mode).Defaults to ``None``.
```
prmukherj, line 63: 👍

src/ansys/fluent/core/rest/client.py
prmukherj: Is DetaModel term required? @seanpearsonuk

src/ansys/fluent/core/rest/client.py
```diff
@@ -304,9 +312,9 @@ def delete_all_child_objects(self, path: str, obj_type: str) -> None:
     # Commands / queries
     # ------------------------------------------------------------------
 
-    def _execute(self, path: str, name: str, force: bool = False, **kwds) -> Any:
+    def _execute(self, path: str, command: str, force: bool = False, **kwds) -> Any:
```
prmukherj, line 315: 👍

src/ansys/fluent/core/services/rest_settings.py
```diff
@@ -0,0 +1,180 @@
...
+# SOFTWARE.
+
+"""REST settings service wrapper."""
+
+from typing import Any
+
+from ansys.fluent.core.rest.client import FluentRestClient
+from ansys.fluent.core.services.settings import BaseSettings, _trace
+
+_REST_STATIC_INFO_KEY_MAP: dict[str, str] = {
```
prmukherj, line 33: Please comment that we are getting the data from server in this format but our flobject expects the snake case for parsing.

src/ansys/fluent/core/services/rest_settings.py
```diff
@@ -0,0 +1,180 @@
...
+
+    @_trace
+    def is_wildcard(self, input: str | None = None) -> bool:
+        """Check whether a name contains a wildcard pattern.
+
+        ``AbstractSettings`` requires this, but the REST API exposes no
+        equivalent of the gRPC ``Settings.IsWildcard`` endpoint, so the
+        fnmatch metacharacters are matched client-side instead.
+        """
+        if input is None:
```
prmukherj, line 126: It is best to raise NotImplementedError here? @seanpearsonuk, @mkundu1

src/ansys/fluent/core/services/rest_settings.py
```diff
@@ -0,0 +1,180 @@
...
+
+def _unwrap_result(response: Any) -> Any:
+    """Extract the ``"result"`` value from a command/query response envelope.
+
+    The REST API returns ``{"result": <value>, "output": <text>}`` for
+    command/query execution. Responses without a ``"result"`` key are
+    returned unchanged (defensive fallback).
+    """
+    if isinstance(response, dict) and "result" in response:
+        return response["result"]
```
prmukherj, line 182: 👍 

This is correct behavior. The fallback might not be needed but fine for now I think.

tests/conftest.py
```diff
@@ -289,14 +294,17 @@ def exhaust_system_geometry_filename():
...
         file_transfer_service = ContainerFileTransferStrategy()
-        container_dict = {"mount_source": file_transfer_service.mount_source}
+        container_dict.setdefault("mount_source", file_transfer_service.mount_source)
         return pyfluent.launch_fluent(
             container_dict=container_dict,
             file_transfer_service=file_transfer_service,
             **kwargs,
         )
+    elif container_dict:
+        return pyfluent.launch_fluent(container_dict=container_dict, **kwargs)
```
prmukherj, line 307: 👍

prmukherj:
@mayankansys, This PR overall looks good to me now. 

One thing would be good to confirm with the team is if we want an env var or skip marker to skip the REST tests as I expect it to increase the time to run the CI as we keep on appending more tests to this,

@mkundu1, @seanpearsonuk

src/ansys/fluent/core/services/rest_settings.py
mayankansys: Done.

src/ansys/fluent/core/rest/client.py
mayankansys: Thanks

src/ansys/fluent/core/rest/client.py
seanpearsonuk: @mayankansys is removing these.

src/ansys/fluent/core/services/rest_settings.py
seanpearsonuk: Wirth this one, it necessary to use global reasoning - it cannot be judeged on local reasoning. If we raise then crucial parts of the API won't function. @mayankansys's code implements almost the same as what's in Fluent's Scheme version. We have just added the `|` character.

mkundu1 (approved):

prmukherj (approved):

seanpearsonuk (approved):

PR #5307, opened 2026-08-03
feat: api performance optimized

### LINKED ISSUE #4924: Consider avoiding eager imports where possible
> It seems that `launch_options` is intended to be a lightweight module, which is potentially contradicted by some of the imports here (not only the session class imports), and that might lead to concrete performance issues (for instance, at start-up). Should this be a concern? Perhaps it would be worthwhile looking into the preferred way to avoid eager imports if that's something we want/need to do. We could investigate that as a part of later work conducted by @mayankansys. 

 _Originally posted by @seanpearsonuk in [#4761](https://github.com/ansys/pyfluent/pull/4761/changes/ab34346aaf3824c48b11eb30d935d87a872ae6b1..c4f07326b25e443a20e10c7b9157112c1eeb9131#r2792274986)_

mayankansys:
## Context

## Change Summary
What changes were made?

## Rationale
Why was this approach taken?

## Impact


Gobot1234 (changes requested):
This in it's current state will completely remove type information from all of the API. You'll need to add an if TYPE_CHECKING block with the imports still

mayankansys:
Check the working of the autocomplete.

mayankansys:
> This in it's current state will completely remove type information from all of the API. You'll need to add an if TYPE_CHECKING block with the imports still

@Gobot1234 Thanks for pointing out, I was currently working on it & it will be resolved in the upcoming commits.

PR #5318, opened 2026-08-12
refactor: updated from using() to ContextVar

### LINKED ISSUE #4928: Consider updating the `using()` context manager implementation
Regarding PyFluent's `using()` context manager, I received this useful insight from @h-krishnan: 

>You should probably use [contextvar](https://docs.python.org/3/library/contextvars.html) which handles both thread locality and async locality and is a core Python feature. Personally I don't like contextvars and I would rather see explicit passing of context rather than some behind the scene magic.

mayankansys:
## Context
The `using()` context manager previously relied on `threading.local` with a manual stack to track the active Fluent session. This only guaranteed thread-local isolation and did not work correctly with `asyncio` tasks, since a thread-local stack is shared across tasks running on the same thread.

## Change Summary
Replaced the `threading.local` stack implementation in context_managers.py with a `contextvars.ContextVar` to store the active session. `using()` now sets the context variable and resets it via a token when the context exits, instead of pushing/popping from a manual stack.

## Impact
* core implementation change from thread-local stack to `ContextVar`.
* documentation updated to reflect thread- and task-local isolation.


src/ansys/fluent/core/utils/context_managers.py
```diff
@@ -24,31 +24,22 @@
...
 __all__ = ("using",)
 
-_thread_local = local()
-
-
-def _get_stack():
-    if not hasattr(_thread_local, "stack"):
-        _thread_local.stack = []
-    return _thread_local.stack
+_active_session: ContextVar = ContextVar("active_session", default=None)
```
Gobot1234:
```suggestion
_active_session: ContextVar[BaseSession] = ContextVar("active_session", default=None)
```
I think it should be like this

src/ansys/fluent/core/utils/context_managers.py
```diff
@@ -24,31 +24,29 @@
...
+if TYPE_CHECKING:
+    from ansys.fluent.core.session import BaseSession
 
+__all__ = ("using",)
 
-def _get_stack():
-    if not hasattr(_thread_local, "stack"):
-        _thread_local.stack = []
-    return _thread_local.stack
+_active_session: ContextVar[BaseSession] = ContextVar("active_session", default=None)
```
prmukherj:
```suggestion
_active_session: ContextVar["BaseSession"] = ContextVar("active_session", default=None)
```

src/ansys/fluent/core/utils/context_managers.py
```diff
@@ -24,31 +24,29 @@
...
+if TYPE_CHECKING:
+    from ansys.fluent.core.session import BaseSession
 
+__all__ = ("using",)
 
-def _get_stack():
-    if not hasattr(_thread_local, "stack"):
-        _thread_local.stack = []
-    return _thread_local.stack
+_active_session: ContextVar["BaseSession"] = ContextVar("active_session", default=None)
```
prmukherj:
```suggestion
_active_session: ContextVar["BaseSession | None"] = ContextVar("active_session", default=None)
```

src/ansys/fluent/core/utils/context_managers.py
```diff
@@ -24,31 +24,31 @@
...
+__all__ = ("using",)
 
-def _get_stack():
-    if not hasattr(_thread_local, "stack"):
-        _thread_local.stack = []
-    return _thread_local.stack
+_active_session: ContextVar["BaseSession | None"] = ContextVar(
+    "active_session", default=None
+)
+_active_session: ContextVar["BaseSession | None"] = ContextVar(
```
mkundu1: duplicate

mkundu1 (approved):

prmukherj (approved):

src/ansys/fluent/core/utils/context_managers.py
prmukherj: same comment.

seanpearsonuk (approved):

PR #5323, opened 2026-08-18
fix: Default ui mode.

### LINKED ISSUE #5280: Default for UI_Mode for container launch
Currently default for UI_Mode is coded as:
```
    def _default(self):
        # Not using NO_GUI in windows as it opens a new cmd or
        # shows Fluent output in the current cmd if start <launch_string> is not used
        return self.HIDDEN_GUI if is_windows() else self.NO_GUI
```
`HIDDEN_GUI` doesn't work within container. We should make `NO_GUI` default in all cases except in standalone launch on Windows where `HIDDEN_GUI` will be used.

prmukherj:
## Context
`HIDDEN_GUI` doesn't work within container.

## Change Summary
Made `NO_GUI` default in all cases except in standalone launch on Windows where `HIDDEN_GUI` will be used.

### DIFF
--- a/src/ansys/fluent/core/launcher/container_launcher.py
+++ b/src/ansys/fluent/core/launcher/container_launcher.py
@@ -55,10 +55,11 @@ from ansys.fluent.core.launcher.fluent_container import (
 )
 from ansys.fluent.core.launcher.launch_options import (
     FluentMode,
+    UIMode,
     _get_argvals_and_session,
     get_remote_grpc_options,
 )
-from ansys.fluent.core.launcher.launcher_utils import ComposeConfig
+from ansys.fluent.core.launcher.launcher_utils import ComposeConfig, is_windows
 from ansys.fluent.core.launcher.process_launch_string import (
     _build_fluent_launch_args_string,
 )
@@ -235,6 +236,14 @@ class DockerLauncher:
         kwargs["insecure_mode"] = insecure_mode
         if certificates_folder is None and not insecure_mode:
             raise ValueError(CERTIFICATES_FOLDER_NOT_PROVIDED_AT_LAUNCH)
+
+        # HIDDEN_GUI is not supported inside Windows containers; downgrade to NO_GUI.
+        if is_windows() and UIMode(kwargs.get("ui_mode")) == UIMode.HIDDEN_GUI:
+            logger.warning(
+                "HIDDEN_GUI is not supported inside Windows containers. "
+                "Downgrading UI mode to NO_GUI."
+            )
+            kwargs["ui_mode"] = UIMode.NO_GUI
 
         self.argvals, self.new_session = _get_argvals_and_session(kwargs)
         if self.argvals.get("cleanup_on_exit") is None:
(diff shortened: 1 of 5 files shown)


seanpearsonuk (approved):

src/ansys/fluent/core/launcher/standalone_launcher.py
```diff
@@ -223,11 +223,17 @@ def __init__(
         """
         import ansys.fluent.core as pyfluent
 
+        ui_mode = kwargs.get("ui_mode")
+        if pyfluent.config.show_fluent_gui:
+            ui_mode = UIMode.GUI
+        elif ui_mode is None and is_windows():
```
mkundu1: I think it is better that the system-override "downgrades" the default (in container launcher) rather than "upgrades" it.

src/ansys/fluent/core/launcher/launch_options.py
```diff
@@ -184,7 +184,9 @@ class UIMode(FluentEnum):
 
     def _default(self):
         # Not using NO_GUI in windows as it opens a new cmd or
-        # shows Fluent output in the current cmd if start <launch_string> is not used
+        # shows Fluent output in the current cmd if start <launch_string> is not used.
+        # Note: HIDDEN_GUI is not supported inside Windows containers and will be
```
mkundu1: We don't have Windows containers.

This is a development side comment. So, we can say along this lines:
```
# Note: HIDDEN_GUI is not supported inside containers and the default value will be
# downgraded to NO_GUI on windows host automatically by the container launcher.
```

src/ansys/fluent/core/launcher/container_launcher.py
```diff
@@ -236,6 +237,14 @@ def __init__(
         if certificates_folder is None and not insecure_mode:
             raise ValueError(CERTIFICATES_FOLDER_NOT_PROVIDED_AT_LAUNCH)
 
+        # HIDDEN_GUI is not supported inside Windows containers; downgrade to NO_GUI.
+        if is_windows() and UIMode(kwargs.get("ui_mode")) == UIMode.HIDDEN_GUI:
+            logger.warning(
```
mkundu1, line 242: We don't need to show any message to users if the `ui_mode` is not specified by user. We can keep a debug level log if that is useful.

What is the current behaviour if user passes `ui_mode=UIMode.GUI` to container launcher? Passing `ui_mode=UIMode.HIDDEN_GUI` should behave or continue to behave in a similar manner.

src/ansys/fluent/core/launcher/launch_options.py
mkundu1: Do we have any user-facing documentation for the default value of `ui_mode`?

prmukherj: > Do we have any user-facing documentation for the default value of `ui_mode`?

No @mkundu1, we only have the api reference.

src/ansys/fluent/core/launcher/container_launcher.py
prmukherj: If the user explicitly passes `GUI` to the container then it will fail. However here it is default value, that's the reason for this guard, but I agree that it will also over-ride user's explicit input value.

@mkundu1, let's discuss this and decide. Thank you.

prmukherj: Created: https://github.com/ansys/pyfluent/issues/5333

mkundu1 (approved):

PR #5339, opened 2026-08-21
fix: issue with interruptible commands on interrupt.

### LINKED ISSUE #5227: [BUG]: iterate() raises RuntimeError when stopped via interrupt() from a solver event callback
### ✅ Before submitting

- [x] I have searched the [existing issues](https://github.com/ansys/pyfluent/issues) and did not find a duplicate.
- [x] I am using a Python virtual environment.
- [x] I have confirmed this is a PyFluent issue, not a problem with Ansys Fluent. See the [posting issues guide](https://fluent.docs.pyansys.com/version/stable/contributing/contributing_contents.html#post-issues).

### 📝 Description of the bug

Calling `session.settings.solution.run_calculation.interrupt()` from within a `SolverEvent.ITERATION_ENDED` callback causes the concurrently-blocking `run_calculation.iterate()` gRPC call on the main thread to raise `RuntimeError: ()` rather than returning normally — even though Fluent itself stops the solver cleanly and prints `Done. Operation stopped.`

The consequence is that any workflow code following the `iterate()` call (writing case files, post-processing, etc.) is silently skipped. This makes event-driven convergence monitoring unreliable in practice. 

### 🔁 Steps to reproduce

```python
from ansys.fluent.core import SolverEvent, IterationEndedEventInfo

session = pyfluent.launch_fluent(mode="solver")
# ... load case, apply settings ...

def on_iteration_ended(session, event_info: IterationEndedEventInfo) -> None:
    if event_info.index >= 10:
        # Signal Fluent to stop — Fluent stops cleanly
        session.settings.solution.run_calculation.interrupt()

session.events.register_callback(SolverEvent.ITERATION_ENDED, on_iteration_ended)

# Raises RuntimeError: () when the interrupt signal is processed
session.settings.solution.run_calculation.iterate(iter_count=1000)

# Never reached — skipped due to unhandled RuntimeError
session.settings.file.write_case_data(file_name="result.cas.h5")
```

### 🎯 Expected behavior

`run_calculation.iterate()` returns normally when stopped by `interrupt()`. Fluent has already stopped cleanly — the gRPC response for a solver-stopped iterate operation should be treated as a successful completion, not an error. Execution should continue to any statements following the `iterate()` call.

Optionally, `iterate()` could return a status indicator (e.g. an `IterationResult` enum with `COMPLETED` and `INTERRUPTED` variants) so callers can distinguish the two outcomes if needed.

### 💥 Actual behavior / traceback

```shell
  File "workflow.py", line N, in __call__
    session.settings.solution.run_calculation.iterate(iter_count=1000)
  File ".../ansys/fluent/core/solver/flobject.py", line 2066, in __call__
    return self.execute_command(**kwds)
  File ".../ansys/fluent/core/solver/flobject.py", line 2006, in execute_command
    ret = self._execute_command(*args, **scmKwds)
  File ".../ansys/fluent/core/solver/flobject.py", line 1977, in _execute_command
    ret = self.flproxy.execute_cmd(self._parent.path, self.obj_name, **kwds)
  File ".../ansys/fluent/core/services/settings.py", line 383, in execute_cmd
    response = self._service_impl.execute_cmd(request)
  File ".../ansys/fluent/core/services/settings.py", line 117, in execute_cmd
    return self._stub.ExecuteCommand(request, metadata=self._metadata)
  File ".../ansys/fluent/core/services/interceptors.py", line 164, in _intercept_call
    raise new_ex from None
RuntimeError: ()
```

### 📦 PyFluent version

0.39.0

### 🌀 Ansys Fluent version

2025 R2 (25.2)

### 🐍 Python version

3.12

### 💻 Operating system

Linux

### 📌 Additional context

_No response_

prmukherj:
## Context
Calling `session.settings.solution.run_calculation.interrupt()` from within a SolverEvent.ITERATION_ENDED callback causes the concurrently-blocking `run_calculation.iterate()` gRPC call on the main thread to raise RuntimeError: () rather than returning normally — even though Fluent itself stops the solver cleanly and prints Done. Operation stopped.

The consequence is that any code following the iterate() call (writing case files, post-processing, etc.) is silently skipped. This makes event-driven convergence monitoring unreliable in practice.

## Change Summary
1. Pass is_interruptible from the solver to the settings root.
2. Use that to check if any interruptible commands raise `RuntimeError: ()`, and if so swallow the error.
3. Added test with the case in the particular bug to check the behavior for all Fluent versions.

## Rationale
Ideally Fluent should not raise an error in this case, but currently it is interrupting cleanly and raising an error with blank message unlike in other cases where the message strings are populated appropriately. Hence using that to handle this particular bug.

## Impact
Users will not face the situation as described, and will be able to develop similar workflows seamlessly.

### DIFF
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -2271,8 +2271,10 @@ class Command(BaseCommand):
             self._root._on_interrupt(self)
             raise KeyboardInterrupt
         except RuntimeError as ex:
+            # TODO: Remove this Fluent specific logic during later refactoring.
             # RuntimeError("()") is Fluent's gRPC signal for a clean solver stop
-            # via interrupt(). Suppress it for interruptible commands so that
+            # via interrupt() or via manual stop from a GUI in connected instances.
+            # Suppress it for interruptible commands so that
             # workflow code following iterate()/calculate() continues normally.
             _is_interruptible = getattr(self._root, "_is_interruptible_command", None)
             if (
@@ -2300,8 +2302,10 @@ class CommandWithPositionalArgs(BaseCommand):
             self._root._on_interrupt(self)
             raise KeyboardInterrupt
         except RuntimeError as ex:
+            # TODO: Remove this Fluent specific logic during later refactoring.
             # RuntimeError("()") is Fluent's gRPC signal for a clean solver stop
-            # via interrupt(). Suppress it for interruptible commands so that
+            # via interrupt() or via manual stop from a GUI in connected instances.
+            # Suppress it for interruptible commands so that
             # workflow code following iterate()/calculate() continues normally.
             _is_interruptible = getattr(self._root, "_is_interruptible_command", None)
             if (
--- a/src/ansys/fluent/core/solver/flobject.py
+++ b/src/ansys/fluent/core/solver/flobject.py
@@ -524,6 +524,10 @@ class Base:
         """Set interrupt method."""
         self._setattr("_on_interrupt", on_interrupt)
 
+    def _set_is_interruptible_command(self, is_interruptible_command):
+        """Set interruptible command checker."""
+        self._setattr("_is_interruptible_command", is_interruptible_command)
+
     def _set_file_transfer_service(self, file_transfer_service):
         """Set file_transfer_service."""
         self._setattr("_file_transfer_service", file_transfer_service)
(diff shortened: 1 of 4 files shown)


src/ansys/fluent/core/solver/flobject.py
```diff
@@ -2266,6 +2270,19 @@ def __call__(self, **kwds):
         except KeyboardInterrupt:
             self._root._on_interrupt(self)
             raise KeyboardInterrupt
+        except RuntimeError as ex:
+            # RuntimeError("()") is Fluent's gRPC signal for a clean solver stop
+            # via interrupt(). Suppress it for interruptible commands so that
+            # workflow code following iterate()/calculate() continues normally.
+            _is_interruptible = getattr(self._root, "_is_interruptible_command", None)
```
prmukherj, line 2279: # TODO: Refactor this method to reduce Fluent's dependency.

prmukherj: Add a comment on the scenarios when this occurs.

mkundu1 (approved):

seanpearsonuk:
@prmukherj Isn't this all done and ready?

prmukherj:
> @prmukherj Isn't this all done and ready?

Ready now @seanpearsonuk

seanpearsonuk (approved):

PR #5371, opened 2026-09-04
docs: Update `from_connection` user guide [skip tests]

### LINKED ISSUE #5291: Problems with Connect to an existing session user guide
- The section is verbose, due to:
  - repetition of a self-evident fact: a running Fluent instance is needed.
  - over-explanation of meshing session types, which can mostly be hyperlinked through the guide's _sessions_ section.
- It misses information about how the info file path/name can eaily be discovered in Fluent. Leaving this out is certain to create frustration. Similarly how it can be specified, where a brief overview would be helpful.
- The recommendation to use the info file versus its contained information needs to be balanced against whether the connection is local or remote. 
- Unintentional line break after "supports solving," followed by tautological "intentionally want".
- "Interface" terminology is not a good fit, and more direct language such as "session type" is more accurate and clearer.

prmukherj:
Refactor and make the connecting to session's user guide more user friendly.

Please refer #5291, for details of the updates required.

### DIFF
--- a/doc/source/user_guide/session/launching_ansys_fluent.rst
+++ b/doc/source/user_guide/session/launching_ansys_fluent.rst
@@ -113,7 +113,7 @@ to enable the Fluent gRPC server and choose the server info file path.
     - Use ``PureMeshing.from_connection()`` if the session was launched for **meshing only**.
     - Use ``Meshing.from_connection()`` if the session supports **meshing and solving**.
 
-    A ``Meshing`` session type is not recommended for a **meshing-only** session.
+    A ``PureMeshing`` session type is recommended for a **meshing-only** session.
 
 
 .. _connect_to_pim_session:
--- a/doc/source/user_guide/session/launching_ansys_fluent.rst
+++ b/doc/source/user_guide/session/launching_ansys_fluent.rst
@@ -78,22 +78,22 @@ Use this method when:
 Connect to an existing session
 ------------------------------
 
-The :meth:`from_connection() <ansys.fluent.core.session.utilities.SessionBase.from_connection>` method connects to a previously launched Fluent session.
-
-Use this method when:
-
-- Fluent was launched externally or earlier.
-- You need to connect from a different process or system.
+The :meth:`from_connection() <ansys.fluent.core.session.utilities.SessionBase.from_connection>` method connects to an externally launched Fluent session from a local or remote Python process.
 
 **Prerequisites**
 
-Before connecting, you must launch Fluent externally with the gRPC server enabled.
+Before connecting, start Fluent with ``<fluent_executable> -sifile=<server_info_file_name>``
+to enable the Fluent gRPC server and choose the server info file path.
 
 .. note::
 
-    After starting the gRPC server, Fluent writes connection information to the server info file. The file contains the IP address, port, and password needed for connection.
-    
-    The recommended approach is to connect using the server info file path, which encapsulates all connection information.
+  Fluent writes the IP address, port, and password to the server info file. If you do not
+  specify the file path with ``-sifile``, Fluent generates the file in the directory where
+  ``fluent.exe`` resides.
+
+  For local connections, pass the server info file path when it is accessible to Python. For
+  remote connections, use the IP address, port, and password from the file, unless the same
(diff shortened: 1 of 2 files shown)


doc/source/user_guide/session/launching_ansys_fluent.rst
```diff
@@ -107,24 +107,13 @@ Before connecting, you must launch Fluent externally with the gRPC server enable
...
+    When connecting to an existing Fluent session:
 
     - Use ``PureMeshing.from_connection()`` if the session was launched for **meshing only**.
     - Use ``Meshing.from_connection()`` if the session supports **meshing and solving**.
-    - You may also use ``PureMeshing.from_connection()`` with a session that supports solving,
-    if you intentionally want access **only to meshing features**.
 
-    A ``Meshing`` interface is not recommended for a **meshing-only** session, because
-    ``switch_to_solver()`` would raise an error in that case.
+    A ``Meshing`` session type is not recommended for a **meshing-only** session.
```
prmukherj, line 115:
```suggestion
    A ``PureMeshing`` session type is recommended for a **meshing-only** session.
```

doc/source/user_guide/session/launching_ansys_fluent.rst
```diff
@@ -78,22 +78,22 @@ Use this method when:
...
-
-- Fluent was launched externally or earlier.
-- You need to connect from a different process or system.
+The :meth:`from_connection() <ansys.fluent.core.session.utilities.SessionBase.from_connection>` method connects to an externally launched Fluent session from a local or remote Python process.
 
 **Prerequisites**
 
-Before connecting, you must launch Fluent externally with the gRPC server enabled.
+Before connecting, start Fluent with ``<fluent_executable> -sifile=<server_info_file_name>``
+to enable the Fluent gRPC server and choose the server info file path.
```
seanpearsonuk: Maybe something like specify rather than choose, because you're actually directing the process to use that file.

prmukherj: Right @seanpearsonuk, Thank you, Done.

seanpearsonuk (approved):

doc/source/user_guide/session/launching_ansys_fluent.rst
```diff
@@ -78,22 +78,22 @@ Use this method when:
...
-    After starting the gRPC server, Fluent writes connection information to the server info file. The file contains the IP address, port, and password needed for connection.
-    
-    The recommended approach is to connect using the server info file path, which encapsulates all connection information.
+  Fluent writes the IP address, port, and password to the server info file. If you do not
+  specify the file path with ``-sifile``, Fluent generates the file in the directory where
+  ``fluent.exe`` resides.
+
+  For local connections, pass the server info file path when it is accessible to Python. For
+  remote connections, use the IP address, port, and password from the file, unless the same
+  server info file path is available from the client machine.
```
mkundu1: I don't think we should have a different recommendation for remote connection. One can copy the server-info file from the remote machine and connect using it from the local machine.

prmukherj: @mkundu1, I think that is a workaround, right? Whether the user copies it from remote to his local system is not what we should specify. I think that is understandable form the rest of the doc. Please let me know if you agree.

prmukherj: Keeping this only then instead of this paragraph.

"The recommended approach is to connect using the server info file path, which encapsulates all connection information."

prmukherj: @seanpearsonuk, sounds good?

mkundu1: Another point is: if the user doesn't have permission to copy the file from the remote machine, usually they won't have permission to read the file-contents. If another user is sending them the ip/port/password, they could just send the server-info file.

prmukherj: Done. Thanks,

mayankansys (approved):

mkundu1 (approved):

PR #5375, opened 2026-09-05
feat: runtime typechecking

mayankansys:
## Context
PyFluent had no runtime type-checking mechanism to catch API misuse early. Type errors propagated downstream as obscure failures, making debugging difficult for users. This PR opt-in runtime type-checking support (e.g., via beartype), configurable via environment variable.

## Change Summary
This PR implements abstraction layer + config, implicit-Optional fixes and decorator fixes.
*  A new module `_type_checking.py` which is a thin wrapper around the type-checking library. (e.g., via beartype)
* Added runtime_type_checking descriptor to Config class, reading from **PYFLUENT_RUNTIME_TYPE_CHECKING** env var ( which must be set before import )
* Annotation cleanups : implicit-Optional parameters, setting up __signatures__ on both deprecation decorators for interospection compatibility. 
* Dependencies etc. 

## Impact
Now Pyfluent will support the runtime typechecking. When disabled (the default), there is no impact on the API surface. When enabled, type violations are reported as PyFluentTypeCheckingError exceptions. The feature is optional and can be installed via `pip install ansys-fluent-core[type-checking]` users who do not install the type-checking extra can continue using PyFluent without beartype.

**#NOTE: There are further upcoming PRs as well.**


tests/test_runtime_type_checking.py
```diff
@@ -0,0 +1,186 @@
...
+    from beartype.roar import BeartypeCallHintParamViolation
+
+    monkeypatch.setattr(_type_checking, "_HOOK_INSTALLED", True)
+
+    @_type_checking.runtime_type_check
+    def fn(x: int) -> int:
+        return x
+
+    assert fn(1) == 1
+    with pytest.raises(BeartypeCallHintParamViolation):
```
mkundu1, line 124: Ideally, we should not directly emit beartype specific exceptions from PyFluent and ask our users to handle them. The backends-specific exception should be kept behind a PyFluent-owned exception.

mayankansys: Correct @mkundu1. I have added it now, initially planner to make it Pyfluent owned exception in the Phase 2.

src/ansys/fluent/core/module_config.py
```diff
@@ -37,12 +37,24 @@
...
+)
+from ansys.fluent.core._type_checking import ENV_VAR as _TYPE_CHECKING_ENV_VAR
+
 __all__ = ("config",)
 
 
 TConfig = TypeVar("TConfig", bound="Config")
 
 
+# ``TConfig`` is bound to a forward reference which cannot be resolved while the
```
mayankansys, line 54: Should we keep the decorator ? If we didn't put the decorator we get the `beartype.roar.BeartypecallHintForwardRefException`.

PR #5380, opened 2026-09-09
fix: Maximum Fluent version's compatibility.

prmukherj:
## Context
While trying to connect PyFluent with newer Fluent versions (more than in the list of compatible versions), we used to throw an error that it is not supported.

## Change Summary
1. Updated FluentVersion._missing_() so versions newer than the highest known version emit PyFluentUserWarning and use the highest known FluentVersion

2. Older or unsupported intermediate versions still raise AnsysVersionNotFound.

3. Added a focused test for FluentVersion("28.2.0")

## Rationale
To support newer Fluent release without updating the compatibility list urgently.

### DIFF
--- a/doc/changelog.d/5380.fixed.md
+++ b/doc/changelog.d/5380.fixed.md
@@ -1,0 +1,1 @@
+Maximum Fluent version's compatibility.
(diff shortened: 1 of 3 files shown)


seanpearsonuk:
@prmukherj Are you re-working this or are we not going forward with this idea?

prmukherj:
> @prmukherj Are you re-working this or are we not going forward with this idea?

https://github.com/ansys/pyfluent/pull/5381
@seanpearsonuk, please review this one. Thanks

PR #5381, opened 2026-09-09
fix: Maximum Fluent version's compatibility.

prmukherj:
## Context
While trying to connect PyFluent with newer Fluent versions (more than in the list of compatible versions), we used to throw an error that it is not supported.

## Change Summary
1. Updated FluentVersion._missing_() so versions newer than the highest known version emit PyFluentUserWarning and use the highest known FluentVersion

2. Older or unsupported intermediate versions still raise AnsysVersionNotFound.

3. Added a focused test for FluentVersion("28.2.0")

## Rationale
To support newer Fluent release without updating the compatibility list urgently.

### DIFF
--- a/src/ansys/fluent/core/utils/fluent_version.py
+++ b/src/ansys/fluent/core/utils/fluent_version.py
@@ -124,7 +124,8 @@ class FluentVersion(Enum):
                     return member
 
             latest = next(iter(cls))
-            if _version_to_integer(requested_version) > latest.number:
+            version_as_int = _version_to_integer(requested_version) # requires no-raise
+            if version_as_int is None or _version_to_integer(requested_version) > latest.number:
                 warnings.warn(
                     f"Fluent version '{requested_version}' is newer than the highest "
                     f"supported version '{latest.value}'; using '{latest.value}' instead.",
--- a/src/ansys/fluent/core/utils/fluent_version.py
+++ b/src/ansys/fluent/core/utils/fluent_version.py
@@ -127,7 +127,8 @@ class FluentVersion(Enum):
             version_as_int = _version_to_integer(requested_version) # requires no-raise
             if version_as_int is None or _version_to_integer(requested_version) > latest.number:
                 warnings.warn(
-                    f"Fluent version '{requested_version}' is newer than the highest "
+                    # should actually handle the two different scenarios separately
+                    f"Fluent version '{requested_version}' is either unrecognized or newer than the highest "
                     f"supported version '{latest.value}'; using '{latest.value}' instead.",
                     PyFluentUserWarning,
                     stacklevel=2,
--- a/src/ansys/fluent/core/utils/fluent_version.py
+++ b/src/ansys/fluent/core/utils/fluent_version.py
@@ -88,8 +88,10 @@ def _version_to_integer(version: Any) -> int:
     try:
         parts = str(version).split(".")
         return int(parts[0] + parts[1]) if len(parts) > 1 else int(parts[0])
-     except Exception:
-        pass # or explicitly return None
+    except Exception:
+        return (
+            None  # Return None instead of raising an exception for unrecognized formats
+        )
 
 
 @total_ordering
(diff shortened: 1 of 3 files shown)


src/ansys/fluent/core/utils/fluent_version.py
```diff
@@ -103,15 +111,29 @@ class FluentVersion(Enum):
...
+            if normalized_version.count(".") == 1:
+                normalized_version += ".0"
             for member in cls:
-                if version == member.value:
+                if normalized_version == member.value:
                     return member
 
-        raise AnsysVersionNotFound(version[:-2])
+            latest = next(iter(cls))
+            if _version_to_integer(requested_version) > latest.number:
```
seanpearsonuk:
```suggestion
            version_as_int = _version_to_integer(requested_version) # requires no-raise
            if version_as_int is None or _version_to_integer(requested_version) > latest.number:
```

src/ansys/fluent/core/utils/fluent_version.py
```diff
@@ -103,15 +111,29 @@ class FluentVersion(Enum):
...
+                warnings.warn(
+                    f"Fluent version '{requested_version}' is newer than the highest "
```
seanpearsonuk:
```suggestion
                    # should actually handle the two different scenarios separately
                    f"Fluent version '{requested_version}' is either unrecognized or newer than the highest "
```

src/ansys/fluent/core/utils/fluent_version.py
```diff
@@ -81,6 +83,12 @@ def get_version_for_file_name(version: str | None = None, session=None):
     return "".join(version.split(".")[0:2])
 
 
+def _version_to_integer(version: Any) -> int:
+    """Convert a Fluent version to its integer representation."""
+    parts = str(version).split(".")
+    return int(parts[0] + parts[1]) if len(parts) > 1 else int(parts[0])
```
seanpearsonuk:
```suggestion
    try:
        parts = str(version).split(".")
        return int(parts[0] + parts[1]) if len(parts) > 1 else int(parts[0])
     except Exception:
        pass # or explicitly return None
```

src/ansys/fluent/core/utils/fluent_version.py
prmukherj: Good Point @seanpearsonuk. It might be unrecognized as well now. Thanks.

seanpearsonuk: BTW I accidentally just committed one of the suggestions!!

seanpearsonuk: Note that even if we don't change the approach regarding whether or not to accept unrecognized version strings, it's nevertheless important that we trap low-level exceptions arising from the str->int converter before they reach the user.

seanpearsonuk (approved):

src/ansys/fluent/core/utils/fluent_version.py
```diff
@@ -103,15 +116,34 @@ class FluentVersion(Enum):
...
 
-        raise AnsysVersionNotFound(version[:-2])
+            latest = next(iter(cls))
+            version_as_int = _version_to_integer(requested_version)  # requires no-raise
+            if (
+                version_as_int is None
+                or _version_to_integer(requested_version) > latest.number
+            ):
+                warnings.warn(
+                    # should actually handle the two different scenarios separately
```
seanpearsonuk: This comment was added in the edit to direct you 😄 

What it means is: we actually know which of the two cases it is but the current message leaves the user to figure it out.

prmukherj: 🙂

src/ansys/fluent/core/utils/fluent_version.py
```diff
@@ -103,15 +116,33 @@ class FluentVersion(Enum):
...
+            if normalized_version.count(".") == 1:
+                normalized_version += ".0"
             for member in cls:
-                if version == member.value:
+                if normalized_version == member.value:
                     return member
 
-        raise AnsysVersionNotFound(version[:-2])
+            latest = next(iter(cls))
+            version_as_int = _version_to_integer(requested_version)  # requires no-raise
```
prmukherj:
```suggestion
            version_as_int = _version_to_integer(requested_version)
```

src/ansys/fluent/core/utils/fluent_version.py
seanpearsonuk: N.b. My comment points to the possibility of handling the two errors separately so that the user receives more precise information.

prmukherj: @seanpearsonuk, got it, Thanks. Updated.

seanpearsonuk (approved):

mkundu1 (approved):

PR #5382, opened 2026-09-09
feat: Add retries to download_file.

### LINKED ISSUE #5031: add retries to download file

prmukherj:
## Context
Some arguments like `max_retries` were missing from `download_file` method in examples.

## Change Summary
Instead of having a local implementation, just use the implementation already in `ansys-tools-common` by flattening out paths wherever required to match PyFluent's style.


tests/test_session.py
```diff
@@ -186,13 +186,103 @@ def test_base_meshing_cannot_be_instantiated_directly():
...
+        return str(target)
+
+    monkeypatch.setattr(examples.downloads.download_manager, "download_file", _fake)
+
+    result = examples.download_file(
+        "elbow.cas.h5",
+        "pyfluent/mixing_elbow",
+        save_path=str(tmp_path),
+        return_without_path=True,
+    )
```
seanpearsonuk: That's an awkward API because the `return_without_path` argument changes the semantics of the return value. It makes sense to introduce a new method that returns the downloaded file as a `pathlib.Path`, e.g. `download_file_path()`, and deprecate `return_without_path` on the existing `download_file()` method. This would give us a cleaner API while preserving backwards compatibility for existing callers.

@prmukherj @mkundu1 
@RobPasMue @AlejandroFernandezLuces

prmukherj: yes, makes sense @seanpearsonuk. This is a pyfluent specific implementation. Will update it. Thank you

seanpearsonuk: Ok, thanks, explains why I didn't remember it from the recent PR.

prmukherj: Resolved.

src/ansys/fluent/core/examples/__init__.py
```diff
@@ -21,6 +21,13 @@
...
-"""Provides a module to download a file."""
+"""Utilities for downloading and locating Fluent example files."""
 
-from .downloads import RemoteFileNotFoundError, download_file, path  # noqa: F401
+from .downloads import delete_downloads, download_file, get_file_without_path, path
+
+__all__ = [
+    "delete_downloads",
+    "download_file",
+    "get_file_without_path",
```
prmukherj:
```suggestion
```

PR #5391, opened 2026-09-15
docs: Minor improvement in docs

### LINKED ISSUE #5111: Meshing doc updates
1. Create a legacy meshing node in the API reference. 
   a. Put tui + legacy mwf items there. 
   b. In the user guide, we should move the legacy workflows to the legacy user guide section.     That's consistent with what we will do with the API reference. We won't even refer to it in meshing workflows.
   c. Every node in the documentation should clearly explain the roles of all its direct children 
2. Make the two meshing_worklow pages adjacent under a single node. Clarify the relationships
between them in a parent page which is also the new landing page linked from the "meshing workflow" on the API reference landing page.
3. ~~Make the (Meshing)Workflow types directly constructible based on a meshing session context.
That will help clarify the meshing documentation and make it fully self-contained.~~
4. ~~Make the meshing contents hierarchy much more shallow. Remove datamodel altogether.~~
5. Remove all technical jargon from the docs where it cannot be justified.
```
        Singleton <group> -> <group> group
        
        Command <command> -> <command> command
```
6. Add API doc links from the user guide code.
7. ~~Consider a shared base class for the meshing sessions.~~
8. Rectify intellisense issues.

 

meshing
  meshing_worklow
    application
  general
  parts
  part_files
  task_object
  meshing_utilities
  preferences

legacy_meshing
  tui
  meshing
  workflow
  etc.

9. Re-word top-level user guide statement: "Workflows are divided into the new Meshing workflows and Classic meshing workflow, which align with the journal syntax."

prmukherj:
Minor improvement in docs

Singleton <group> -> <group> group
Command <command> -> <command> command
Query <query> -> <query> query

Improve the landing page docs.

Rectify intelli-sense issues with enhanced meshing workflows: Retrieve the all possible tasks under each workflow type by initializing each workflow type once within datamodelgen.py.
Generated a `meshing_workflow_tasks.pyi` with all the task items under particular workflow type.


prmukherj:
1. Put the static data into documentation.
2. Runtime class generation is not required for the given set of data now.

prmukherj:
This will only add intellisense support. Further work to be carried out later as part of: https://github.com/ansys/pyfluent/issues/5393

doc/source/user_guide/meshing/meshing_contents.rst
```diff
@@ -3,9 +3,9 @@
 ================
 Meshing workflow
 ================
-The guided workflows in Fluent meshing are designed to make it easier to generate a
-volume mesh starting from CAD geometries. You can use PyFluent to access the guided workflows.
-Workflows are divided into the new Meshing workflows and Classic meshing workflow, which align with the journal syntax.
+The new Meshing workflows in PyFluent are guided processes for generating a volume
+mesh from CAD geometry. PyFluent recommends using these workflows. PyFluent also
```
seanpearsonuk: The first sentence doesn't work, because it implies that there are some new workflows that generate mesh and older workflows didn't do that. The sentence falls down because it tries to do too much.

We should avoid calling them "new" workflows. This is just the way that the user is directed to do meshing in PyFluent. Calling it "new" dwells on the history and is unclear. Besides, it says "new ... workflows", but there aren't any new workflows!

prmukherj: PyFluent recommends using guided workflows to generate volume meshes from CAD
geometry. It also provides the classic meshing workflow, which aligns with Fluent
journal syntax.

prmukherj: @seanpearsonuk, this is better?

prmukherj: I think it is short and simple and conveys the message properly.

seanpearsonuk: This is closer, but there's a gap: "It also provides the classic meshing workflow" introduces "classic" without ever establishing what it's classic *relative to*. A reader landing here won't know there are two interfaces at all yet, so "classic" has nothing to anchor to.

There's also a framing issue worth fixing while we're at it: this isn't new workflows vs. old workflows. It's the same underlying Fluent meshing workflows, exposed through two different interfaces. Calling them "new" and "classic" workflows implies the *workflows* differ, when really it's the *interface* that differs. ("New" is also a timestamp that'll age badly in docs, so worth avoiding regardless.)

Proposal:

> PyFluent provides guided workflows for generating a volume mesh from CAD geometry. These workflows are accessible through two interfaces: a PyFluent-native interface, and the classic Python interface, which matches Fluent's journaling syntax. PyFluent recommends using the PyFluent-native interface.

This sets up "classic" properly by naming what it's contrasted with first, keeps the new/old distinction off the workflows entirely, and drops "new" in favour of "-native," which won't go stale.

prmukherj: Thanks @seanpearsonuk. This sounds good. Will use this.

prmukherj: Done.

mkundu1 (approved):

mayankansys (approved):

doc/source/user_guide/meshing/meshing_contents.rst
```diff
@@ -3,9 +3,10 @@
 ================
 Meshing workflow
 ================
-The guided workflows in Fluent meshing are designed to make it easier to generate a
-volume mesh starting from CAD geometries. You can use PyFluent to access the guided workflows.
-Workflows are divided into the new Meshing workflows and Classic meshing workflow, which align with the journal syntax.
+PyFluent provides guided workflows for generating a volume mesh from CAD geometry.
+These workflows are accessible through two interfaces: a PyFluent-native interface,
```
seanpearsonuk: I like the **PyFluent-native interface** phrasing as it nails the meaning down so clearly. It's ideal if we use this precise naming everywhere: "PyFluent-native interface", and "PyFluent-native meshing workflow interface" wherever the expansion helps.

I notice that words like "new" and "enhanced" are already absent in `new_meshing_workflows.rst` which is good. That file could possibly lead by mentioning that it's about the "PyFluent-native meshing workflow interface" - there is some spare space to fill under the main heading there, and it wouldn't hurt to reinforce what that page is all about right there.

We should also check on usage of "new" and "enhanced" in public docstrings and other RST files.

The natural flipside to the "PyFluent-native interface" is the "Fluent-native interface". We should replace "Classic" with that because like "PyFluent-native interface", it describes exactly what it is. "Enhanced" and "Classic" don't achieve the same clarity.

src/ansys/fluent/core/meshing/__init__.py
```diff
@@ -57,42 +57,42 @@ def _validate_meshing_session(session: BaseMeshing) -> bool:
 
 
 def _fallback_check(session: PureMeshing | Meshing, legacy: bool | None) -> bool:
-    """Determine whether to use legacy workflow implementation.
+    """Determine whether to use the Fluent-native interface.
```
prmukherj:
```suggestion
    """Determine whether to use the legacy PyFluent-native interface.
```

seanpearsonuk (approved):

PR #5394, opened 2026-09-16
docs: Hide private members from toctree in api docs [skip tests]

prmukherj:
## Context
Some private members were leaking to the tree of the api reference docs although they were hidden by default from the rendered docs due to sphinx settings to hide private members. The toctree is generated before rendering docs via sphinx and thus this manual handling is required for toctree generation.

## Change Summary
Hide private members from toctree in api docs.

## Impact
User to get a cleaner doc. experience.

For example:

Earlier:
image

Now:
image


doc/rstgen.py
```diff
@@ -373,6 +372,8 @@ def _write_doc(menu: type, mode: str, is_datamodel: bool):
             f.write(".. toctree::\n")
             f.write("   :hidden:\n\n")
             for member in _get_sorted_members(menu["with_members"]):
+                if member.startswith("_"):
+                    continue
```
seanpearsonuk, line 376: What's the overall effect of this change? I don't know the entire context.

prmukherj: @seanpearsonuk, it only hides all private members from toctree.

mkundu1 (approved):

seanpearsonuk (approved):

PR #5398, opened 2026-09-16
fix: Support if command args are arbitrary dicts.

prmukherj:
## Context
We used to get this below warning because the command argument `setup.physics.restore_topology.state` is of generic dict-type.

Unable to find base class for 'state' (type = 'dict'). Falling back to String.

## Change Summary
Based on a speculation that we would have situations in which a command argument is an arbitrary dict, we are supporting that.

We remove "map" and support "dict" instead. 

## Rationale
To support command args when it is an arbitrary dict.


mkundu1 (approved):

mayankansys (approved):

src/ansys/fluent/core/solver/flobject.py
```diff
@@ -2042,6 +2042,8 @@ def set_state(self, state: StateT | None = None, **kwargs):
 class Map(SettingsBase[DictStateType]):
```
seanpearsonuk: @prmukherj 

There's also no reason not to rename this `Dict`, given that it lives safely within a namespace, so doesn't clash with e.g. `typing.Dict`. It's the default choice, like  `typing.Dict`, `numba.types.Dict`, etc. `Map` evokes the Python `map` function rather than the C++ `std::map` container template.

tests/test_flobject.py
```diff
@@ -612,6 +612,17 @@ def test_get_cls_base_resolution_and_fallback(caplog):
     cls, _ = flobject.get_cls("create", {"type": "command"}, version="271")
     assert issubclass(cls, flobject.CommandWithPositionalArgs)
 
+    dict_cls, _ = flobject.get_cls("state", {"type": "dict"}, version="271")
+    assert issubclass(dict_cls, flobject.Map)
```
prmukherj:
```suggestion
    assert issubclass(dict_cls, flobject.Dict)
```

seanpearsonuk (approved):

